PR #8, opened 2026-06-16
feat: migrate GUI spike to the full GUI chat protocol

snakajima:
## Summary

Upgrades the GUI spike from its stripped-down, parallel protocol to a **high-fidelity mirror of MulmoClaude's real GUI chat protocol**, so the eventual MulmoClaude M3 port is closer to copy-paste than translation. (The earlier mis-port — M3 first carried the spike's stripped-down shapes and had to be redone — is exactly what this prevents.)

Three shifts:

1. **Two config-driven plugins.** The single hard-coded MCP server is gone. Each tool is a self-contained plugin package under `plugins/<name>/` — a server half (`meta.js`, `definition.js`, `server.js`, plain JS for Node) and a frontend half (`View.vue`, `index.ts` for Vite). **`plugins/plugins.json`** (`{ "enabled": [...] }`) is the configuration data that drives loading on both sides: `server/plugins-registry.js` and `src/plugins-registry.ts` (via `import.meta.glob` filtered by the config). Flip the list → claude is offered fewer tools and the panel renders fewer views, no code change.

2. **Full toolResult model + per-plugin REST + broker.** `server/mcp/broker.js` replaces `present-markdown.js`: it registers one MCP tool per enabled plugin and is a thin HTTP bridge — on a call it POSTs args to the plugin's REST route (`/api/<namespace>`), gets back an envelope `{ data, message, instructions }`, and (only when `data` is set — *data gates rendering*) POSTs a toolResult to `POST /api/agent/toolResult`. That route stores it **keyed by session id** and publishes on `session:<id>`. The panel renders each result via `getPlugin(toolName).viewComponent` — no hard-coded `type` switch.

3. **GUI→LLM feedback types into the PTY.** The blocking `/api/gui/answer` long-poll (`pendingForms`, waiters, timeouts) is **deleted**. Plugin views get a `sendTextMessage(text)` prop; the form view builds a markdown summary on submit and calls it. `App.vue` wires that to `Terminal.sendText()`, which writes into the PTY over the existing `/ws` input channel, followed by a **separate delayed `\r`** (~60 ms, so the TUI submits instead of pasting a newline). A submitted form is just the user's next turn — exactly like MulmoClaude. `presentForm` no longer blocks.

## Persistence

Chat + message history already live in the terminal and Claude's `.jsonl` (resume), so the only GUI-side store is the **list of toolResults per session id** (`toolResultsBySession`, in-memory for the spike), replayed from `GET /api/agent/toolResults/:id` on revisit. A view's state change (e.g. a submitted form's `viewState`) is persisted by re-POSTing the same `uuid`, which upserts in place — so a revisited session shows the form already submitted.

## Verification

- ✅ `npm run build` (vue-tsc + vite) and `npm run lint` clean.
- ✅ End-to-end via an MCP client: `tools/list` returns the two config-driven tools; `tools/call` for both flows through the per-plugin REST routes + broker into the toolResult store; `GET …/toolResults/:id` replays both.
- ✅ Re-POSTing a result's `uuid` upserts (no duplicate, `viewState` preserved).
- ✅ Flipping `plugins.json` to `["markdown"]` drops `presentForm` from both the tool list and `--allowedTools`.
- ✅ Confirmed working in the running app, **including persistence of GUI views** across session revisits.

## Files

- **New:** `plugins/plugins.json`, `plugins/{markdown,form}/*`, `server/plugins-registry.js`, `src/plugins-registry.ts`, `server/mcp/broker.js`.
- **Modified:** `server/index.js`, `src/App.vue`, `src/components/Terminal.vue`, `src/components/GuiPanel.vue`, `tsconfig.app.json`, `docs/gui-protocol-spike.md` (Phase III).
- **Removed:** `server/mcp/present-markdown.js`, `src/components/GuiForm.vue` (its logic moved into `plugins/form/View.vue`).

## Out of scope (kept lean)

No sidebar `previewComponent` thumbnails, no codegen plugin barrels, no workspace dirs / PDF generation. The runtime *contract* is faithful; the build-time plumbing stays minimal.


snakajima:
Thanks for the review — addressed in c80ca64. All four points:

1. **Form locks before confirming PTY delivery** (github-actions; Codex Terminal.vue) — `Terminal.submitText` now returns whether the text was actually written to an OPEN socket, and the form only sets `submitted`/persists `viewState` when delivery succeeds. On a disconnected/reconnecting socket the submit is a no-op, so the form stays editable and shows *"Couldn't reach the terminal (is the session connected?). Please try again."*

2. **Delayed `\r` can hit the wrong session** (github-actions; Codex verdict) — the text + delayed CR are now both sent from inside `Terminal.submitText`, pinned to the socket captured at call time (`const sock = ws`). The CR's `setTimeout` checks `sock === ws && sock.readyState === OPEN` before firing, so a session switch/reconnect in the 60 ms window skips the CR instead of submitting a stray turn elsewhere.

3. **Stale form on cross-panel submit** (Codex P2, form View.vue) — added a watch on `selectedResult.viewState?.submitted` (the `uuid` watch can't fire when the same result is re-published). When another open panel submits the same form, this instance reseeds and locks. Guarded with `!submitted.value` so a non-submit update can't clobber in-progress typing.

Build / lint / typecheck all green. The stacked tools-pane branch (#9) has been rebased onto these fixes.

PR #9, opened 2026-06-16
feat: add a tools pane (available tools + full tool-call history)

snakajima:
Now targets `main` (PR #8 is merged). Three commits:

Mirrors the GUI toolResults to disk so rendered views survive a reboot — one JSON file per session under `<workspace>/.toolresults/<sessionId>.json`, via a reusable `createSessionStore(dirName)` helper (in-memory Map working copy; file rewritten on each change; lazy-loaded on first access with an in-flight load promise to dedupe concurrent loads). `storeToolResult` + the toolResult routes are now async.

### 2. `feat: add a tools pane (available tools + full tool-call history)`
A gear button in the GUI panel header toggles a third column (`ToolsPane.vue`, visibility persisted in `localStorage`):
- **Available Tools** — enabled GUI plugin tools + collapsible descriptions (`GET /api/tools`).
- **Tool Call History** — **every** tool call for the session (Bash, Read, other MCP, and GUI tools), with name, status/duration, time, collapsible args + result.

The history is **hook-driven, not broker-driven**: the spawn's `--settings` registers `PreToolUse`/`PostToolUse` hooks (matcher `""` = all tools) → `/api/hook` → per-session history keyed by `tool_use_id` → published on `toolcalls:<id>`, replayed from `GET /api/tool-calls/:id`. Same shape as MulmoClaude's `toolCallHistory`.

### 3. `feat: persist tool-call history to <workspace>/.toolcalls`
Persists the tool-call history to disk too, reusing the same `createSessionStore` helper, so it survives a reboot.

## Verification
- ✅ `npm run build`, `npm run lint`, `npm run typecheck` clean.
- ✅ Both stores verified to **persist across an actual process restart** (toolResults: count + a submitted form's `viewState`; tool-call history: completed Bash entry with output + duration), loading from disk on first access.
- ✅ Hook-driven history captures non-GUI calls (Bash) and GUI calls alike; config-gating still drops disabled plugins.

### DIFF
--- a/docs/gui-protocol-spike.md
+++ b/docs/gui-protocol-spike.md
@@ -285,6 +285,30 @@ interactive `claude`** is the remaining manual check (the unit under it —
   the spike now exercises the *actual* GUI→LLM mechanism the migration depends on,
   and removes ~70 lines of parked-response/timeout bookkeeping.
 
+### Tools pane (gear toggle)
+
+Mirrors MulmoClaude's right sidebar (`RightSidebar.vue`): a gear button in the GUI
+panel header toggles a third column (`src/components/ToolsPane.vue`, visibility
+persisted in `localStorage`) with two sections:
+
+- **Available Tools** — the enabled GUI plugin tools (name + collapsible
+  description) from `GET /api/tools` (registry `toolSummaries`). The full set claude
+  can call (built-ins, other MCP) isn't enumerable server-side, so it isn't listed
+  here — but it *does* appear in the history below.
+- **Tool Call History** — **every** tool call for the session (Bash, Read, other
+  MCP, and the GUI plugin tools), each row showing name, status/duration, time, and
+  collapsible arguments + result.
+
+**Key point — the history is hook-driven, not broker-driven.** The MCP broker only
+sees GUI-protocol tool calls, so it can't be the history source. Instead the spawn's
+`--settings` registers **`PreToolUse` + `PostToolUse` hooks (matcher `""` = all
+tools)** that `curl` each event to `/api/hook`; the server keys a per-session history
+by **`tool_use_id`** (PreToolUse → "running", PostToolUse → completes it with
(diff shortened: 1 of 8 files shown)


snakajima:
Thanks for the review — all five points addressed in 8285452.

1. **Failed tool executions never finalized** (Codex P2) — confirmed `PostToolUseFailure` is a distinct event and `PostToolUse` does *not* fire on failure. `hookSettingsJson()` now registers `PostToolUseFailure` too, and `/api/hook` routes both to `recordToolCallEnd` — `PostToolUse` → `completed`, `PostToolUseFailure` → `failed`. So a failed tool no longer stays stuck on `running…`; the pane shows a red **failed** badge and labels the output **error**.

2. **Hook payload size** (Codex P2) — raised the parser limit to `express.json({ limit: "25mb" })` so a large `tool_output` (verbose Bash/Read) isn't 413'd before the handler runs. Also cap the *stored* output at 20k chars (`capToolOutput`) so one chatty tool can't bloat the on-disk history/pane — the raw output still reaches the LLM via the terminal, this only trims the history copy.

3. **ToolsPane session-switch race** (Actions) — `loadHistory` now drops a stale response when `id !== props.sessionId` after the await (before assigning). Applied the same guard to `GuiPanel.loadHistory`, which had the identical latent bug.

4. **Gear toggle accessibility** (Actions) — added `aria-label="Toggle tools pane"` and `:aria-pressed="toolsOpen"` to the button.

5. **Coverage gap** — added `src/components/ToolsPane.spec.ts` (following `Sidebar.spec.ts`): renders available tools + running/completed/failed badges, upserts a running call to completed in place by `tool_use_id` via a pub/sub push, and verifies the session-switch race guard drops a stale response. `npm test` → 7 passing; build/lint/typecheck green.

One note on #2: I went with a generous global body limit + output cap rather than a per-route parser, since the toolResult route (large `presentDocument` markdown) benefits from the same headroom.

PR #11, opened 2026-06-16
feat: session-history UI — spinner, unread filter, horizontal layout, stable order

snakajima:
## Summary
Ports mulmoclaude's session-history presentation into mulmoterminal's session list (Vue 3, scoped-CSS dark theme — no Tailwind/material-icons/i18n).

- **Spinning "thinking" icon** for working sessions, replacing the static dot. Suppressed while a session is waiting for input and on the active session (its progress is already visible in the terminal), so only background work spins.
- **Bold "unread" rows** (mulmoterminal's `waiting` flag) plus an **All / Unread** filter via a new `FilterChip` pill.
- **Horizontal `SessionTabBar` layout** as an alternative to the vertical `Sidebar`, toggled from either layout and persisted to `localStorage`. Tabs flex to fit without scrolling, capped at the 8 most recent.
- **Stable list order**: the server sorts by mtime, so switching sessions (which bumps the previous session's file mtime) used to reshuffle rows under the user. Pub/sub refreshes now merge without moving existing rows; the ⟳ button performs an explicit recency re-sort and lives with the filter controls.

## Implementation notes
- Extracted the `/api/sessions` fetch + pub/sub subscription into a shared `useSessions` composable so both layouts share one source of truth.
- "Unread" maps onto the existing `waiting` flag — the only attention signal the backend exposes. Backend unchanged. mulmoclaude extras with no backend equivalent (bookmarks, AI summaries, origin badges) were intentionally not ported.

## Test plan
- `vue-tsc` clean, eslint clean, `vitest` 12/12 passing (spinner states, active/waiting suppression, unread filter, stable order, recency re-sort).


snakajima:
Addressed both review points in 9b0c028:

**Shared session store (Codex CHANGES REQUESTED):** Hoisted a single `useSessions()` instance plus the `filter` state up to `App.vue`; `Sidebar` and `SessionTabBar` are now presentational (props + `v-model:filter`). Toggling vertical/horizontal no longer unmounts the store, so it no longer refetches the list or resets the filter — one shared source of truth across both layouts.

**Horizontal pane height (Codex P2):** The tab bar takes 40px but the panes were still `height: 100vh`, overflowing under `.app { overflow: hidden }` (and letting xterm fit to off-screen rows). Bounded `.main` with `min-height: 0` and switched Terminal/GuiPanel/ToolsPane to `height: 100%` so they fill the leftover height exactly in both layouts.

Also moved the stable-order / recency-resort logic into the exported `mergeStable` with direct unit tests (`useSessions.spec.ts`), and rewrote `Sidebar.spec.ts` to drive the presentational component via props/emits. tsc + eslint + 16 tests green.

snakajima:
Addressed the accessibility feedback in e27b115:

**Splitter (mouse-only → keyboard + AT):** The divider now has separator semantics — `role="separator"`, `aria-orientation="vertical"`, `aria-label="Resize terminal"`, `aria-valuenow/valuemin/valuemax`, and `tabindex="0"` — plus keyboard handlers: ←/→ nudge by 16px, Home/End jump to the min/max, with a `:focus-visible` style. I also track the viewport so `aria-valuemax` stays accurate and the saved width re-clamps when the window shrinks.

**Icon-only controls (title → explicit name):** Added `aria-label` to each new icon-only button — `⟳` (sort), `⇥`/`⇤` (layout toggles), and `+` (new session in the tab bar). The text controls (`+ New session`, session tabs) already have accessible names.

Note: the shared-store point from earlier reviews was resolved in 9b0c028 (single `useSessions` + filter hoisted to `App.vue`); the later comment re-flagging it predated that commit. tsc + eslint + 16 tests + build all green.

PR #19, opened 2026-06-17
chore: stricter lint — typescript-eslint strict + eslint-plugin-sonarjs

isamu:
## Summary

Make lint stricter: add **`eslint-plugin-sonarjs`** (recommended, ~269 rules) and bump
**`typescript-eslint` `recommended` → `strict`**, then fix all **13** resulting
violations with real code changes (no `any`, no `as`).

The codebase was already clean — only 13 problems surfaced.

## Items to Confirm / Review

- **`no-non-null-assertion`** (PluginFrame.vue, Terminal.vue): `ref.value!` replaced
  with a `const x = ref.value; if (!x) return;` guard inside `onMounted`.
- **`@typescript-eslint/unified-signatures`** (Sidebar.vue, SessionTabBar.vue): the three
  payload-less emits merged into `(e: "new" | "toggle-layout" | "refresh"): void`.
- **`no-invalid-void-type`** (ToolsPane.spec.ts): `deferred<void>()` → `deferred<undefined>()`
  (`resolve(undefined)`).
- **`sonarjs/cognitive-complexity`** (server/index.ts, 3 functions): `readSessionMeta`, the
  `/api/hook` handler, and the WebSocket connection handler were split into small named
  helpers (`userPromptText`, `parseJsonl`, `handleActivityHook`, `handleToolHook`,
  `reattachPty`, `spawnClaudePty`, `handleClientFrame`, `handleClientClose`).
  **Behaviour is unchanged** — please sanity-check the connection-handler split (the
  spawn/reattach/message/close logic), since it's the core PTY path and has no unit tests.

## User Prompt

- そして lint rule も厳し目に。
- eslint-plugin-sonarjs（を使って）。

## Implementation

See `plans/chore-stricter-lint.md`.

## Verification

- `yarn lint` ✅ · `yarn typecheck` ✅ · `yarn typecheck:server` ✅ · `yarn test` ✅ (16/16)
  · `yarn build` ✅ · booted via `node --import tsx … server/index.ts` ✅.

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -1,12 +1,14 @@
 import js from "@eslint/js";
 import tseslint from "typescript-eslint";
 import pluginVue from "eslint-plugin-vue";
+import sonarjs from "eslint-plugin-sonarjs";
 
 export default [
   { ignores: ["dist/", "node_modules/"] },
   js.configs.recommended,
-  ...tseslint.configs.recommended,
+  ...tseslint.configs.strict,
   ...pluginVue.configs["flat/recommended"],
+  sonarjs.configs.recommended,
   {
     files: ["**/*.vue"],
     languageOptions: {
--- a/package.json
+++ b/package.json
@@ -51,6 +51,7 @@
     "@vue/tsconfig": "^0.9.1",
     "concurrently": "^10.0.3",
     "eslint": "^10.5.0",
+    "eslint-plugin-sonarjs": "^4.0.3",
     "eslint-plugin-vue": "^10.9.2",
     "jsdom": "^29.1.1",
     "tailwindcss": "^4.3.1",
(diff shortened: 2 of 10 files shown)


isamu:
Addressed the Codex review:

- **`userPromptText` nullish regression (server/index.ts:420)** — Fixed in b464f41.
  Non-object array blocks now map through `String(x ?? "")`, so `null`/`undefined`
  content blocks join to `""` (matching the pre-refactor behavior) instead of leaking
  `"null"`/`"undefined"` into sidebar titles. Good catch — thanks.

Not changed:
- **Socket: `eslint-plugin-sonarjs@4.0.3` license score (70)** — it's LGPL-3.0, a
  dev-only lint tool that isn't bundled/shipped, so no runtime/license exposure.
  Leaving as-is.

`yarn lint` / `yarn typecheck` / `yarn typecheck:server` / `yarn test` (16/16) /
`yarn build` all green after the fix.

PR #20, opened 2026-06-17
feat: copy the whole tool-call history from the Tools pane

snakajima:
Adds a **Copy all** button to the **Tool Call History** header in the Tools pane — mirrors MulmoClaude's RightSidebar copy-history button. When a run goes sideways (the LLM keeps retrying a tool, an unexpected error, …), you can grab the entire history — every call's arguments + results — as pretty JSON in one click to paste into a bug report or share.

- Copies `JSON.stringify(toolCalls, null, 2)` via the clipboard.
- Disabled when there are no calls; shows `✓ Copied` for 2s.
- Styled for the pane's dark theme; routed through `window.*` (navigator / set/clearTimeout) to satisfy the eslint globals allowlist.

## Verification
`vue-tsc`, `vite build`, eslint, ToolsPane spec (16/16).

### DIFF
--- a/src/components/ToolsPane.vue
+++ b/src/components/ToolsPane.vue
@@ -120,6 +120,26 @@ function formatValue(v: unknown): string {
     return String(v);
   }
 }
+
+// Copy the WHOLE tool-call history (arguments + results) as pretty JSON — handy
+// to paste into a bug report / share when a run goes sideways. Mirrors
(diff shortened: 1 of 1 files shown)


snakajima:
The Copy-all button is already in main via #23; closing this standalone PR.

PR #30, opened 2026-06-17
feat: run claude sessions in an optional Docker sandbox

snakajima:
## What

When Docker is reachable (and `DISABLE_SANDBOX` is unset), every `claude` session now runs **inside the `mulmoterminal-sandbox` container** instead of directly on the host. **Auto-detected at startup — no flag.** Falls back to the host path when Docker is absent or the image build fails.

Builds on #29 (MCP-on-host): because the GUI MCP server and activity hooks live on the host, the sandboxed agent reaches them via an outbound HTTP/curl call to `host.docker.internal`, so we don't ship server code into the container — it only needs `claude` + the CLI toolset.

## How

- **`Dockerfile.sandbox` + `sandbox-entrypoint.sh`** — image with the same toolset as mulmoclaude (git, ripgrep, gh, python+pandas/numpy/matplotlib, pandoc, ffmpeg, poppler, imagemagick, graphviz, libreoffice). Entrypoint ported verbatim; in the default `--user` path it just execs `claude` (the SSH-forwarding root-setup branch stays dormant until wired up).
- **`server/system/docker.ts`** — `isDockerAvailable()` liveness probe + `ensureSandboxImage()` (sha256-of-Dockerfile stored as an image label → auto-rebuild on change).
- **`server/system/sandbox.ts`** — builds the `docker run` argv. node-pty drives the container TTY via `-i -t` (resize propagates via SIGWINCH). **Workspace is mounted at its identical host path + `-w`** so claude's project-session encoding matches the host's — the **sidebar and resume keep working** with no path translation. Mounts `~/.claude(.json)`; forwards an allowlist of credential env vars.
- **`server/index.ts`** — branch `spawnClaudePty` on the sandbox flag; route MCP + hooks to `host.docker.internal` when sandboxed; **guard `/mcp/:sessionId` with a per-session bearer token** (the endpoint is now container-reachable); run the sandboxed agent with `--permission-mode bypassPermissions` so **all tools (built-in AND MCP) are available without prompts** (Docker is the safety boundary).
- **`.dockerignore`** keeps the build context to just the entrypoint.

## Key decisions

- **Image toolset = mulmoclaude parity** (per request). Dropped only `tsx` — that was for mulmoclaude's in-container broker; ours is on the host.
- **Identical-path workspace mount** so session storage/sidebar/resume are unaffected.
- **`bypassPermissions` in the sandbox** to satisfy "all MCP tools incl. built-in must be available" — safe because the container is the boundary. `--strict-mcp-config` stays so we don't try to spawn the user's own stdio MCP servers (absent from the image).

## Verification

- ✅ Image builds (~720MB); `claude 2.1.179` runs in-container as host uid `501` with `~/.claude` mounted; toolset present (git/python+pandas/pandoc/gh/rg).
- ✅ Server boots in sandbox mode: detect → sha-mismatch rebuild (cached) → "sessions run sandboxed".
- ✅ Host path still works under `DISABLE_SANDBOX=1`.
- ✅ `/mcp/:sessionId` returns 401 without / with a wrong token.
- ✅ Arg builder unit-tested (6 cases); full suite 22/22; typecheck + lint clean.
- ⚠️ **Not yet driven end-to-end in a live browser session** (needs node-pty + a browser) — every component is verified independently, but a manual interactive smoke test is recommended before merge.

## Follow-ups (deferred)

- SSH-agent / `gh` credential forwarding (mulmoclaude's opt-in entrypoint machinery — the entrypoint already supports it; just needs the spawn-side wiring).
- Optionally authenticate `/api/hook` with the same per-session token.


snakajima:
It is quite hard to do it right. Because we can not use "natively installed" Claude Code, we need to use rpm version, which causes various warnings.

PR #32, opened 2026-06-17
fix: handle port-in-use gracefully + random free-port fallback (#31)

isamu:
## Summary

`npx mulmoterminal` crashed with an unhandled `EADDRINUSE` when the port was busy
(and printed a misleading "ready" banner just before). Fixes the crash and makes
port selection collision-proof.

Two root causes:
- The launcher probed `127.0.0.1` (IPv4) while the server binds `::` (dual-stack), so a
  port held only on `::` looked free — the server then crashed binding it.
- The server had no `EADDRINUSE` handler, so the failure surfaced as an unhandled
  `'error'` event / stack trace.

## Changes

- **`server/index.ts`**: `server.on("error")` prints a clear message and `exit(1)` on
  `EADDRINUSE` (and other bind errors) instead of crashing. Reuses `hasErrnoCode` / `messageOf`.
- **`bin/mulmoterminal.js`**:
  - The free-port probe now binds without a host (`probe.listen(port)`), matching the
    server's `::` bind so probe and real bind agree.
  - Fallback when the preferred port (3456) is busy is now an **OS-assigned ephemeral
    (random) free port** (`listen(0)`) instead of a sequential `+1` scan — no clashes
    across concurrent instances. Explicit `--port` still errors cleanly when taken.

## Verification

- With 3456 occupied → launcher prints `Port 3456 busy → using <random> instead` and
  boots; **no crash**.
- `PORT=3456 node --import tsx server/index.ts` while busy → `Port 3456 is already in use …`
  and exits 1, **no stack trace**.
- 3456 free → uses 3456 as before. `--port <busy>` → clean error.
- `yarn lint` / `typecheck` / `typecheck:server` / `test` (16/16) / `build` / `format:check` ✅.

## User Prompt

- `npx mulmoterminal` がポート衝突で `EADDRINUSE` クラッシュする。修正して。
- ポートもランダムな方が良いかも。

## Note

Publishing is a separate step — this would ship as `0.1.1` via `/publish`.

See `plans/fix-port-in-use.md`.

### DIFF
--- a/bin/mulmoterminal.js
+++ b/bin/mulmoterminal.js
@@ -18,6 +18,10 @@ const PKG_DIR = join(__dirname, "..");
 const SERVER_ENTRY = join(PKG_DIR, "server", "index.ts");
 const DEFAULT_PORT = 3456;
 const READY_TIMEOUT_MS = 15_000;
+const MAX_BIND_RETRIES = 5;
+// Server exit code meaning "port taken at bind time" — keep in sync with
+// server/index.ts (PORT_IN_USE_EXIT_CODE).
+const PORT_IN_USE_EXIT_CODE = 75;
 
 // Single source of truth: read the version from the shipped package.json so
 // `--version` never drifts from the published version.
@@ -58,8 +62,9 @@ function isPortFree(port) {
 }
 
 // Ask the OS for a free port (listen on 0) and return the one it assigned, or
-// null. A collision-proof, effectively-random fallback when the preferred port
-// is taken — no two instances clash.
+// null. An effectively-random fallback when the preferred port is taken; the
+// bind-retry in main() closes the small probe-to-bind race so concurrent starts
+// don't clash.
 function findEphemeralPort() {
   return new Promise((resolve) => {
     const probe = createServer();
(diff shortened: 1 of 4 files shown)


isamu:
Addressed the Codex review (d074591) — the probe-to-bind TOCTOU window is now closed properly instead of overclaiming "collision-proof":

- **server/index.ts**: on `EADDRINUSE` the server exits with a sentinel code **75** (other bind errors → 1).
- **bin/mulmoterminal.js**: `runServer()` detects an early exit with code 75 (port taken before the server became ready) and `main()` **retries on a fresh OS-assigned port** (up to `MAX_BIND_RETRIES`). So even if another process grabs the probed port in the race window, the launcher just moves to the next free port instead of failing. An explicit `--port` is still honored (clean error, no silent reassignment). Signal handlers are registered once and target the live child across retries. The `findEphemeralPort` comment no longer claims collision-proofness on its own.

Verified:
- 3456 busy → launcher boots on a random free port (no crash).
- `PORT=3456 node --import tsx server/index.ts` while busy → clean message + exit code **75**.
- `yarn lint` / `typecheck` / `typecheck:server` / `test` (16/16) / `build` / `format:check` ✅.

isamu:
Addressed the follow-up Codex review (a2ce7b5):

- **`ready` was unreliable** — it came from polling `http://127.0.0.1:${port}/`, so another process winning the bind race could answer the probe, flip `ready` true, and the `!ready && code === 75` guard would skip the retry. Fixed by taking Codex's suggested route: **treat exit code 75 as retriable regardless of probe readiness**. A `75` exit means the child hit `EADDRINUSE` and never served, so retrying on a fresh port is always correct.
- **`waitUntilReady` is now cancelable** — when a child exits, the launcher cancels its poll, so an abandoned/raced attempt can't fire a stale "ready" banner or open the browser to a port we're no longer using.

Verified: 3456 busy → fallback to a random free port; direct `PORT=3456` server → clean message + exit 75; `lint` / `typecheck` / `typecheck:server` / `test` (16/16) / `build` / `format:check` ✅.

PR #34, opened 2026-06-17
fix: node-pty spawn-helper perms on hoisted installs (posix_spawnp crash)

isamu:
## Summary

Published `npx mulmoterminal` (0.1.1) crashed with `Error: posix_spawnp failed.` on the
first terminal connection (WebSocket → `pty.spawn(claude)`). Root cause + fix:

- `server/fix-pty-perms.js` (postinstall, makes node-pty's `spawn-helper` executable)
  looked for node-pty at `<pkg>/node_modules/node-pty` — only true in this repo. Consumer
  installs (`npx`/`npm i`) **hoist node-pty to the top-level `node_modules`**, so the
  postinstall missed it and the helper stayed `644` → `posix_spawnp failed`. Reproduced
  locally by installing the tarball into a hoisted layout.
- Now resolves node-pty via `createRequire().resolve("node-pty/package.json")` (works
  nested or hoisted) and chmods every `prebuilds/*/spawn-helper` to `755`.
- Hardening: the WebSocket spawn and `spawnBackgroundChat` spawn are wrapped so a failed
  spawn closes just that connection (error frame) instead of crashing the whole server.

## Why CI didn't catch it

`package-smoke` only checks `GET /` (HTTP 200) — it never opens a WebSocket, so the PTY
path was untested. (Follow-up idea: add a WS+PTY check to the smoke job.)

## Verification (local, hoisted install — as requested)

- After fix: `spawn-helper` is `755`; node-pty resolves to `lib/index.js` under tsx.
- Isolated `pty.spawn('claude','--version')` → exit 0.
- **Real WebSocket to `/ws` → `[pty] spawned claude (pid=…)` + output streams (PTY-OK)**, no `posix_spawnp`.
- Forced spawn failure (`CLAUDE_BIN=/nonexistent`) → connection closed, **server still serves `GET / → 200`**.
- `yarn lint` / `typecheck` / `typecheck:server` / `test` (16/16) / `build` / `format:check` ✅.

## User Prompt

- `npx mulmoterminal` 起動後、ターミナル接続で `posix_spawnp failed` クラッシュ。直して。
- ちゃんと local でテストして。

Bumps version to 0.1.2. See `plans/fix-pty-helper-perms.md`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,19 +50,22 @@ jobs:
 
       - run: yarn install --frozen-lockfile
 
-      # The launcher's pre-flight requires `claude --version` to succeed. The
-      # smoke test only needs the server to boot and serve "/", so a tiny stub
-      # on PATH passes the check without the real CLI + auth.
+      # The launcher's pre-flight requires `claude --version` to succeed, and the
+      # PTY check spawns `claude` — so the stub answers --version AND, for any
+      # other invocation, prints a line and exits cleanly (a real PTY lifecycle)
+      # without needing the real CLI + auth.
       - name: Stub Claude Code CLI
         run: |
           mkdir -p "$HOME/.local/bin"
-          printf '#!/usr/bin/env bash\n[ "$1" = "--version" ] && { echo "claude 0.0.0 (smoke-stub)"; exit 0; }\nexit 1\n' > "$HOME/.local/bin/claude"
+          printf '#!/usr/bin/env bash\nif [ "$1" = "--version" ]; then echo "claude 0.0.0 (smoke-stub)"; exit 0; fi\necho "smoke-stub claude started"\nexit 0\n' > "$HOME/.local/bin/claude"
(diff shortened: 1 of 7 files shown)


isamu:
Added the regression coverage (cd444ac). The `package-smoke` job now exercises the PTY path on a hoisted install:

1. **Helper-exec guard** — fails if `node_modules/node-pty/prebuilds/linux-x64/spawn-helper` isn't executable after install (the exact #33 regression).
2. **Live `/ws` PTY check** (`scripts/ci-ws-smoke.mjs`) — connects to `/ws` and requires a real PTY lifecycle (`session` + `output`/`exit`), failing if it gets an `error` frame (spawn failed) or times out, then confirms the server is still serving.

The stub `claude` now prints a line and exits 0 so the PTY produces a frame.

Verified locally that the WS check **fails (exit 1)** when the spawn-helper is `chmod 644` (the regression), and passes with it `755` — so a future perms regression would turn CI red.

PR #50, opened 2026-06-19
feat: integrate @mulmoclaude/collection-plugin (read-side foundation)

snakajima:
## What

First increment of the collection-plugin integration: **`presentCollection` chat cards render real collections from the shared workspace** — table/kanban/custom (sandboxed-iframe) views, with images. Read-side foundation toward the collections toolbar (Tier 2).

**Needs zero MulmoClaude changes** — everything required ships in `@mulmoclaude/collection-plugin@0.5.1`.

## Changes

**Server**
- `server/backends/collections.ts` — `configureCollectionHost` over the shared workspace (`CLAUDE_CWD`) with MulmoClaude's exact path layout; read routes `GET /api/collections/list` and `/:slug/detail`; and the read-only custom-view surface: `view-file` (HTML), `view-token` (scoped token), `view-data` (enriched `{ items }`, token-guarded + CORS for the opaque-origin iframe).
- `server/backends/viewToken.ts` — HMAC scoped-token mint/verify/require/clamp (port of MulmoClaude's), keyed by a per-process secret.
- `server/backends/files.ts` — `GET /api/files/raw?path=` (workspace-file serving for image/file fields + custom-view `` URLs): path containment, MIME, size caps, Range, and MulmoClaude's `sandbox`+`nosniff` security headers.
- `server/plugins-registry.ts` — `loadPackage` resolves a sole `execute*` export so the collection core loads like the others.

**Frontend**
- `src/composables/collectionUi.ts` — the `CollectionUi` binding: real `fetchCollectionDetail`/`listCollections`/`confirm`/`localeTag`, read-only custom views (`mintViewToken`/`fetchViewHtml`/`buildViewSrcdoc`), and `imageSrc`/`fileAssetUrl` → the raw-file route; write/feeds/favorites/chat stubbed; teleport-target stack for the Shadow-DOM modal.
- `src/utils/customViewSrcdoc.ts` — sandboxed `srcdoc` builder + CSP policy.
- `src/components/CollectionCardView.vue` — wraps the package View, registers its shadow root as the modal teleport target, and fills the fixed-height frame.
- `PluginFrame`/`GuiPanel`/registry — a fixed frame `height` (`100vh` for the collection card) threaded through the Shadow DOM so the View's `h-full` layout resolves (matches MulmoClaude's StackView).
- Registered the ToolPlugin in `plugins-registry.ts` + `plugins.json`; binding configured in `main.ts`.
- **Icons** — shadow-scoped rule mapping `.material-icons`/`.material-symbols-outlined` to the loaded "Material Symbols Outlined" font (the plugin's glyphs rendered as text otherwise).

**Dev** — `vite.config.ts`: `server.cors: false` + `changeOrigin` so the sandboxed iframe's `view-data` preflight reaches Express (matches MulmoClaude).

**Deps** — `@mulmoclaude/collection-plugin@^0.5.1`, `vue-i18n@^11`.

## Verification

- client + server typecheck, `yarn build`, `yarn lint`, 29 tests — all pass
- live: collections list/detail, custom-view round-trip (`view-token`→`view-file`→`view-data`), and `/api/files/raw` (200 + full bytes, traversal → 403, missing → 404); preflight through the Vite proxy → 204 with ACAO; icon + height fixes verified via Puppeteer

## Not in this PR (next)

Tier 1 (write routes / interactive editing) and **Tier 2 — the collections toolbar**: shared favorites (`/api/shortcuts` over `config/shortcuts.json`), a toolbar launcher, and a state-based browse panel. See `docs/collection-plugin-integration.md`.

### DIFF
--- a/docs/collection-plugin-integration.md
+++ b/docs/collection-plugin-integration.md
@@ -1,0 +1,230 @@
+# Integrating `@mulmoclaude/collection-plugin` into MulmoTerminal
+
+Status: **the cross-repo blocker is gone.** The package shipped `0.5.0`/`0.5.1` with exactly the
+host-contract changes this doc originally said were still needed (ToolPlugin entry, router-optional
+nav, configurable teleport target, server engine). What remains is **MulmoTerminal-side wiring only** —
+**no MulmoClaude changes are required.**
+
+This doc was originally written against `0.4.1` and assumed a future `0.5.0` was on the critical path.
+It has been updated against the **published `@mulmoclaude/collection-plugin@0.5.1`** and the current
+MulmoTerminal tree.
+
+---
+
+## TL;DR verdict
+
+**Can this be done without any changes to MulmoClaude? Yes.** Everything MulmoTerminal needs is already
+published in `@mulmoclaude/collection-plugin@0.5.1`:
+
+- `./vue` exports a `plugin: ToolPlugin` (chat-result `View` + `Preview`) — the runtime registration
+  shape MulmoTerminal uses, just like chart/form/markdown.
+- Navigation is **router-optional**: `CollectionEmbedView` no longer renders a bare `<router-link>`;
+  refs/embeds go through `collectionUi().navigateToRecord(slug, recordId?)` (`src/vue/refLink.ts`), and
+  `recordHref`/`navigate` are optional so a router-less host returns `undefined`.
+- The record modal teleports to a host-supplied target: `modalTeleportTarget?: () => string | HTMLElement`
+  (defaults to `"body"`), so a Shadow-DOM host points it at an in-shadow node.
+- `./server` exports the full read-side engine: `configureCollectionHost({ workspaceRoot, log, paths,
+  isPresetSlug })` plus `discoverCollections` / `loadCollection` / `toSummary` / `toDetail` /
+  `listItems` / `validateCollectionRecords`.
+
+The **only** MulmoClaude artifact that is NOT packaged is the favorites/shortcuts format
+(`Shortcut` type + `shortcuts-io`), which still lives in MulmoClaude's app tree. MulmoTerminal must
+**reimplement that format verbatim** for the Tier 2 toolbar (see the drift warning). That is a
+MulmoTerminal-side change — it requires no MulmoClaude edit. Extracting a shared shortcuts package
+later (to eliminate drift risk) is optional and deferrable.
+
+---
+
+## Correction to the original plan: even the read-only card needs the server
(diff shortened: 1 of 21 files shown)


snakajima:
Thanks for the review — addressed in `8870900`:

**#4 (P1) — view-file served same-origin HTML without sandbox.** Fixed. `GET /api/collections/:slug/view-file` now sets `Content-Security-Policy: sandbox` + `X-Content-Type-Options: nosniff`, so direct navigation runs the LLM-authored HTML in an opaque origin (no access to `/api/*`). The iframe render path is unaffected — it `fetch()`es the body and injects it as `srcdoc`, and a response-level CSP doesn't block reading bytes. (This is actually stricter than MulmoClaude, which relies on its global bearer there; MulmoTerminal has no bearer.)

**#2 — no route-level tests.** Added coverage, 29 → 64 tests:
- `viewToken.spec` — mint/verify round-trip, tamper / bad-sig / expiry rejection, `clampCapabilities`, `requireViewToken` (401 paths, slug & capability mismatch).
- `files.spec` — `/api/files/raw` 200 + hardening headers, 400 (no path), 403 (traversal / absolute escape), 404, 206 (Range).
- `collections.spec` — `/list`, `/detail` (+404), `view-token` (read-only clamp, 400 missing viewId), `view-file` (404 + sandbox/nosniff), `view-data` (401 without token, 200 records with token) — against a tmpdir fixture workspace.
- `customViewSrcdoc.spec` — connect-src locked to origin, `__MC_VIEW` injection, `<`-escaping, head-inject vs fragment-wrap.

**#1 — `imageSrc` returns `""`.** Already resolved earlier in this PR (commit `ce54880`); the bot reviewed a stale commit. `imageSrc`/`fileAssetUrl` now resolve to `/api/files/raw?path=…` (data: URIs pass through), matching MulmoClaude's `resolveImageSrc`.

**#3 — broad `img-src`/`media-src https:` in the custom-view CSP.** Intentional. This is a verbatim port of MulmoClaude's `buildCustomViewCsp` and its documented threat model: `connect-src` stays locked to the server origin (the channel that matters for bulk/bidirectional exfil), while `img-src`/`media-src https:` is a known one-way, GET-only, response-unreadable channel accepted so feed/collection records can render external thumbnails and media. The views are authored by the user's own agent over the user's own data. A test now pins `connect-src` closed to `https:`. Kept for parity per the "match MulmoClaude" direction — happy to tighten beyond MulmoClaude (e.g. proxy images through the origin) if preferred.

PR #51, opened 2026-06-19
feat: Tier 2 collections toolbar (favorites + full-screen browse)

snakajima:
## What

The actual goal of the collection-plugin work: a **collections toolbar** like MulmoClaude's, adapted to MulmoTerminal's no-router, state-driven host. Builds on the merged read-side foundation (#50).

Two design decisions (per maintainer): **favorites = reimplement verbatim + fixture test**; **browse panel = full-screen overlay**.

**Favorites are SHARED with MulmoClaude via the workspace** — a collection pinned in either app shows up in the other (same `<workspace>/config/shortcuts.json`, same format). Still zero MulmoClaude changes.

## Changes

**Server**
- `server/backends/shortcuts.ts` — `GET`/`PUT /api/shortcuts` over the shared `config/shortcuts.json`, in MulmoClaude's exact **object-wrapper** format `{ shortcuts: Shortcut[] }`, normalized + atomically written. `shortcuts.spec.ts` pins the on-disk contract.

**Frontend**
- `src/types/shortcuts.ts` — the `Shortcut` type (verbatim from MulmoClaude).
- `src/composables/useShortcuts.ts` — singleton favorites store over `/api/shortcuts`: optimistic mutation + rollback, serialized replace-all PUTs (ported).
- `src/components/PinToggle.vue` — the ★ toggle (inline-styled, since it renders inside the plugin shadow root); wired into the binding's `pinToggle`.
- `src/composables/useCollectionBrowse.ts` — reactive view-state (`closed | index | detail`). The binding's nav fns (`gotoIndex`/`gotoDetail`/`navigateToRecord`/`routeSlug`/`routeSelectedId`/`isFeedRoute`/`setSelectedId`) map onto it; `unpin`/`reconcileShortcuts` onto `useShortcuts`.
- `src/components/CollectionsBrowseOverlay.vue` — full-screen overlay rendering `CollectionsIndexView` (index) or standalone `CollectionView` (detail) in a `PluginFrame` shadow root; registers its shadow root as the record-modal teleport target.
- `App.vue` — the toolbar launcher (a "Collections" button + one per favorite) + the overlay mount.
- `src/collectionShadowCss.ts` — shared shadow CSS (icon rule + `style.css`), now used by both the chat card and the overlay.

## Verification

- client + server typecheck, `yarn build`, `yarn lint`, **71 tests** (7 new shortcuts tests)
- Puppeteer smoke: MulmoClaude's pinned collections appear in MulmoTerminal's toolbar from the shared file; the "Collections" button opens the overlay; the index renders (27 cards).

## Known limits / next

- `startChat` is still a no-op — collection "create"/action buttons that seed a chat are inert (needs a hook into the terminal session).
- A dedicated `/feeds` index (`FeedsView`) is deferred; the launcher/overlay handle `kind: "feed"` shortcuts but the collections index filters feeds out.
- Tier 1 (write routes / interactive editing) still pending.

See `docs/collection-plugin-integration.md` (resolved decisions + Tier 2 status).

### DIFF
--- a/docs/collection-plugin-integration.md
+++ b/docs/collection-plugin-integration.md
@@ -215,15 +215,27 @@ MulmoTerminal equivalent needs three pieces, adapted to a no-router, state-drive
 
 ---
 
-## Open questions for the maintainers
+## Resolved decisions
 
-- **Shortcuts sharing**: shared package (recommended; the only optional MulmoClaude change) vs
-  reimplement-verbatim + format test?
-- **Browse panel placement**: replace the GuiPanel content, a new tab/pane, or overlay?
-- **CSS isolation**: keep the shadow frame (and the teleport-target wrapper) or treat the collection
-  plugin as first-party in light DOM?
-- **Raw-file route**: add one so `image`/`file` fields render, or accept the v1 gap?
-- **`/feeds`**: wanted, or collections-only?
+- **Shortcuts sharing** → **reimplement verbatim + fixture test** (`server/backends/shortcuts.ts` +
+  `src/types/shortcuts.ts`, `shortcuts.spec.ts` pins the `{ shortcuts: [...] }` wrapper). Keeps
+  MulmoTerminal MulmoClaude-change-free; the shared file is the contract.
+- **Browse panel placement** → **full-screen overlay** (`CollectionsBrowseOverlay.vue`), driven by
+  `useCollectionBrowse` view-state; the binding's nav fns map onto it.
+- **CSS isolation** → keep the shadow frame (the overlay reuses `PluginFrame` + `collectionShadowCss`,
+  and registers its shadow root as the modal teleport target like the card).
+- **Raw-file route** → added (`/api/files/raw`), so `image`/`file` fields render.
+- **`/feeds`** → the launcher/overlay support `kind: "feed"` shortcuts, but the collections index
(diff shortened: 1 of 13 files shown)


snakajima:
Thanks — addressed in `b65c037`:

**`shortcuts.ts` fixed temp path race** → fixed. `writeShortcuts` now uses a unique temp file per write (`${file}.${randomUUID()}.tmp`) and removes the temp on failure, so concurrent `PUT /api/shortcuts` can't collide on the temp path or ENOENT on `rename`. Last writer wins (fine for a replace-all endpoint; the client also serializes its own PUTs via the mutation chain). Added a regression test that fires 8 concurrent PUTs and asserts all 200 + the on-disk file stays a valid wrapper.

**Overlay `aria-modal` without focus trap** → already resolved in `e3519bd` (after the reviewed commit `1517643`). Per follow-up UX feedback the overlay is no longer a modal dialog — it's a `role="region"` panel that fills the page *below* the toolbar, and the toolbar (Chat / Collections / favorites tabs) is intentionally reachable while it's open. So there's nothing to trap focus into; `aria-modal` is gone. Escape-to-close is a window-level listener, so it doesn't depend on dialog focus.

typecheck (client+server), build, lint, 72 tests all pass.

PR #54, opened 2026-06-19
feat: collection action routes — action buttons spawn a chat

snakajima:
## What

Wires the per-record and collection-level **`kind: "chat"` action routes** — the consumers of the `startChat` hook in #53. Schema-defined **action buttons** (e.g. Repair / enrich a record) now fetch a seed prompt and spawn a visible chat.

> **Stacked on #53** (`startChat`). Base is `feat/collection-startchat`; GitHub will retarget to `main` once #53 merges. Review #53 first.

## Changes

**Server** (`server/backends/collections.ts`)
- `POST /api/collections/:slug/items/:itemId/actions/:actionId` — per-record action: load the record, enforce the action's `when` predicate (the authorization rule), read the `templates/` file, return `{ prompt: buildActionSeedPrompt(record, template), role }`.
- `POST /api/collections/:slug/actions/:actionId` — collection action: read the template + all records, return `{ prompt: buildCollectionActionSeedPrompt(items, schema, template), role }`.

Uses the package's `readItem`/`readSkillTemplate`/`buildActionSeedPrompt`/`buildCollectionActionSeedPrompt`/`actionVisible`; maps 404 (collection/action/item) and 409 (out-of-state action).

**Frontend** (`collectionUi` binding) — `runItemAction`/`runCollectionAction` now POST those routes; `CollectionView` feeds the `{ prompt, role }` into `startChat`, which spawns a visible chat.

## Behavior

Records are edited by the spawned agent session **directly** (file edits) — the intended model, per the user. (An already-open collection view won't live-update after a background edit; that's a separate follow-up.)

## Verification

- 4 new action-route tests (per-record + collection seed prompts, 404 unknown action, 404 missing item) against the fixture + real engine. **84 tests** total.
- client + server typecheck, `yarn build`, `yarn lint` pass.

## Still stubbed (follow-ups)

Feeds (refresh + a feeds index), collection/view deletion, and collection live-refresh on file change.

### DIFF
--- a/server/backends/collections.spec.ts
+++ b/server/backends/collections.spec.ts
@@ -23,6 +23,8 @@ const SCHEMA = {
     name: { type: "string", label: "Name" },
   },
   views: [{ id: "v1", file: "views/v1.html", label: "Custom", capabilities: ["read"] }],
+  actions: [{ id: "enrich", label: "Enrich", kind: "chat", role: "general", template: "templates/enrich.md" }],
+  collectionActions: [{ id: "audit", label: "Audit", kind: "chat", role: "general", template: "templates/audit.md" }],
 };
 
 let server: Server;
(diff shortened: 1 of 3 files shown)


snakajima:
Re: the Codex P2 (route embedded action prompts through `startChat`) — **intended, not changing.**

This is the collection plugin's own two-context design: `CollectionView` does `if (props.sendTextMessage) props.sendTextMessage(prompt) else startChat(prompt, role)`.

- **In a `presentCollection` chat card** the collection is embedded in a live session, so `GuiPanel` passes `sendTextMessage` and an action **continues that conversation** — same as MulmoClaude.
- **In the standalone browse overlay / index** there's no active chat, so the action **spawns a visible chat** via `startChat`.

#54 added the missing overlay path; the card path already worked. Making embedded card actions also spawn a fresh chat would diverge from MulmoClaude and change the card UX (every action button would open a new tab instead of continuing the current chat), which we don't want. So both paths are working as designed.

PR #57, opened 2026-06-20
feat: integrate @mulmoclaude/html-plugin (presentHtml)

snakajima:
## What

Adds **`presentHtml`** to MulmoTerminal by consuming the just-published **`@mulmoclaude/html-plugin@0.2.2`** (Option B from `mulmoclaude/plans/feat-presenthtml-mulmoterminal.md`). This closes the gap where HTML-page skills (e.g. `lessons-biology`'s `learn.md`) made the agent hunt for a non-existent `presentHtml` and fall back to `presentDocument` (markdown), losing the styled page.

## How

**Server** (`server/backends/html.ts`)
- `mountHtmlDispatchRoute` — intercepts the View's source-editor dispatch (`loadHtml`/`saveHtml`) on `/api/plugin/presentHtml` **before** the generic plugin catch-all, routing to the package's `executeHtmlDispatch` over the artifacts `FileOps`. A tool-call (no `kind`) falls through to the package's `executeHtml`. Publishes `plugin:html:file:<path>` after a save for live-refresh.
- `mountHtmlPreviewRoute` — serves `GET /artifacts/html/<rest>` (the View's iframe `src`, `htmlArtifactPreviewUrl`) from the workspace with an **HTML preview CSP** (inline scripts + curated CDN allowlist, `connect-src 'none'`), `nosniff`, and path containment.
- The `presentHtml` tool-call itself is handled by the generic package loader (`plugins.json` `packages` → artifacts `FileOps` context) — no bespoke server execute.

**Frontend**
- Register the html `ToolPlugin` in `plugins-registry.ts`, wrapped with `wrapWithPluginRuntime("html", …)` (so the View's `dispatch`/`pubsub` work) + `style.css?inline` + a fixed `80vh` frame (the View's iframe is `h-full`).
- `plugins.json` `packages` += `@mulmoclaude/html-plugin` (drives both server load + frontend).
- `vite.config.ts` — proxy `/artifacts` to the backend in dev (the iframe src).

## Verification

- 7 new route tests (loadHtml / saveHtml / containment / tool-call fall-through; preview serving CSP / 403 / 404). **94 tests** total.
- client + server typecheck, `yarn build`, `yarn lint` pass.
- Live smoke: tool-call save → `artifacts/html/…html`; the page serves at `/artifacts/html/…` with the preview CSP + content; `loadHtml`/`saveHtml` dispatch round-trips; `../../etc/passwd` blocked.

## Notes

- Live-refresh fires on the View's own `saveHtml`. A background agent editing the file *directly* (Write tool) won't publish the change (same broad live-refresh gap as collections) — a follow-up could forward workspace file-change events onto `plugin:html:file:<path>`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,6 +54,7 @@
     "@mulmoclaude/chart-plugin": "^0.1.1",
     "@mulmoclaude/collection-plugin": "^0.5.1",
     "@mulmoclaude/form-plugin": "^0.1.1",
+    "@mulmoclaude/html-plugin": "^0.2.2",
     "@mulmoclaude/markdown-plugin": "^0.1.4",
     "@mulmoclaude/x-plugin": "^0.1.1",
     "@xterm/addon-attach": "^0.12.0",
--- a/plugins/plugins.json
+++ b/plugins/plugins.json
@@ -1,5 +1,5 @@
 {
-  "packages": ["@mulmoclaude/markdown-plugin", "@mulmoclaude/form-plugin", "@mulmochat-plugin/generate-image", "@mulmoclaude/chart-plugin", "@mulmoclaude/collection-plugin"],
+  "packages": ["@mulmoclaude/markdown-plugin", "@mulmoclaude/form-plugin", "@mulmochat-plugin/generate-image", "@mulmoclaude/chart-plugin", "@mulmoclaude/collection-plugin", "@mulmoclaude/html-plugin"],
   "servers": ["@mulmoclaude/x-plugin"],
   "local": []
 }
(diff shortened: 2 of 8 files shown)


snakajima:
Addressed in the latest push — valid finding, fixed.

`mountHtmlPreviewRoute`'s CSP now starts with **`sandbox allow-scripts`**, so a presentHtml page loaded by **direct navigation** to `/artifacts/html/…` is opaque-origin and its inline scripts can't reach the app origin's cookies / `/api/*`. This matches the hardening already on the collection `view-file` route. The View's iframe render is unchanged — it's already a `sandbox="allow-scripts"` opaque-origin frame, so the effective sandbox (intersection) is identical. `connect-src 'none'` still blocks fetch/XHR. Test updated to assert the `sandbox` directive.

PR #48, opened 2026-06-19
feat(grid): grid-first multi-terminal with zoom, status, reload-resilience (#46)

isamu:
## Summary

First slice of the multi-terminal plan (#46, `plans/feat-multi-terminal-grid.md`).

**Design note (changed from the original P0):** this is **grid-first with a zoom**, NOT a Single⇄Grid toggle. Per the maintainer's direction ("基本はgrid"), the 2×2 grid is the base view and you **zoom one cell to fullscreen and back** (a zooming interface). This is an intentional UX change — the old single view (sidebar/history + GUI panel + Tools pane + splitter) is not rendered at the top level for now; the **GUI panel + Tools return inside the *expanded* cell in a later phase** (decision: GUI/Tools only when zoomed).

## What's in this PR

- **Grid base view** — fixed 2×2 of independent cells (`TerminalGrid` → `TerminalCell`, reusing `Terminal.vue`).
- **Zoom** — ⤢ expands a cell to fill the area, ⤡ restores; animated via grid-track transition (Mac-like). Other cells stay **mounted** while zoomed so their PTYs keep running.
- **Lazy launch** — empty cells start nothing; "New terminal" spawns on demand.
- **Cell header** — status dot (idle / working=pulsing / waiting=attention, from the activity pub/sub), workspace dir basename, and the latest prompt.
- **Reload resilience**:
  - grid state (each cell's session id + zoom) persisted to `localStorage`;
  - idle sessions kept for a 30s grace window on disconnect so a reload **re-attaches the live PTY** (full scrollback), instead of being reaped;
  - dropped sockets auto-reconnect (resuming the known id);
  - a requested id that can't be served (no transcript, no live pty) mints a fresh id;
  - `superseded` message stops the reattach ping-pong when a session is open in two tabs;
  - the cell's ✕ sends `terminate` for immediate reap.
- **Server**: `GET /api/config` (`{cwd}`), `GET /api/session/:id` (initial status+prompt), `lastPrompt` in the activity publish (pruned on reap).
- **dev**: `node --watch` so the dev server auto-restarts on change.
- **Tests**: `TerminalGrid.spec` + `TerminalCell.spec` (localStorage persist/restore, zoom toggle, close/reset, pub/sub status/prompt incl. null-clear, foreign-session filtering, Windows-path basename). 29 → 41.

## Verification

- `lint` / `typecheck` / `typecheck:server` / `test` (41) / `build` ✅.
[...]

PR #59, opened 2026-06-20
feat(grid): layout selection (2x2/3x2/4x2/3x3) + per-terminal working dir (#46)

isamu:
## Summary

Two grid features requested after #48:

### 1. Layout selection
- Toolbar picker for **2x2 / 3x2 / 4x2 / 3x3** (persisted).
- The grid generalizes cols/rows/cellCount from the layout; the Mac-like zoom
  (track-collapse) works for any layout. Persisted cell arrays are kept at
  `MAX_CELLS` (9) so a session/cwd survives switching to a smaller layout and back;
  a zoomed cell that the new layout hides un-zooms.
- Layout math is in `gridLayout.ts` (unit-tested).

### 2. Per-terminal working directory
- An empty cell shows a **dir field (prefilled with the server default) + Start**.
  The chosen dir is sent as `?cwd=<abs>` and used as the PTY cwd (and the project
  to `--resume` from). Each cell can run in a **different directory**, shown in its
  header; the dir is **persisted per cell** so a reload restores it.
- Server `resolveWorkspace()` validates the dir (absolute, existing directory) and
  **falls back to `CLAUDE_CWD`** for anything else — raw client input is never used
  in a path/flag position unchecked.

## Verification

- `lint` / `typecheck` / `typecheck:server` / `test` (52) / `build` ✅.
- End-to-end: two cells launch claude in `/tmp/celldir-a` and `/tmp/celldir-b`;
  an invalid `?cwd=` falls back to `CLAUDE_CWD` (server log confirms).
- Tests: `gridLayout` (dims/trackStyle/isLayout), grid per-layout cell counts +
  layout-change un-zoom + per-cell cwd persist/restore, cell launch form emits cwd
  and passes it to the terminal.

## Out of scope (later)

dir **presets** in a settings screen (this PR is manual entry, prefilled with the
default); sound/notifications; tabs + 20-session cap; GUI/Tools inside the
expanded cell.

base: `dev_tool`.

### DIFF
--- a/server/index.ts
+++ b/server/index.ts
@@ -1031,8 +1031,10 @@ wss.on("connection", (ws, req) => {
   const sessionId = reattachId ?? resume ?? randomUUID();
   const live = reattachId ? ptys.get(reattachId) : undefined;
 
-  // Tell the browser which session this is (it learns the id of new sessions).
-  ws.send(JSON.stringify({ type: "session", id: sessionId }));
+  // Tell the browser which session this is (it learns the id of new sessions) and
+  // the EFFECTIVE cwd — which may differ from what was requested (resolveWorkspace
+  // falls back to CLAUDE_CWD), so the cell shows/persists where claude really runs.
+  ws.send(JSON.stringify({ type: "session", id: sessionId, cwd }));
 
   let entry: PtyEntry;
   try {
--- a/server/index.ts
+++ b/server/index.ts
@@ -7,7 +7,7 @@ import path from "path";
 import os from "os";
 import fs from "fs/promises";
 import { randomUUID } from "crypto";
-import { existsSync } from "node:fs";
+import { existsSync, statSync } from "node:fs";
 import { fileURLToPath } from "url";
 import { StreamableHTTPServerTransport } from "@modelcontextprotocol/sdk/server/streamableHttp.js";
 import { createPubSub } from "./pubsub.js";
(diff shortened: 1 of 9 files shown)


isamu:
Addressed the Codex review:

**cwd mismatch (server may fall back to CLAUDE_CWD)** — the server is now authoritative. The `session` message carries the **effective** cwd; `Terminal` emits it and the cell **adopts/persists that**, not the typed value (which is only an optimistic placeholder until confirmed). So the header + localStorage always match where the PTY actually runs.

Verified: `?cwd=/tmp/eff-ok` → `session.cwd:/tmp/eff-ok`; `?cwd=relative/bad` → `session.cwd:<CLAUDE_CWD>`. Added a test that a bad typed path results in the cell persisting/showing the server's effective cwd.

CI was already green (lint/typecheck/typecheck:server/test 53/build/package-smoke).

isamu:
Addressed the follow-up Codex review:

- **Reattach reported the request's cwd, not the live PTY's** — a `PtyEntry` now stores its cwd, and the `session` message on reattach reports that (the reattaching `?cwd=` is ignored by the live process). Verified: reattaching a session with a different `?cwd=` still reports the original dir.
- **Cell kept stale cwd/dirInput after close** — `close()` now resets `cwd`/`dirInput` to the default (the cell isn't remounted), so the empty launch form no longer shows the closed session's directory. Added a test.

All gates green: lint / typecheck / typecheck:server / test (54) / build.

isamu:
Addressed the a11y nit (9d4b11d): the launch-form dir input is now wrapped in its `<label>` (caption in a span), giving it a proper accessible name with no per-cell id collisions. lint/typecheck/test (54)/build green.

PR #60, opened 2026-06-20
feat(grid): directory presets — settings screen + one-click launch (#46)

isamu:
## Summary

Saved working-directory **presets**, so launching a terminal in a known project is one click.

- **⚙ Settings** in the toolbar opens a modal to add / edit / remove presets (`{label, path}`).
- The cell launch form shows **preset chips**; clicking one launches a terminal in that dir immediately. Manual entry still works.
- Persisted server-side at `~/.mulmoterminal/config.json` (survives restart).

## Changes

- **server**: `GET /api/config` → `{ cwd, cwdPresets }`; `POST /api/config` validates (trims, drops rows missing label/path, caps at 50) and persists.
- **App.vue**: owns config (default cwd + presets), the ⚙ button, and `SettingsModal`; passes them to the grid.
- **TerminalGrid**: forwards `defaultCwd` + `presets` to cells (config fetch moved up to App).
- **TerminalCell**: preset chips in the launch form (`launchPreset`).
- **SettingsModal.vue** (new) + `presets.ts` (shared type).

## Verification

- `lint` / `typecheck` / `typecheck:server` / `test` (60) / `build` ✅.
- End-to-end: POST trims/filters junk rows, persists to `config.json`, and reloads after a server restart.
- Tests: SettingsModal (add/remove/save-trims/cancel/no-prop-mutation), grid forwards presets, cell one-click preset launch.

## Note

Current dir + latest user prompt are already shown in each cell header (from #59). This PR adds the presets + settings screen.

base: `dev_tool`.

### DIFF
--- a/server/cwd-presets.spec.ts
+++ b/server/cwd-presets.spec.ts
@@ -1,0 +1,54 @@
+import { describe, it, expect } from "vitest";
+import { mkdtempSync, writeFileSync, readFileSync, rmSync } from "node:fs";
+import { tmpdir } from "node:os";
+import path from "node:path";
+import { sanitizePresets, loadPresets, savePresets } from "./cwd-presets";
+
+const tmp = () => mkdtempSync(path.join(tmpdir(), "mt-presets-"));
+
+describe("sanitizePresets", () => {
+  it("keeps valid {label,path}, trims, and drops incomplete/junk rows", () => {
+    expect(sanitizePresets([{ label: " a ", path: " /a " }, { label: "", path: "/b" }, { label: "c", path: "" }, { nope: 1 }, "x"])).toEqual([
+      { label: "a", path: "/a" },
+    ]);
+  });
+
+  it("returns [] for non-array input", () => {
+    expect(sanitizePresets(null)).toEqual([]);
+    expect(sanitizePresets({ cwdPresets: [] })).toEqual([]);
+  });
+
+  it("caps the count", () => {
+    const many = Array.from({ length: 60 }, (_, i) => ({ label: `l${i}`, path: `/p${i}` }));
+    expect(sanitizePresets(many, 50)).toHaveLength(50);
+  });
+});
+
+describe("savePresets / loadPresets", () => {
+  it("round-trips through a file", () => {
+    const dir = tmp();
+    const file = path.join(dir, "nested", "config.json"); // nested → mkdir is exercised
+    expect(savePresets(file, [{ label: "x", path: "/x" }])).toBe(true);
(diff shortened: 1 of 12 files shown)


isamu:
Addressed the Codex review:

- **Modal closed unconditionally on save** — `savePresets()` now closes **only on a successful POST**; on failure the modal stays open (draft preserved) and shows an inline error. Added `saving`/`error` state; Save shows "Saving…" and is disabled while in flight.
- **Modal keyboard a11y** — `Escape` closes the dialog, `Tab` is trapped within it, and the first field is focused on open.

Tests added (Escape → close; error shown + Save disabled while saving). All gates green: lint / typecheck / typecheck:server / test (62) / build.

isamu:
Addressed the Codex review: `saveCwdPresets()` now returns a boolean and `POST /api/config` responds **500** on a disk-write failure instead of a false 200 — the client (App.savePresets) then keeps its error state and the modal stays open. Verified POST still returns 200 on success. lint / typecheck:server / test (62) / build green.

isamu:
Addressed the Codex review:

- **Stage-then-commit**: the POST handler now sanitizes into `next`, persists it, and commits `cwdPresets` in-memory **only on a successful write**. A failed write returns 500 *and* leaves GET reporting the previous (actually-persisted) presets — no divergence between memory and disk.
- **Testability + regression test**: extracted the logic to `server/cwd-presets.ts` (`sanitizePresets`/`loadPresets`/`savePresets`) with unit tests, including the **write-failure → `savePresets` returns false** path (the 500 case). `tsconfig.server.json` excludes `*.spec.ts`.

68 tests; lint / typecheck / typecheck:server / build green.

isamu:
Addressed the Codex review: `SettingsModal` now `watch`es `props.presets` and resyncs the editable rows when they change — so opening Settings before `/api/config` resolves no longer edits stale/empty data (and Save can't wipe real presets with `[]`). Added a regression test for the early-open race. lint / typecheck / test (69) / build green.

isamu:
Addressed the Codex a11y nit: the preset `label`/`path` inputs now have `aria-label` ("Preset label" / "Preset directory path") + a test asserting the accessible names. lint / typecheck / test (70) / build green.

isamu:
Addressed the Codex review: SettingsModal now tracks a `dirty` flag (set on add/remove/input) and the `props.presets` watch only hydrates `rows` **while the form is pristine** — late-arriving config no longer clobbers in-progress edits. Added a regression test for 'type before config resolves' (the pristine early-open resync test is kept too). lint / typecheck / test (71) / build green.

PR #67, opened 2026-06-20
feat(grid): resume an existing session from the cell launch picker (#46)

isamu:
## Summary

The cell launch form could only start a **new** session. Now it also lets you **resume an existing** one — the "new vs existing" choice from the plan.

An empty cell shows:
- **New**: dir field + preset chips + `＋ New terminal` (unchanged).
- **or resume here**: a list of existing sessions for the entered directory (title + relative time); clicking one resumes that session in that dir.

## Changes

- **server**: `GET /api/sessions` accepts an optional **`?cwd=<abs>`** that scopes the on-disk session list to that project. Without it, the classic single view's behavior is unchanged (CLAUDE_CWD + in-memory pending sessions). Validated via `resolveWorkspace`.
- **TerminalCell**: fetches sessions for the form's dir (debounced on change, on mount, after close), renders the resume list, and `resume(s)` launches with that session id + cwd (server `--resume`s it).

## Verification

- `lint` / `typecheck` / `typecheck:server` / `test` (145) / `build` ✅.
- Tests: resume list renders + click resumes (correct sessionId + cwd); no list when the dir has no sessions.
- `/api/sessions?cwd=<dir>` returns `cwd` + that project's sessions; default `/api/sessions` unchanged (verified live).

## Note

The resume list is **directory-scoped, on-disk** sessions (resumable ones). It excludes in-memory pending sessions (those aren't tracked per directory).

base: `dev_tool`.

### DIFF
--- a/server/index.ts
+++ b/server/index.ts
@@ -800,9 +800,15 @@ app.get("/api/session/:id", (req, res) => {
 
 // List the chat sessions for the current project (CLAUDE_CWD), including
 // newly-created sessions that aren't persisted to disk yet.
-app.get("/api/sessions", async (_req, res) => {
+app.get("/api/sessions", async (req, res) => {
   try {
-    const dir = projectSessionsDir(CLAUDE_CWD);
+    // Optional ?cwd= scopes the list to that project's on-disk sessions (the grid
+    // cell's resume picker). Without it, the classic single view's behavior is
+    // unchanged: CLAUDE_CWD + in-memory pending sessions.
+    const cwdParam = typeof req.query.cwd === "string" ? req.query.cwd : null;
+    const cwd = cwdParam ? resolveWorkspace(cwdParam) : CLAUDE_CWD;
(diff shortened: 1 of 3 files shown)


isamu:
Addressed the Codex review:

- **Stale/out-of-order `loadResumable()`** — added a request token (`resumableReq`); a response is applied only if it's still the latest request, so a slow earlier-dir response can't overwrite the current dir's rows.
- **`resume()` cwd mismatch** — the listed sessions now carry the **resolved cwd** the server returned (`resumableCwd`), and `resume()` uses that (not the live `dirInput`). So `session=<id>` and `cwd` always match the clicked row, and the server's resolved/fallback `cwd` is honored.
- **Tests** — out-of-order response keeps the latest dir's rows; resume uses the API-resolved cwd, not the typed input. (147 tests.)

lint / typecheck / typecheck:server / test / build green.

PR #130, opened 2026-06-28
feat: scheduled feed refresh (register shared feedRefreshTaskDef)

snakajima:
Follow-up to #129 (manual refresh). **Standalone MulmoTerminal** (no MulmoClaude running) now refreshes due feed / agent-ingest collections **on schedule**, using the shared `feedRefreshTaskDef` exported from `@mulmoclaude/core/feeds` (`0.2.10`, MulmoClaude PR #1843) — no host-duplicated task definition.

## Changes
- **Bump `@mulmoclaude/core` → `^0.2.10`**.
- **`server/backends/scheduler.ts`** — `initUserTaskScheduler` now accepts an optional `systemTasks?: TaskDefinition[]`, registers them alongside user tasks on the same task-manager, and starts the tick loop whenever **any** task exists (so a system task alone drives its schedule).
- **`server/index.ts`** — pass `systemTasks: [feedRefreshTaskDef({ workspaceRoot: CLAUDE_CWD })]`. The feeds host is configured earlier at boot (`initFeedsBackend`), so `refreshDue` can run.

## Double-scheduling (shared workspace)
When both apps run against `~/mulmoclaude`, dedup relies on the engine's shared `lastFetchedAt` (per the plan's decision — **soft dedup, no lock**): whoever refreshes first stamps it; the other host's `isFeedDue` skips. The standalone case (the actual target) has no conflict at all.

## Verification
- Typecheck (client + server), production build, 177 server tests — all green.
- Boot smoke test against an empty workspace: `[scheduler] scheduler started { userTasks: 0, systemTasks: 1 }` — confirms the task registers and the tick loop starts with zero user tasks.

## Notes
- The hourly tick is the **polling cadence**; each feed's own `schedule` (hourly/daily/weekly) is still enforced by `isFeedDue`.
- **Forward-firing** (no catch-up): the first scheduled refresh is up to an hour after boot; the manual Refresh button covers immediate needs.
- Carry-over (plan open-Q4): scheduled **agent-ingest** dispatches hidden workers but doesn't yet invoke the `onComplete` failure-reconciliation — declarative feeds are fully correct. Flagged as a follow-up, not a blocker.

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,7 +54,7 @@
     "@mulmoclaude/accounting-plugin": "^0.2.0",
     "@mulmoclaude/chart-plugin": "^0.1.3",
     "@mulmoclaude/collection-plugin": "^0.5.11",
-    "@mulmoclaude/core": "^0.2.6",
+    "@mulmoclaude/core": "^0.2.10",
     "@mulmoclaude/form-plugin": "^0.1.3",
     "@mulmoclaude/html-plugin": "^0.2.5",
     "@mulmoclaude/markdown-plugin": "^0.1.7",
--- a/server/backends/scheduler.spec.ts
+++ b/server/backends/scheduler.spec.ts
@@ -1,11 +1,20 @@
 // @vitest-environment node
-import { describe, it, expect, beforeEach, afterEach } from "vitest";
+import { describe, it, expect, beforeEach, afterEach, vi } from "vitest";
 import express from "express";
 import { mkdtempSync, mkdirSync, writeFileSync, rmSync } from "node:fs";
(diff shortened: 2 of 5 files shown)


snakajima:
Addressed the review: added `initUserTaskScheduler` unit tests (mocking the shared task-manager) covering (1) system tasks are registered and (2) the tick loop starts when there are zero user tasks but ≥1 system task — plus the no-tasks-at-all (no start) and both-system-and-user-tasks cases. `server/backends/scheduler.spec.ts` now 12 tests; lint + typecheck + full suite green.

PR #152, opened 2026-06-29
fix: don't reap working/waiting chat sessions on switch-away

snakajima:
## Problem

Running a long task in a chat session, switching away to do something else, then coming back — the terminal had been closed and the task aborted ("タスクはやめました").

A chat session's PTY is reaped when its WebSocket closes (which happens on every switch-away) unless the hook-driven `working` flag is set. The killer case: a long multi-step task that **ends its turn to ask the user a question** fires the `Stop` hook → `working=false`. The old `setWorking(false)` then reaped the detached PTY **immediately, with no grace window** — so returning to answer the question found the session gone.

The fragile assumption was "running == `working` flag". A session `waiting` on a permission/question prompt is also an unfinished task, but counted as reapable.

## Fix

Replace the working-only check with a 3-tier policy in a new `armReapForDetached()`, used by both the socket-close path (`handleClientClose`) and the turn-end path (`setWorking(false)`):

| State | Meaning | Behaviour |
|---|---|---|
| `working` | actively thinking | never reaped |
| `waiting` | needs the user (permission / question / unseen Stop) | long grace `WAIT_REAP_GRACE_MS` (default 30 min) |
| neither | finished + viewed, or idle | short `REAP_GRACE_MS` (30s) — auto-close |

- Removes the **immediate** reap in `setWorking(false)`; a turn that ends while you're away now gets a grace matching its state instead of an instant kill.
- `WAIT_REAP_GRACE_MS=0` keeps needs-the-user sessions alive until reattached or explicitly closed.
- Genuinely idle sessions still auto-close after 30s, so the user isn't burdened with closing finished small tasks by hand.

## Testing

- `typecheck:server`, `eslint`, and all 459 tests pass.
- Manually verified: long task → switch away → return → session still alive.

### DIFF
--- a/server/index.ts
+++ b/server/index.ts
@@ -386,7 +386,17 @@ const REAP_GRACE_MS = 30_000;
 // unfinished task: reaping it loses work. So it gets a much longer grace than an
 // idle one, long enough that you can switch away, do other things, and come back
 // to answer it. Override with WAIT_REAP_GRACE_MS=0 to never auto-close these.
-const WAIT_REAP_GRACE_MS = process.env.WAIT_REAP_GRACE_MS !== undefined ? Number(process.env.WAIT_REAP_GRACE_MS) : 30 * 60_000;
+const WAIT_REAP_GRACE_MS = (() => {
+  const def = 30 * 60_000;
+  const raw = process.env.WAIT_REAP_GRACE_MS;
+  if (raw === undefined) return def;
+  const n = Number(raw);
+  if (!Number.isFinite(n)) {
+    console.warn(`[pty] ignoring non-numeric WAIT_REAP_GRACE_MS=${JSON.stringify(raw)}; using default ${def}ms`);
+    return def;
+  }
+  return n; // a non-positive value means "never auto-reap waiting sessions" (see scheduleReap)
+})();
 const reapTimers = new Map<string, ReturnType<typeof setTimeout>>();
 
 function cancelReap(id: string) {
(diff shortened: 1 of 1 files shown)


snakajima:
Good catch — fixed in 69a90bf.

The `idle -> waiting` escalation no longer leaves a stale short timer:
- `armReapForDetached()` now `cancelReap()`s and recomputes the grace from current state, instead of relying on `scheduleReap`'s "already armed" early-return.
- `setWaiting(true)` re-arms a detached session, so a `Notification` arriving after disconnect escalates the 30s idle timer to the long `WAIT_REAP_GRACE_MS` grace.

typecheck:server, eslint, and all 459 tests pass.

snakajima:
Addressed in e972f96.

- `WAIT_REAP_GRACE_MS` is now parsed with a `Number.isFinite` check that falls back to the 30m default (and warns) on non-numeric input, so `WAIT_REAP_GRACE_MS=abc` no longer yields `NaN`.
- Hardened `scheduleReap` to ignore any non-finite delay (`!Number.isFinite(delayMs) || delayMs <= 0`) as defense-in-depth, so a bad delay can never schedule a near-immediate `setTimeout`.

The two earlier escalation comments were against commit `441fc2f` and were already fixed in `69a90bf` (cancel+recompute in `armReapForDetached`, re-arm in `setWaiting`).

typecheck:server, eslint, and all 459 tests pass.

snakajima:
Addressed in 349c99b.

`scheduleReap` now clamps the delay to `MAX_TIMER_MS` (`2147483647`, ~24.8 days) before calling `setTimeout`, so a large `WAIT_REAP_GRACE_MS` (e.g. `3000000000`) no longer overflows the 32-bit timer and fires at ~1ms. Combined with the earlier `Number.isFinite`/`<= 0` guards, the delay is now bounded on both ends before scheduling.

typecheck:server, eslint, and all 459 tests pass.

PR #155, opened 2026-06-29
feat: remember recent working directories in the cell launcher

snakajima:
## Summary

The multi-terminal launcher can now **remember the last few working directories** you started sessions in. They appear as one-click chips under the curated preset chips, so you don't have to retype paths.

Built on the request: *"The multi-terminal mode allows me to specify the current folder … It would be great if it can remember last few locations."*

## What changed

- **Recents (last 4 dirs)** — recorded from the **server-confirmed (effective) cwd**, so only directories that actually ran are remembered. Persisted in `localStorage` via a module-level singleton, so every empty cell shares one list. New composable: `src/composables/useRecentDirs.ts`.
- **One-click chips, consistent behavior** — both preset and recent chips are now segmented pills:
  - **Main button** → launches a fresh session in that dir.
  - **Tiny end button** (pencil) → only *fills* the field and loads the resume / script / worktree lists, so you can resume an existing session there without launching.
- **Removed the redundant "＋ New terminal" button** — chips launch on click, and the path field launches on Enter.
- **Prefill from history** — a fresh launcher prefills the Working-directory field with your most recent location instead of the server default (`~/mulmoclaude`).
- **Top-aligned the launch form** — was vertically centered, which looked adrift with few/no resume rows.

## Testing

- `npm run test` — **463 passed**. Added coverage for: recording on server cwd, chip launch, the fill-without-launch button, prefill-from-recent, and the 4-item cap. Also reset the recents singleton + `localStorage` in `beforeEach` to prevent cross-test pollution (one test emits a server cwd that previously leaked into later cells' prefill).
- `npm run lint` — clean.
- `npm run typecheck` — clean.
- `npm run build` — succeeds (the >500 kB chunk warning is pre-existing and unrelated).

### DIFF
--- a/src/components/TerminalCell.spec.ts
+++ b/src/components/TerminalCell.spec.ts
@@ -134,6 +134,25 @@ describe("TerminalCell", () => {
     expect(term.props("cwd")).toBe("/home/me/picked");
   });
 
+  it("launches via the go button next to the field (alternative to Enter)", async () => {
+    const w = mountCell(null, { defaultCwd: "/home/me/default" });
+    await flushPromises();
+    await w.find(".cell-dir-input").setValue("/home/me/picked");
+    await w.find(".cell-dir-go").trigger("click");
+    const term = w.findComponent({ name: "TerminalView" });
+    expect(term.exists()).toBe(true);
+    expect(term.props("cwd")).toBe("/home/me/picked");
+  });
+
+  it("disables the go button when the field is empty", async () => {
+    const w = mountCell(null, { defaultCwd: null });
+    await flushPromises();
+    await w.find(".cell-dir-input").setValue("   ");
+    expect((w.find(".cell-dir-go").element as HTMLButtonElement).disabled).toBe(true);
+    await w.find(".cell-dir-input").setValue("/home/me/picked");
+    expect((w.find(".cell-dir-go").element as HTMLButtonElement).disabled).toBe(false);
+  });
+
   it("shows a cancel ✕ on a cancellable launcher that emits close, but not otherwise", async () => {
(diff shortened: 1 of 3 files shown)


snakajima:
Addressed the review feedback in 70eb589:

**Accessibility (github-actions)** — Added `aria-label` to the icon-only `.cell-chip-fill` buttons and `type="button"` to all launcher buttons (`.cell-chip-main`, `.cell-chip-fill`, `.cell-dir-go`) so screen readers announce a clear action and the buttons can't accidentally submit if ever wrapped in a form.

**Record recents only for user-initiated launches (Codex P2)** — `recordDir` is now gated on a fresh-launch flag (`recordNextCwd`, set in `launchIn`) instead of firing on every server cwd report. Reconnecting/restoring a persisted session also reports a cwd, and recording those rewrote `recentDirs` in mount/socket order on reload — so a fresh launcher could prefill an old viewed session's dir instead of the last one actually launched. Added regression tests covering both a restored session and a resume-list resume (neither records a recent).

**Clickable launch for typed paths (Codex CHANGES REQUESTED)** — already resolved by the play button next to the working-directory field (commit 3858663), which launches on click in addition to Enter, and is disabled when the field is blank.

All 467 tests pass; lint and build clean.

snakajima:
Good catch on the stale-flag race — fixed in c7e3bba.

`recordNextCwd` is now also cleared in `resume()` and `teardown()`, so a fresh launch that's torn down or resumed before its `cwd` event arrives can't leak the pending record onto the next session's cwd. Added a regression test: launch → close before any cwd → resume an existing session → asserts nothing is recorded (it would record `/home/me/proj` without the fix).

468 tests pass; lint and build clean.

PR #158, opened 2026-06-30
feat: persist terminal connections across UI navigation

snakajima:
## Problem

When you leave a terminal in multi-terminal (grid) mode — or switch away from the chat view — the Vue component unmounts, its WebSocket closes, and the server reaps the Claude PTY after a grace period. Coming back re-mounts the component, which cold-restarts the session via `claude --resume`. Because the Anthropic prompt cache has only a 5-minute TTL, this reprocesses the full context at the uncached rate and triggers Claude Code's "restoring this session will consume a lot of tokens" warning.

VS Code doesn't have this problem because its terminal apps keep running while hidden.

## Approach (client-only, zero server change)

Decouple the WebSocket + xterm lifetime from the Vue component lifetime. A module-singleton **connection manager** (`useTerminalConnections.ts`) owns each terminal's `Terminal` + `FitAddon` + a once-`open()`'d host `<div>` + `WebSocket`, keyed by a stable slot id:

- `attach(key, …, el)` — on first acquire, creates the terminal and connects; on later acquires, **re-parents the existing host div** into the new mount point (Variant A) instead of reconnecting a live slot.
- `detach(key, el)` — re-parents the host out of the DOM but keeps the socket alive.
- `release(key)` / `terminate(key)` — for ephemeral and command slots, fully tears down on unmount.

Because an open WebSocket keeps the server PTY alive indefinitely, navigating away no longer reaps it. `Terminal.vue` is slimmed to a thin view over the manager; its public props/emits/exposed methods are unchanged.

Persist keys:
- single chat view → `"single"`
- grid cells → `"cell-<uid>"`
- ephemeral / command terminals → released on unmount (unchanged behavior)

## Backstop

Closing the browser tab still tears down the TCP socket cleanly, so the server's existing reap grace tiers (`working`=never, `waiting`=30 min, `idle`=30 s) remain the backstop. "Dirty disappearance" (laptop sleep / crash leaving a zombie socket) is a pre-existing condition and explicitly out of scope.

See `plans/feat-persistent-terminal-connections.md` for the full design.

## Testing

- `yarn lint`, `yarn format:check` — clean
- `yarn typecheck` + `yarn typecheck:server` — clean
- `yarn build` — succeeds
- `yarn test` — 477 passed

### DIFF
--- a/plans/feat-persistent-terminal-connections.md
+++ b/plans/feat-persistent-terminal-connections.md
@@ -1,0 +1,254 @@
+# Plan: Keep PTYs alive by persisting terminal connections client-side
+
+## Problem
+
+When the user navigates away from the terminal UI for a while and comes back,
+Claude Code warns that restoring the session will consume a lot of tokens. This
+happens because:
+
+1. Navigating away (or flipping to an off-page grid tab, or switching the whole
+   view) **unmounts** `Terminal.vue`, whose `onUnmounted` closes the WebSocket.
+2. The server treats a socket close as a detach and arms the reap grace timer
+   (`armReapForDetached`): idle = 30s, waiting = 30min, working = never.
+3. An **idle** session is reaped after 30s. Coming back then spawns
+   `claude --resume <id>`, which replays the transcript from disk and triggers
+   the "restoring old session consumes a lot of tokens" warning.
+
+VS Code never shows this because the terminal process and its connection stay
+alive across tab switches. We want the same: a session stays connected as long
+as it is **logically open**, and only goes cold when explicitly closed or pulled
+from history/resume.
+
+## Goal
+
+The backend Claude PTY stays alive (no `--resume`, no token warning, VS Code
+parity) for as long as a session's terminal UI is **logically open** — i.e. it
+would be re-mounted when the user navigates back, NOT re-opened from the history
+list. Concretely, "logically open" = the union of:
+
+- every grid cell that has a session id (the `grid_v2` roster), regardless of
+  which page is currently rendered, and
+- the single/chat view's currently-active session.
+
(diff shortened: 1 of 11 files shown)


snakajima:
Fixed in ca9b461. The connection now tracks the server-resolved cwd (`knownCwd`) alongside `knownSessionId`, and `attach()` replays both to the freshly-bound handlers. So a slot that learned its session id/cwd while detached forwards them on remount, and the parent's `grid_v2`/session state catches up — no more `session: null` persisted for a live connection. `knownCwd` is reset on `retarget()`.

snakajima:
Both follow-ups addressed in c3785b3:

**Unload guard (App.vue / GridView.vue P2):** Confirmed and fixed. `reportActiveTerminals` now keys counts by source (`"single"` / `"grid"`) and sums them instead of overwriting a single shared ref. So a single session kept alive after switching to the grid still contributes to the `beforeunload` guard — and symmetrically, grid cells (which now persist too) keep counting after switching back to the single view. Added regression cases in `useUnloadGuard.spec.ts` for "only the hidden single terminal is live" and "blocks while either source is live, clears only when both are zero".

**Regression test for the detach-before-`session` race:** Added `useTerminalConnections.spec.ts`, which mocks xterm + WebSocket and drives the exact lifecycle: fresh launch (no id) → socket open → `detach()` before `{type:"session"}` → server assigns `id`/`cwd` while detached → `attach()` with fresh handlers. It asserts the reattached handlers receive `sess-123` and `/resolved` (and a second test asserts no bogus id is replayed before the server assigns one). Verified it fails without the replay fix.

481 tests pass; lint/format/typecheck (client+server)/build all green locally.

PR #161, opened 2026-06-30
feat: adopt vue-router for top-level navigation

snakajima:
## Summary

Phase 2 of `plans/feat-vue-router.md` (Phase 1 was #160, the `/api/mcp` relocation). MulmoTerminal had **no router** — top-level navigation was a hand-rolled `viewMode` ref + `localStorage` plus the `browse*`/`accounting` reactive stores, which already admitted to being a router stand-in. This promotes those to real **history-mode URLs**:

| Path | Surface |
|---|---|
| `/` | chat (single view) |
| `/terminals` | multi-terminal grid |
| `/collections` · `/collections/:slug` | collections index / page |
| `/feeds` · `/feeds/:slug` | feeds index / page |
| `/accounting` | accounting overlay |

Per the locked decisions: **records stay non-history modals**, **sessions are not addressable**, **settings stays a modal**.

## Approach

- **`src/router/index.ts`** — the router singleton + a `Stub` component. App.vue renders by `route.name` (**not** `<router-view>`), so the existing 3-pane shell and full-screen overlays stay exactly where they are. Catch-all redirects unknown paths to `/`.
- **`useCollectionBrowse` / `useAccountingView`** — reimplemented on the router singleton with **unchanged export names**, so `collectionUi`, the toolbar, and the overlays come along untouched. The open *record* is the only retained state — **keyed by its slug** so a slug change implicitly drops the modal (no watcher racing navigation), and it never touches the URL.
- **`AppToolbar` / `GridView`** — launcher buttons push routes directly; the now-redundant `go-single` / `go-grid` / `exit` plumbing is removed; grid-state derives from `route.name`.
- **Server SPA fallback** (`server/spa-fallback.ts`) — `index.html` for any non-`/api` path. Safe because every server HTTP endpoint is under `/api` (Phase 1), static assets + the `/artifacts/html` preview are registered earlier (so they win), and WebSocket upgrades bypass Express via `server.on("upgrade")`.

## Verification

Built the server and probed it live:
- `/terminals`, `/collections/foo`, `/` → **200 SPA shell** (`id="app"` present)
- `/api/bogus` → **404** (Express default, `id=app` count: 0 — never the SPA)
- `/assets/index-*.js` → **200 `text/javascript`** (static not shadowed)

## Test plan

- [x] `yarn format` / `yarn lint`
- [x] `yarn typecheck` + `yarn typecheck:server`
- [x] `yarn build`
- [x] `yarn test` — **491 passed** (+10 new: router table/slug/redirect, `useCollectionBrowse` over the router, SPA matcher)
- [x] live deep-link / `/api` 404 / asset probes (above)

## Out of scope
`/wiki` routes (deferred to the wiki port), chat/terminal sessions as URLs, collection records as URLs, `/settings` route — all intentionally not done.


snakajima:
@codex review

Addressed both findings in 5452366:
- **Bare `/api`**: `SPA_FALLBACK_RE` now reserves the whole `/api` prefix via `(?!api(?:/|$))`, so `GET /api` 404s instead of serving the shell. Added matcher tests for `/api` (reserved) and `/apidocs` (not over-reserved). Verified live.
- **Record kind-leak**: the open-record modal is now keyed by the exact detail **path** (not slug), so `/collections/foo` and `/feeds/foo` never share a record. Also clear record state on every programmatic non-record navigation, which restores the original 'leaving the page drops the modal' semantics and fixes the same-page-return revive (the P2 inline note). Added regression tests for the kind-leak hop and the same-page return.

snakajima:
@codex review

Fixed in d8c8ec9. Replaced the per-handler clears with a single **sync** watcher on the route path in `useCollectionBrowse`: leaving a detail page by any means (toolbar Chat/Grid push, browser Back/Forward, hand-typed URL) drops the open record, so it can't revive on return. `browseNavigateToRecord` re-applies its record in the push's `.then` — `flush:"sync"` guarantees the watcher's clear runs first. Added a regression test for the bare-push leave + return path.

snakajima:
@codex review

Addressed both in 18da165:
- **main.ts**: now mounts only after `router.isReady()`, so a deep-link/reload to `/terminals` honors the URL before any component mounts — no stray 'single' PTY attaches during the start-location render.
- **navigateToRecord**: the push's `.then` now always (re)assigns record state (`recordId ? targetPath : null` / `recordId ?? null`), so a bare ref-hop to the current collection page closes any open record instead of leaving it stale.

The earlier AppToolbar `showChat`/`showGrid` Back-revive comments are resolved by the d8c8ec9 sync path watcher (covered by the bare-push regression test) — leaving a detail page by any route change drops the record.

PR #167, opened 2026-06-30
Wire the custom-view write tier (PUT /view-data)

snakajima:
## Problem

Custom collection views could **read** but never **write**. A write-capable view — e.g. the vocabulary flashcard's ▲/▼ grade buttons — failed in the browser with:

```
Access to fetch at '.../api/collections/vocabulary/view-data' has been blocked by
CORS policy: Method PUT is not allowed by Access-Control-Allow-Methods in preflight response.
```

The CORS error was just the visible symptom. The host had never wired the write tier at all — three gaps in `server/backends/collections.ts`:

1. `Access-Control-Allow-Methods` was `GET, OPTIONS` (no `PUT`).
2. The view-token mint force-clamped every token to `["read"]`, so even a PUT route would 401.
3. There was no `PUT /view-data` route.

The working MulmoClaude host has all three; this ports them.

## Changes

- **CORS** — allow `GET, PUT, OPTIONS` on view-data.
- **Token mint** — grant the view's *declared* capabilities instead of forcing `["read"]`. `clampCapabilities` still guarantees a read-only view can never obtain a write token.
- **`PUT /view-data`** — new route gated by `requireViewToken("write")`. Honors `mode: "merge"` by layering each partial record onto the existing one (so grading a word writes only `proficiency` without clobbering `word`/`meaning`/`example`), and returns the `{ items, rejected }` contract views check.

## Tests

`server/backends/collections.spec.ts`: **25 → 30**. New cases cover the write-token grant, `PUT` in the CORS preflight, read-token `PUT` → 401, merge-preserves-untouched-fields, and missing-primary-key rejection.

- `tsc --noEmit`: clean
- `vitest run server/backends/collections.spec.ts`: 30 passed

### DIFF
--- a/server/backends/collections.spec.ts
+++ b/server/backends/collections.spec.ts
@@ -39,6 +39,7 @@ const SCHEMA = {
   fields: {
     id: { type: "string", label: "ID", primary: true, required: true },
     name: { type: "string", label: "Name" },
+    status: { type: "enum", label: "Status", values: ["open", "closed"] },
   },
   views: [
     { id: "v1", file: "views/v1.html", label: "Custom", capabilities: ["read"] },
@@ -223,10 +224,13 @@ describe("custom view routes", () => {
       body: JSON.stringify({ items: [{ id: "item1", note: "graded" }], mode: "merge" }),
     });
     expect(res.status).toBe(200);
-    const body = (await res.json()) as { items: Array<{ id: string; name?: string; note?: string }>; rejected: unknown[] };
+    const body = (await res.json()) as { written: string[]; rejected: unknown[] };
     expect(body.rejected).toEqual([]);
+    expect(body.written).toEqual(["item1"]);
     // name survived the partial write; note was added.
-    expect(body.items[0]).toMatchObject({ id: "item1", name: "Foo", note: "graded" });
+    const detail = await fetch(`${base}/api/collections/testcol/detail`);
+    const item1 = ((await detail.json()) as { items: Array<{ id: string; name?: string; note?: string }> }).items.find((i) => i.id === "item1");
+    expect(item1).toMatchObject({ id: "item1", name: "Foo", note: "graded" });
   });
 
   it("rejects an item missing its primary key", async () => {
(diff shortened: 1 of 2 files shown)


snakajima:
Addressed in b3c3aeb:

- **Singleton invariant** — `PUT /view-data` now rejects any id other than `schema.singleton` on a singleton collection, matching `PUT /items/:itemId`. The per-record write moved into a `writeViewItem()` helper that performs the guard, so a write-capable custom view can no longer mutate a singleton with an arbitrary id. Added tests for both the rejection and the fixed-id accept path.
- **Lint** — the same extraction flattens the route loop, resolving the `sonarjs/cognitive-complexity` failure.

Local: full `eslint .` clean, 32 tests passing.

snakajima:
Addressed in the latest commit: `mode: "merge"` is now update-only. A `readItem` miss rejects the item as `not found (merge is update-only)` instead of falling back to `{}` and writing a half-populated record; `mode: "replace"` still writes the record as given. Added a test that a merge to a missing id is rejected **and** creates no record file.

Local: `eslint` clean, 33 tests passing.

snakajima:
Both contract-compliance points addressed in the latest commit:

- **Modes** — `PUT /view-data` now accepts `"merge" | "upsert" | "create"` (default `"upsert"`) per `custom-view.md`, and 400s an unknown mode. `create` uses `refuseOverwrite` so an id collision is rejected (`already exists`) instead of silently overwriting; `merge` stays update-only; `upsert` writes as given. Dropped the invented `"replace"`.
- **Envelope** — now returns `{ written, rejected }` (`written` = stored ids), matching what views read. Was `{ items, rejected }`.

Also gated every row on `validateRecordObject` (required fields, enum values, id↔primaryKey) so invalid rows come back in `rejected` with an actionable `problem` rather than persisting.

Added tests: create-collision reject, default-upsert, unknown-mode 400, bad-enum reject, plus the updated envelope across existing cases. Local: full `eslint` + `tsc` clean, 37 tests passing.

PR #169, opened 2026-07-01
fix: hide multi-terminal grid sessions from the chat sidebar

snakajima:
## Problem

While a multi-terminal (grid) cell was running Claude, its session appeared as a **clickable row in the chat sidebar**. Selecting that row in the chat view reattached the session's live PTY and **superseded** the grid cell — so the chat "hijacked" the terminal and the multi-terminal page showed it as *disconnected*. Reproduced twice by the user.

### Root cause

- `/api/sessions` returns **every** session. A grid cell's session enters the list either as an in-memory *pending* entry (cwd-independent, right at launch) or, if it ran in the default workspace, as an on-disk transcript — indistinguishable from a chat session.
- Selecting any live session in chat hits the server's `reattachPty()` → `superseded` handoff, which detaches whoever held it first (the grid cell).
- The grid *does* guard against this internally (`openSessionIds` + a confirm dialog in `TerminalCell`), but that state is local to `GridView`, which is unmounted while you're in chat — so the chat sidebar has no such protection.
- `cwd` can't distinguish them: grid cells default to the same `{workspace}` folder as chat.

The only reliable discriminator — grid cells connect as **dev terminals** (`gui=0`, no GUI MCP) — was read at connect time (`attachGuiMcp`) and then thrown away.

## Fix

Record each dev-terminal (grid) session id and exclude it from the chat sidebar:

1. **Persisted set** `devTerminalSessions` → `~/.mulmoterminal/dev-terminal-sessions.json`, hydrated at boot. Persisted so the exclusion survives a reboot and outlives the live PTY (a reaped grid session's transcript shouldn't reappear as a chat).
2. **Marked at the single WS connection choke point** when `!attachGuiMcp` — covers new, resumed, and reattached grid cells, and re-records after a reboot when cells reconnect.
3. **Filtered from `/api/sessions` only for the unscoped chat query** (`includePending`). The grid's own resume picker uses `?cwd=…` (`includePending=false`) and keeps listing these, so grid sessions stay resumable there.

Background chats (feeds / scheduler / translation / drafts) spawn with the GUI MCP attached (`attachGuiMcp=true`), so they're never marked — only real grid dev terminals are hidden. No frontend changes needed.

## Verification

- `tsc -p tsconfig.server.json --noEmit` — clean
- `eslint server/index.ts` — clean
- `vitest run` — **555 passed / 55 files**

The inline `/api/sessions` + WS logic lives in the untested `index.ts` monolith, so this was verified via type/lint/full-suite rather than a targeted spec. Happy to extract the filter into a testable helper + add a spec if preferred.

### DIFF
--- a/server/index.ts
+++ b/server/index.ts
@@ -154,6 +154,39 @@ const pendingTranslations = new Map<string, { resolve: (translations: string[]) 
 // a value that claude re-parses as a flag) into the spawned process.
 const SESSION_ID_RE = /^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$/i;
 
+// Session ids that belong to the multi-terminal GRID — dev terminals, spawned with
+// gui=0 (no GUI MCP; see the ?gui handling in the WS connection handler). They're
+// FILTERED OUT of the chat sidebar's /api/sessions so a grid terminal never surfaces
+// as a clickable chat row: selecting it in chat would reattach its live PTY and
+// SUPERSEDE the grid cell (the "chat hijacked my multi-terminal session" bug). The
+// set is persisted so the exclusion survives a reboot and outlives the live PTY — a
+// reaped grid session's on-disk transcript still shouldn't reappear as a chat. NOTE:
+// the exclusion applies ONLY to the unscoped (chat) query; the grid's OWN cwd-scoped
+// resume picker (/api/sessions?cwd=…) must keep listing these so they stay resumable.
+const devTerminalSessions = new Set<string>();
+const DEV_TERMINAL_SESSIONS_FILE = path.join(MULMOTERMINAL_HOME, "dev-terminal-sessions.json");
+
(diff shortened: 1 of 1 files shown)


snakajima:
Addressed the Codex review in `b2974ac`:

**1. Write race (fire-and-forget persistence)** — persistence is now serialized into a single promise chain (`persistDevTerminalSessions`). Only one `writeFile` runs at a time, so concurrent marks can't interleave and clobber the file with an older snapshot. Each link stringifies the *current* set at execution time, so the last write always reflects the complete set.

**2. Hydration race** — the boot load is now an awaitable promise (`devTerminalSessionsHydrated`) instead of a detached IIFE:
- `/api/sessions` awaits it before filtering the chat query, so a request racing boot can't leak previously-hidden grid transcripts.
- Each persist waits on it too, so a mark that lands before hydration finishes never writes a set missing the on-disk ids (they're merged in first).

Verification: `tsc` clean, `eslint` clean, `vitest run` — 555 passed / 55 files.

PR #209, opened 2026-07-05
chore: adopt CI security checks from mulmoclaude (gitleaks + actionlint/zizmor)

isamu:
## Summary

Adopts two proven, self-contained CI jobs from the sibling repo **receptron/mulmoclaude** (as of 2026-07-05), plus the minimal hardening to `ci.yml` needed to pass them cleanly:

- **`secret-scan.yml`** — gitleaks full-history secret scan. Pulls a version-pinned CLI binary and SHA256-verifies it (no third-party Action). Honours `.gitleaksignore` automatically.
- **`workflow-lint.yaml`** — actionlint (+ embedded shellcheck) and zizmor (workflow supply-chain) on the workflow YAML itself, only when `.github/**` changes. `.github/zizmor.yml` relaxes `unpinned-uses` to `ref-pin` for first-party `actions/*` only (tag pins OK), still requiring a hash pin for any future third-party action.
- **`ci.yml` hardening** (so zizmor/actionlint stay green): top-level `permissions: contents: read`, `persist-credentials: false` on both checkouts, and the smoke-stub `printf` rewritten as a quoted heredoc (root-cause fix for the SC2016 false positive — the generated stub is byte-equivalent).
- **`.gitleaksignore`** — header-only template documenting the fingerprint format (MT has no findings today).

All three checks were run **locally at the pinned tool versions** and pass before pushing:

| check | result |
|-------|--------|
| gitleaks 8.30.1 (full history) | ✅ 0 leaks / 375 commits |
| actionlint 1.7.12 | ✅ 0 issues |
| zizmor 1.25.2 (regular, offline) | ✅ 0 findings (13 suppressed = actions/* ref-pin policy) |

## Items to Confirm / Review

- **Zero runtime impact** — CI-only. No app/server/client code changed; default behavior is untouched.
- **Smoke-stub rewrite** (`ci.yml`): please sanity-check the quoted-heredoc form. I verified it functionally (both `claude --version` and a bare invocation behave identically to the old `printf`), but it's the one behavioral line in the diff.
- **zizmor online audits**: I verified locally with `--offline` (no token). CI runs it with `GITHUB_TOKEN` for the few online ref-resolution audits — same config mulmoclaude runs in CI, so low risk, but the PR's own `workflow-lint` run is the real proof.
- **secret-scan triggers**: kept mulmoclaude's `pull_request` (unfiltered) + `push: [main]`. Every PR is scanned regardless of target branch; pushes are scanned on main.

## User Prompt

> 他、mulmoclaudeで使っているもので問題なく取り込めるものは進めて
> (From MulmoClaude, find and adopt whatever can be brought in without problems.)

This is the first, clearly-safe batch. Deferred as follow-ups (need a real run / would be disruptive): a Windows/multi-OS CI matrix (tests may need path fixes first) and `eslint-plugin-import` (would flag existing code).

## Source

Adopted from `receptron/mulmoclaude` at 2026-07-05. Tool versions and pinned SHAs copied verbatim; refresh them alongside any upstream bump.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,11 +6,19 @@ on:
   pull_request:
     branches: [main, dev_tool]
 
+# Least privilege: every job here only reads the checked-out repo
+# (lint / build / test / pack-and-boot). No pushes, no PR/issue writes.
+permissions:
+  contents: read
+
 jobs:
   lint-and-build:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v6
+        with:
+          # Read-only CI; don't persist the token in .git/config (zizmor artipacked).
+          persist-credentials: false
 
       - uses: actions/setup-node@v6
         with:
@@ -42,6 +50,9 @@ jobs:
     timeout-minutes: 15
     steps:
       - uses: actions/checkout@v6
+        with:
+          # Read-only CI; don't persist the token in .git/config (zizmor artipacked).
+          persist-credentials: false
 
       - uses: actions/setup-node@v6
         with:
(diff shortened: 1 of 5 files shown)


isamu:
### Bot triage — all green ✅

- **Codex**: `CODEX VERDICT: LGTM` — no findings (CI/workflow-only scope). The earlier red `codex-review` was a spurious **10-min `timeout-minutes` cancellation** (the codex CLI got throttled mid-inference on Azure; "Terminate orphan process: pid (codex)"), not a verdict and unrelated to this diff. Re-ran the job → passed in 30s.
- **Socket Security**: pass (both jobs).
- No inline review comments were posted by any bot; nothing to address or skip.

The two new jobs (`secret-scan`, `workflow-lint`) and the hardened `ci.yml` all passed in real CI, matching the local runs at the pinned tool versions.

isamu:
Closing: this was CI/dev-infra, but the intended scope is Docker-sandbox-specific adoption from mulmoclaude. Reorienting there instead.

PR #219, opened 2026-07-05
feat(remote-view): inline image-field thumbnails in mobile views

snakajima:
## Summary

First of the two deferred image features. A mobile view's declared `imageFields` were coming back as raw workspace paths — unrenderable on the phone (it can't reach the host) and in the sandboxed preview (its CSP forbids `localhost`) — so they counted as `omitted`. Now the host inlines them as downscaled JPEG `data:` URLs.

## What changed

- **`server/backends/thumbnailStore.ts`** (port of MulmoClaude's): `sharp` decode → downscale to the view's `imageMaxEdge` (EXIF-rotated, never enlarged) → JPEG `data:` URL, with an mtime-keyed LRU cache so repeated pages / "load more" scrolls don't re-decode. The workspace root comes from the configured collection host (`getWorkspaceRoot`); reads are containment-guarded (`containedPath`) against `..` escapes.
- **`remoteView.ts`**: swapped the `noThumbnails` stub for the real resolver, so **both** the remote-host channel (`getRemoteViewItems` + the mutate-update response) and the **desktop phone-frame preview** now show thumbnails.
- **`sharp@^0.35.3`**, dynamic-imported so non-thumbnail code paths never load the native binary.

## Verification

- `thumbnailStore.spec` (injected stub resize, no native binary): data-URL output, missing / path-escape / empty → `null`, and the mtime cache (a repeat doesn't re-decode).
- Smoke with **real sharp**: a 200×120 PNG → a 64×38 JPEG `data:` URL (aspect preserved, capped to `maxEdge`).
- `yarn typecheck` / `lint` / `build` green; **649 tests pass** (+3).

Next: chat image attachments from the phone (`ingestAttachments`) — the second deferred feature.

### DIFF
--- a/package.json
+++ b/package.json
@@ -80,6 +80,7 @@
     "material-symbols": "^0.45.5",
     "node-pty": "^1.1.0",
     "puppeteer": "^25.3.0",
+    "sharp": "^0.35.3",
     "socket.io": "^4.8.3",
     "socket.io-client": "^4.8.3",
     "tsx": "^4.23.0",
--- a/package.json
+++ b/package.json
@@ -58,7 +58,7 @@
     "@mulmochat-plugin/ui-image": "^0.4.0",
     "@mulmoclaude/accounting-plugin": "^0.3.2",
     "@mulmoclaude/chart-plugin": "^0.1.3",
-    "@mulmoclaude/collection-plugin": "^0.7.0",
+    "@mulmoclaude/collection-plugin": "^0.7.1",
     "@mulmoclaude/core": "^0.9.0",
     "@mulmoclaude/form-plugin": "^0.1.4",
     "@mulmoclaude/html-plugin": "^0.2.5",
--- a/server/backends/collections.spec.ts
+++ b/server/backends/collections.spec.ts
@@ -1,6 +1,7 @@
 // @vitest-environment node
 import { describe, it, expect, beforeAll, afterAll, beforeEach, vi } from "vitest";
 import express from "express";
+import sharp from "sharp";
 import { mkdtempSync, mkdirSync, writeFileSync } from "node:fs";
 import { tmpdir } from "node:os";
 import path from "node:path";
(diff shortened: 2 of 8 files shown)


snakajima:
Pushed `bb923e8` — completes the desktop-preview half.

The original commit only inlined thumbnails for the **phone** (mulmoserver calls `getRemoteViewItems`, which the host inlines). The **desktop phone-frame preview** stayed broken: collection-plugin@0.7.0's preview paged the already-loaded raw detail items (CSP-blocked paths) because it had no `fetchRemoteViewItems`.

Now that **collection-plugin@0.7.1** ships `fetchRemoteViewItems`, this wires the host side:
- bump `@mulmoclaude/collection-plugin` → `^0.7.1`
- `collectionUi.ts`: add the `fetchRemoteViewItems` binding
- `collections.ts`: add `GET /api/collections/:slug/remote-view/:viewId/items`, reusing the same `remoteViewItems` builder the channel handler uses — so preview === phone.

New test: a dedicated `photoscol` fixture (mobile view with `imageFields` + a real image record) drives the route and asserts the image field comes back as a `data:image/jpeg` thumbnail (real sharp) + a 404 for an unknown view. Typecheck / lint / build green; 651 tests pass.

⚠️ This now depends on `@mulmoclaude/collection-plugin@0.7.1` being published (it is).

snakajima:
Addressed both review threads in `34a08c9` / `bb923e8`:

- **Symlink containment (P2, `thumbnailStore.ts`)** — fixed. The resolver now runs `realContainedWithin` on the lexically-contained path, resolving symlinks and re-checking containment before `stat`/`readFile`. A symlink under the workspace whose target escapes it (e.g. `data/link.png -> /etc/passwd`) now resolves to `null` instead of being inlined. Added a test that creates exactly that symlink and asserts `null`.

- **"Route preview pages through thumbnail inlining" (codex, on `remoteView.ts:267`)** — this was flagged against the earlier commit `97d8c3f`, before the desktop-preview fix. `bb923e8` addresses it directly: bumped to `@mulmoclaude/collection-plugin@0.7.1` (which added `fetchRemoteViewItems`), wired the `fetchRemoteViewItems` binding + the `GET …/remote-view/:viewId/items` route, so the preview now pages through the same `remoteViewItems` builder (real `data:` thumbnails) the phone channel uses. New `photoscol` route test asserts an inlined thumbnail end-to-end.

652 tests pass; typecheck / lint / build green.

PR #220, opened 2026-07-05
feat(remote-host): chat image attachments from the phone

snakajima:
## Summary

The second deferred image feature: the phone can now **attach photos / PDFs to a chat**. Full-res bytes go to Firebase Storage (the 1 MiB Firestore command doc can't carry them); the phone sends only `storage_id`s on `startChat`, and the host — signed in as the same user — ingests them.

## What changed (all under `server/backends/remoteHost/`)

- **`firebase.ts`**: expose `storage` (dropped in phase 1).
- **`attachmentStore.ts`**: a lean `saveAttachment` — base64 → `data/attachments/YYYY/MM/<id>.<ext>` (atomic temp+rename), returns the workspace-relative path. (A trimmed port of MulmoClaude's — no hooks/companions/loaders.)
- **`ingestAttachments.ts`**: download each staged object from `users/{uid}/uploads/{storage_id}` → `saveAttachment` → best-effort delete staging → path-only `Attachment[]`. `storage_id` is regex-guarded (no `/` or `..`); the batch rejects on the first hard failure so the remote never gets a chat missing a file.
- **`handlers.ts` `startChat`**: ingest **before** spawning, then reference the saved paths in the seeded prompt.

## The delivery wrinkle

MulmoTerminal drives `claude` through an **interactive PTY** (typed input), which can't carry image *content blocks* the way MulmoClaude's stream-json spawn does. So instead the seeded prompt **names the saved files** and `claude` reads them with its Read tool (its cwd is the workspace):

```
what is in this photo?

Attached file(s) — read them from the workspace:
data/attachments/2026/07/d78a5b38.png
```

Text-only `startChat` is unchanged — `ingest([])` short-circuits before any Storage/auth touch. Result shapes match MulmoClaude's, so the same mulmoserver client drives either host.

## Verification

- `attachmentStore.spec` (real save): partitioned path + exact bytes on disk, `.bin` fallback.
- `ingestAttachments.spec`: download/save/delete order, not-signed-in, malformed `storage_id`, delete-failure tolerance.
- `handlers.spec`: `startChat` ingests + references paths, rejects a malformed `attachments` param.
- **Integration smoke** (real `saveAttachment` inside ingest inside `startChat`): the image landed on disk with correct bytes and the prompt referenced its path.
- `yarn typecheck` / `lint` / `build` green; **654 tests pass** (+8).

> Note: the actual Firebase Storage download requires a real phone upload + the bucket's read/delete rules for `users/{uid}/uploads/*` (same rules MulmoClaude relies on) — that boundary is exercised manually, like the connect flow.

### DIFF
--- a/server/backends/remoteHost/attachmentStore.spec.ts
+++ b/server/backends/remoteHost/attachmentStore.spec.ts
@@ -1,0 +1,34 @@
+// @vitest-environment node
+import { describe, it, expect, afterEach } from "vitest";
+import { mkdtempSync, rmSync, readFileSync, existsSync } from "node:fs";
+import { tmpdir } from "node:os";
+import path from "node:path";
+
+import { createSaveAttachment } from "./attachmentStore.js";
+
+describe("createSaveAttachment", () => {
+  let ws = "";
+  afterEach(() => ws && rmSync(ws, { recursive: true, force: true }));
+
+  it("saves base64 bytes under data/attachments/YYYY/MM and returns the workspace-relative path", async () => {
+    ws = mkdtempSync(path.join(tmpdir(), "mt-att-"));
+    const fixed = new Date(Date.UTC(2026, 6, 5)); // → 2026/07
+    const save = createSaveAttachment(ws, () => fixed);
+    const bytes = Buffer.from([0xde, 0xad, 0xbe, 0xef]);
+
+    const saved = await save(bytes.toString("base64"), "image/png");
+
+    expect(saved.mimeType).toBe("image/png");
+    expect(saved.relativePath).toMatch(/^data\/attachments\/2026\/07\/[0-9a-f]{8}\.png$/);
+    const abs = path.join(ws, saved.relativePath);
+    expect(existsSync(abs)).toBe(true);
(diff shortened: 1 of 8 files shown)


snakajima:
Fixed in `84bf4c2`.

`ingestAttachments` now runs in two phases:
1. **Fetch + save every attachment** into the workspace. A hard failure (bad id, download/save error) throws here — **before any staged object is deleted** — so a remote retry with the same `storage_id`s still finds every upload intact.
2. Only once all bytes are safely in the workspace, best-effort delete the staged objects (a failed delete still just logs + leaves an orphan for the TTL sweep).

Added a test: a mid-batch fetch failure leaves **zero** staged objects deleted.

On the "ideally until spawn succeeds" part: I kept the delete at the end of ingest (right after the workspace saves) rather than deferring past `spawnChat`. Once the bytes are in the workspace they're durable, and `spawnChat` here is a synchronous `spawnClaudePty` — threading the storage paths back out to delete post-spawn would add coupling for a spawn step that doesn't realistically fail after ingest. Happy to revisit if you'd prefer the stricter ordering.

655 tests pass; typecheck / lint / build green.

Issue #236, opened 2026-07-07
feat: multi-agent support — run Codex (then Antigravity) as first-class sessions


isamu:
## Summary

Make MulmoTerminal able to host **non-Claude agent CLIs as first-class sessions** — starting with **OpenAI Codex**, then **Google Antigravity (`agy`)** — not just as a raw TTY. "First-class" means the same experience we have for Claude today: session resume across reload, the single-mode GUI panel (charts / forms / collections / images) driven over MCP, and skill/collection runs.

This is a **tracking issue** for a stacked-PR effort landing on an integration branch (`feat/multi-agent`), merged to `main` at the end.

## Motivation

Today only `claude` is a first-class agent. A generalized launcher already runs arbitrary commands (`codex` works as a raw TTY *now*), but it lacks session tracking, GUI-MCP, and skill runs. We want Codex (and later Antigravity) to be selectable per session/cell and behave like Claude.

## Feasibility (verified against docs)

| Layer | Claude (today) | Codex | Antigravity (`agy`) |
|---|---|---|---|
| **L1** TTY launch / reattach | ✅ | ✅ existing launcher | ✅ |
| **L2** Session resume across reload | server mints id (`--session-id`), `--resume` | id is self-minted → discover from `~/.codex/sessions/**/rollout-*.jsonl` `session_meta.id`, resume `codex resume <id>` (or `-c experimental_resume=<path>`) | `agy --conversation <id>`; **prints resume cmd on exit**; `agy -c` for last |
| **L3** GUI-MCP (single mode) | `--mcp-config … --strict-mcp-config --allowedTools` | `[mcp_servers.x] url=…` + `default_tools_approval_mode="approve"`, injectable per-session via `-c key=value` (TOML). **Broker unchanged.** | `mcp_config.json` `url` server (same JSON shape as Claude) + `--dangerously-skip-permissions` |
| **L4** Skill / collection run | slash-seed + draft-typing (`DRAFT_READY_MARKER=/shift+tab to cycle/`) | codex has skills (`~/.codex/skills/*/SKILL.md`); needs codex-specific marker + seed syntax | `/skills`; needs agy-specific marker + seed |

[...]

PR #238, opened 2026-07-07
feat: Codex first-class sessions (L1+L2 — server + Claude/Codex toggle)

isamu:
## Summary

**PR#1a of the multi-agent stack (tracking issue #236) — the SERVER half of "Codex L1+L2".**
Makes MulmoTerminal able to host **codex as a first-class session** (like Claude): spawn it,
learn its session id, and resume it. Claude's `/ws` path is **untouched**.

New `/ws/codex` endpoint, mirroring the launcher's persistent + reattachable lifecycle, but:
- **Fresh spawn** snapshots `$CODEX_HOME/sessions`, runs `codex`, then discovers the rollout
  file codex creates at startup and reads the **id codex minted for itself** (codex has no
  `--session-id`, so the server can't force one). The id is kept in a `key → rolloutId` map.
- **Reconnect** reattaches a live PTY / surviving tmux session (running codex picked up, no
  resume); otherwise **cold-resumes** a known rollout id via `codex resume <id>`.

The browser-facing session id is a mulmoterminal-minted key (stable across reload); codex's own
rollout id is internal, used only to build `codex resume`.

**Split note:** the client half (agent picker UI + `ConnTarget`/wsUrl wiring + grid-cell
persistence) is **PR#1b**, kept separate because it's UI I want to live-verify once the local
`codex` binary is repaired. This PR is reachable/reviewable via a manual `/ws/codex?cwd=…`
connection and is fully unit-tested at the logic layer.

## Items to Confirm / Review

- **Session model** — browser key ↔ codex rollout id mapping (in-memory), reattach-vs-cold-resume
  decision in `resolveCodexSession`. In-memory map means **cold-resume after a full server
  restart** falls back to fresh (tmux still reattaches a surviving live codex); persisting the
  map is a deliberate follow-up. OK for L2's primary "hot reload" goal?
- **Isolation choice** — codex is a separate `/ws/codex` endpoint (not a branch inside `/ws`), so
  Claude carries zero regression risk. Agree with that over unifying the handlers?
- **NEEDS LIVE VERIFICATION** (local `codex` is currently broken — vendor `ENOENT` after a Node
  upgrade; unit tests + build don't need it, but these do):
  - `codex resume <id>` behavior: does it continue the same rollout or fork to a new id? The code
    handles both (discovery overwrites the mapping on a fork), but the real behavior is unconfirmed.
  - Arg order `codex [--model x] resume <id>` (global flag before subcommand — codex's clap layout).
  - Rollout-file appears fast enough after spawn for discovery (5s timeout, 150ms poll).

## User Prompt

> I want MulmoTerminal to use Codex (then Antigravity), not just Claude — Codex first. Running it
> in a raw TTY isn't enough; I need first-class sessions (resume across reload), the single-mode
> MCP/GUI panel, and skill runs. Merge PR#0 and proceed to PR#1 (Codex L1+L2). I'll review on the PR.

(Full plan + PR breakdown: #236 and `plans/feat-multi-agent-support.md`.)

## What this PR changes

| File | Change |
|---|---|
| `server/codex-args.ts` (+spec) | Pure argv builder: fresh vs `resume <id>`, optional `--model`. |
| `server/codex-session.ts` (+spec) | Snapshot `$CODEX_HOME/sessions`, find the rollout that appears after spawn, parse `session_meta` (id + cwd). |
| `server/agents/{types,claude,codex,registry}.ts` (+spec) | `AgentKind` gains `codex`; `codexAdapter` (CODEX_BIN); `draftReadyMarker` now optional (codex omits it); adapters use `satisfies`. |
| `server/index.ts` | `/ws/codex` endpoint + `spawnCodexPty`/`resolveCodexSession`/discovery wiring; `codexRolloutIds` map; `CODEX_BIN`/`CODEX_MODEL`. Claude path unchanged. |

## Verification

`yarn format` / `yarn lint` (0 errors) / `yarn typecheck:server` / `yarn build` / `yarn test` — all
green (**693 tests**, +18: codex-args 4, codex-session 13, registry codex +1). The two pure modules
are the tested core; the `/ws/codex` handler mirrors the already-proven launcher path. A codex-stub
integration test (like the Claude `/ws` stub in `package-smoke`) is planned for PR#5.

Refs #236

### DIFF
--- a/server/agents/claude.ts
+++ b/server/agents/claude.ts
@@ -1,8 +1,10 @@
 import type { AgentAdapter } from "./types.js";
 
 // "shift+tab to cycle" is the hint Claude paints once its input box is ready for a paste.
-export const claudeAdapter: AgentAdapter = {
+// `satisfies` (not a `: AgentAdapter` annotation) keeps draftReadyMarker's concrete type so
+// callers that rely on it (the draft-injection scanner) don't see it as possibly-undefined.
+export const claudeAdapter = {
   kind: "claude",
   bin: () => process.env.CLAUDE_BIN || "claude",
   draftReadyMarker: /shift\+tab to cycle/,
-};
+} satisfies AgentAdapter;
--- a/server/agents/codex.ts
+++ b/server/agents/codex.ts
@@ -1,0 +1,7 @@
+import type { AgentAdapter } from "./types.js";
+
+// Draft injection isn't wired for codex yet, so it omits draftReadyMarker.
+export const codexAdapter = {
+  kind: "codex",
+  bin: () => process.env.CODEX_BIN || "codex",
+} satisfies AgentAdapter;
--- a/server/agents/registry.spec.ts
+++ b/server/agents/registry.spec.ts
@@ -1,12 +1,19 @@
 import { describe, it, expect, afterEach } from "vitest";
 import { getAgentAdapter } from "./registry.js";
 import { claudeAdapter } from "./claude.js";
+import { codexAdapter } from "./codex.js";
+
+function restoreEnv(key: string, value: string | undefined): void {
+  if (value === undefined) Reflect.deleteProperty(process.env, key);
+  else process.env[key] = value;
+}
 
 describe("agent registry", () => {
-  const originalBin = process.env.CLAUDE_BIN;
+  const originalClaudeBin = process.env.CLAUDE_BIN;
+  const originalCodexBin = process.env.CODEX_BIN;
   afterEach(() => {
-    if (originalBin === undefined) delete process.env.CLAUDE_BIN;
-    else process.env.CLAUDE_BIN = originalBin;
+    restoreEnv("CLAUDE_BIN", originalClaudeBin);
+    restoreEnv("CODEX_BIN", originalCodexBin);
   });
 
   it("defaults to the Claude adapter", () => {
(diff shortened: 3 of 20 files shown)


isamu:
Update: PR now covers the full **L1+L2** — the client landed on this branch, so it's one coherent PR (not split a/b).

**Added since the server half:**
- Client protocol: `buildCodexWsUrl` + `ConnTarget.codex` routed in `connUrl`.
- UI: a **Claude / Codex** toggle in the grid cell's new-session form. Picking Codex launches via `/ws/codex`; the choice persists on the cell (`Cell.agent`) so a reload reconnects as codex.

**Live findings that shaped it (codex 0.142.5):**
- `codex` persists a session's rollout only AFTER the first user turn (like claude's transcript), tracked in `state_5.sqlite` threads + rollout JSONL — NOT at startup. So the initial 5s-at-spawn discovery was replaced with a lifetime watch (`watchForCodexSession`, cwd-disambiguated, stops on reap).
- Arg order `codex --model x resume <id>` confirmed (global flags precede the subcommand; `resume [SESSION_ID]` positional).
- `-c key=value` (dotted TOML) confirmed → unblocks GUI-MCP injection for PR#2.

**Verified end-to-end:** a `/ws/codex` connection spawns codex, streams its TUI, and reports a session id. Interactive resume (chat → reload → continue; cold `codex resume`) is the remaining manual check in-browser.

698 tests green; lint/typecheck/build clean.

isamu:
Addressed the two CHANGES REQUESTED findings in c0b6c22:

1. **Resume-overwrite (index.ts)** — the discovery watcher now runs **only for a fresh spawn**. On resume the rollout id is already known, so the watcher can no longer overwrite it with a mis-attributed concurrent rollout.

2. **Ambiguous attribution (codex-session.ts)** — `pickFreshSession` now attributes a rollout **only when unambiguous**: exactly one fresh rollout, or a *unique* cwd match among several. The old 'newest wins' tiebreak is gone (it was exactly the wrong guess the review flagged). When it can't attribute confidently it returns null, so an unattributable session simply stays unresumable-by-id and a cold reconnect starts fresh — it will never reopen a *different* codex conversation.

**Residual (documented in code):** two codex cells launched in the *identical* cwd before either sends its first turn can't be told apart from the rollout files alone (a snapshot-timing race), so neither gets a cold-resume id — safe degradation, not a wrong resume. A full fix needs a codex-side spawn↔session correlation (its `state_5.sqlite` threads are also written lazily, at first turn); deferred. Note this only affects **cold resume by codex id** — the primary hot-reload continuity (reattach + tmux) doesn't use the id at all.

isamu:
d1d2232 adds a **claimed-rollout set**: once a rollout is mapped to a session key it's excluded from every other watcher (serialized by the single event loop), so a single codex conversation can **never** be cold-resumed from two different keys — the concrete harm from the review.

**Residual, deferred with rationale:** two codex cells in the *identical* cwd, launched before either sends its first turn, still can't be told apart from the rollout files alone (a snapshot-timing race). A true one-to-one guarantee needs a codex-side spawn↔session correlation, but codex 0.142.5 persists BOTH the rollout JSONL and its `state_5.sqlite` thread row **lazily, at the first turn**, with no handle tying either back to our specific PTY spawn — so there's no reliable file/DB signal to key on. This case now degrades to *unresumable-by-id* (a cold reconnect starts fresh), never a wrong resume, and never touches the primary hot-reload continuity (reattach + tmux, which don't use the id). Tracking the full fix in #236 as a follow-up.

Merging L1+L2 on that basis; L3 (GUI-MCP) is next.

PR #303, opened 2026-07-09
feat: live-reload .mulmoterminal.json without filesystem watchers

### LINKED ISSUE #302: feat: live-reload .mulmoterminal.json without filesystem watchers
## 課題

`.mulmoterminal.json` を書き換えても、ブラウザをリロードするまで反映されない。

```ts
// src/composables/useDirConfig.ts
// ...the config is stable for the page's lifetime (changes to the file take
// effect on the next page load — MVP, no live watch).
const cache = new Map<string, Promise<DirConfig>>();
```

`mulmoterminal-config` skill は色を提案 → 適用 → 確認、を繰り返す設計だが、**プレビュー手段が「実際のセルを見る」しかない**ため、毎回の手動リロードがワークフローの律速になっている。

## なぜ fs.watch にしないか

cwd はディスク上に散在するため、開いているターミナルの数だけ watcher が要る（集約できない）。**書いた側が通知する push 型**にする。

## 設計（watcher ゼロ・追加エンドポイント無し）

サーバは既に、全セッションの `PreToolUse/PostToolUse` フックを `/api/hook` で受け取り、`tool_name` と `tool_input`（＝`file_path`）を見ている（`server/index.ts` の `handleToolHook`、ツール履歴の生成に使用中）。**「.mulmoterminal.json が書かれた」ことは既に検知できる。**

1. `server/dir-config.ts` — 純粋関数 `dirConfigWriteTarget(toolName, toolInput)` を追加。書き込み系ツールが `<dir>/.mulmoterminal.json` を書いたなら `<dir>` を返す。
2. `server/index.ts` — `handleToolHook` の `PostToolUse` でそれを見て `pubsub.publish("dir-config", { cwd })`。
3. `src/composables/useDirConfig.ts` — `dir-config` チャネルを購読し、該当 cwd のキャッシュを捨てて再取得、その cwd を見ている全 ref を更新。

クライアント側の反映は既存で足りる：`Terminal.vue` は既に `watch([themeId, () => props.dirTheme, () => props.dirColors], …)` でパレットを貼り替える。

## トレードオフ

検知できるのは Claude のツール経由の書き込みのみ（vim での直接編集は拾わない）。skill のフローはこれで完全にカバーされるので、手編集用の明示エンドポイントは surface を増やすだけなので作らない。

## 受け入れ条件

- skill が設定を書いた瞬間、リロード無しでセルの色・バッジが変わる
- バッチで複数 dir を設定しても全て反映される（Write ごとにフックが飛ぶ）
- ファイル watcher を 1 つも作らない

isamu:
## Summary

Changing a directory's `.mulmoterminal.json` now recolours its terminals **immediately** — no page reload, no server restart, and **not a single filesystem watcher**.

The trick: the server already sees every write. Claude's `PostToolUse` hook reports `tool_name` + `tool_input` (it builds the tool-call timeline today), so a write to `<dir>/.mulmoterminal.json` is already observable. The writer announces the change; nothing polls.

## Items to Confirm / Review

- **The hook is now load-bearing for a second purpose.** `handleToolHook` publishes on `PostToolUse` only — a `PostToolUseFailure` wrote nothing and must not trigger a re-read. Worth a look that this is the right seam.
- **Tool-name allowlist** is `Write` / `Edit` / `MultiEdit`. If Claude renames a tool we silently lose live-reload (graceful — you just reload the page). Gating on `tool_input.file_path` alone would also fire on `Read`, which is why it's an allowlist.
- **Trade-off, deliberate:** a `vim` edit is not detected. That covers the skill flow completely, so `POST /api/dir-config/reload` was **not** added — it's surface for a case we don't have. Say the word if hand-editing turns out to be common.
- **Ref lifetime:** `useDirConfig` registers each `config` ref in a module-level `Map` keyed by cwd and unbinds on `onScopeDispose`, so a closed cell can't keep receiving invalidations.

## User Prompt

> live reload ほしいけど、これって watch が terminal の数だけ増えるよね。skill のあとに呼び出す関数とかで出来ない？

Correct on both counts — working directories are scattered, so a watcher can't be shared across terminals. It turned out we don't even need the skill to call anything: the tool hook already fires on the write.

## Implementation

| File | Change |
|---|---|
| `server/dir-config.ts` | pure, testable `dirConfigWriteTarget(toolName, toolInput) -> dir \| null` |
| `server/index.ts` | on `PostToolUse`, `pubsub.publish("dir-config", { cwd })` |
| `src/composables/useDirConfig.ts` | subscribe once; invalidate that cwd's cache, refetch, push into every bound ref |
| `SKILL.md` | drop the now-stale "tell the user to reload the page" |

No client rendering work was needed — `Terminal.vue` already repaints the xterm palette via `watch([themeId, () => props.dirTheme, () => props.dirColors], …)`.

## Verification

Unit tests cover `dirConfigWriteTarget` (each write tool, relative path, wrong tool, wrong filename, malformed payload).

Then **driven in a real browser**, because unit tests don't prove the pubsub → client → xterm path:

```
STEP1 initial xterm bg: rgb(23, 18, 16)    (= the configured #171210)
STEP3 bg WITHOUT reload: rgb(11, 61, 11)   ✅ live-reloaded
STEP5 second change:     rgb(122, 31, 162) ✅ again
🔍 unrelated file write      -> ignored     ✅
🔍 PostToolUseFailure        -> ignored     ✅
page navigations (0 = never reloaded): 0 | page errors: none
```

(The background lands on `.xterm-scrollable-element`, not `.xterm-viewport` — my first probe read the wrong element and reported a false negative.)

Local: 949 tests pass, typecheck 0, lint 0 errors, build ok.

Plan: `plans/feat-dir-config-live-reload.md`. Closes #302.

### DIFF
--- a/plans/feat-dir-config-live-reload.md
+++ b/plans/feat-dir-config-live-reload.md
@@ -1,0 +1,63 @@
+# feat — live-reload `.mulmoterminal.json` without filesystem watchers
+
+Issue: #302
+
+## Problem
+
+Editing `<cwd>/.mulmoterminal.json` does nothing until the browser reloads:
+
+```ts
+// src/composables/useDirConfig.ts
+// ...the config is stable for the page's lifetime (changes to the file take
+// effect on the next page load — MVP, no live watch).
+const cache = new Map<string, Promise<DirConfig>>();
+```
+
+The `mulmoterminal-config` skill's colour loop is *apply → look at the real cell → adjust* (there is no
+way to preview colour inside the conversation — Claude Code doesn't render ANSI, and a Bash child has
+no controlling terminal). So the manual reload is the rate limiter of the whole flow.
+
+## Why not `fs.watch`
+
+Working directories are scattered across the disk, so a watcher can't be shared: one per open cwd.
+Instead, make the **writer announce the change** — push, not poll.
+
+## Design — zero watchers, no new endpoint
+
+The server *already* sees every write. Claude's `PreToolUse`/`PostToolUse` hooks POST to `/api/hook`
+with `tool_name` and `tool_input` (used today to build the per-session tool-call timeline, see
+`handleToolHook` in `server/index.ts`). A write to `<dir>/.mulmoterminal.json` is therefore already
+observable — no new machinery.
+
+1. **`server/dir-config.ts`** — pure, testable helper:
+   `dirConfigWriteTarget(toolName, toolInput): string | null` → the directory whose config a
+   file-writing tool (`Write` / `Edit` / `MultiEdit`) just wrote, else null.
+2. **`server/index.ts`** — in `handleToolHook`, on `PostToolUse` only (a `PostToolUseFailure` wrote
+   nothing), publish `pubsub.publish("dir-config", { cwd })`.
+3. **`src/composables/useDirConfig.ts`** — subscribe once to the `dir-config` channel; on a message
+   drop that cwd from `cache`, refetch, and push the result into every live `config` ref bound to it.
+
+Nothing else is needed on the client: `Terminal.vue` already re-applies the xterm palette via
+`watch([themeId, () => props.dirTheme, () => props.dirColors], …)`, and the chrome colours are plain
+reactive props.
+
+## Preserved behaviour
+
+- Channel naming follows the existing convention (plain literal, e.g. `"sessions"`), duplicated
+  client-side like `useSessions` / `useAttentionSound` do.
+- `useDirConfig` keeps its shape (`{ config }`), still dedupes one fetch per cwd across cells, and
+  still ignores a stale resolve after a fast directory switch.
+- Refs unregister on scope dispose, so a closed cell leaks nothing.
+
+## Trade-off (accepted)
+
+Only writes made through Claude's tools are detected; editing the file in vim will not live-reload.
+That covers the skill flow completely, so an explicit `POST /api/dir-config/reload` is not worth the
+extra surface. Revisit if hand-editing turns out to be common.
+
+## Verification
+
+- Unit: `dirConfigWriteTarget` — matching write tool + path, wrong tool, wrong filename, non-record
+  input, relative path, `PostToolUseFailure` not published.
+- Live: drive the app, change a directory's palette, and confirm the terminal recolours **without a
+  reload** (read the xterm background via computed style).
--- a/server/dir-config.spec.ts
+++ b/server/dir-config.spec.ts
@@ -2,7 +2,7 @@ import { describe, it, expect } from "vitest";
 import { mkdtempSync, writeFileSync, mkdirSync, rmSync, symlinkSync } from "node:fs";
 import { tmpdir } from "node:os";
 import path from "node:path";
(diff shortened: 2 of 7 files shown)


isamu:
Both findings were real. Fixed.

**1. Relative `file_path` resolved against the wrong base** — `path.resolve(file)` used the *server process* cwd. `dirConfigWriteTarget` now takes the session cwd and resolves against it:

```ts
if (path.isAbsolute(file)) return path.dirname(path.resolve(file));
return sessionCwd ? path.dirname(path.resolve(sessionCwd, file)) : null;
```

Call site passes `ptys.get(sessionId)?.cwd ?? null` (`PtyEntry` carries `cwd`). With **no known session cwd we publish nothing** rather than guess — a wrong publish both invalidates a directory nobody is watching *and* misses the real one, so silence is strictly better. Worth calling out that my own spec had enshrined the bug (`expect(...).toBe(path.resolve("."))`); that test is now replaced by three: relative-with-session-cwd, relative-without (→ null), and absolute-ignores-session-cwd.

**2. `bound` map leaked empty Sets** — `unbind()` deleted the callback but left the key. Now:

```ts
targets.delete(apply);
if (!targets.size) bound.delete(boundCwd);
```

Added `src/composables/useDirConfig.spec.ts` (4 tests): live update on announcement, announcement for an unwatched dir is ignored, unbind-on-dispose leaves **no empty key** (via a test-only `boundDirCount()`), and a cwd switch releases the old directory instead of accumulating.

**Re-verified by driving the app**, since the server path changed:

```
initial:                                  rgb(23, 18, 16)
after absolute-path hook:                 rgb(11, 61, 11)  ✅ live recolour
after RELATIVE path, unknown session cwd: rgb(11, 61, 11)  ✅ nothing published (new guard)
restored:                                 rgb(23, 18, 16)
```

Local: 955 tests pass, typecheck 0, lint 0 errors, build ok.

isamu:
Real bug, fixed — and it was worse than reported: **the same race exists on the mount path**, not just in `invalidateDirConfig`. A cached in-flight promise from mount can resolve *after* an invalidation already applied a newer read, clobbering it. The `cwd.value === c` check doesn't catch that, because the cwd never changed.

Both paths now gate on a per-cwd generation counter:

```ts
const seq = generationOf(cwd) + 1;
generation.set(cwd, seq);
fetchDirConfig(cwd).then((config) => {
  if (generationOf(cwd) !== seq) return; // a newer invalidation superseded this response
  targets.forEach((apply) => apply(config));
});
```

and on mount:

```ts
const seq = generationOf(c);
const resolved = await fetchDirConfig(c);
if (cwd.value === c && generationOf(c) === seq) config.value = resolved;
```

`unbind()` drops the `generation` key alongside the `bound` key, so I didn't reintroduce the unbounded-key-growth bug you caught last round.

**The regression test has teeth** — I checked it fails without the guard rather than assuming:

```
plan: mount "first" (0ms) -> invalidate A -> "older" (60ms) -> invalidate B -> "newer" (5ms)

with guard:    5 passed
guard removed: AssertionError: expected 'older' to be 'newer'   ← the exact race
restored:      5 passed
```

Local: 956 tests pass, typecheck 0, lint 0 errors, build ok.

PR #340, opened 2026-07-13
feat: send a Web Push when a background task finishes

### LINKED ISSUE #339: feat: send a Web Push when a task finishes (settings toggle)
## Request

Send a Web Push to the user's registered devices each time a task finishes. The actual delivery server is separate (`mulmoserver`'s `sendPush` Cloud Function) — MulmoTerminal only makes the send call. Gate it with a Settings on/off toggle; for now, fire on every (background) task finish.

Reference: mulmoserver `docs/web-push-sending.md` — `POST .../sendPush`, `Authorization: Bearer <Firebase ID token>`, body `{ data: { title, body } }`; target devices resolved from the signed-in uid.

## Approach

- **Server-side send** (chosen over browser-side so it works even when the browser tab is closed, as long as the machine is up).
- Auth comes from the **RemoteHost** channel's Firebase sign-in (`server/backends/remoteHost/firebase.ts`, project `mulmoserver`) — so push only sends while RemoteHost is connected. `auth.currentUser.getIdToken()` → `sendPush`.
- Trigger: `handleActivityHook` on `event === "Stop" && !active` — one push per finished background turn (matches the attention-chime semantics; avoids the double-publish of a background Stop).
- New `pushEnabled` global config field (`~/.mulmoterminal/config.json`), read live at the hook; a Settings toggle persists it via `POST /api/config`.
- New `server/web-push.ts` calls `sendPush` with an `AbortController` timeout; no-ops (returns null, never throws/fetches) when not signed in.

## Notes

- Prerequisite: **RemoteHost connected** + at least one registered device (else the send is a no-op).
- Actual push delivery needs the user's Google auth + a device, so it's verified manually by the user; the pure/config/no-op paths are unit-tested.

isamu:
## Summary

Opt-in **Web Push on task finish**. When enabled in Settings, the server calls the separate `mulmoserver` `sendPush` Cloud Function each time a **background** task finishes, so the user's phone/devices are notified for the panes they aren't watching. MulmoTerminal only makes the send call — device registration, delivery, and dead-token pruning are the server's job.

- **Server-side send** (chosen over browser-side) so it works even with the browser tab closed, as long as the machine is up.
- **Auth** = the RemoteHost channel's Firebase sign-in (project `mulmoserver`, same as `sendPush`). `auth.currentUser.getIdToken()` → `sendPush`. ⇒ push only fires **while RemoteHost is connected**; otherwise a silent no-op.
- **Trigger** = `handleActivityHook` on `event === "Stop" && !active` — one push per finished background turn (the viewed pane is `active` → skipped, like the attention chime; also avoids the double-publish a background Stop causes on the raw stream).
- **Toggle** = new `pushEnabled` global config field, read live at the hook (no restart), persisted via `POST /api/config`. Mirrors the existing `soundFile` plumbing.
- **New `server/web-push.ts`** calls `sendPush` with an `AbortController` timeout; pure `buildSendPushBody` / `parseSendPushResult`; no-ops (returns null, never throws/fetches) when not signed in.

## Items to Confirm / Review

- **Prerequisite is intentional:** push requires **RemoteHost connected** + at least one registered device. With either missing, the toggle is a no-op. Confirm that's acceptable for v1 (browser-side fallback is a listed follow-up).
- **Trigger scope:** fires on `Stop` for **background** panes only (not the pane you're actively viewing, and not on `Notification`/blocked-on-input). Matches the attention-chime semantics and "notify me about the ones I'm not watching". Confirm this is the intended "タスクが終わるたび".
- **Notification content:** title = project dir basename, body = the last prompt (falls back to "タスクが完了しました"). Tune if you'd prefer the AI summary or a different shape.

## Verification

- **Unit-tested:** `sanitizePushEnabled`; config load/save round-trip incl. `pushEnabled`; `buildSendPushBody` / `parseSendPushResult`; and `sendWebPush` **no-ops without fetching when not signed in**.
- Gates: format / lint (0 errors) / typecheck / build / test (**1070**).
- **Manual (you):** actual push delivery needs your Google sign-in (RemoteHost connected) + a registered device — please verify a real push arrives after connecting RemoteHost and enabling the toggle.

## User Prompt

> [mulmoserver docs/web-push-sending.md] これ、web pushの実装方法。サーバは別途あるのでここでは送信だけしたい。送信するかどうかは、設定画面でon/offしたい。とりあえず通知は、タスクが終わるたびに送りたい。できる?

Plan: `plans/feat-web-push-on-task-finish.md`. Closes #339.

### DIFF
--- a/README.md
+++ b/README.md
@@ -302,6 +302,7 @@ The Settings modal (⚙) persists per-user UI choices to `~/.mulmoterminal/confi
 | `userMcpServers` | `{ id, url }` HTTP MCP servers merged into the **single-view** Claude session's `--mcp-config` (a `localhost` URL is reached over `host.docker.internal` in the Docker sandbox). Takes effect on the next session. |
 | `buttons`    | Header action buttons — see [Header buttons](#header-buttons). Omit to keep the defaults; set to replace them. |
 | `chips`      | Header info chips. Omit to keep the default set; `[]` hides all built-ins. |
+| `pushEnabled` | `true` to send a **Web Push** to your registered devices when a background task finishes. Off by default; only sends while the **RemoteHost** channel is connected (see below). |
 
 #### Header buttons
 
@@ -324,6 +325,14 @@ concerns. To use your own sound, set `soundFile` in Settings (Browse / Test / Us
 chime) or in the config file; the server streams that file at `GET /api/sound` and
 the client decodes it (falling back to the chime if it's missing or not audio). It's
 your own local file referenced by absolute path — nothing is added to the package.
+
+**Web Push on task finish.** Enable `pushEnabled` in Settings to have the server send a
+push (title = the project dir, body = the last prompt) to your registered devices each
+time a **background** task finishes — the same signal as the attention chime, but for the
+panes you're not watching. Delivery is handled by the separate `mulmoserver` `sendPush`
+Cloud Function; MulmoTerminal only makes the call, and only while the **RemoteHost**
+channel is connected (its Google sign-in supplies the notification auth). With RemoteHost
+disconnected, or with no device registered, the toggle is a no-op.
 
 ### Per-directory settings (`<project>/.mulmoterminal.json`)
 
--- a/plans/feat-web-push-on-task-finish.md
+++ b/plans/feat-web-push-on-task-finish.md
@@ -1,0 +1,55 @@
+# feat: Web Push on task finish (#339)
+
+## Goal
+
+Send a Web Push to the user's registered devices when a background task finishes,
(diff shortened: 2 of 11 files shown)


isamu:
### Codex review round 1 対応

**指摘（妥当）:** hidden background worker / translation worker も `/api/hook` を `active=false` で通るため、`notifyTaskFinished` が実タスクと無関係な Web Push を出してしまう。

**対応:** `notifyTaskFinished` の先頭で `hiddenSessions` / `translationWorkerIds` に属するセッションを除外（実ユーザーセッションのみ push）。

全ゲート green（typecheck 0 / lint 0 / build ✓ / test 1079）。

Issue #452, opened 2026-07-20
refactor: remove jscpd-reported duplicate code


isamu:
`jscpd` (min 5 lines / 50 tokens) over `src server common plugins test` reports **173 clones**, 1666 duplicated lines (2.43%). 79 are in production code.

Remove the duplication that represents a real missing abstraction; leave alone clones that are merely coincidental shape.

Plan: [`plans/refactor-jscpd-duplicates.md`](plans/refactor-jscpd-duplicates.md)

Shipped as several small PRs, one per cluster:

- [ ] `server/backends/collections.ts` (107 dup lines)
- [ ] `CommandCell.vue` ↔ `LauncherCell.vue` (105 CSS + 13 HTML + 11 TS)
- [ ] `RunMenu.vue` ↔ `SkillMenu.vue` (50 CSS + 25 TS)
- [ ] `GuiPanel.vue` ↔ `ToolsPane.vue`, `NotificationBell.vue` ↔ `RemoteHostControl.vue`
- [ ] `server/config/dir-config.ts` ↔ `src/composables/useDirConfig.ts`
- [ ] `server/index.ts` (65 dup lines)

Out of scope: the 94 `*.spec.ts` clones (test duplication is deliberate — each test should read top-to-bottom), and sub-15-line clones between unrelated modules.

Constraint: pure refactor, no behavior change. Baseline to hold: 131 test files / 1438 tests.

isamu:
All six clusters are up as separate PRs.

| PR | Cluster | Status |
|---|---|---|
| #453 | `server/backends/collections.ts` | ✅ merged |
| #455 | `CommandCell` ↔ `LauncherCell` chrome | ✅ merged |
| #454 | `RunMenu` ↔ `SkillMenu` dropdown | open |
| #456 | `GuiPanel`/`ToolsPane` + `NotificationBell`/`RemoteHostControl` | open — **stacked on #454** |
| #460 | `dir-config.ts` ↔ `useDirConfig.ts` | open |
| #461 | `server/index.ts` PTY socket plumbing | open |

## Combined effect

Measured by merging all six branches together and re-running jscpd:

| | before | after |
|---|---|---|
| production clones | 79 | **43** |
| production duplicated lines | 896 | **484** (−46%) |
| all clones (incl. specs) | 173 | 142 |

Verified on the merged state: `yarn test` **131 files / 1438 tests**, `eslint .` **0 errors**.

## Note on the spec-clone count

[...]

PR #466, opened 2026-07-20
chore: let no-unused-vars ignore rest siblings

isamu:
## Summary

Adds `ignoreRestSiblings: true` to `@typescript-eslint/no-unused-vars`.

`const { secret, ...rest } = obj` is the idiomatic way to drop a field by construction — the named siblings exist precisely to be *excluded* from `rest`, not used. Without this option, the rule flags them, and the only two escapes are re-listing every remaining field by hand or an `eslint-disable`.

This is the small lint-policy follow-up noted on #452. It unblocks the cleaner form of `publicDirConfig()` (which currently spells out all 11 fields to avoid the error), though that cleanup is left for a separate PR.

## Verification

- **Confirmed the rule actually changes behavior**, tested against the exact case that motivated it:
  ```ts
  const { secret, other, ...rest } = source;
  ```
  → **2 errors** on the current `main` config, **0** with this change.
- `yarn lint` on the full tree: **0 errors** (the 4 pre-existing warnings are unrelated).

No source code changed — config only.

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -47,6 +47,9 @@ export default [
       "max-depth": ["error", 4],
       "max-params": ["warn", 6],
       "max-nested-callbacks": ["error", 4],
+      // `const { secret, ...rest } = obj` is how you drop a field by construction —
+      // the named siblings are the point, not dead code.
+      "@typescript-eslint/no-unused-vars": ["error", { ignoreRestSiblings: true }],
     },
   },
   {
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -47,8 +47,14 @@ export default [
       "max-depth": ["error", 4],
       "max-params": ["warn", 6],
       "max-nested-callbacks": ["error", 4],
-      // `const { secret, ...rest } = obj` is how you drop a field by construction —
-      // the named siblings are the point, not dead code.
+    },
+  },
+  {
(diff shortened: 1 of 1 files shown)


isamu:
Addressed in the latest commit — scoped `@typescript-eslint/no-unused-vars` to `**/*.{ts,tsx,mts,cts}` + `**/*.vue`, so it no longer redeclares a TS rule on plain `.js`.

One correction on the diagnosis, though: I checked `eslint --print-config` before changing anything, and **base `no-unused-vars` is already `off` (0)** on `.js` files — `tseslint.configs.strict` disables it wherever it enables its own. So the base-vs-extension overlap this review predicted doesn't actually occur; there was no duplicate `no-unused-vars` pair from my change.

The duplication that *does* exist is `@typescript-eslint/no-unused-vars` + **`sonarjs/no-unused-vars`**, and it's pre-existing on `main` (both `sonarjs.configs.recommended` and `tseslint.configs.strict` are applied unscoped) — my one-line change didn't introduce it, and it's orthogonal to unused-vars scoping. Happy to open a separate PR to silence one of that pair if you'd like, but I kept it out of this one.

The scoping you asked for is still the right call regardless — declaring the TS rule with the other TS rules is cleaner — so I made that change. Verified: `.ts`/`.vue` now resolve to `ignoreRestSiblings: true`, `.js` falls back to the plugin default.

PR #471, opened 2026-07-20
refactor: share the theme-id list across the build boundary

isamu:
## Summary

The four theme ids were declared **twice**: `server/config/config-schema.ts` had `THEME_IDS = [...] as const` (feeding `z.enum`), and the client `src/composables/useTheme.ts` hardcoded the same union as a type. Adding a theme meant editing both, and nothing flagged it if you forgot one.

Both now derive from `common/themeIds.ts` — the server enumerates it in `z.enum`, the client types and validates against it. Same pattern as `THEME_COLOR_KEYS`, which already lives in `common/` for exactly this reason.

This is the `ThemeId` drift-risk follow-up noted on #452. jscpd doesn't flag it (the list is too short to trip the token threshold), but it's the precise class of client/server divergence `common/` exists to prevent.

## Items to Confirm / Review

- **Proved single-source, didn't just assert it.** I temporarily added a 5th id (`"sunset"`) to `common/themeIds.ts` and checked both sides in one edit:
  - server `tsc` → the `z.enum(THEME_IDS)` schema extended automatically (0 errors)
  - client → `const x: ThemeId = "sunset"` accepted, `"notatheme"` rejected

  So a new theme is now genuinely one edit.
- **Server keeps `type ThemeId = z.infer<typeof themeIdSchema>`** rather than re-exporting the common type. Since `themeIdSchema = z.enum(THEME_IDS)` and `THEME_IDS` is now the shared list, the inferred union equals the common one — and this keeps the server consistent with its sibling types (`ViewTarget`, `RunType` are also `z.infer`). The six server files importing `ThemeId` from `config-schema` are unchanged.
- **`isThemeId` now checks `THEME_IDS.some(...)` instead of `THEMES.some((t) => t.id === value)`.** Behavior is identical (the `THEMES` array's ids are exactly `THEME_IDS`), but it now validates against the shared source rather than the client-only theme table. `THEMES` is still used elsewhere in the file (4 refs), so nothing is orphaned. Used `.some` rather than `.includes` to avoid a cast when narrowing `unknown`.

## User Prompt

> 他は進める  ·  並行して次のrefactoringできる？

(from the #452 follow-up list)

## Verification

- `yarn typecheck` · `yarn typecheck:server` · `yarn typecheck:test` → **0 errors each**
- `yarn test` on config-schema + theme specs → **15 passed**; full suite unaffected
- `eslint` clean; `prettier` clean

### DIFF
--- a/common/themeIds.ts
+++ b/common/themeIds.ts
@@ -1,0 +1,6 @@
+// The app's theme ids, shared across the build boundary: the server's zod schema
+// (config-schema.ts) enumerates them and the client (useTheme.ts) types + validates
+// against them. Kept in common/ so adding a theme is one edit, not two that can
+// drift — same reason THEME_COLOR_KEYS lives here (see themeColors.ts).
+export const THEME_IDS = ["midnight", "nord", "daylight", "solarized"] as const;
+export type ThemeId = (typeof THEME_IDS)[number];
--- a/server/config/config-schema.ts
+++ b/server/config/config-schema.ts
@@ -13,10 +13,10 @@
 import { z } from "zod";
 // Shared with the client dir-config parser so the two can't drift — see common/themeColors.ts.
 import { THEME_COLOR_KEYS } from "../../common/themeColors.js";
+import { THEME_IDS } from "../../common/themeIds.js";
 
 // ---- shared constants ---------------------------------------------------------------------
 
-export const THEME_IDS = ["midnight", "nord", "daylight", "solarized"] as const;
 export const VIEW_TARGETS = ["diff", "prs", "wiki", "collections", "accounting"] as const;
 export const RUN_TYPES = ["shell", "input", "open"] as const;
 export const BUILTIN_CHIPS = ["dir", "git", "ctx", "usage", "status", "diff", "tools"] as const;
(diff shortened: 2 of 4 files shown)


isamu:
Good catch — addressed in the latest commit.

You're right: switching `isThemeId` to `THEME_IDS` widened what it accepts. If an id were added to `common/themeIds.ts` without a matching `THEMES` entry, the client would accept it, set `data-theme`, and then silently fall back to `THEMES[0]` for the terminal palette. Reverted `isThemeId` to validate against `THEMES` so an id is only usable when it actually has a palette object.

Kept `THEME_IDS` as the shared **type + server-enum** source (the drift risk this PR targets), and added the invariant you suggested — `test/src/composables/useTheme.spec.ts` asserts `THEMES.map(t => t.id)` equals `[...THEME_IDS]` in order. So the two can't drift in either direction: a `THEME_IDS` entry without a theme object, or a `THEMES` id not in the list, now fails CI.

Verified the guard fires: adding `"sunset"` to `THEME_IDS` alone makes the lockstep test fail (and `isThemeId("sunset")` returns `false`).

PR #495, opened 2026-07-20
feat: enable Tailwind for the app + map theme tokens + styling guide

isamu:
## Summary

Tailwind was only ever built for the **plugin Shadow DOM** (`plugin-tailwind.css`, injected `?inline`) — the app's own components had **no Tailwind available** and used scoped `<style>` everywhere. This enables Tailwind for the app, mapped onto the existing theme tokens, and converts two buttons as a proof of concept.

**Why:** less CSS, and — the real pain point — **markup that carries its own styling**, so copy-pasting or extracting a component never leaves CSS behind (no orphaned scoped rules, no `<style scoped src>` move-omissions).

## What's here

- **`src/tailwind.css`** — imports Tailwind **theme + utilities only, NOT preflight** (the app keeps its own reset in `style.css`; adding Tailwind's base reset would restyle every existing component). Maps the design tokens via `@theme inline`, so utilities resolve to the same `var(--…)` and **theme-switch across all four themes**: `bg-elevated`, `text-secondary`, `hover:bg-hover`, `border-border`, `text-fg`, `bg-accent`, `font-sans`/`font-mono`, …
- **`main.ts`** loads it globally.
- **PoC:** `TerminalCell`'s `.cell-dir-pick` + `.wt-reuse` buttons converted from scoped CSS to utilities.
- **`docs/styling.md`** — the rules + gotchas.

## Items to Confirm / Review

- **This is a foundational front-end change.** Please sanity-check the no-preflight decision: I import `tailwindcss/theme.css` + `tailwindcss/utilities.css` but **not** `preflight.css`, so nothing existing is restyled. Verified the built CSS has **zero** universal-reset rules (`::file-selector-button` etc.) — only the harmless `--tw-*` property declarations.
- **Pixel-identity was verified statically, not just by eye.** Every utility I used resolves to the byte-identical original declaration in the built CSS — e.g. `.bg-elevated{background-color:var(--bg-elevated)}`, `.rounded-md{border-radius:var(--radius-md)}` (=6px), `.px-2.5{…* 2.5}` (=10px). Screenshot across all four themes attached below.
- **One real gotcha the PoC surfaced (now in the doc):** Tailwind's `text-xs` bundles a `line-height`; the original set only `font-size: 12px`. I used `text-[12px]` (font-size only) to stay pixel-identical. Worth knowing before anyone reaches for `text-sm`/`text-xs` in a migration.
- **The "CSS move omission" hazard, demonstrated:** removing `.cell-dir-pick` orphaned a `.cell-dir-pick .material-symbols-outlined { font-size: 18px }` descendant rule. Converted it to `text-[18px]` on the icon element itself — exactly the failure mode utilities prevent.
- **Test selectors changed.** The 3 specs that selected `.cell-dir-pick` / `.wt-reuse` (removed classes) now use `[aria-label=…]` and a new `data-testid="worktree-reuse"` (data-testid is already used in the repo — `PinToggle.vue`). This is the "don't couple tests to styling classes" rule.
- **`@theme inline` naming** drops the property prefix (`bg-elevated` not `bg-bg-elevated`; `text-fg` for `--text`; `border-border` per the shadcn convention). If you'd prefer different names, now's the cheap time to change them.

## Migration is incremental

Only these two buttons are converted. The rest of the app keeps its scoped CSS and migrates component-by-component as they're touched, per `docs/styling.md`. The variant-heavy buttons (`.btn`+`.btn-primary`, `.ccx-btn`+`.ccx-keep/remove`) need the base→utilities + variant handling described in the doc — deliberately left for follow-up.

## On the `~/.claude/CLAUDE.md` pointer

`docs/styling.md` lives in the repo (the Tailwind setup is repo-specific). I did **not** touch your global `~/.claude/CLAUDE.md` — a "prefer Tailwind" rule there would apply to *every* project, including ones that don't use it. A project-scoped pointer (repo `AGENTS.md`, or a line in the README) is the right home; happy to add one. Flagging it as your call.

## Verification

- `vue-tsc -b --force` → 0 · `yarn typecheck:test` → 0 · `vite build` → succeeds
- `vitest run` → **1476 passed** (incl. the 3 re-selectored TerminalCell specs)
- `eslint` clean · `prettier` clean
- built-CSS static equivalence + 4-theme screenshot

### DIFF
--- a/docs/styling.md
+++ b/docs/styling.md
@@ -1,0 +1,84 @@
+# Styling: Tailwind-first, tokens shared, CSS minimal
+
+The app is styled with **Tailwind utility classes** driven by the theme's design
+tokens. Prefer utilities on the element over per-component `<style scoped>`. The
+goal: less CSS, and markup that carries its own styling so copy-pasting or
+extracting a component never leaves CSS behind.
+
+## How it's wired
+
+- **`src/style.css`** — the design tokens. Four themes (`midnight`/`nord`/`daylight`/`solarized`)
+  each redefine the same `--bg-*` / `--text-*` / `--border` / `--accent` / status
+  vars under `:root[data-theme="…"]`. Switching `data-theme` on `<html>` recolors
+  everything. This is the single source of color truth.
+- **`src/tailwind.css`** — enables Tailwind for the app and maps those tokens into
+  Tailwind's namespace with `@theme inline`, so the utilities resolve to the same
+  vars and theme-switch for free. It pulls in **theme + utilities only — no preflight**:
+  the app has its own reset in `style.css`, and adding Tailwind's base reset would
+  restyle every existing component. (The plugin Shadow DOM has a separate Tailwind
+  build — `src/plugin-tailwind.css` — don't confuse the two.)
+- **`src/main.ts`** imports both, globally.
+
+## The utilities you get
+
+Names drop the CSS-property prefix so they read naturally:
+
+| token (`style.css`) | utilities |
+|---|---|
+| `--bg-base` / `-deep` / `-panel` / `-subtle` / `-elevated` / `-input` | `bg-base`, `bg-elevated`, … |
+| `--bg-hover` / `--bg-selected` | `hover:bg-hover`, `bg-selected` |
+| `--border` | `border-border` |
+| `--accent` (`--accent-bg`, `--on-accent`) | `bg-accent`, `text-accent`, `border-accent`, `bg-accent-bg`, `text-on-accent` |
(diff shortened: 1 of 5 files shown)


isamu:
**Visual + static proof:** https://claude.ai/code/artifact/4167ff4a-63ac-4503-9da8-0484f3614b57

The two converted buttons rendered across all four themes (they clearly re-color per theme), plus the table showing each Tailwind utility resolves to the byte-identical original CSS declaration. (The missing panel backgrounds on 3 quadrants are a harness quirk — the tokens are `:root[data-theme]`-scoped — not part of the change.)

PR #505, opened 2026-07-20
refactor: convert RunMenu/SkillMenu to Tailwind, drop headerMenu.css

isamu:
## Summary

Sixth Tailwind migration. `RunMenu` and `SkillMenu` shared `headerMenu.css` via `<style scoped src>`. Both now carry utilities directly and **`headerMenu.css` is deleted**.

## Items to Confirm / Review

- **The `.active` state became the `aria-expanded:` variant.** The trigger already sets `:aria-expanded="open"`, so `aria-expanded:bg-hover aria-expanded:text-fg` reproduces `.menu-trigger.active` — and the `:class="{ active: open }"` binding is gone. Verified the built CSS: `[aria-expanded=true]{background-color:var(--bg-hover)}`.
- **Specs select by ARIA, not the removed classes** — `[aria-haspopup="menu"]` (trigger), `[role="menu"]` (panel), `[role="menuitem"]` (items). These roles were already on the elements. Applied the #504 rule: grepped the whole repo; only these two specs referenced the classes.
- **Arbitrary values** for dimensions only, per the doc: `top-[calc(100%+4px)]`, `min-w-[180px]`, `shadow-[0_6px_20px_rgba(0,0,0,0.35)]`. Colors stay tokens (`bg-base`, `bg-panel`, `text-secondary`, `hover:bg-hover`).
- `text-[12px]` (not `text-xs`) on both, to avoid the bundled line-height; the trigger adds `leading-none` (its original `line-height: 1`), the item inherits.

## Verification

- `vue-tsc -b --force` → 0 · `eslint`/`prettier` clean
- `vitest` RunMenu + SkillMenu specs → 12 passed; full suite → **1483 passed**

Follows `docs/styling.md`.


isamu:
Addressed in #507 — the sed-based selector rename collapsed `.menu-trigger` and `.menu-root` into the same `[aria-haspopup="menu"]` check, duplicating the empty-state assertion. #507 restores the second condition as `[role="menu"]` absent. Good catch.

PR #587, opened 2026-07-22
feat: pick a provider/model when launching a session (#584)

### LINKED ISSUE #584: feat(ui): ターミナル起動時にプロバイダ／モデルを選ぶ + 設定ヘルプ + ドキュメント
#579 / #580 でディレクトリ単位のプロバイダ指定が入りました。その上に UI と導線を載せます。

## やること

1. **ターミナル起動時にプロバイダ／モデルを選べる**（使えるプロバイダがあるときだけ出す）
2. **モデルは preset から選ぶ**。preset は**リポジトリ側で組み込んでメンテ**し、ユーザーは設定で追加できる
3. **未設定のときは、何をどこに設定すればよいか分かるヘルプ**をクリックで出す
4. **ドキュメント**（`docs/guide/{ja,en}`）
5. **設定支援スキル**（`mulmoterminal-config`）を拡張し、LLM がプロバイダ／モデル設定を書けるようにする

## モデル一覧は動的取得しない — 実測に基づく判断

当初 OpenRouter の API から動的に取る案を検討し、実際に叩いて確かめました:

| URL | 結果 |
|---|---|
| `https://openrouter.ai/api/v1/models` | **200**（認証不要・342 モデル・530 KB） |
| `https://api.moonshot.ai/anthropic/v1/models` | **404** |
| `https://api.moonshot.ai/v1/models` | 401（別パスに存在） |

つまり **カタログの URL は `ANTHROPIC_BASE_URL` から導出できません**。OpenRouter と LiteLLM がたまたま `${baseUrl}/v1/models` で当たるだけで、Moonshot は Anthropic 互換の `/anthropic` の下にカタログを持ちません。

動的取得にすると、プロバイダごとの `modelsUrl`、キャッシュ、失効時の表示、530 KB の射影……と付帯物が増えます。**組み込み preset ＋ ユーザー追加**にすれば、それが全部消えます。

## preset の置き場所

**`common/modelPresets.ts`** に置きます。`common/` は既にクライアントとサーバの共有境界（`themeIds.ts` / `dirChrome.ts` と同じ役割）なので、

- **UI のピッカー**がそのまま読める
- **設定支援スキル**が「いま使える preset」の唯一の出典として読める（既存スキルが `palettes.json` を使っているのと同じ関係）

という 1 ファイル共有になります。ユーザー追加分は config の `providers[].models` で、preset とマージして表示します。

## preset は「動く」と主張せず、実測値を持つ

モデルを実際に走らせて分かったこと: **`-p "say routed"` が通っても、ツールを使うタスクで壊れるモデルがある**。しかも**単発では判定できません** — 同じモデルが試行間で PASS / FAIL を行き来しました。

そこで preset の各エントリに、**どう測ったか**を持たせます:

```jsonc
{
  "provider": "openrouter",
  "id": "moonshotai/kimi-k2.6",
  "label": "Kimi K2.6",
  "contextLength": 262144,
  // 主張ではなく実測: ツールを使うタスクを N 回中 何回完了したか
  "toolTrials": { "passed": 3, "of": 3, "measuredAt": "2026-07-22" }
}
```

UI のピッカーはこれを出すので、ユーザーは**選ぶ前に信頼度が分かります**。

### 測り方

`input.txt` を読み、大文字にして `output.txt` に書き、その語を答える — **Read と Write の両方を必要とするタスク**を、本番と同じ経路（settings ファイルの env ブロック → `claude -p`）で実行し、`output.txt` が実際に書かれたかで判定します。

実測で落ちた例:

- `meta-llama/llama-4-maverick` — **0/4**。ツール呼び出しを**テキストとして出力**する（`Read(file_path=...)` と書くだけで呼ばない）
- `mistralai/devstral-2512` — **ERROR ×3**、5 秒で即死
- `deepseek/deepseek-v4-flash` / `openai/gpt-oss-120b` — **3/4**（1 回だけ失敗）

registry (`models.yaml`) 側の注記とも整合します。あちらは `provider.order` のピン留め・`reasoning_exclude`・`max_tokens_cap` で安定させていますが、**それらは OpenRouter のリクエストボディのパラメータで Claude Code からは送れません**。だから我々の経路では別途測る必要があります。

## 選択の寿命

**そのセッションだけ**にします。`.mulmoterminal.json` の `provider` / `model` は「既定値」で、起動時に選ばなければそれが使われます。

理由: `.mulmoterminal.json` は**ユーザーが手で管理しているファイル**（名前・色・テーマ・スキル許可リストが同居）で、mulmoterminal が書き戻すとコメントの消失・整形の変化・複数インスタンスの同時書き込みを扱うことになります。まず安全側で出し、必要なら「このディレクトリの既定にする」を後から足せます。

## ヘルプ

プロバイダが 1 つも設定されていない、または鍵が解決できないときに、起動フォームから開けるヘルプを出します。内容は**そのとき足りていないものだけ**を具体的に:

- `~/.mulmoterminal/config.json` に `providers` を足す例（コピー可能）
- 鍵はサーバの環境変数（`.env`）に置くこと、**設定ファイルには書かない**こと
- `.mulmoterminal.json` に既定を書く例
- 「LLM に設定してもらう」導線（`mulmoterminal-config` スキル）

`resolveProvider` は既に理由つきで拒否するので（「`OPENROUTER_API_KEY` が要る」等）、その文言をヘルプに繋げます。

## 対象外

- OpenRouter 以外のプロバイダを preset に載せるかは、まず OpenRouter で形を固めてから
- Docker sandbox 対応（#579 で明示的に拒否したまま）

isamu:
## Summary

#579/#580 landed the ability to run a session on an Anthropic-compatible backend, but the
only way to reach it was hand-editing a directory's `.mulmoterminal.json`. This adds the
way in: **choose at launch, see what is running, and find out what to configure when
nothing is.**

- **A curated, measured model list** — `common/modelPresets.ts`, 27 models. Each one was
  actually launched and made to complete a real task, and the entry records how many
  attempts succeeded.
- **A MODEL select in the launch form**, shown only when a provider is genuinely reachable.
  The pick lasts for that session; the directory's default is never rewritten.
- **The header's model badge now knows provider models** — name and a real `ctx %` instead
  of an id tail with no percentage.
- **Help** that leads with the exact sentence the spawn would have refused with.
- **Docs** (ja + en + README) and the `mulmoterminal-config` skill.

Closes #584.

## Items to Confirm / Review

1. **No visual pass was done.** Browser tooling wasn't available in the session that wrote
   this, so the picker and the help dialog are covered by component tests only
   (`test/src/components/ModelPicker.spec.ts`) — not by looking at them. Please eyeball the
   launch form and the modal, especially on a narrow cell and on mobile.
2. **`spawnClaudePty`'s tail arguments became an options object.** It was already at the
   7-parameter lint cap, so the launch choice had nowhere to go; a side-channel keyed by
   session id read worse than the refactor. Six call sites changed — all mechanical, but
   it is the riskiest diff here.
3. **A 0/4 model is kept in the list.** `meta-llama/llama-4-maverick` reaches fine and never
   once called a tool. It is listed, sorted to the bottom, and labelled
   `0/4 — never used a tool`, on the same reasoning as the unreachable entries: answering
   "what about X?" with a measurement beats answering it with silence. Say the word and it
   goes.
4. **The measurements are from one account, one day (2026-07-22).** Three models are marked
   `not reachable from this account` — that is an OpenRouter **privacy setting**, not a
   defect, and another account may run them fine. The wording tries to make that clear;
   worth a second read.
5. **`provider` / `model` were missing from the writable dir-config schema** — a gap from
   #579, not something this PR introduced. The config skill generates its JSON Schema from
   it, so it could not write the two keys the runtime already honoured. Fixed here with a
   regression test.

## User Prompt

> Claude Code 起動時に model を指定できるようにしたい。Claude Code で OpenRouter や
> Moonshot の API（Anthropic 互換）を利用して、それらのモデルを使ったセッションを起動でき
> るようにしたい。環境変数の指定・保存方法を含めて、どうすれば安全か、Docker 利用時も考慮
> して考えてほしい。

> `.mulmoterminal.json` で default を指定できるとして、terminal を起動するときに指定でき
> るようにしてほしい。これは OpenRouter が有効な場合ね。で、モデルも preset で多くのモデル
> から選べるようにしてほしい。

> それらが設定されないときは、設定方法がわかるようにヘルプを用意して、クリックすれば何を
> どこに設定すればよいかわかるようにしてほしい。

> ヘルプも必要だけど document も必要。

> やっぱり動的にすると問題がありそうなので、こっちで preset を組み込んでメンテする。その
> うえでユーザが明示的に設定で追加できるようにしよう。その設定も LLM でサポートすれば良い。

> header の 1 行目の方に、モデル名を出せるようにしてほしい。でも、全部テストしてね。動くか
> どうか。skill で動かさないときちんと動くかわからないよ。

## Why the numbers, and why they are ratios

A model that answers `-p "say routed"` correctly can still be unusable: the reply comes
back while the tool loop never fires — one model printed `Read(file_path=…)` as prose
instead of calling the tool. So the probe is a task that **cannot** be completed by
talking: read `input.txt`, write the word uppercased into `output.txt`, and the verdict is
whether that file exists with the right contents.

It is also not a single-shot judgement. Two models flipped between pass and fail across
runs, so each was tried 3–4 times and the preset records the ratio. The picker shows it:

```
Kimi K2.7 Code · 3/3 · 14s · 262k
Llama 4 Maverick · 0/4 — never used a tool · 1M
Qwen3.7 Plus · not reachable from this account · 1M
```

`scripts/model-trials.ts` is the harness, committed so the list can be re-measured:

```bash
yarn tsx scripts/model-trials.ts --provider openrouter --trials 3 <model-id>
```

Prices and context lengths are read from the provider's catalog, not estimated.

## Why not fetch the catalog dynamically

It was tried first. `${baseUrl}/v1/models` works for OpenRouter but **the catalog URL is
not derivable from `ANTHROPIC_BASE_URL`** — Moonshot's Anthropic-compatible `/anthropic`
has nothing under it (measured: 404). Supporting it properly means a per-provider
`modelsUrl`, a cache, and a staleness story, all to fetch a list whose useful content is
the part a catalog does not carry: whether the model can actually drive Claude Code.

## How the choice travels

`?provider=` / `?model=` on the existing `/ws` query, beside `cwd` and `gui`. Both values
end up in `claude --model` argv and `ANTHROPIC_MODEL`, so they are validated rather than
trusted (`launchChoiceFromParams`): shaped like a model id, length-capped, no leading dash
argv would read as a flag, no whitespace or control characters.

The launch choice replaces the directory's default **as a pair** (`effectiveChoice`) —
taking the picker's model while keeping the directory's provider would aim one vendor's id
at another vendor's endpoint.

## Verification

- 2006 tests / 164 files pass; lint 0 errors; `typecheck`, `typecheck:server`,
  `typecheck:test`, and `build` all clean (after merging `origin/main`).
- **End-to-end against a running server**, not just unit tests: opening `/ws` with
  `?provider=openrouter&model=z-ai/glm-5.2` produced a session whose settings carry
  `ANTHROPIC_BASE_URL=https://openrouter.ai/api` and `ANTHROPIC_MODEL=z-ai/glm-5.2`, with
  `--model z-ai/glm-5.2` in argv and **the token in a 0600 file rather than argv**. The
  same socket with no params produced a plain Anthropic session with no provider
  environment at all.
- `GET /api/launch-options` was checked for token leakage: the payload names
  `OPENROUTER_API_KEY` and never contains its value.
- New tests were mutation-checked — removing the param validation turns 7 red, and breaking
  the picker's "only when reachable" guard turns 3 red.

## Not in scope

- **Writing the pick back to `.mulmoterminal.json`.** That file is hand-maintained and
  holds colours, themes and skills; having the app rewrite it is a separate decision.
- **The Docker sandbox.** Still refused outright: a container inherits no environment, so
  such a session would silently run against Anthropic instead of the chosen backend.

### DIFF
--- a/README.md
+++ b/README.md
@@ -225,6 +225,17 @@ today — **Claude Code** (the default) and **Codex**.
 cell's launch form and the Collections browser carry a **Claude / Codex** toggle (your
 choice is remembered).
 
+**Other models.**
+Claude Code can run against any **Anthropic-compatible** backend (OpenRouter, Moonshot, a
+LiteLLM gateway). Backends are listed in `~/.mulmoterminal/config.json` under `providers`,
+and their **keys are read from the server's environment** — never from a file the app
+serves. A directory sets its default in `.mulmoterminal.json` (`provider` / `model`), and
+each grid cell's launch form has a **MODEL** select that overrides it for one session,
+listing ~27 curated models with the measured pass rate of a real tool-using task beside
+each. A provider whose token can't be resolved **refuses to start** rather than falling
+back to Anthropic, and providers can't be combined with the Docker sandbox. See
+[Configuration → providers](https://receptron.github.io/mulmoterminal/guide/en/config.html#providers).
+
 **Skills for Codex.** Codex has no `/<slug>` slash commands, so on session setup
 MulmoTerminal **mirrors the workspace's `.claude/skills` into `~/.codex/skills`** (each
 mirrored directory carries a `.mt-mirror` marker so a re-sync overwrites what MulmoTerminal
@@ -668,8 +679,10 @@ Each grid cell's header shows two badges for its session, refreshed when a turn 
 
 - **Context badge** — e.g. `Opus · ctx 35%`: the model family plus how full its context
   window is (the *last* turn's input + cache tokens ÷ the model's window — **1M** for
-  current-gen Opus / Sonnet / Fable, **200k** otherwise; an unknown model shows the label
-  with no %).
+  current-gen Opus / Sonnet / Fable, **200k** otherwise). A session running on a
+  [provider model](#agents-claude--codex) shows that model's name and its published window
+  (`Kimi K2.7 Code · ctx 12%`); a model in neither list keeps the label and hides the %,
+  since the window is never guessed.
 - **Token badge** — `⇡<in> ⇣<out>`: cumulative input (fresh + cache-read + cache-creation)
   and output tokens for the session, k/M-formatted, with a full breakdown in the tooltip.
 
@@ -938,8 +951,9 @@ same-origin-guarded.
 
 | Endpoint | Purpose |
 | -------- | ------- |
-| `GET\|POST /api/config` | User UI config (`cwdPresets`, `soundFile`, `prRepos`, `launchers`, `userMcpServers`). |
+| `GET\|POST /api/config` | User UI config (`cwdPresets`, `soundFile`, `prRepos`, `launchers`, `userMcpServers`, `providers`). |
 | `GET /api/sound` · `/api/dir-sound?cwd=` · `/api/dir-config?cwd=` | Custom / per-directory attention sound + per-dir config. |
+| `GET /api/launch-options` | The Anthropic-compatible backends this server can reach, each with its models and — when it can't — the reason. Reports the **name** of the env var a key is read from, never the key. |
 | `GET /api/notifications`(`/history`) · `POST /api/notifications/:id/clear` | Notification feed. |
 | `POST /api/transcribe`(`/model`…) | Voice-input transcription (Whisper, macOS). |
 | `POST /api/translation` | Runtime UI-string translation. |
--- a/common/modelIds.ts
+++ b/common/modelIds.ts
@@ -1,0 +1,17 @@
+// The shape a provider id or model id is allowed to take.
+//
+// Shared because the same string is checked in three places that must agree: the config
+// schema that accepts it, the launch picker that offers it, and the ws query that carries
+// it to `claude --model` and ANTHROPIC_MODEL. When those disagree, a config-accepted id is
+// dropped at launch — and a dropped *provider* with its model kept would start the session
+// on Anthropic instead, which is exactly the silent-wrong-backend this feature exists to
+// prevent.
+//
+// Vendor ids in the wild: `moonshotai/kimi-k2.7-code`, `gpt-5.6-luna-pro`, `z-ai/glm-5.2`.
+export const MODEL_ID_MAX_LENGTH = 120;
+
+// No leading dash (argv would read it as another flag), no whitespace, no control
+// characters, and no `|` — the picker joins provider and model with it.
+const MODEL_ID_RE = /^[A-Za-z0-9][A-Za-z0-9._:/-]*$/;
+
+export const isUsableModelId = (value: string): boolean => value.length <= MODEL_ID_MAX_LENGTH && MODEL_ID_RE.test(value);
--- a/common/modelPresets.ts
+++ b/common/modelPresets.ts
@@ -1,0 +1,311 @@
+// The models MulmoTerminal offers when launching a session on an Anthropic-compatible
+// backend, and how well each one actually drove Claude Code when it was measured.
+//
+// Shared across the build boundary (like themeIds / dirChrome): the launch picker reads
+// it in the browser, and the mulmoterminal-config skill reads it to know what it may
+// suggest. One list, one source of truth.
+//
+// WHY THE NUMBERS: a model that answers a plain prompt can still be unusable, because the
+// reply arrives while the tool loop never fires. Measuring that needs a task no model can
+// complete by talking — so `scripts/model-trials.ts` makes each one read a file and write
+// another, through the real spawn path, and records how many attempts actually produced
+// the file. Two models flipped between pass and fail across runs, so this is a ratio, not
+// a boolean, and the picker shows it. Re-measure with that script when adding entries.
+
+export interface ModelTrialsMeasured {
+  status: "measured";
+  // Attempts that completed a tool-using task, out of attempts made.
+  passed: number;
+  of: number;
+  // Typical wall-clock for that task, over the attempts that passed. The spread is wide
+  // and worth showing: the same probe ran in 11s on one model and 69s on another. Null
+  // when nothing passed — a model can be reachable, answer in prose, and never once call
+  // a tool, and that is a different thing from being slow.
+  medianSeconds: number | null;
+  measuredAt: string;
+}
+
+// Could not be reached from the account that ran the measurement — NOT a defect in the
+// model. OpenRouter answers 404 "No endpoints available matching your guardrail
+// restrictions and data policy" when the account's privacy settings exclude every
+// provider serving it. Kept in the list, clearly marked, because another account (or a
+// settings change) may run it fine.
+export interface ModelTrialsUnreachable {
+  status: "unreachable";
+  reason: string;
+  measuredAt: string;
+}
+
+// A model the USER added in their own config. We have no numbers for it — saying so is
+// better than implying it was checked.
+export interface ModelTrialsUnmeasured {
+  status: "unmeasured";
+}
+
+export type ModelTrials = ModelTrialsMeasured | ModelTrialsUnreachable | ModelTrialsUnmeasured;
+
+export interface ModelPreset {
+  // The `id` of a provider in config.json's `providers`.
+  provider: string;
+  // Passed to `claude --model`, verbatim.
+  id: string;
+  label: string;
+  contextLength: number;
+  // US dollars per million tokens, as published by the provider's catalog. Shown because
+  // the range across this list is two orders of magnitude.
+  pricePerMTok: { input: number; output: number };
+  trials: ModelTrials;
+}
+
+const MEASURED_AT = "2026-07-22";
+
+const measured = (passed: number, of: number, medianSeconds: number | null): ModelTrialsMeasured => ({
+  status: "measured",
+  passed,
+  of,
+  medianSeconds,
+  measuredAt: MEASURED_AT,
+});
+
+const unreachable = (reason: string): ModelTrialsUnreachable => ({ status: "unreachable", reason, measuredAt: MEASURED_AT });
+
+const ACCOUNT_PRIVACY = "OpenRouter returned no endpoints for this account's privacy settings — see openrouter.ai/settings/privacy";
+
+export const MODEL_PRESETS: readonly ModelPreset[] = [
+  // ── Open models: cheapest first ──────────────────────────────────────────────
+  {
+    provider: "openrouter",
+    id: "nvidia/nemotron-3-super-120b-a12b",
+    label: "Nemotron 3 Super 120B",
+    contextLength: 1_000_000,
+    pricePerMTok: { input: 0.08, output: 0.45 },
+    trials: measured(3, 3, 18),
+  },
+  {
+    provider: "openrouter",
+    id: "qwen/qwen3-235b-a22b-2507",
+    label: "Qwen3 235B A22B",
+    contextLength: 262_144,
+    pricePerMTok: { input: 0.09, output: 0.55 },
+    trials: measured(3, 3, 16),
+  },
+  {
+    provider: "openrouter",
+    id: "minimax/minimax-m2.7",
+    label: "MiniMax M2.7",
+    contextLength: 204_800,
+    pricePerMTok: { input: 0.25, output: 1.0 },
+    trials: measured(3, 3, 16),
+  },
+  {
+    provider: "openrouter",
+    id: "deepseek/deepseek-v3.2",
+    label: "DeepSeek V3.2",
+    contextLength: 163_840,
+    pricePerMTok: { input: 0.269, output: 0.4 },
+    trials: measured(3, 3, 42),
+  },
+  {
+    provider: "openrouter",
+    id: "minimax/minimax-m3",
+    label: "MiniMax M3",
+    contextLength: 1_048_576,
+    pricePerMTok: { input: 0.3, output: 1.2 },
+    trials: measured(3, 3, 14),
+  },
+  {
+    provider: "openrouter",
+    id: "deepseek/deepseek-v4-pro",
+    label: "DeepSeek V4 Pro",
+    contextLength: 1_048_576,
+    pricePerMTok: { input: 0.435, output: 0.87 },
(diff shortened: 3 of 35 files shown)


isamu:
## Codex review — both findings applied

Both were real, and the second one is the failure this feature exists to prevent. Fixed in `40ecd1a`.

### 1. A stale pick could reach a resume — MUST-FIX

Confirmed by reading the route: the cell re-sends its choice on every reconnect, and `/ws` passed it to *any* spawn, including one resuming an on-disk transcript. A cell holding an earlier pick could therefore aim a session it did not start at a provider that session never ran on.

A resume now ignores the query entirely and uses **what the session was actually started on** — `registry.launchChoices`, recorded at spawn and dropped in `reap()` — falling back to the directory's default when this server never started it.

That also closes a hole the review did not name and which was worse: before this, a provider session that got resumed **slid back to the directory's default mid-conversation**, silently continuing on a different backend.

Verified against a running server, not just in unit tests:

```
resumed id   : 11111111-…-555555555555 (the on-disk one)
provider env : (none — stayed on Anthropic)     ← stale pick did not leak

started on   : moonshotai/kimi-k2.7-code
resumed on   : moonshotai/kimi-k2.7-code        ← kept its backend, no params sent
```

### 2. Validation contract mismatch — MUST-FIX

Correct, and the consequence was the sharp part: `providerSchema.id` took any non-empty string while `launchChoiceFromParams` enforced an id shape, so a config-valid provider id could be **dropped in transit while its model survived** — resolving to Anthropic running another vendor's model id.

- Parsing is now **all-or-nothing**: if either half is unusable the whole pick goes, and the directory's own default decides. Keeping the good half is the one outcome that must not happen.
- Both sides share one shape (`common/modelIds.ts`), pinned by a round-trip test: every id `sanitizeProviders` keeps must survive the ws query.
- A malformed entry in a provider's `models` is dropped on its own rather than taking the list with it.

### Checks

2019 tests pass (164 files), lint 0 errors, `typecheck` / `typecheck:server` / `typecheck:test` / `build` clean, after merging `origin/main` (#588). Both new safety rules were mutation-checked — 4 tests go red when either is reverted.

The **visual pass is still outstanding** (item 1 in the PR description): browser tooling was unavailable, so the picker and help dialog are covered by component tests only.

isamu:
## Codex review, iteration 2 — applied

### `POST /api/config` could silently wipe `providers` — MUST-FIX

Confirmed, and it is worse than a validation gap: the merge reads *present* as *replace*, and every sanitizer answers a non-array with an empty array. So `{"providers": {}}` was **not** an error — it deleted the user's configured backends and returned `200`.

It predates this PR (#579 added the field without adding it to the guard), but this PR is what makes `providers` load-bearing, so it is fixed here rather than deferred.

- `providers` joins the array-shape guard, so a malformed body gets a `400` like every other list.
- The guards moved to `server/config/config-body.ts` **to be testable at all** — `config-routes.ts` calls `loadAppConfig` at import time, so a test importing it would read the developer's real home config.
- The regression test states the hazard executably instead of describing it: it shows the merge *would* wipe a saved provider, and that the guard rejects that same body. Removing `providers` from the list turns **7 tests red**.

### Merge with main

`origin/main` moved twice during this iteration (#588, #589). #589 introduced `spawnModeFor` in the same lines this PR had changed to the options object — resolved by keeping main's new policy and this PR's call shape.

### Checks

`166 files / 2060 tests` pass, lint 0 errors, `typecheck` / `typecheck:server` / `typecheck:test` / `build` all clean.

One honest note: during this iteration a single test failed once and did not reproduce across two subsequent full-suite runs. I could not identify it from the output and am not claiming it was unrelated — flagging it so a red CI on this SHA is not mistaken for a new regression.

Still outstanding from the PR description: **the visual pass** — browser tooling was unavailable, so the picker and help dialog have component tests but no one has looked at them.

isamu:
## Codex review, iteration 3 — respectfully declining this one

> persist the effective choice for all spawned sessions (not just launch-picked ones)

The concern is legitimate — a conversation should not change backend under itself — but I checked the claim against history before acting on it, and the proposed fix would make things less predictable, not more.

**It is not a regression in this PR.** Before the picker existed, every spawn — resume included — re-read the directory config:

```
2ab00be:server/session/spawn-claude.ts
  56  const dir = loadDirConfig(cwd);
  57  const resolved = requireResolution(resolveProvider({ provider: dir.provider, model: dir.model }, …));
```

So "edit `.mulmoterminal.json`, resume, get the new backend" is #579's behaviour. This PR leaves that path equivalent: for a dir-default session `remembered` is absent, so `effectiveChoice` returns `dir` — the same value the old code passed.

**Three reasons not to change it here:**

1. **The dir file is live by contract.** Theme, colours, the skill allowlist — every field of `.mulmoterminal.json` is read fresh and takes effect. Making `provider`/`model` the only sticky keys would break that file's whole model, and would surprise a user who edits it precisely *because* they want the change to apply.
2. **The memory is in-process.** A sticky dir default would apply before a server restart and not after — two users doing the same thing get different results depending on server uptime. That is strictly less predictable than "the file is live".
3. **The asymmetry is the point.** The picker's choice is remembered because it has **nowhere else to live**: lose it and the session lands on a backend its user never chose *for that session*. The directory's default **is** that backend, so there is nothing to lose.

Also worth noting it is not silent: the header badge shows the running model, so a resumed session that picked up an edited default displays it.

I have recorded this as a decision rather than leaving it to fall out of the code — `79c0f0f` adds the rationale to `launch-choice.ts` and a test that pins the behaviour (`lets a directory's edited default apply on resume`), so a future change here has to be deliberate.

**If you want the stronger guarantee** — a conversation pinned to its backend for life, surviving restarts — that is per-session persistence on disk and a change to #579's contract, and I would rather do it as its own PR than smuggle it in here. Happy to open the issue; say the word.

### Also in this iteration

The two inline comments on the diff are the iteration-1 findings (id-shape mismatch, stale pick on resume) — both already fixed in `40ecd1a`; they predate that push.

CI on `53fc104` was fully green, ubuntu included. `166 files / 2061 tests` pass locally.

isamu:
## Codex review, iteration 4 — applied

### The model list never retried after one failed fetch — MUST-FIX, and mine

Correct, and it is a real dead-end rather than a theoretical one. `inFlight ??= fetchOptions()` was never cleared, so a single failed request was remembered as *already asked*: `options` stayed `EMPTY`, no later mount retried, and the picker stayed hidden for the rest of the page session even once the server was fine. The first attempt is precisely the one that can lose a race with a server still starting up, so this would have shown up as "the model select just isn't there sometimes".

Fixed in `8069e63`:

- Only a **successful** fetch is remembered. A failure leaves the module ready to try again on the next mount.
- `inFlight` is cleared inside `fetchOptions`' own `finally`, not in a `.finally()` on the returned promise, so it is already null by the time anything awaiting the call resumes.
- Concurrent callers still dedupe — a grid mounting a dozen empty cells asks the server once.

New tests in `test/src/composables/useLaunchOptions.spec.ts` cover: the retry after a rejection, the retry after a non-2xx, dedupe across 12 simultaneous mounts, no re-ask after success, and the explicit reload. Restoring the old behaviour turns the two retry tests red.

### Iteration 3 finding — still declining, now confirmed by the author

The "remember the effective choice for every session, not just launch-picked ones" finding was discussed with the repo owner, who confirmed no change is wanted. The reasoning is in the previous comment and recorded in `launch-choice.ts` plus a test that pins the behaviour, so it is a decision rather than an oversight. Short version: the directory default is live by contract like every other `.mulmoterminal.json` field, this PR leaves that path equivalent to #579's, and an in-process-only stickiness would behave differently either side of a server restart.

### Checks

`167 files / 2068 tests` pass, lint 0 errors, `typecheck` / `typecheck:server` / `typecheck:test` / `build` clean. CI on the previous SHA (`53fc104`) was fully green including ubuntu.

Still outstanding: **the visual pass** on the picker and help dialog — component tests only, nobody has looked at them.

isamu:
## Iteration 5 — Codex LGTM taken, one test gap filled

Codex signed off on `8069e63`. Reviewing that fix independently, the logic is right; one case its tests do not reach:

**Added in `e141f6d`** — an empty provider list must count as an answer. `loaded` is the only thing separating "the server says nothing is configured" from "the request failed": both leave `options` at `{ providers: [], anyReady: false }`. Reading the empty one as a failure sends a grid back to the server once per mounted cell, and that is exactly what a fresh install returns. Making the flag depend on the list being non-empty turns this new case red and nothing else, so it is not covered by the existing ones.

## Observed, deliberately not changed

**`reloadLaunchOptions` has no production caller.** Only `useLaunchOptions.spec.ts` and `ModelPicker.spec.ts` import it — the settings screen its comment describes does not exist yet. Left alone here because the two specs use it as their priming hook and rewriting them is churn unrelated to this PR, but it is worth deciding: either the settings-screen caller lands with it, or the export goes and the specs prime via `vi.resetModules()`.

That absence is also why one race is currently unreachable, and it will not stay that way. `fetchOptions` writes `options.value = EMPTY` unconditionally on failure, so a first load still hanging when a reload answers would, on giving up at the 8s timeout, wipe the list the reload had just brought in — and since the failure path leaves `loaded` at `true`, nothing would re-ask. Today nothing starts a second concurrent fetch, so it cannot happen; **whoever wires the settings screen needs a newest-request-wins guard at the same time.** Not adding it now on the reasoning that a guard for an unreachable path is code no one can test against real behaviour.

Checks: `167 files / 2069 tests` pass, lint 0 errors, `typecheck` / `typecheck:test` / `build` clean.

isamu:
Follow-up filed for the deferred finding above: #590 — the directory config's `provider`/`model` are the fourth door of the `isUsableModelId` invariant this PR introduced, and the only one not held to it. Pre-existing since #579, so out of scope here.

PR #594, opened 2026-07-22
fix: validate a directory's model id before it becomes argv (#590)

### LINKED ISSUE #590: fix: .mulmoterminal.json の provider/model だけが id 検証を通らず argv に届く
## 何が起きているか

#587 で `isUsableModelId`（`common/modelIds.ts`）が「同じ文字列を検査する3箇所を一致させる」不変条件として導入された。コメント自身がその3箇所を挙げている:

> the config schema that accepts it, the launch picker that offers it, and the ws query that carries it to `claude --model` and ANTHROPIC_MODEL

しかし**ディレクトリ設定は4つ目の入口**で、ここだけ検査されていない。`server/config/config-schema.ts`:

```ts
export const dirProviderField = z.string().trim().min(1).nullable().catch(null);
export const dirModelField = z.string().trim().min(1).nullable().catch(null);
```

非空文字列なら何でも通る。しかもこれは**明示的にピッカーで選ばない限り常に使われる既定値の経路**で、最もよく通る道。

## 経路（確認済み）

`loadDirConfig(cwd)` → `dir.model` → `effectiveChoice()` → `resolveProvider()` → `buildClaudeArgs()` の `guiArgs.push("--model", input.model)` → argv。

`.mulmoterminal.json` は `<cwd>/.mulmoterminal.json`、つまり**開いたディレクトリから読む**ので、クローンしてきたリポジトリが持ち込める。

## シェル注入ではない

- `spawnPty` は `pty.spawn(bin, args, ...)` の配列渡し
- tmux 経路も `tmuxNewSessionArgs` が `["--", file, ...args]` を組んで同じく配列で渡す

問題はシェルではなく **argv のフラグ混入**。`common/modelIds.ts` が正規表現の先頭を `[A-Za-z0-9]` に縛っている理由がまさにこれで、コメントにもそう書いてある:

> No leading dash (argv would read it as another flag)

`--model` の直後の値を `claude` の引数パーサがどう扱うかに依存させないためのガードなので、依存させている現状を直したい。

## 由来

#579 で `provider`/`model` がディレクトリ設定に入った時点からの既存挙動。#587 のレビュー中に発見。#587 自体は ws クエリ側を正しく検証しているので、その PR の回帰ではない。

## 修正時の注意 — ペア単位の全否定にすること

フィールドごとに独立して `.catch(null)` すると、**provider だけ落ちて model が残る**状態が作れてしまう。これは他社のモデル id を Anthropic のバックエンドで走らせるということで、`server/session/launch-choice.ts` が ws クエリ側で全否定を選んでいる理由そのもの:

> a rejected `provider` whose `model` survives resolves to Anthropic running another vendor's model id, which is the silent wrong-backend this whole feature exists to prevent

`launchChoiceFromParams` と同じ「両方有効か、両方捨てるか」に揃える必要がある。捨てた場合はディレクトリ既定 = Anthropic の既定にフォールバックする。

## テスト

`test/server/config/` に、少なくとも:

- 先頭ダッシュの `model`（例 `--mcp-config=/tmp/x.json`）がペアごと落ちる
- 空白・制御文字・`|` を含む id が落ちる
- `MODEL_ID_MAX_LENGTH` 超過が落ちる
- **provider が不正で model が正当なとき、model も一緒に落ちる**（ペア単位の全否定の回帰テスト）
- 正当なペアはそのまま通り、ディレクトリの他の設定は不正な値があっても読み込みを止めない（現行の lenient な意図の維持）

isamu:
## Summary

`.mulmoterminal.json` was the **fourth** way a provider/model reaches `claude --model`, and the only one without the id check #587 introduced — while being the **most travelled**, since it decides every session that does not use the picker. The file is also read from whatever directory was opened, so a cloned repository brings its own.

Reproduced through the real path (`loadDirConfig` → `effectiveChoice` → `resolveProvider` → `buildClaudeArgs`) **before** writing the fix:

```
先頭ダッシュ   → argv: ["--model","--mcp-config=/tmp/evil.json"]
空白入り       → argv: ["--model","kimi k2 --dangerously-skip-permissions"]
改行入り       → argv: ["--model","kimi\nrm -rf /"]
```

Not shell injection — the spawn is array-based — but the guard exists precisely so the value the CLI ends up reading never depends on how its argument parser treats a leading dash.

Same repro after the fix: all three refuse, with the value quoted in the message.

Closes #590.

## Items to Confirm / Review

1. **Refused, not dropped — deliberately different from what the issue proposed.** The issue asked for pair-wise all-or-nothing (drop both halves). Dropping them would start the session on the **Anthropic default while the directory plainly asked for OpenRouter** — silently running somewhere the user did not choose is the failure this module exists to prevent. It is also not a new contract: a directory naming a provider typo, or omitting its model, is **already** refused rather than downgraded (verified, not assumed). The pair property the issue wanted still holds: a bad provider takes its model with it, because an unusable id matches no configured provider — covered by a test.
2. **The check lives in `resolveProvider`, not in the file parser.** That is the single gate every entry point already passes through, so the ws query, the directory file, and whatever comes next are covered once instead of three times drifting apart. This is the same invariant `common/modelIds.ts` was introduced for — the issue's real point is that the invariant had a fourth member nobody added.
3. **The directory's other settings still load.** The lenient contract for colours/theme/header is unchanged; only the launch is refused. Covered by a test.
4. **Message hygiene.** The offending value is `JSON.stringify`'d into the reason — it reaches a log line and the session's terminal, so a newline must not be able to forge one. The pre-existing unknown-provider message got the same treatment.

## Tests

`test/server/config/dir-model-choice.spec.ts` — 14 cases driving the **real chain** (write the file → load → resolve → build args), not the check in isolation, so they fail if any link stops applying it:

- a usable id reaches the session; a bare model with no provider still means "another Anthropic model"
- leading dash / whitespace+second flag / newline / NUL / `|` are refused
- over `MODEL_ID_MAX_LENGTH` refused, exactly at the limit accepted
- **never falls back to Anthropic when the directory named a provider** (the regression this PR is about)
- **an unusable provider does not leave its model behind** (the pair property)
- the other dir settings still load
- what actually lands in argv

Removing the check turns **7 of them red**.

## Checks

`168 files / 2083 tests` pass, lint 0 errors, `typecheck` / `typecheck:server` / `typecheck:test` / `build` clean.

Docs updated in both languages: the per-project section now states the id rule and that a malformed value refuses the launch rather than quietly running something else.


isamu:
## Iteration 1 — Codex LGTM, but my own review found something it missed

Codex returned `LGTM` and all 9 checks were green. I ran an independent pass anyway, and it turned up a real defect — one that reasoning about the diff would not have surfaced, so I checked the shape against the **live OpenRouter catalog** instead.

### The id shape rejected 10 legitimate model ids

OpenRouter's "always the latest" aliases begin with `~`:

```
~anthropic/claude-opus-latest   ~moonshotai/kimi-latest   ~google/gemini-flash-latest
~anthropic/claude-sonnet-latest ~openai/gpt-latest        ~x-ai/grok-latest        …
```

10 of its 342 ids. The shape introduced in #587 rejected every one of them, so a perfectly valid configuration was unlaunchable — and this PR **makes that visible**, by refusing at the gate instead of quietly passing the value through to `--model`.

`~` is safe where `-` is not: no argument parser reads it as a flag. Fixed in `8e24978`; re-checked against the whole catalog:

```
カタログ 342 件中、弾かれる id: 0
`-` 始まりは依然として拒否: true
空白・改行も拒否: true true
```

Pinned by tests in both the directory path and the ws-query path — reverting the shape turns 4 red. Docs updated in both languages.

### Where this leaves the PR

- **Codex**: LGTM (iteration 1), no inline comments
- **My own review**: one defect found and fixed, described above; no others outstanding
- **Checks**: `168 files / 2087 tests`, lint 0 errors, `typecheck` / `typecheck:server` / `typecheck:test` / `build` clean
- The design decision that differs from the issue (**refuse, don't drop**) is in the PR description with its reasoning, and verified against how a provider typo already behaves

Also worth recording: `dir.model` has exactly **one** consumer (`spawn-claude.ts`), and `publicDirConfig` does not expose it — checked, so the single gate really is the only path.

PR #734, opened 2026-07-24
fix: auto-restart the dev backend on crash so terminals self-heal

snakajima:
Follow-up to #732 — after merging the process guards, the **"all terminals disconnected"** symptom recurred right after a server restart.

## Investigation

- When it happens, the mulmoterminal dev processes (`concurrently`, the `node --watch` parent, `vite`) are **all alive**, but **nothing listens on 34567** — the backend's `index.ts` child crashed during boot and `node --watch` is sitting idle.
- Booting the **exact same code** directly comes up fine → the crash is **intermittent** *and* **escapes the merged guards**. Escaping the guards means it's an **import-time error** (before `installProcessGuards()` runs) or an **explicit `process.exit`** — e.g. an `EADDRINUSE` bind race against a not-yet-dead previous instance on restart.
- Verified empirically that **`node --watch` does not restart on a crash** — it prints `Failed running... Waiting for file changes before restarting` and just waits. That's precisely why the backend stays down and every terminal disconnects for good.

## Fix

Replace `node --watch server/index.ts` with **`scripts/dev-server.mjs`**, a small supervisor that:

- **Restarts the backend on any exit** (crash-loop backoff) — so an import-time crash, a `process.exit`, or an `EADDRINUSE` race all **self-heal in ~1s**. Because sessions are tmux-backed, the fresh backend transparently reattaches every terminal, turning a permanent dead-end into a blip.
- **Restarts on source changes** too, preserving what `node --watch` did. (`node --watch` re-executes the whole process per change anyway, so full restart-on-save costs the same here.)
- Inherits stdio, so the crash stack still prints on the server pane — now preceded by a clear `[dev-server] backend exited (...)` line for diagnosis.

## Test

`test/scripts/dev-server.spec.ts` drives the supervisor with a stub entry that crashes on boot and asserts a **second boot** (i.e. it restarted). `DEV_SERVER_ENTRY` / `DEV_SERVER_WATCH` are test-only overrides.

Manually verified: crash → restart with escalating backoff (500→1000→2000ms), `.ts` change → graceful restart, non-source change → ignored, SIGINT → clean shutdown.

Analysis appended to `plans/fix-backend-crash-disconnects-all-terminals.md` (Step C).

Note: the pre-existing `docPath.spec.ts` failure is unrelated (a local-timezone bug that also fails on clean `main`; not run in CI's path).


snakajima:
Both Codex P2s addressed in b6e3982:

1. **Watch set** — now watches `server/`, `common/`, and `bin/` (the repo dirs the backend imports from: `common/modelIds.ts`, `bin/update-check.js`), matching what `node --watch` covered.
2. **Debounce/crash race** — every (re)start now routes through a single idempotent `scheduleBringUp`/`bringUp` guarded by one `restartTimer`, so a crash landing inside the file-change debounce collapses to one spawn instead of racing two backends onto 34567. A `killedForRestart` flag keeps a deliberate reload from being mis-counted as a crash.

Re-verified manually (crash-restart w/ backoff, reload on .ts, non-source ignored, clean shutdown); existing test still green.

isamu:
Addressed the remaining Codex review (test coverage for the two invariants). The code fixes were already in b6e3982; this adds the tests it asked for.

Extracted the two decisions into `scripts/dev-server-config.js` (behavior in `dev-server.mjs` unchanged) and pinned them:

- **Watch set** — `resolveWatchDirs` unit test asserts the default set is `server/` + `common/` + `bin/` (not only `server/`), so editing `common/modelIds.ts` / `bin/update-check.js` reloads. Plus an integration test that a watched source change actually restarts the backend.
- **Single-spawn guard** — `shouldSchedule` unit test asserts a schedule request is a no-op when a restart is already pending, so an overlapping crash + file-change collapses to one spawn instead of racing two backends onto 34567.

Both invariant tests are mutation-checked (dropping `common/bin` from the set, and dropping the `restartPending` guard, each turn a test red). Also merged latest main. Full suite green; lint/typecheck×3/build green by exit code.

isamu:
レビュー3点は commit 70fdf24（"pin the watch set and single-spawn guard"）で対応済みです。旧コミット e326f8c 宛の bot コメントなので、内容を確認しました。

**1. watch 対象が `server/**` に狭い（dev-server.mjs:25）** → 対応済み。
`scripts/dev-server-config.js` の `resolveWatchDirs` が `["server", "common", "bin"]` を返します。指摘どおりバックエンドは `server/config/config-schema.ts:17 → common/modelIds.ts` と `server/config/update-status.ts:9 → bin/update-check.js` を import しており、その両ディレクトリを watch 対象に含めています。テスト: `dev-server-config.spec.ts` の "watches server/, common/ and bin/ by default (not just server/)"。

**2. debounce 中のクラッシュで2重起動（dev-server.mjs:101）** → 対応済み。
全ての (再)起動を `scheduleBringUp` に一本化し、`shouldSchedule({ restartPending: restartTimer !== null })` で「既に予約済みなら no-op」にしています。クラッシュの exit ハンドラが先に `restartTimer` をセットするため、その後に発火する file-change debounce の `scheduleBringUp` は no-op になり、2つ目のバックエンドは起動しません（EADDRINUSE 回避）。テスト: `dev-server-config.spec.ts` の "skips when a restart is already pending — collapses an overlapping crash + file-change to one spawn"。

**3. 高リスク境界のテスト不足（dev-server.spec.ts:27）** → 対応済み。
(a) watch 対象は `resolveWatchDirs` のユニットテストで固定＋`dev-server.spec.ts` の "restarts the backend when a watched source file changes" で実 supervisor の restart-on-change を検証。(b) 2重起動ガードは `shouldSchedule` のテストで固定（タイミング依存のレースを実プロセスで再現するとフレークになるため、決定ロジックを pure 関数で固定する方針）。

全 dev-server テスト（9件）パス。bot スレッドは resolve して問題ありません。

PR #738, opened 2026-07-24
fix: wheel scrolls the TUI again instead of spinning input history

### LINKED ISSUE #737: fix: 拡大ターミナルでホイールが履歴めくりになる（#729 の回帰）— ホイールはスクロールに戻す
## 症状

expand したターミナル上でマウスホイールを回すと、テキストがスクロールせず**入力履歴が回る**。ホイールはスクロールだけでよい（履歴はキーボードの上下で操作する）。

## 原因（コード確認済み）

#729 / PR #730 で、ドラッグを選択に保つためにマウストラッキングの SET（`CSI ? 1000/1002/1003/1006… h`）を swallow するようにした。その結果:

1. Claude TUI は alt buffer で動作し、以前はトラッキング有効時に**ホイール＝マウスレポート**が TUI に届き、TUI 自身がトランスクリプトをスクロールしていた。
2. swallow 後は xterm から見ると「alt buffer（scrollback なし）+ マウスプロトコル無効」なので、`CoreBrowserTerminal` の wheel ハンドラが**ホイールを ↑/↓ 矢印シーケンスに変換して送信**する（`node_modules/@xterm/xterm/src/browser/CoreBrowserTerminal.ts` の `if (!this.buffer.hasScrollback)` ブロック）。
3. Claude は ↑/↓ を履歴に割り当てているため、ホイール＝履歴めくりになる。

## 修正方針

`term.attachCustomWheelEventHandler` を使い、**「アプリがマウストラッキングを要求していた（swallow 済みで記録）」かつ「alt buffer」**のときだけ、ホイールを SGR ホイールレポート（`CSI < 64/65 ; col ; row M`）として合成して PTY へ送る。

- 通常バッファ・トラッキング未要求時は今までどおり xterm のスクロールバックが動く（return true）。
- swallow 済みモードの記録は SET で追加・RESET（`l`）でクリア。#729 の「RESET は素通し」の設計は維持。
- SGR（1006）を要求していないアプリはスコープ外（現状の矢印変換のまま）。
- 判定・シーケンス生成は純関数に切り出してテスト（#729 の mouseTrackingModes と同じ流儀）。

isamu:
## Summary

Closes #737 — a regression from #729/#730. Scrolling the wheel over a (zoomed) Claude cell was cycling the input history instead of scrolling the transcript.

**Mechanism:** #729 swallows the mouse-tracking DECSETs so drags stay text selections. But that also hides the wheel: to xterm the alt-buffer app now looks mouseless, and its no-scrollback fallback (`CoreBrowserTerminal`, `if (!this.buffer.hasScrollback)`) converts wheel events into ↑/↓ arrow sequences — which Claude's TUI binds to input history. Before #729, the wheel reached the TUI as mouse reports and the TUI scrolled its own transcript.

**Fix:** remember which tracking modes were swallowed (per terminal; SET records, RESET clears and still passes through to xterm as before), and in `attachCustomWheelEventHandler`, when the app is in the **alternate buffer** and had asked for **SGR wheel tracking** (a 1000–1003 mode + 1006), synthesize the SGR wheel report it asked for (`CSI < 64/65 ; 1 ; 1 M`) via `term.input()` — it flows through `onData` to the PTY like a keystroke. Everything else returns `true` so xterm's normal behavior (scrollback scrolling in the normal buffer) is untouched. Drag selection (#729) is unaffected.

The rules are pure functions in `src/composables/wheelReports.ts` (mutation-checked tests); the CSI + wheel wiring is extracted into `guardMouseTracking()` (also resolving the `max-lines-per-function` limit on `ensure()`).

## Items to Confirm / Review

- **Please manually verify in the running app**: expand a Claude cell → wheel should scroll the transcript, not the history; keyboard ↑/↓ still cycles history; text drag still selects; a plain shell cell (normal buffer) still scrolls xterm scrollback. I could not drive a live Claude TUI in this environment — unit tests cover the rules, not the end-to-end feel.
- The synthesized report uses fixed cell coordinates `1;1` — transcript scrolling doesn't depend on pointer position and the position isn't exposed at this layer. Flag if any target TUI scrolls per-pane by coordinate.
- Non-SGR encodings (X10/urxvt) are deliberately out of scope — every current target requests 1006; without it the old arrow-conversion behavior remains.
- One wheel event → one report (matches xterm's own simplified fallback granularity).

## User Prompt

> あーー、別件で、expandしているterminal上でスクロールすると、文字がスクロールするんじゃなくて、入力の履歴が回る。スクロールだけで良い。履歴はキーボードで上下で動くので。

## Implementation notes

- Diagnosis verified against xterm source (`CoreBrowserTerminal.ts` wheel listener: custom handler hook → `hasScrollback` check → arrow sequence) and the #729 swallow in `mouseTrackingModes.ts`.
- `useTerminalConnections.spec.ts`: the Terminal mock gained `attachCustomWheelEventHandler`, and the "DECSET only" wiring assertion now expects the DECRST observer too (it must keep returning `false` — asserted rationale in the comment).
- Plan: `plans/fix-alt-buffer-wheel-scroll.md`. Gates: full suite 3531 passed, lint 0 errors (4 pre-existing warnings), typecheck ×2, build — all green.


isamu:
Addressed the Codex review: the swallowed-mode record lived in `guardMouseTracking`'s closure and survived `connect()`'s `term.reset()`. An alt-buffer app that died without sending DECRST would leave the slot believing the **next** app wanted mouse reports, so that app's wheel would receive synthesized SGR bytes instead of scrolling — the #729 noise one layer over. Confirmed real, not theoretical.

The record now belongs to the connection and is cleared alongside `term.reset()`. Added a reconnect test that drives the real wheel handler through a socket drop: it synthesizes while the app holds 1002+1006, and stops after the session is replaced. Mutation-checked — removing the `clear()` turns it red.

Full suite 3532 passed; lint/typecheck/build green.

PR #866, opened 2026-07-26
feat: configurable terminal font size, globally and per directory (#860)

### LINKED ISSUE #860: feat: ターミナルのフォントサイズを設定できるようにしてほしい
## やりたいこと

ターミナルのフォントサイズを変更したい。

## 現状

`src/composables/useTerminalConnections.ts` で xterm.js の `Terminal` コンストラクタに
`fontSize: 14` が固定値で渡されており、`~/.mulmoterminal/config.json` にも
`<dir>/.mulmoterminal.json` にも該当する設定キーが存在しない。

回避策としてブラウザの全体ズーム（Ctrl+/Ctrl-）を試したが、xterm.js 内部のセルグリッド計算と
ズレて表示が微妙に崩れる（カーソル位置・折り返し位置のズレ）。

## 提案

`.mulmoterminal.json` または `~/.mulmoterminal/config.json` にターミナルのフォントサイズを
指定できる設定キー（例: `fontSize`）を追加する。

isamu:
## Summary

`fontSize` was hardcoded in the xterm constructor. This adds it in two places, following exactly the shape `theme` already has:

| | Global | Per-directory |
|---|---|---|
| Where | Settings modal stepper → `localStorage["terminalFontSize"]` | `.mulmoterminal.json` → `fontSize` |
| Range | 8–32, default 14 | same; out-of-range **clamped**, non-number ignored |
| Precedence | fallback | **wins**, per terminal |

**The part that isn't plumbing:** a size change alters the *cell metrics*, so `cols`/`rows` change and the PTY has to be told. `setFontSize()` re-fits and syncs; `attach()` applies a changed size *before* its own fit. Setting the option alone would reproduce the exact bug that makes browser zoom useless here — xterm's grid and the PTY disagreeing, so the cursor and wrap points drift. `Conn.fontSize` remembers the applied value so a rebuilt terminal (#846) doesn't snap back to 14.

Closes #860.

## Items to Confirm / Review

- **localStorage, not `config.json`.** The issue body proposed `~/.mulmoterminal/config.json`; the user chose "same as color", which is localStorage + Settings UI. It is also the better grain for a *size* — per-display, so a phone and a desktop on one server keep their own, which one shared value can't express. Flagging it because it deviates from the issue text.
- **Clamp vs reject on the lenient path.** `fontSize: 99` becomes 32 rather than falling back to 14, so the setting reads as working rather than ignored. The strict schema (`writableDirConfigSchema`) still rejects out-of-range, so an editor flags it at authoring time. Both behaviours are pinned by tests — worth a sanity check that this is the trade you want.
- **`fontSize` went into `DirChrome`** (`common/dirChrome.ts`), whose stated invariant is "fields whose type is identical on both sides" — `theme`/`colors` are excluded only because each side types them differently, which doesn't apply to a number. All three `DirConfig`s extend it, so one side can no longer omit the field. I retitled the comment accordingly; say the word if you'd rather it were declared per side.
- **Bounds are a judgement call.** 8–32 px: below ~8 the canvas glyphs stop being legible, above ~32 a grid cell holds too few columns to be useful. Neither is an xterm limit.
- **i18n.** `SettingsModal.vue` has no `$t()` and the app wires no `createI18n` (`vue-i18n` is in package.json but unused anywhere in `src/`), so the new labels are plain strings like every other label in that file. This departs from the global "use vue-i18n" rule — raised separately rather than fixed here.

## User Prompt

> https://github.com/receptron/mulmoterminal/issues/860
> 試してみたところ確かにサイズは変わる。色と同様に、global 設定とリポジトリごとの設定ができると良いかも。リポジトリごとが難しいなら一先ず global だけで。

Per-directory turned out **not** to be hard — the `.mulmoterminal.json` plumbing already ran end to end for `theme`/`colors`, so `fontSize` rides the same route. Both are included.

## Verification

Gates: `format`, `lint` (0 errors), `typecheck`, `typecheck:server`, `typecheck:test`, `build`, `test` — **4442 passed** (13 new).

**Driven in the running app** (own instance on a spare port), not just unit-tested:

| Check | Result |
|---|---|
| Stepper | 14 px → +10 → 24 px |
| Ceiling | stops at 32 px, ＋ disabled |
| Persistence | survives reload via localStorage |
| **PTY re-fit** | **67×47 @14px → 38×27 @24px**, one resize frame per press |

The screenshot at 24 px shows Claude Code's own output **re-wrapped to 38 columns** — the process received the new geometry and re-rendered. That is the behaviour browser zoom cannot produce.

Per-directory verified at the wire with `curl /api/dir-config`: `22` → `22`, `999` → `32` (clamped), `"22"` → `null`.

**Not verified:** the dir pin clicked through in the browser. The cwd-picking launcher lives in the grid cell and each attempt spawns a real session, so I stopped rather than spawn more. That path is covered by the wire check above plus 4 parser unit tests; the resolution itself is `props.dirFontSize ?? fontSize.value`, passed at both call sites (single view + grid cell).

## Docs

`useTerminalConnections.ts` carries a header requiring terminal changes to update the docs in the same commit — so: `docs/terminal-notes.md` (the resolution order and the must-re-fit rule), `README.md`, and `docs/guide/{en,ja}/config.md` + `features.md` in both languages. `server/skills/mulmoterminal-config/SKILL.md` gained the key and a trigger for "the text is too small", since that skill writes these files and would otherwise refuse to write a key the runtime honours.

## Summary by CodeRabbit

* **New Features**
  * Added adjustable terminal font size controls in Settings.
  * Added per-directory font size overrides through configuration.
  * Font sizes are validated, rounded, and constrained to supported limits.
  * Terminal layout and cursor alignment now remain synchronized when font size changes.

* **Documentation**
  * Updated English and Japanese guides with configuration details and usage guidance.

* **Tests**
  * Added coverage for font-size validation, overrides, persistence, and configuration handling.

### DIFF
--- a/README.md
+++ b/README.md
@@ -434,7 +434,7 @@ The Settings modal (⚙) persists per-user UI choices to `~/.mulmoterminal/confi
 
 ![The Settings modal — theme, notification sound, PR repos, launch commands, and MCP servers](https://raw.githubusercontent.com/receptron/mulmoterminal/main/docs/guide/images/settings.png)
 
-*Open it from the ⚙ button in the toolbar. Pick a **theme**, set a custom **attention sound**, list the repos the cross-repo **PRs & Issues** view should aggregate, add **launch commands** for grid cells, and register your own **MCP servers** — no need to hand-edit the config file.*
+*Open it from the ⚙ button in the toolbar. Pick a **theme**, set the **terminal font size**, set a custom **attention sound**, list the repos the cross-repo **PRs & Issues** view should aggregate, add **launch commands** for grid cells, and register your own **MCP servers** — no need to hand-edit the config file.*
 
 | Field        | Meaning |
 | ------------ | ------- |
@@ -527,6 +527,7 @@ malformed file is ignored.
   "buttonColor": "#c7cdf0",             // header icon buttons (hex #rrggbb)
   "theme": "nord",                      // terminal palette: midnight | nord | daylight | solarized
   "colors": { "background": "#190a23", "cursor": "#ff2e63" }, // per-key palette overrides
+  "fontSize": 16,                       // terminal font size in px (8–32); overrides Settings
   "sound": "./.mulmoterminal/alert.mp3" // attention sound, RELATIVE to this directory
 }
 ```
@@ -547,6 +548,7 @@ malformed file is ignored.
 | `buttonColor` | Header **icon button** color (`#rrggbb`) — expand / close / attach / folder / etc., across both header rows. |
 | `theme`      | xterm palette for terminals in this directory (one of the built-in theme ids). |
 | `colors`     | Per-key xterm palette overrides applied on top of `theme` (or the app theme when `theme` is unset). Keys are xterm `ITheme` names (`background`, `foreground`, `cursor`, `selectionBackground`, the 16 ANSI colors, …); values are hex (`#rgb` / `#rrggbb` / `#rrggbbaa`). Unknown keys / bad values are dropped. |
+| `fontSize`   | Terminal font size in px for this directory (8–32), overriding the Settings value. A size outside the range is clamped; a non-number is ignored. Changing it re-fits the terminal, so the PTY learns the new width — unlike browser zoom, which leaves the two disagreeing. |
 | `sound`      | Attention sound for this directory's sessions, a path **relative to the directory** (served at `GET /api/dir-sound`). |
 
 **Security.** `sound` is a directory-relative path only — absolute paths and any
--- a/README.md
+++ b/README.md
@@ -434,7 +434,7 @@ The Settings modal (⚙) persists per-user UI choices to `~/.mulmoterminal/confi
 
 ![The Settings modal — theme, notification sound, PR repos, launch commands, and MCP servers](https://raw.githubusercontent.com/receptron/mulmoterminal/main/docs/guide/images/settings.png)
 
-*Open it from the ⚙ button in the toolbar. Pick a **theme**, set the **terminal font size**, set a custom **attention sound**, list the repos the cross-repo **PRs & Issues** view should aggregate, add **launch commands** for grid cells, and register your own **MCP servers** — no need to hand-edit the config file.*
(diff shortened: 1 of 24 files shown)


isamu:
## Bot review triage — iteration 1 (`fde9199` → `fbac2e1`)

| Reviewer | Signal |
|---|---|
| Codex | `CODEX VERDICT: LGTM`, no findings |
| CodeRabbit | 2 inline findings + 1 pre-merge warning |
| Socket Security / knip / jscpd | pass |

### Applied

**1. Blank stored value started the app at 8px** — `useTerminalFontSize.ts`. CodeRabbit was right, and I verified the mechanism before touching anything:

| stored | before | after |
|---|---|---|
| `""` | **8** | 14 |
| `"   "` | **8** | 14 |
| `"abc"` | 14 | 14 |
| `null` | 14 | 14 |

`Number("")` is `0` — a *finite* number — so it clamped to the minimum rather than falling through to the default, while every other unusable value already did. Blank means "unset", so it is now screened out before `Number()`. Covered by a regression test, plus 8 more around startup/persistence/bounds for that composable, which previously had no spec of its own.

**2. README overstated where settings live** — the paragraph implied everything in the ⚙ modal lands in `config.json`. It now says which are per-browser. I named `theme` as well as font size: it was already localStorage-backed and that sentence already covered it, so singling out font size would have left the same wrong impression about theme.

**3. Found during my own review, not flagged by either bot** — the stepper paired a fullwidth `＋` (U+FF0B) with a U+2212 `−`, which renders at inconsistent widths in some fonts. Now `+` / `−`.

### Skipped, with reason

**CodeRabbit pre-merge: "Docstring Coverage 42.86%, threshold 80%."** Declining — this repo's `CLAUDE.md` is explicit in the opposite direction:

> **Default to writing no comments.** … **NEVER explain WHAT the code does** … **Don't write multi-paragraph docstrings or multi-line comment blocks.**

Adding docstrings to clear a generic threshold would violate the project's own stated style. The comments in this diff follow the repo rule instead: they record *why* — why the size must re-fit, why blank is screened before `Number()`, why localStorage rather than `config.json` — and nothing that merely restates the code. Flagging it here rather than disagreeing silently.

### Verification after the fixes

`format`, `lint` (0 errors), `typecheck`, `typecheck:server`, `typecheck:test`, `build` all clean; `yarn test` **4450 passed** (up from 4442).

PR #870, opened 2026-07-26
feat: make the terminal font family configurable, and put CJK faces in the default stack

### LINKED ISSUE #864: feat: ターミナルのフォント(フォントファミリー)を設定できるようにしてほしい
## やりたいこと

ターミナルのフォント（フォントファミリー）を変更したい。

## 現状

`src/composables/useTerminalConnections.ts` で xterm.js の `Terminal` コンストラクタに
`fontFamily: 'JetBrains Mono', 'Fira Code', 'Menlo', monospace` が固定値で渡されており、
`~/.mulmoterminal/config.json` にも `<dir>/.mulmoterminal.json` にも該当する設定キーは
存在しない。

## 提案

`.mulmoterminal.json` または `~/.mulmoterminal/config.json` にターミナルのフォント
ファミリーを指定できる設定キー（例: `fontFamily`）を追加する。

## 関連

フォントサイズについては #860 で同様の要望を起票済み。今回はフォントファミリーの変更に
ついての要望。

isamu:
## Summary

The terminal's font is now configurable, and the built-in stack finally names CJK faces.

- **`fontFamily` in `~/.mulmoterminal/config.json`** — the font every terminal renders in. No
  Settings UI (chosen with the user): edit the file, reload the tab.
- **`fontFamily` in `<dir>/.mulmoterminal.json`** — pins one directory's terminals, winning over
  the global value.
- **The default stack now ends in CJK fonts** (JP → KR → SC → TC). Latin is untouched, so existing
  ASCII rendering is identical; only CJK, which previously fell through to the browser's pick,
  changes.

Closes #864. Follows #866 (`fontSize`, merged yesterday) and reuses its shape.

## Items to Confirm / Review

1. **Not verified in a browser.** I had no browser automation in this session. The whole
   server-side chain *is* verified at runtime (dir pin / dir garbage / global / unset all resolve
   correctly, and the generated `dir-config.schema.json` carries the pattern), and the re-fit
   mechanism is confirmed by reading the installed xterm bundle — `CharSizeService` re-measures on
   exactly `["fontFamily","fontSize"]`, the same pair — but **nobody has watched a terminal
   actually re-wrap after a font change.** Worth one manual check: put a `fontFamily` in
   `config.json`, reload, confirm the glyphs change and the PTY re-fits (no drifting cursor).
2. **`Consolas` is new in the default stack**, and it is a visible change for Windows users who
   have neither JetBrains Mono nor Fira Code — they were falling all the way to Courier New. I
   think this is an improvement; say the word if you'd rather the stack stay strictly additive.
3. **`monospace` is auto-appended** when a user's stack names no generic family. Rationale below —
   it prevents a silently proportional terminal — but it does mean the stored value isn't
   byte-identical to what was typed.
4. **`mergeConfigUpdate` was refactored** (`server/config/app-config.ts`). Adding a 16th field
   pushed it over the `sonarjs/cognitive-complexity` limit, so the repeated
   `body.x !== undefined ? sanitize(body.x) : base.x` ternary became a single `merge()` helper.
   Behaviour is identical — the `!== undefined` semantics (so `chips: []` and `soundFile: null`
   stay real edits) are preserved and still covered by the existing specs.
5. **`setFontSize` → `setFont`**, and `Conn.fontSize` → `Conn.font`. Rationale below; it touches
   code merged yesterday, so worth a look that I didn't undo anything from #866.

## User Prompt

> https://github.com/receptron/mulmoterminal/issues/864 これでの実装と、実装時に日本語などのメジャーなフォントをdefaultに追加して入れたい

Two follow-up decisions, made by the user when asked:

- **Settings UI shape** → *設定ファイルのみ*. No Settings-modal control; the keys live in
  `.mulmoterminal.json` **and** `~/.mulmoterminal/config.json`.
- **Scope of the default font additions** → *CJK 全般 (JP/KR/SC/TC)*, not Japanese alone.

## Where the value lives, and why it differs from `fontSize`

`fontSize` (#860) went to **localStorage + a Settings stepper**, because a size is a per-display
preference — a phone and a desktop viewing one server want their own.

A family is a different kind of thing. It is per-**machine** (which fonts are installed), long and
fiddly to type, and no bounded control fits it. So it is **global config, one value per host**, and
`terminalSubmit` is the precedent for that shape: a `config.json` key with no UI, served by
`GET /api/config`. `fontSize` never needed that half, so this PR adds the `AppConfig` plumbing.

| | Global | Per-directory |
|---|---|---|
| file | `~/.mulmoterminal/config.json` → `fontFamily` | `<dir>/.mulmoterminal.json` → `fontFamily` |
| precedence | fallback | wins |
| unset | `TERMINAL_FONT_FAMILY_DEFAULT` | — |

One wrinkle `terminalSubmit` doesn't have: it is a plain module value because the keydown handler
reads it imperatively and nothing renders it. Hydration here is **async and visible** — a terminal
can mount before `/api/config` resolves — so the holder is a `ref` (like `activeKeymap`), and
`Terminal.vue`'s watcher re-applies the family when it lands.

## Changing the family changes the cell metrics

This is the load-bearing part, and it is the same rule #860 established for the size: a different
face has a different advance width, so `cols`/`rows` change and **the PTY has to be told**. Setting
`term.options.fontFamily` alone would reproduce the browser-zoom bug from #860 — xterm's grid
disagreeing with what the shell believes the window to be, so the cursor and wrap points drift.

Rather than a second copy of that logic, `setFontSize` became **`setFont(key, { size, family })`**
and `Conn.fontSize` became `Conn.font: TerminalFont`. Both fields re-fit through one path, a change
to both now costs one fit instead of two, and `attach()` doesn't grow a seventh positional
parameter.

Confirmed against the installed xterm bundle rather than assumed:

```js
this._register(this._optionsService.onMultipleOptionChange(["fontFamily","fontSize"], () => this.measure()))
```

`CharSizeService` invalidates on either — they are registered as literally the same pair.

## The default stack

```
'JetBrains Mono', 'Fira Code', Menlo, Consolas,
'Noto Sans Mono CJK JP', 'Hiragino Sans', 'BIZ UDGothic', 'MS Gothic', IPAGothic,
'Noto Sans Mono CJK KR', 'Malgun Gothic',
'Noto Sans Mono CJK SC', 'Microsoft YaHei',
'Noto Sans Mono CJK TC', 'Microsoft JhengHei',
monospace
```

Latin first, so ASCII rendering is unchanged. JP before the other CJK locales — a stack reaching
`…CJK SC` first would draw kanji with mainland glyph shapes.

**Why name them at all**, when generic `monospace` already triggers per-glyph fallback? Because the
face the browser picks for that fallback is not required to be em-square, and **xterm reserves
exactly two cells for a fullwidth character**. A face whose fullwidth advance isn't exactly twice
the Latin one tears every box-drawing frame — which is most of what an agent TUI draws. Every
family named above is em-square.

## Validating a font-family string

`normalizeFontFamily(input: unknown): string | null` in `common/terminalFontFamily.ts` — pure, in
`common/` because the server validates config files with it and the client re-validates what comes
off the wire.

- Accepts font names in **any script** — `游ゴシック` and `ＭＳ ゴシック` are real font names, and a
  Japanese user naming the font as their font book shows it must not be told it's invalid.
- **Rejects the whole stack** on any bad entry, where `normalizeFontSize` clamps. A size is a
  continuous quantity, so honouring the direction asked for beats ignoring it; a stack is one
  intent, and silently dropping half of it renders in a font the author never named.
- Rejects CSS syntax (`;{}()<>\/@!`) and unbalanced quotes. An unbalanced quote or stray `;`
  invalidates the whole CSS declaration, and xterm's way of reporting that is to render in the
  browser's *proportional* default — i.e. the setting reads as having broken the terminal.
- **Appends `monospace`** when no generic family is named, for the same reason: three fonts none of
  which are installed otherwise lands on a proportional face with nothing on screen to explain it.

### A finding worth recording

`z.toJSONSchema` **silently drops a `.refine`**. Expressing the exact rule in
`writableDirConfigSchema` would therefore have vanished from the shipped
`dir-config.schema.json` with nothing to show for it. The strict field carries
`TERMINAL_FONT_FAMILY_SAFE_RE` instead — a `pattern`, which survives — and since a JSON Schema
pattern is ECMA-262 *without* the `u` flag it can't use the `\p{L}` classes the exact rule needs.
So it is deliberately a **subset** check, and a spec pins the invariant: it must never reject what
`normalizeFontFamily` accepts, or the config skill would refuse a stack the server would load
happily.

## Tests

`yarn format`, `yarn lint`, `yarn typecheck`, `yarn typecheck:server`, `yarn typecheck:test`,
`yarn build`, `yarn test` — all green (**4473 passed**).

New coverage: `normalizeFontFamily` (scripts, quoting, generic-tail appending, CSS injection,
control characters, length cap, the subset invariant), `dirFontFamilyField`, the generated JSON
Schema, `loadDirConfig`/`publicDirConfig` round-trip, `sanitizeAppConfig`/`mergeConfigUpdate`/
`saveAppConfig` round-trip, `useDirConfig` off-the-wire parsing, and the global ref's hydration and
fallback.

## Docs

`README.md` (both config tables + the example), `docs/guide/{en,ja}/config.md` (a **Terminal font**
section covering how to choose a CJK face and what to do when it doesn't take, plus the per-dir
key), `docs/guide/{en,ja}/features.md`, `docs/terminal-notes.md`, and
`server/skills/mulmoterminal-config/SKILL.md` (including its frontmatter, so the skill triggers on
"Japanese text in the terminal looks wrong").


isamu:
@coderabbitai review

isamu:
@coderabbitai full review

isamu:
## Review round-up

**Bot signals**

| Reviewer | Result |
|---|---|
| Codex auto-review | `LGTM` ×3, latest on the final commit — no findings raised at any point |
| CI / dead-code-scan / duplication-scan | green on every push |
| CodeRabbit | **never reviewed** — blocked by the account's fair-usage limit, not by the code |

CodeRabbit was rate-limited on the first push, then answered both `@coderabbitai review` and
`@coderabbitai full review` with "finished" followed by *"Your next included review will be
available in 50 minutes"*. So there is **no CodeRabbit signal on this PR at all** — please don't
read the absence of comments as a pass.

**Findings from my own review (no bot flagged either)**

Both were in the "looks fine, silently does nothing" category, which is the class worth a human
double-check:

1. **The global key needed a restart, and the docs said the opposite.** `config-routes.ts:31`
   loads `~/.mulmoterminal/config.json` once at boot and serves that cached object; only a
   `POST /api/config` refreshes it. `fontFamily` has no Settings UI, so nothing ever POSTs it — a
   hand-edit does nothing until the server restarts. Both guides said "reload the tab" /
   「再起動は不要」. Fixed in both languages, the README, the troubleshooting lists and the config
   skill. The **per-directory** key genuinely is live (`.mulmoterminal.json` is watched), so the
   two are now documented separately. — 8b88fc1
2. **A quoted generic family isn't a generic family.** `normalizeFontFamily` appends `monospace`
   so a stack of uninstalled fonts can't land on the browser's *proportional* default, but the
   has-a-generic check unquoted each entry first, so `'monospace'` counted. In CSS a generic
   family is a keyword — quoting it asks for a font literally named "monospace", which resolves
   nowhere and gives no fallback. The guard was skipping the append in exactly the case it exists
   for. — 62eab79

**Added after the fact, at the author's request** (4d42cd2): Tailwind's `--font-mono` turned out
to be a *third* hardcoded monospace stack with no CJK either, so the same defect existed in the
header, diff view and inputs. Only the CJK tail was added — the Latin head stays
`ui-monospace, "JetBrains Mono"`, so ASCII chrome renders byte-identically, and the chrome
deliberately does **not** follow the user's `fontFamily` setting.

**Still open for a human — see "Items to Confirm / Review" above**

- **Not verified in a browser.** The server-side chain is verified at runtime and the re-fit
  premise was confirmed by reading xterm's own source (`CharSizeService` re-measures on exactly
  `["fontFamily","fontSize"]`), but nobody has watched a terminal re-wrap after a font change.
- **`Consolas` is new in the default stack** — a real if small change for Windows users who had
  neither JetBrains Mono nor Fira Code.

`yarn format` / `lint` / `typecheck` / `typecheck:server` / `typecheck:test` / `build` / `test`
all pass (4478 tests). Exit codes were checked individually — an earlier `| tail` in my own
command was swallowing them.

PR #875, opened 2026-07-26
Replace every emoji in the UI with Material Symbols

snakajima:
## Why

The Material Symbols convention was set in June but only reached the app chrome. The grid cells still drew their toolbars with emoji and unicode glyphs, so one header row mixed `attach_file` and `folder_open` with a literal folder emoji sitting right beside them.

Noticed while reading a grid cell's toolbar: two Material Symbols, four emoji, and an inline SVG, all in the same cluster.

## What changed

Every button keeps its `title` and `aria-label` — only the glyph changed.

- **`DEFAULT_BUTTONS`** (`server/config/header-config.ts`) — `emoji` to `icon` for reveal / terminal / pr / gh. The `emoji` field stays in the schema for end-user configs and still wins over `icon` when both are set; nothing this repo ships uses it now.
- **Grid cells, all three types** — ask, timeline, move, exchange, expand/restore/close, the diff panel's commit / push / open-PR, re-run, summarize, worktree add/delete. Converting only `TerminalCell` would have left chevrons in one cell and the old `◀▶` in the command cell next to it.
- **Overlays and modals** — files (incl. the tree twisty and folder/file icons), PRs, timeline, settings, cockpit row menu.
- **Run / Skill dropdowns** — carets now flip `expand_more` / `expand_less` with the menu rather than drawing a static triangle.
- **Text that named a button by its emoji** — one UI string ("the paperclip button") and ~24 source comments ("restore via the expand button"), which otherwise pointed at glyphs that no longer exist.
- **`CLAUDE.md`** — new **No emojis** section. Its own changelog convention prescribed a book emoji under each entry heading; that is now a `> **Setup guide:**` line.

## Deliberately not converted

Documented in `CLAUDE.md` so a later sweep doesn't break them:

| Kept | Why |
|---|---|
| `/^\s*[❯›]\s/u` in `screen-rows.ts` | Parses Claude Code's real terminal output |
| The favicon's `❯` chevron | Drawn on canvas as the app's identity mark |
| CLI doctor's `✓ / ✗ / ○` | A terminal can't render an icon font |
| `⎇ main ●3 ↑2`, `●` dots, `−12` counts | Dense notation; icons are bigger and slower to scan |

## Testing

- `yarn test` — 4428 pass. One pre-existing failure, `docPath.spec.ts` "zero-pads the month", which is a timezone bug (`2026-03-01T00:00:00Z` lands on Feb 28 locally) and **also fails on a clean tree** — confirmed by stashing.
- `yarn typecheck`, `yarn typecheck:server`, `yarn typecheck:test` — clean.
- `yarn lint` — 0 errors (6 pre-existing warnings in untouched files).
- Verified in the running app after `yarn build`.

One spec updated: `CellChromeButtons.spec.ts` asserted the literal `⤢`/`⤡` text.

## Follow-up, not in this PR

`docs/`, `plans/`, and `README.md` still hold ~453 emoji, and `docs/ChangeLog.md` has existing `> 📘` lines from the old convention. The new rule implicates them, but sweeping prose is a separate change.

## Summary by CodeRabbit

* **Style**
  * Replaced emoji and text glyphs throughout the interface with consistent Material Symbols icons.
  * Updated terminal, grid, file, settings, timeline, menu, launcher, and action controls with clearer visual indicators.
  * Improved button spacing and alignment for icon-and-label controls.
  * Refined labels and tooltips to describe actions such as closing, reordering, expanding, and file insertion more clearly.

* **Documentation**
  * Updated guidance and release documentation to follow the no-emoji convention and use clearer icon terminology.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -21,6 +21,22 @@ anything not covered here.
 - `yarn test` — **Vitest** (`test/**/*.spec.ts`). Mock external APIs; tests must run without API keys.
 - `yarn dev` — server + Vite together (local development).
 
+## No emojis
+**Never use emojis anywhere in this project** — UI, source comments, docs, changelog, commit
+messages, skills, CLI output. Icons are **Material Symbols (outlined)**, self-hosted via the
+`material-symbols` npm package: `<span class="material-symbols-outlined">icon_name</span>`.
+A global rule in `src/style.css` gives them `font-size: inherit`, so size them on the parent.
+
+- A header button in config (`server/config/header-config.ts`) takes **`icon`**, not `emoji`.
+  The `emoji` field still exists for end-user configs and wins over `icon` when both are set —
+  don't use it in anything this repo ships.
+- Three deliberate exceptions, all functional. Don't "fix" them:
+  - `server/session/screen-rows.ts` — `/^\s*[❯›]\s/u` parses Claude Code's real terminal output.
+  - `src/composables/useDynamicFavicon.ts` — the `❯` chevron drawn on canvas as the favicon mark.
+  - `bin/mulmoterminal.js` — the CLI doctor's `✓ / ✗ / ○` (a terminal can't render an icon font).
+- Compact status **notation** stays text, not icons: `⎇ main ●3 ↑2`, `●` unsaved dots, `−12` diff
+  counts. Icons there are bigger and slower to scan.
+
 ## Layout
 - `server/` — backend (PTY sessions, config, agents, backends). Ships user-facing skills in `server/skills/`.
 - `src/` — Vue web UI (App.vue, components, composables, router).
@@ -64,7 +80,8 @@ procedure: open this file, paste this, restart what, how to tell it worked, what
   guide holds the reference — do not duplicate the reference.
 - A fix-only release still gets a page: "nothing to configure", what was broken, and **how to
   tell you have the fix**. That is what an upgrader actually wants to know.
-- **Link it from the changelog entry** (a `> 📘` line right under the heading). Before this
+- **Link it from the changelog entry** (a `> **Setup guide:**` blockquote line right under the
+  heading — the old convention used a book emoji, dropped per **No emojis** above). Before this
   existed the changelog had one link into the guide in 717 lines, which is why nobody found the
   manual.
 - **Verify before committing**: every internal link resolves to a real page *and anchor*, and any
--- a/server/config/header-config.ts
+++ b/server/config/header-config.ts
@@ -38,13 +38,13 @@ export interface HeaderConfig {
 // the branch has no open PR, so they self-hide where they don't apply.
 export const DEFAULT_BUTTONS: HeaderButton[] = [
   { id: "pick-file", icon: "attach_file", label: "Insert a file path", run: "open", open: { pickFile: true } },
-  { id: "reveal", emoji: "📂", label: "Reveal in the file manager", run: "open", open: { reveal: "${dir}" } },
+  { id: "reveal", icon: "folder", label: "Reveal in the file manager", run: "open", open: { reveal: "${dir}" } },
   { id: "files", icon: "folder_open", label: "Browse files in the app", run: "open", open: { files: "${dir}" } },
-  { id: "terminal", emoji: "🖥", label: "New terminal here", run: "open", open: { terminal: "${dir}" } },
-  { id: "pr", emoji: "🔗", label: "Open this branch's PR", run: "open", when: "isGitRepo", open: { pr: true } },
+  { id: "terminal", icon: "terminal", label: "New terminal here", run: "open", open: { terminal: "${dir}" } },
+  { id: "pr", icon: "merge", label: "Open this branch's PR", run: "open", when: "isGitRepo", open: { pr: true } },
   // `repo != ` gates on a resolvable GitHub owner/repo (ctx.repo is null for non-GitHub or remoteless
   // repos), so this never renders a broken `https://github.com/` link.
-  { id: "gh", emoji: "🌐", label: "Open on GitHub", run: "open", when: "repo != ", open: { url: "https://github.com/${repo}" } },
+  { id: "gh", icon: "public", label: "Open on GitHub", run: "open", when: "repo != ", open: { url: "https://github.com/${repo}" } },
 ];
 
 // The live context a header is resolved against — all trusted server-side session state.
--- a/server/index.ts
+++ b/server/index.ts
@@ -478,7 +478,7 @@ startCollectionCompletionWatchers().catch((err) => {
 // nudge). The run-binding spawns a VISIBLE session so the user sees the result.
 // Non-fatal: a scheduler failure must never abort startup.
 //
-// Nobody ever presses ✕ on a scheduled session, and one blocked on a permission prompt
+// Nobody ever presses close on a scheduled session, and one blocked on a permission prompt
 // never finishes a turn, so the hook-driven reap can miss it entirely — hence the
 // registry, which bounds them by count and age whatever their hooks did (#541).
 
--- a/server/infra/tmux-routes.ts
+++ b/server/infra/tmux-routes.ts
@@ -31,7 +31,7 @@ export function orphanReapable(resumable: boolean, attachedCount: number | null)
 }
 
 export function mountTmuxRoutes(app: Express, deps: TmuxRouteDeps): void {
-  // Explicit close (the cell's ✕): reap NOW — kill the pty AND its tmux — instead of
+  // Explicit close (the cell's close button): reap NOW — kill the pty AND its tmux — instead of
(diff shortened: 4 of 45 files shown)


snakajima:
Good catch on the ligature text — fixed in 3539a0c.

Material Symbols spans carry the ligature name as real text content, so the Run button announced as "play_arrow Run expand_more", and even the icon-only buttons leaked "forum" / "history" next to their `aria-label`.

Every decorative icon span is now `aria-hidden="true"`. That was already the convention in four places (`App.vue`, `AppToolbar`, `NotificationBell`, `Terminal`) — the new spans just didn't follow it. I applied it across `src/` rather than only the buttons this PR touched: same defect, and a half-converted app would still announce ligatures from the sidebar and toolbar. 75 spans in 23 files.

Audited the obvious risk of hiding the only child — **0 icon-only buttons are left without an accessible name**. Every one already carries an `aria-label`, and the icon-plus-text buttons keep their visible label:

| Pattern | Accessible name before | After |
|---|---|---|
| `▶ Run ▾` dropdown | "play_arrow Run expand_more" | "Run" |
| Ask (icon-only, has aria-label) | "Bring another terminal's last turn here forum" | "Bring another terminal's last turn here" |
| Commit / Push / Open PR | "check Commit" | "Commit" |

`yarn test` 4428 pass, three typechecks clean, lint clean. The one `docPath.spec.ts` failure is the pre-existing timezone bug noted in the PR description and fails on a clean tree too.

snakajima:
Right on the skill — fixed in 20678ee. Stripping the palette labels left the *guidance* intact, which was the more consequential half: the skill is shipped to end users and was still teaching the deprecated shape.

There were **five** emoji examples, not two:

| Line | Was | Now |
|---|---|---|
| 68 | "add a rough emoji swatch (🟫 🟦 🟩 ⬛)" | apply it and look at the real cell — the only accurate preview |
| 182 | `"emoji": "🔨"` | `"icon": "build"` |
| 191 | "`emoji` or `icon` (a material-symbol name)" | `icon` first, with a note that `emoji` still exists and wins when both are set |
| 390–392 | `"emoji": "🗜️" / "📊" / "📁"` | `"icon": "compress" / "difference" / "folder"` |

Two more turned up alongside: `⚙ Settings` in prose, and `"name": "✳ my-project"` — an example that taught putting a glyph in the directory badge, which is how emoji would have crept back in one config at a time.

`server/skills/` now scans emoji-free.

On the `FilesOverlay.vue:256` comment — same finding as Codex's, fixed in 3539a0c: all 75 icon spans across `src/` are `aria-hidden="true"`, verified that no icon-only button is left without an accessible name.

The SkillSpector warnings (AS3 skill enumeration, EA2, RA2) are pre-existing behaviour of that skill and unrelated to this PR — the skill lists a directory's `.claude/skills` because configuring the `skills` allowlist requires knowing what's installed. Happy to look at them separately if you want.

snakajima:
Note on the `CODEX VERDICT: CHANGES REQUESTED` at 19:49 — **that finding was fabricated, and so was its retraction two minutes later.** Flagging it so nobody reads the thread as "a regression was introduced and then fixed."

It claimed this PR deleted a "Terminal font size" settings section plus `useTerminalFontSize`, `dirFontSize`, `effectiveFontSize()`, a `conn.setFontSize(...)` watcher, and a `:dir-font-size` prop. **None of those identifiers have ever existed in this repository:**

```
$ grep -rn "useTerminalFontSize|effectiveFontSize|dirFontSize|dir-font-size|setFontSize" src/
(0 results in this branch)

$ git grep "useTerminalFontSize|effectiveFontSize|dir-font-size" main -- src/
(0 results in main)

$ git diff main...HEAD -- src/ | grep "^-" | grep -i "font"
(0 deleted lines mention font)
```

The only `fontSize` in the codebase is a hardcoded xterm option, `fontSize: 14` in `useTerminalConnections.ts:257`, byte-identical in `main` and at HEAD. There is no font-size settings UI, no composable, and no per-directory override to have removed.

The 19:51 "resolved" verdict is equally wrong in the other direction — it reports that `SettingsModal` "still exposes terminal font-size controls" and `TerminalCell` "still passes `dir-font-size`". It does not, and never did.

Net effect on this PR: nothing to change. This diff removes no functionality; every change is a glyph swap, an `aria-hidden` attribute, or documentation. The Codex findings worth keeping are the original P2 ligature/accessibility one (real, fixed in 3539a0c) and CodeRabbit's skill-guidance one (real, fixed in 20678ee).

isamu:
## Review — no functional defects found

Verified independently of the bot threads. Two things worth acting on, neither blocking the code.

### 1. The "that finding was fabricated" note is itself wrong

The 20:12 comment says none of `useTerminalFontSize` / `dirFontSize` / `effectiveFontSize` / `setFontSize` "have ever existed in this repository". All four exist on `main` right now:

```
useTerminalFontSize   origin/main: 3 files      this branch: 0 files
introduced in         fde9199  "feat: make the terminal font size configurable, globally and per directory"  (#866)
this branch's merge base   cf1a14a  →  does NOT contain fde9199
```

`git merge-base --is-ancestor fde9199 <branch>` fails: **this branch was cut before #866 landed.** So Codex compared the branch's files against a newer `main` and correctly observed the identifiers were absent from the branch; only its *conclusion* — that this PR deleted them — was wrong.

No impact on the merge: GitHub reports `MERGEABLE CLEAN`, and **#866 added no emoji in its added lines**, so nothing it introduced escapes this sweep. The reason to correct the record is that "the bot invented it" invites dismissing the next finding of the same shape, when the actual lesson is "the branch is behind `main`; compare against the merge base."

### 2. `CLAUDE.md` lands a rule the repo violates, with the caveat only in the PR description

The new section says never use emoji **anywhere** — "UI, source comments, docs, changelog, commit messages, skills, CLI output". Meanwhile `docs/`, `plans/` and `README.md` still hold ~453, and `docs/ChangeLog.md` still has `> 📘` lines.

"Sweeping prose is a separate change" is in the PR description, which no future reader of `CLAUDE.md` sees. An agent reading the rule will either sweep `docs/` inside an unrelated PR, or stall on the contradiction. One sentence in the section closes it — e.g. *"`docs/`, `plans/` and `README.md` still carry legacy emoji; they are swept separately — don't fix them as a side effect of an unrelated change."*

### Verified clean (so nobody re-checks these)

- **All 33 icon names resolve in the shipped font.** One trap for whoever checks this next: `material-symbols@0.45.9`'s `index.d.ts` name union is **missing `auto_awesome`**, so validating against the type list produces a false positive. Decompressing `material-symbols-outlined.woff2` and reading the ligature table shows it *is* in the font. Check the font, not the types.
- **No control lost its accessible name.** Every icon-only `<button>` still has `aria-label`, `title`, or visible text.
- **87 `aria-hidden` attributes, none wrapping focusable content** — the usual way this sweep goes wrong.
- **`emoji` still beats `icon`** (`Terminal.vue` `v-if`/`v-else`), and a user-config `icon` survives `sanitizeButton`, so a custom button can't lose its glyph while the shipped defaults keep theirs.
- **No sizing regressions from unsized spans.** `src/style.css` declares `.material-symbols-outlined { font-size: inherit }` in the `base` layer, so the 20+ spans with no explicit size follow surrounding text rather than the package's 24px default.
- **No UI emoji left in `src/`** — remaining matches are `→` in comments and the documented `❯` exceptions.
- "the paperclip button" is accurate: `attach_file` is a paperclip glyph.
- CI green on every check.

isamu (approved):

snakajima:
Correcting my own correction: **I was wrong, and Codex was right.**

I claimed the 19:49 font-size finding was fabricated because none of those identifiers existed in the repo. They do. They arrived in #866 (`fde9199`, Jul 27 03:20 JST), which merged *after* this branch's base (`cf1a14a`, Jul 26 20:24 JST). Codex was diffing this branch against a newer `main`, so code this PR never touched appeared as deletions — a stale-base artifact, which is a normal review false positive and not an invention. My "0 results in main" check was run against the branch's `main` ref before I pulled, so it confirmed my own mistake rather than testing it.

The 19:51 "resolved" verdict was also right for the same reason, once it re-diffed.

No harm done to the code — the merge preserved the feature. Verified on merged `main` (`31d2daa`):

```
src/composables/useTerminalFontSize.ts        present
src/components/Terminal.vue                   useTerminalFontSize / effectiveFontSize present
src/components/TerminalCell.vue:1118          :dir-font-size="dirConfig.fontSize"
src/App.vue:362                               :dir-font-size="singleDirConfig.fontSize"
SettingsModal.vue                             4 font-size references
```

I also checked that #866's new icon buttons didn't drift from the accessibility rule this PR established: **0 icon spans on `main` are missing `aria-hidden`**, so the two PRs composed cleanly.

Apologies for the noise — the confident phrasing was unwarranted, and I should have pulled `main` before asserting anything about what does or doesn't exist in it.

PR #881, opened 2026-07-26
feat: order the grid by each directory's orderPriority (#876)

isamu:
## Summary

A repo declares `orderPriority` in its `.mulmoterminal.json`, and the grid arranges cells by it:

```json
{ "name": "PROD", "orderPriority": 10 }
```

Lowest first; any integer including negatives; **directories that declare nothing sort last**, keeping their existing order, so adding the key to one project doesn't shuffle the rest. Equal ranks stay in their current order — which is also what happens when several cells share a directory, since the rank belongs to the *directory*.

Added as a **third sort mode**, so the toolbar button now cycles `auto → manual → priority`.

## Items to Confirm / Review

- **Why a third mode and not a change to the existing two.** The grid already had `auto` (attention-first) and `manual` (drag with `◀ ▶`). Folding priority into `auto` would silently change what today's users see the moment any directory grew the key; making it the base order would **fight drag-to-reorder** (a dragged cell whose directory sets a rank would snap back). As its own mode, a user who writes no `orderPriority` sees byte-identical behaviour. This was the one design call I asked about before building.
- **The toolbar button changed shape.** It was `autoSort?: boolean` with `:aria-pressed="autoSort"`. Three states can't ride a boolean, and `aria-pressed` describes a *binary* toggle — so the prop is now `sortMode`, `aria-pressed` is gone, and the accessible name carries the state instead (`"Grid cell ordering: priority (click for auto)"`). Worth a look if you have a view on the a11y trade: a 3-way cycle button has no standard ARIA pattern.
- **`useDirPriorities` is new plumbing, not a wrapper.** The rank is needed for cells on pages that are **not mounted**, and an unmounted cell runs no `useDirConfig` — the same constraint the attention sort already documents for status. It subscribes to the whole *set* of directories via the existing per-cwd invalidation channel, so editing a `.mulmoterminal.json` re-sorts the grid live with no watcher. Its bind/release now shares a helper with `useDirConfig` rather than being a second copy of that bookkeeping.
- **`orderPriority` is not clamped**, unlike `fontSize`. Every finite integer is a usable rank, so there is no out-of-range to pull back; `1.5` and `"3"` are rejected to null (= unset) instead.
- **Launch cells** are held last by their own comparison level rather than by their sort key — in priority mode every unset directory already ranks at `Infinity`, so a key alone could not place a launch cell after them.

## User Prompt

> 各レポジトリの `.mulmoterminal.json` に `orderPriority` を数字で書くと、grid view のときにその順番で並べる機能あると良いね

Design question I raised before implementing, and the answer: how should this relate to the existing `auto`/`manual` modes? → **add it as a third mode**, leaving both existing modes untouched.

## Verification

Gates: `format`, `lint` (0 errors), `typecheck`, `typecheck:server`, `typecheck:test`, `build`, `test` — **4469 passed** (+19 new).

**Driven in the running app**, with three directories seeded into the grid in deliberately the wrong order — `gamma`(30), `beta`(unset), `alpha`(10):

| mode | on-screen order |
|---|---|
| manual | gamma, beta, alpha — the seeded order, untouched |
| **priority** | **alpha, gamma, beta** — 10, 30, then unset |
| auto | gamma, beta, alpha — untouched (all idle → stable manual order) |

The button cycled `manual → priority → auto → manual`, and all three modes were reached.

Also checked on the wire: `GET /api/dir-config` returns `orderPriority` `10` / `30` / `null` for the three directories.

One note on that run: my first verification script reported a false negative because it assumed priority was two clicks away when the grid had started in `manual` (one click). The re-run keys off the mode the button reports rather than a click count.

## Docs

`README.md`, `docs/guide/{en,ja}/config.md` (with the ordering rules) and `{en,ja}/basics.md` (the button now has three modes) — both languages. `server/skills/mulmoterminal-config/SKILL.md` gained the key, plus a note to space ranks 10/20/30 so a project can be slotted in later without renumbering, and a warning to tell the user the key only takes effect in the priority mode.

## Summary by CodeRabbit

* **New Features**
  * Added a 3-way grid ordering cycle: **auto**, **manual**, and **priority**.
  * Introduced per-directory `.mulmoterminal.json` **`orderPriority`** (integer, including negatives) to drive **priority** sorting.
  * Directories without `orderPriority` sort last; equal ranks keep existing order. Config changes now apply live when written via tools.
* **Documentation**
  * Updated English and Japanese guides and configuration reference to explain `orderPriority` and the updated ordering-mode behavior.


isamu:
## Bot review triage — iteration 1 (`4b0613a` → `cef2613`)

| Reviewer | Signal |
|---|---|
| Codex | `CODEX VERDICT: LGTM` — no findings, explicitly covering accessibility and the new sort mode |
| CodeRabbit | **rate-limited — it did not review this PR.** The green check is a limit notice, not a sign-off |
| CI | 11/11 green |

No bot findings to apply. **One issue found by my own reading**, which is the reason for the follow-up commit:

### `visibleOrdered` didn't pass the priority map

```ts
const ordered = orderCells(state.cells, statusByUid, state.sortMode);  // no priorityByCwd
```

Defaulted to `{}`, every directory reads as **unset**, so in the new mode this returned a *different order from the grid* — the exact drift #720 exists to prevent.

**It is currently unreachable**: nothing in the app calls `visibleOrdered` (GridView computes its own `orderedCells`), only its specs do. knip doesn't flag it either, since `test/**/*.spec.ts` is a knip entry point, so a test-only export counts as used. So this was a trap for the next caller rather than a live bug — which is why it became a parameter instead of a deletion.

Pinned in both directions, so a future caller that forgets the map fails a test instead of silently disagreeing with the grid:

```ts
expect(visibleOrdered(s, {}, { "/y": 1 }).map((c) => c.uid)).toEqual([1, 0]);
expect(visibleOrdered(s, {}).map((c) => c.uid)).toEqual([0, 1]);
```

### After the fix

`format`, `lint` (0 errors), all three typechecks, `build`, `test` — **4470 passed**.

isamu:
## Bot review triage — iteration 2 (`cef2613` → `0d13698`)

**Codex: `CODEX VERDICT: CHANGES REQUESTED`** — one finding, and it was right:

> `src/composables/useDirConfig.ts:70` currently accepts any finite numeric `orderPriority` from the API payload. The feature contract/schema is integer-only, so this boundary parser should reject non-integers (e.g. `1.5`) as unset (`null`).

Verified: the client used `Number.isFinite`, the server `z.number().int()`. So `1.5` would have sorted on the client while the server read it as unset.

### Fixed at the root, not at the symptom

The one-character change (`isFinite` → `isInteger`) would have closed this instance and left the cause: **two boundaries validating the same field with two hand-written rules**. Both sides now share `normalizeOrderPriority` in `common/` — exactly how they already share `normalizeFontSize`. I wrote that pattern for the font size and then failed to apply it here, which is what let the two drift apart in the first place.

Pinned both places: the pure function directly (integers kept, `1.5` / `-0.5` / `"3"` / `NaN` / `Infinity` rejected), and a fraction arriving over the wire reading as unset in the client parser.

### Also in this push: main was merged

`origin/main` had advanced (#879, `prWorkdirFooter`) while this PR was open, so it is merged in and every check re-run on the result — per the loop's own rule that a main-side change can move a contract under you. No conflicts.

### Verification

`format`, `lint` (0 errors), all three typechecks, `build`, `test` — **4496 passed** (up from 4474; the jump includes main's new specs).

isamu:
## Bot review triage — iteration 3 (`0d13698` → `f9ce848`)

CodeRabbit **did** review this round (its earlier rate-limit notices were from previous iterations). Two findings, both applied.

### 1. README claimed dir-config changes need a reopen

> `orderPriority` is reloaded through the directory-config invalidation channel, but the nearby note says changes take effect only when a terminal is reopened.

Correct — and checking it, **the line was wrong for the whole mechanism, not just the new key**:

> ~~Changes take effect when the terminal is next opened (no live file watch).~~

`dir-config.ts` says so in its own comment: *"Claude's tool hooks already report every write, so they double as the live-reload signal."* Colors, palette and font size have all been live for as long as that channel has existed.

So rather than bolting an exception onto a wrong sentence, it now states the actual rule: **a write through Claude's tools (including this skill) applies live; an edit made outside a session waits for a reopen**, because there genuinely is no filesystem watcher. Both halves matter — the second is why the original sentence looked true.

### 2. A test that didn't check what its name claimed

> The final count is checked only after `scope.stop()`, so this passes even if `/g/y` remains bound until disposal.

Exactly right, and the more useful kind of finding: `it("drops a directory that leaves the set, and releases its binding")` counted bindings **only after disposal**, which releases everything regardless — so it would have passed even if leaving the watched set released nothing at all. The name asserted a guarantee the body never tested.

Now counted at three points — `before + 2` bound, `before + 1` the moment `/g/y` leaves, `before` after disposal. It passes, which means `untrack` really does release at set-change; that behaviour is now pinned instead of assumed.

### Also: main merged again

`origin/main` had advanced by 4 more commits (#877 cockpit lines) — merged, no conflicts, all checks re-run on the result.

### Verification

`format`, `lint` (0 errors), all three typechecks, `build`, `test` — **4507 passed**.

PR #940, opened 2026-07-27
feat: copy-on-select in the terminal, as an opt-in config setting (#900)

isamu:
## Summary

Closes the half of #900 that #933 left open: **copy-on-select**. With
`{ "copyOnSelect": true }` in `~/.mulmoterminal/config.json`, a drag over terminal output reaches
the clipboard the moment the selection settles, with no key pressed. **Off by default.**

It looks like the continuation of #933 and isn't. There a copy key had been pressed, so the browser
was already going to copy — returning `false` from xterm's key handler and standing back was the
whole implementation, and no clipboard API was touched. **A selection asks the browser for nothing**,
so this is the first place in the app that writes the clipboard itself. Every constraint below comes
from that one difference.

## Items to Confirm / Review

1. **Not verified in a browser.** No automation in this session, so the evidence is unit tests plus
   reading xterm's source — the same standing as #933. What most wants a real check: does a drag
   actually land on the system clipboard, and does it still land when the app is opened over plain
   `http://<lan-ip>:PORT` (the `execCommand` fallback path)?
2. **The `execCommand` fallback requires the terminal to still hold keyboard focus, and it does not
   take focus.** Reaching that line means the user just dragged in that terminal, so xterm has
   already focused its textarea — stealing focus back from wherever it went instead seemed clearly
   worse. If that reasoning is wrong, this is where it shows: a silent no-op in an insecure context.
3. **The 150 ms settle window is a judgement call.** Long enough to collapse a drag into one write,
   short enough not to feel laggy — but it is a guess against human drag speed, not a measurement.
4. **`copyOnSelect` as its own setting rather than a `keymap` value.** `{ "keymap": { "copy":
   "select" } }` was considered and rejected (reasoning below). Worth a second opinion since it
   permanently shapes the config surface.
5. **What is deliberately NOT copied** — whitespace-only selections, and the same text twice. Both
   protect the clipboard the user already had; both are silent when they trigger.

## User Prompt

- #900 を読んでほしい。`copyOnSelect` も必要とのこと
- （`copyOnSelect` とは何か、Windows 固有の機能か、という確認。X11 の PRIMARY selection の挙動は既知）
- キーのイベントではなく、選択（値）のイベントの方に `copyOnSelect` を足せるということか
- まずはそのことを丁寧に issue に書いてから plan
- 良くわからない約束とかやめよう。事実だけ伝えて作ろう

設計メモは実装前に
[#900 へコメント](https://github.com/receptron/mulmoterminal/issues/900#issuecomment-5089518433)
として残してあります。

### 1. `navigator.clipboard` is secure-context-only

At `http://<lan-ip>:PORT` — an ordinary way to reach this app from a second machine — it does not
exist at all. The existing OSC 52 provider swallows the same hole in a `catch`, which is survivable
there because a terminal program asked for the copy and can ask again. Here nothing was pressed, so
a silent no-op is indistinguishable from the feature being off.

Fallback: focus is checked on `.xterm-helper-textarea`, then `document.execCommand("copy")` fires
xterm's own `copy` listener, which writes the current selection. Works in insecure contexts, and is
the route the reporter's own userscript used.

That fallback is usable **only because the wanted text is the terminal's live selection**. It is not
a general "copy this string" primitive, and the comment on it says so.

### 2. `onSelectionChange` fires throughout the drag, not at its end

Verified in `@xterm/xterm` 6.0.0 — the selection service fires whenever start/end coordinates
differ:

```js
e[1]===this._oldSelectionStart[1] && t[0]===this._oldSelectionEnd[0] && ... || this._fireOnSelectionChange(e,t,i)
```

One drag across twenty cells is up to twenty events. Writing on each is not merely wasteful: on
Windows with clipboard history on, **every intermediate write becomes a Win+V entry**, so one drag
buries the user's history under partial selections. The timer restarts per event; only the last one
writes.

Same shape as the reporter's own `setTimeout(…, 0)` fix for reading the selection too early after
`mouseup`, arrived at for a different reason.

### 3. A whitespace-only selection must not clobber the clipboard

Dragging across empty terminal space selects spaces. Silently replacing the user's clipboard with a
run of them is this feature's worst failure mode. The `copy` keymap action remains for anyone who
genuinely wants indentation.

## The bug found while reviewing my own first pass

"Don't write the same text twice" is right during a drag and wrong across drags: between two of them
the user may have copied something else, so re-selecting the same text is a **fresh intent**, and
remembering it forever leaves them holding the wrong thing with nothing to show for it.

A cleared selection now retires the remembered text — and that has to be observed **per event**
rather than at settle, because a click that clears and the drag that follows can both land inside
one settle window, leaving only the new selection visible by the time the timer runs.
`hasSelection()` is used rather than `getSelection()` for this: the cheap half of the API, since
`getSelection()` rebuilds the text from the buffer.

There is a test that fails without the fix.

## Config surface: why not a `keymap` value

`{ "keymap": { "copy": "select" } }` would keep "when does copy happen" in one place. Rejected
because every `keymap` value is currently a **key spelling** that `parseKeyBinding` parses as one,
so `"select"` becomes a special case inside it — and one action holds one value, so there would be
no way to have **both** a key and the selection. `copyOnSelect` composes with the `copy` binding
instead.

Config-file-only, no Settings UI, matching `prWorkdirFooter` / `worklogEnabled` / `fontFamily`.

## Also in here

The bundled `mulmoterminal-config` skill was still missing `copy` / `paste` from #933 — its action
table now lists them, alongside the new `copyOnSelect` section.

## Checks

`yarn format` / `lint` / `typecheck` / `typecheck:server` / `typecheck:test` / `build` clean.
`yarn test` 4889 passed, 38 skipped. (The one lint warning is pre-existing in `ws-routes.ts`.)

## Summary by CodeRabbit

- **New Features**
  - Added an optional “copy on select” feature for terminal text selections.
  - Settled mouse selections can be copied automatically without pressing a key.
  - Avoids copying whitespace-only or unchanged selections.
  - Works alongside the existing copy shortcut and includes a clipboard fallback when browser access is limited.

- **Documentation**
  - Added configuration guidance in English and Japanese.
  - Documented setup, reload requirements, and HTTP clipboard limitations.


isamu:
## レビュー状況（`6a8cfea` 時点）

| | 結果 |
|---|---|
| Codex auto-review | **LGTM**（`e4098b2` と `6a8cfea` の 2 回とも。指摘なし） |
| CodeRabbit | **未レビュー** — レート制限。「レビューできなかった」であって承認ではありません |
| CI | 全 green（lint-and-build macos / ubuntu、package-smoke、knip、jscpd、Socket、triage） |

## ボットの指摘ではなく、自分で読み直して直した 2 件（`6a8cfea`）

1. **OFF のときも動いていた。** `onSelectionChange` は座標が変わるたびに発火するので、
   `copyOnSelect` を有効にしていない人（＝既定）でも、ドラッグするたびにセルの数だけ
   タイマーの設定と取り消しが走っていました。早期 return を入れ、OFF のときは本当に何も
   しないようにしています。
2. **README の設定表に行が無かった。** 追加しました。なお同じ表に `keymap` と
   `cockpitLines` の行も元から無く、これは既存の漏れとして**この PR では触っていません**。

## 人に見てほしいこと

**実機のブラウザで確認していません。** 根拠はユニットテストと xterm のソース確認までです
（#933 と同じ状態）。特に確認したいのは以下の 2 つです。

- ドラッグした内容が実際にシステムのクリップボードに入るか
- `http://<LAN の IP>:PORT` で開いた場合にも入るか（`execCommand` フォールバックの経路）

isamu:
@coderabbitai review

isamu:
## CodeRabbit の指摘 4 件、すべて対応しました（`fe16f77`）

いずれも実際のコードに当てて確認した上で適用しています。**4 件とも実在**していました。

### Major — 書き込みが重なると、古い方の選択がクリップボードに残りうる

これは本物でした。settle タイマーはイベントごとに張り直しているので同時に 1 本しか無いのですが、
そこから始まる**書き込みは非同期**です。前の書き込みが飛んだままの状態で次の settle が来ると、2 つが
重なります。

窓が空くのは机上の話ではありません。**クリップボードの許可を尋ねるブラウザでは、`writeText` は
ユーザが答えるまで pending のまま**で、これに上限はありません。重なった 2 つはブラウザの都合の順で
解決し、`lastCopied` もその順で代入されます。

書き込みを直列化しました。`.catch` は見た目より重要で、1 つ reject するとチェーンが死んで、その
ターミナルの copy-on-select が黙って終わります。

テストは、hung な書き込みに対して settle を 2 回起こす形にしました。**修正を戻すと落ちることを確認済み**です。

### MD018（plan ファイル）

`#933` が行頭に来ていて、markdownlint が見出しの書き損じと解釈していました。修正。

### テストの後始末 2 件

- フォーカスのテストが `<input>` をテスト本体で削除していたので、アサーションが落ちるとフォーカスされた
  ノードが残り、`document.activeElement` を見る他 2 つのテストに波及していました。`afterEach` に移動
- `vi.unstubAllGlobals()` は `document` への `defineProperty` を戻しません。jsdom には元々
  `execCommand` が無いので、このファイルの後続 suite が「常に true を返す実装」を引き継いでいました。
  `afterEach` で削除

最後の 1 点だけ、提案されていた `as Partial<Document>` ではなく `Reflect.deleteProperty` を使いました。
このリポジトリのルールが `as` キャストを禁じているためです。

isamu:
## 実機で確認しました（Chrome / macOS / `http://localhost`）

ドラッグして離すだけで選択テキストがクリップボードに入り、**別アプリへの貼り付けで実際に取り出せる**ことを
確認しました。これで「ブラウザで未検証」という状態は解消です。

### 確認の途中で踏んだこと

**1. 有効化したあと、タブの再読み込みが要ります。**

設定はページ読み込み時に一度だけ `/api/config` から読まれるので、サーバを再起動しても**開いたままのタブは
古い値を掴んだまま**です。最初「効かない」と見えたのはこれでした。ドキュメントには「タブを再読み込み」と
書いてありますが、実際に踏むと**原因として真っ先には思い浮かびません**。

**2. 古いバージョンが設定を書くと、このキーが消えます。**

`~/.mulmoterminal/config.json` は全インスタンス共有で、`copyOnSelect` を知らない 2.2.0 が設定を書くと
（プリセット記録や設定保存のたびに）自分の知るフィールドだけで書き直します。検証中に実際に消えました。

この PR 固有の話ではなく、**追加されるすべての設定項目に同じ性質**があります（`keymap` も同様）。今回は
指摘に留めます。

**3. `writeText` は document が非フォーカスだと `NotAllowedError` で弾かれます。**

つまり、**ドラッグを離してから settle の 150ms 以内にウィンドウを離れると、そのコピーは落ちます。** 実際には
ボタンを離して別アプリへ移るまでに 150ms 以上かかるので、上の確認では踏みませんでした。塞ぐとすれば
「フォーカスが戻ったら書き直す」ですが、静かに直るぶん挙動が読みにくくなるので、**今回は入れていません**。

### まだ確認していないこと

`http://<LAN の IP>:PORT` 経由での `execCommand` フォールバック。macOS 側は `localhost` なので
`navigator.clipboard` の経路しか通っていません。こちらは Windows / LAN で運用されている方の確認に委ねます。

PR #1071, opened 2026-07-29
feat: launcher offers a switch for every GUI tool group, not just the Canvas two

snakajima:
## Why

`data` (presentCollection, manageCollection, manageAccounting) and `external` (google, readXPost, searchX) have been routed, gated and pre-approved on the server since the groups were defined in `common/toolGroups.ts` — the group URL, the tool gate, `allowedToolNames`, and the codex `-c mcp_servers.*` args all take any of the four. Only the launcher was narrower: it drew a row per `CANVAS_TOOL_GROUPS`, so the only way to hand a grid cell a collection or a google tool was to type `claude mcp add -s local …` by hand against the url template.

## What

- **`common/toolGroups.ts`** — `TOOL_GROUP_HEADINGS`, a `Record<ToolGroup, string>`: render/media → "Canvas", data → "Workspace data", external → "External accounts". A Record rather than a lookup with a fallback, so adding a group without deciding what to call it fails to compile instead of shipping a switch labelled "external".
- **`TerminalCell.vue`** — the row block iterates `TOOL_GROUPS`; the state record, the read-back in `loadMcpGroups` and the worktree carry-over all cover four groups. Each row still names the MCP server it registers (`Workspace data (data MCPs)`), because one switch is one server.
- `CANVAS_TOOL_GROUPS` / `hasCanvasGroup` are untouched — they still answer "which groups DRAW", and are no longer the set that gets a switch. The Canvas pane's availability check does not widen.
- Renames, since the state is no longer Canvas-specific: `canvas*` → `mcpGroup*`, testids → `cell-mcp-toggle-<group>`, `canvasWriteQueue.ts` → `mcpWriteQueue.ts`. The queue's behaviour is unchanged and still module-level — the file being serialized is Claude Code's, shared by every directory.
- The checkbox aria-label loses "so the agent can draw", which is false for data/external.

No server change was needed.

## Known limitation (pre-existing)

The hover lists a group's tools statically from `toolGroups.ts`, so if a backend plugin isn't present (accounting not configured, say) the group registers with fewer tools than the tooltip names. Already true for render/media; not addressed here.

## Test plan

- New spec: all four rows render, a `data` registration reads back ticked, flipping `external` POSTs `{group:"external", enabled:true}` to the directory it was flipped in.
- New spec pinning `TOOL_GROUP_HEADINGS` covers every group and keeps "Canvas" for exactly the drawing ones.
- `yarn format`, `yarn lint`, `yarn typecheck`, `yarn typecheck:server`, `yarn typecheck:test`, `yarn test` (6041 passed) all green.
- Not yet exercised in a live browser session — the rows are covered by component tests only.

work in mulmoterminal

## Summary by CodeRabbit

- **New Features**
  - Added configurable MCP tool-group switches for each supported group, including Render, Media, Workspace data, and External accounts.
  - Sessions now carry enabled MCP tool-group settings into newly created or reused worktrees.
  - Added clear labels and contextual information for each tool group.

- **Bug Fixes**
  - Improved tool-group setting updates to prevent conflicting writes and restore state when an update fails.
  - Refreshed tool-group settings correctly when changing directories or resetting sessions.

### DIFF
--- a/common/toolGroups.ts
+++ b/common/toolGroups.ts
@@ -37,6 +37,20 @@ export const CANVAS_TOOL_GROUPS: readonly ToolGroup[] = ["render", "media"];
 export const hasCanvasGroup = (groups: unknown): boolean =>
   Array.isArray(groups) && groups.some((group) => isToolGroup(group) && CANVAS_TOOL_GROUPS.includes(group));
 
+// What the launcher's switch for a group calls it. The group NAME is shown too (it is the MCP
+// server id the switch registers), so this line answers the other question: what does the agent
+// gain. Two groups share a heading on purpose — render and media both end up in the Canvas pane,
+// which is why CANVAS_TOOL_GROUPS above holds exactly those two.
+//
+// A Record, not a lookup with a fallback: adding a group to TOOL_GROUPS without deciding what to
+// call it should fail to compile, not ship a switch labelled "external".
+export const TOOL_GROUP_HEADINGS: Record<ToolGroup, string> = {
+  render: "Canvas",
+  data: "Workspace data",
+  media: "Canvas",
+  external: "External accounts",
+};
+
 // Which group each GUI tool belongs to.
 //
 // A tool that is absent belongs to NO group and is therefore reachable only through the
--- a/src/components/TerminalCell.vue
+++ b/src/components/TerminalCell.vue
@@ -554,10 +554,12 @@ const mcpGroupTitle = (group: ToolGroup): string =>
 async function loadMcpGroups() {
   const dir = dirInput.value.trim() || props.defaultCwd;
   const reqId = ++mcpGroupReq;
-  if (launched.value || !dir) {
-    mcpGroupDir.value = null;
-    return;
-  }
+  // Cleared BEFORE the fetch, not only on the failure path: the switches showing while the answer
+  // is in flight are the PREVIOUS directory's, and a flip during that gap writes to the previous
+  // directory (applyMcpGroup captures mcpGroupDir, which is what keeps a queued write honest).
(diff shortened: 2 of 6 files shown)


snakajima:
Both review findings addressed in e29d79be.

**codex — stale worktree registrations.** Confirmed, and it is the `external` case that makes it bite: uncheck a group in the repo, reuse a worktree it was once enabled on, and the session gets tools the launcher shows as off. `carryMcpGroups` is now `syncMcpGroupsInto` and syncs all of `TOOL_GROUPS` in both directions.

One deviation from the suggestion: it writes only the groups that **differ**, not all four unconditionally. Each write shells out to the `claude` CLI (`add` is a remove+add, so two spawns) and the launch awaits them, so four blind writes would put ~8 process spawns in front of every worktree launch. The target's current state is one GET, which reads the config file without shelling out. If that read fails, every group is written rather than assumed — a wrong assumption there is exactly the stale registration this exists to clear. Regression test: repo has `render`, worktree has `render` + `external`, reuse posts exactly `{group: "external", enabled: false}`.

**coderabbit — toggles live during the debounce.** Also confirmed. Cleared on the directory change itself and again at the top of `loadMcpGroups` before its fetch, so the in-flight window after `fillDir()` is covered too. The rows disappear rather than staying at the previous directory's positions, and return ~300ms later with the new directory's. Regression test asserts the row is gone immediately after the field changes and comes back correctly ticked after the debounce.

One existing spec changed: `reuses an existing worktree…` now flushes before asserting the terminal mounted, because the launch waits on one more async hop (the read).

`yarn format` / `lint` / all three typechecks / `yarn test` (6043 passed) green.

work in mulmoterminal

PR #1095, opened 2026-07-29
feat: Antigravity (agy) as a first-class agent, with per-project GUI MCP

snakajima:
## Summary

Adds **Antigravity (`agy`)** as a third first-class agent, alongside Claude and Codex: its own
WebSocket endpoint, conversation resume across reload, the GUI panel over MCP, and the agent
pickers in the single view, grid cells, sidebar/tab bar and the Collections browser.

Tracking issue: #236. Plan: `plans/feat-multi-agent-support.md` (this is PR#4 of that stack).

**The session plumbing** mirrors codex: `/ws/antigravity`, `ANTIGRAVITY_BIN` / `ANTIGRAVITY_MODEL`
/ `ANTIGRAVITY_HOME`, and — because agy mints its own conversation id — a watcher over
`~/.gemini/antigravity-cli/brain/` that claims the new conversation directory, attributed only when
unambiguous, so a later cold reconnect can `agy --conversation <id>`.

**The GUI MCP works differently, and that is the interesting part.** `agy` takes no MCP flag: it
reads its servers from a FILE. The only project-scoped file it reads is `.agents/mcp_config.json`,
which is per DIRECTORY and shared by every session running there. So:

- the registration is written from the directory's **Canvas switches** — the same switches Claude's
  cells read (`infra/gui-mcp-registration.ts` stays the only registry) — and rewritten whenever a
  switch flips or an agy session starts;
- the TOOL GROUP is a property of the directory, so it rides the entry's own `env`;
- the SESSION is not, so it is **never written to disk**. It reaches the bridge through the spawned
  agy process's environment and only there;
- servers in that file that MulmoTerminal did not write are left alone, the file is removed once no
  group is on, and it is kept out of the user's `git status` via `.git/info/exclude` rather than
  their `.gitignore` — the same reasoning as `claude mcp add -s local`;
- the entry runs `server/mcp/bridge.mjs`, a stdio-to-HTTP shim onto the same in-process GUI MCP
  server the other agents call.

Verified against a live server — agy's own `/mcp` screen reports
`mulmoterminal-render  Tools: presentDocument, presentForm, presentChart, presentHtml` — and in a
real session in another workspace with `render` + `data` switched on.

### Structural changes this pulled in

Adding a THIRD agent exposed that the codebase was shaped for exactly two, in ways that all
typecheck and all silently mislabel the new one. Fixed here rather than left to the next agent:

- **One `agent: TerminalAgent` value** replaces the parallel `codex?: boolean` /
  `antigravity?: boolean` flags through `ConnTarget`, `ConnTargetUrlInput`, `SlotCandidate` and the
  Terminal props. `connWsUrl`'s branch is one line; `isClaudeTarget` stops growing a
  `&& !t.<agent>` per agent.
- `TerminalAgent` (`common/sessionAgent.ts`) replaces the `"claude" | "codex" | "antigravity"`
  union that had been written out at 14 call sites, server included.
- `buildCodexWsUrl` + `buildAntigravityWsUrl` (identical but for the path) → `buildAgentWsUrl` over
  a `Record`, so an agent without an endpoint fails to compile instead of building a 404 URL.
- The codex and antigravity spawners' PTY relays were the same function modulo the log word →
  `session/pty-relay.ts`. Deliberately NOT shared with claude's, which drives hooks, transcripts
  and the draft injector off the same stream.
- `codexResumeId` was agent-neutral in everything but its name → `agentResumeId`, shared.
- `skillSeed` takes the agent instead of an `isCodex` boolean, with claude as the special case.

### Bugs found and fixed while reviewing the branch

Each of these accepted an antigravity agent in the types and then did something else:

| Where | Was |
|---|---|
| Collections browser toggle | spawned **Claude** (`parseBackgroundChat` coerced non-codex to claude) |
| Phone "new antigravity terminal" | opened a **shell** (`cellForAgent` if-chain fell through) |
| Surviving `agy` pane after restart | read as **shell** (`AGENT_COMMANDS` had no `agy`) |
| Push notifier's last reply | read a **claude transcript**, leaving `sessionLastTurn`'s antigravity branch unreachable |
| Cell copy-code-block / activity timeline | treated antigravity as **claude** |
| Cold resume | resumed **unconditionally** — a stale key reached `agy --conversation`, which answers a conversation it cannot find by starting a fresh one under the old session's id |

Antigravity background chats now really run, seeded through `--prompt-interactive` (verified
against the CLI: it runs the turn and stays interactive).

## Items to Confirm / Review

- **`.agents/mcp_config.json` is written into the user's project.** This was a deliberate choice
  over the global `~/.gemini/config/mcp_config.json` (agy offers no third location). It is
  merge-written, removed when empty, and excluded via `.git/info/exclude` — but it *is* a file we
  create in someone's repo. Say the word if you'd rather it were the global file.
- **A single-view agy session gets the tools its DIRECTORY has switched on**, unlike Claude and
  Codex which carry every tool in the single view regardless. Forced by agy's per-directory config;
  documented in README rather than papered over. Flagging it because it is a real asymmetry.
- **agy's own trust prompt.** The first agy session in a new directory asks "Do you trust the
  contents of this project?" and `--dangerously-skip-permissions` does not skip it — one Enter.
  Left alone; automating a trust dialog seemed like the wrong instinct.
- **Auto-approval.** Antigravity spawns with `--dangerously-skip-permissions`, matching what the
  branch already did and the `--permission-mode auto` claude gets. Same owner call as codex's
  per-server auto-approve (see `session/mcp-config.ts`).
- **`AgentMark`'s antigravity glyph** is a filled four-point star — a suggestion of the mark, not a
  copy. Same compromise as codex's two loops.
- **Sidebar / tab-bar labels** are "AGY" and "ag" next to "Codex" and "cx". Not obviously right;
  easy to change.
- **Docker sandbox is NOT generalized** (plan PR#5). `Dockerfile.sandbox` installs
  `@google/antigravity-cli`, but `buildDockerRunArgs` still hardcodes `"claude"` at the tail, so
  the sandbox remains claude-only. Deliberately out of scope here.
- **Global-config cleanup.** An earlier iteration of this branch wrote into `~/.gemini` (global
  `mcp_config.json`, a plugin dir, `plugins.json`) and into workspaces (`.mcp.json`, `.gemini/`).
  None of that ships. The maintainer's machine has been cleaned up by hand; anyone who ran an
  earlier build of this branch may want to check `~/.gemini/config/`.

## User Prompt

> Take a look at plans/feat-multi-agent-support.md

> This branch allows the user to start an antigravity in terminal, but MCP does not work. Review all the changes and find out why it does not work.

> Fix those problems, and remove unnecessary and redundant code. Keep it as simple as possible. Do not rely on global settings.

> It means we need to update that project-local file when the user enable those MCPs.

> We also remove changes we made to global settings or per-plugin settings.

> Re-review all the changes in this branch again to make it sure that this is simple, well-architected, and easy-to-maintain code.

## 実装メモ

- Why the MCP was dead before the fix, for the record: the bridge was a `.ts` file run as
  `node --import tsx <path>`, and node resolves a bare `--import tsx` against the CHILD's cwd — the
  user's project, which has no tsx — so it died with `ERR_MODULE_NOT_FOUND` in every directory but
  this repo, and agy sat on `MCP: 1 server(s) still connecting` for the life of the session. Plain
  `.mjs` run by `process.execPath` now, and every failure answers with a JSON-RPC error rather than
  silence.
- Dead code removed from the branch: `agents/antigravity-activity.ts` (imported only by its own
  spec — nothing ever tracked antigravity activity), the `--effort` flag (no caller), and the SSE
  MCP transport + `GET`-into-the-stateless-handler change in `mcp-routes.ts` (nothing ever wrote an
  `sse` URL, and the GET variant marked a session's tool groups on a bare GET).
- Also fixes a break the branch already had: `yarn typecheck:test` was failing because `SpawnDeps`
  grew two fields and `tool-group-reattach.spec.ts` was never updated. CI runs that script.
- No emojis anywhere (repo rule).
- `yarn lint`, `yarn typecheck`, `yarn typecheck:server`, `yarn typecheck:test`, `yarn build` and
  `yarn test` (6121 passed) all green on the tip.

work in mulmoterminal

## Summary by CodeRabbit

* **New Features**
  * Added Antigravity as a supported terminal agent.
  * Added Antigravity session creation in single-view, tabs, sidebar, collections, and grid layouts.
  * Added Antigravity WebSocket connectivity, session resume support, configurable executable/model settings, and background chat launching.
  * Added GUI tool integration through per-project MCP configuration.
* **Documentation**
  * Updated setup, agent selection, configuration, and environment variable guidance for Antigravity.
* **Tests**
  * Added coverage for Antigravity launching, sessions, routing, MCP tools, and UI behavior.


snakajima:
## Review round 1 — what changed

Thanks, both bots found real things. `b9c33543` and `191a8be5`:

**P1 — Antigravity reattach ignored input.** Confirmed and fixed. The handler returned early after
`reattachPty`, which only swaps the socket and replays the buffer; the `message`/`close` listeners
are installed by `startAndWire`, and the early return skipped it. After a page reload the terminal
printed output, took no keystrokes, and never detached or reaped on close. Reattach now goes
through `startAndWire` like codex's, via a `startAntigravityEntry` that mirrors `startCodexEntry`.

**P2 — background spawns dropped `mcpGroups`.** Worse than reported: the spawner syncs
`.agents/mcp_config.json` from that value, and the file is shared by every session in the
directory, so a collection action with the default `[]` CLEARED the entries other sessions were
using. The option is now required with no default — a caller that has not looked the groups up
fails to compile — and `spawnBackgroundChat` resolves them before spawning. The ws route also skips
the sync on a reattach, where rewriting a shared file would speak for sessions it knows nothing
about.

**CodeRabbit — encode the MCP path segments.** Done, and it surfaced a bug next to it: the bridge
required `MULMOTERMINAL_TOOL_GROUP` but never checked it, so an entry without one built a URL with
the literal text `undefined` as a path segment. Checked with the others now, and both segments are
percent-encoded.

**CodeRabbit — honour `ANTIGRAVITY_BIN` in pane recovery.** Done, and applied to all three agents
rather than only the new one: `claude` and `codex` had the same latent bug, since a pane reports
the running program's own name. The default name stays in the map either way, so setting the
variable never un-recognises sessions started before it.

**Also caught by CI, not by review:** `Dockerfile.sandbox` had grown `@google/antigravity-cli`,
which does not exist on npm (`agy` ships as a standalone binary), so the sandbox image failed to
build — for **claude**, the only agent the sandbox can currently run. Reverted in `191a8be5`.

## Deferred, on purpose

**P1 — session list / cold resume** → #1096. Not a regression (it has never worked on this branch),
and it needs more than wiring: agy records no cwd per conversation, so listing needs MulmoTerminal
to keep its own `session -> conversation -> cwd` record. A grid cell still reattaches on reload and
a tmux-backed session still survives a restart; what is missing is finding a single-view session
again, and resuming one whose tmux session is gone. `resolveAntigravitySession` already checks the
brain directory before resuming, so it declines rather than handing a stale id to
`agy --conversation`. Owner's call to land this PR first.

work in mulmoterminal

PR #1101, opened 2026-07-29
feat: tool-call history for codex and agy, fed by the MCP broker

snakajima:
## What this fixes

The tools pane's **Tool Call History** had exactly one writer: `POST /api/hook`, driven by the `Pre/PostToolUse` settings that only `claude` carries (`session/hook-settings.ts`). codex has no hook mechanism and neither does agy — the note at the top of `agents/codex-activity.ts` says so, which is why codex turn boundaries are read by tailing the rollout instead.

So for those two the history was empty however much they did, while **Available Tools** directly above it populated normally (that half comes from broker registration, not hooks). Half-alive reads as broken, not as missing.

## How

The broker is the one place their work reaches us: every GUI tool they call arrives at `/api/mcp/:sessionId` with the session id in the URL. `broker.ts` now records start/end around the dispatch, and the existing store, pub/sub channel and pane render it with no changes.

- **Refused tools** (a tool outside the URL's group) are recorded as failed — that is precisely what someone reading the history is looking for.
- **A failed dispatch** closes the entry too, for the same reason claude registers `PostToolUseFailure` alongside `PostToolUse`: otherwise it sits on \`running…\` for the life of the session. The error still propagates to the agent.
- The recorder is **fire-and-forget**: it sits on the tool-call path, and an agent waiting on `presentDocument` must not also wait on our history write.

This is a **partial** history — their Bash, their edits and their other MCP servers never touch us. An empty GUI-only history looks exactly like an agent that ran nothing, so the pane says which kind it is showing. `/api/tools` answers that, computed from the same predicate as the gate below, so the disclaimer and the history can never disagree.

## Not doubling claude — and why the gate is not an agent check

Claude's hooks match `\"\"`, so they already report `mcp__*` calls. Recording in the broker as well would list every GUI call twice, and not as duplicates anything could collapse: the broker mints its own uuid while the hook carries claude's `tool_use_id`.

The first version of the gate was \`agent === \"codex\" || agent === \"antigravity\"\`. That shipped and **half-worked**: agy recorded, codex did not. A codex **launcher chip** runs

\`\`\`
tmux … -- /bin/zsh -lc exec codex -c 'mcp_servers.mulmoterminal-render.url=\"…/api/mcp/render/<id>\"' …
\`\`\`

— `ws-routes.ts` hands it the same GUI MCP URLs, but the PTY comes from `spawnLauncherPty`, so its `PtyEntry` is honestly labelled `\"shell\"`. No agent name identifies a session started by a command the user wrote.

So the gate asks the real question instead: **is something else already recording these calls?** Exactly one thing is — claude's `--settings` hooks — and that is now recorded per spawn in `registry.hookedSessions`, with the agent name kept as a second, independent backstop for a claude session that outlived a restart and was never reattached in this process. Two signals, because a doubled entry is read as the agent having called the tool twice, while a missing one is merely missing.

## Also in here

The pane asked `/api/tools` while the agent was still starting — before its MCP client had connected and the server had learned any groups — and let that first \"none\" stand until something remounted it. That was the \"No GUI plugin tools enabled.\" a freshly launched session showed until it was redrawn. It now re-asks on the `tool-groups` announcement, the same fix the grid's Canvas button already carries.

## Verification

- `yarn format` / `yarn lint` / `yarn typecheck` / `yarn typecheck:server` / `yarn typecheck:test` clean; `yarn test` 6332 passing.
- New specs: the broker path driven through a real MCP client over an in-memory transport (completed / dispatch-failed / refused / no-recorder), the gate including the launcher-chip case that regressed, and the pane's re-ask and disclaimer. The re-ask spec was confirmed to fail without its fix.
- The diagnosis was made against the running server: the same harmless JSON-RPC probe sent to a live agy session and a live codex session added a row for one and no file at all for the other.

Not yet verified in a live browser session — the dev server needs a restart to pick this up.

work in mulmoterminal

## Summary by CodeRabbit

* **New Features**
  * Added GUI tool-call history recording for MCP sessions, including per-session control for worker vs non-worker flows.
  * Updated the tools API and Tools pane to surface whether a session’s history is GUI-only, with improved refresh behavior as tool data arrives.
* **Bug Fixes**
  * Prevented duplicate GUI call history entries and ensured refused/failed tool calls are captured correctly.
  * Made Tools pane and tool-group announcements more robust against out-of-order or malformed data.
* **Tests**
  * Added/expanded coverage for GUI call-history recording, broker behavior, and Tools pane race/GUI-only UI states.

### DIFF
--- a/server/index.ts
+++ b/server/index.ts
@@ -429,6 +429,7 @@ mountAppRoutes(app, {
   reap,
   // Defined further down; reached only from a request, which cannot arrive before listen().
   registerBackgroundSession: (id: string) => scheduledSessions.register(id),
+  agentOfSession: (id: string) => agentOfSession(id),
   setWorking,
   setWaiting,
   publishActivity,
--- a/server/mcp/broker.ts
+++ b/server/mcp/broker.ts
@@ -31,6 +31,8 @@ import { offeredTools, routeToolCall, SUBMIT_TRANSLATION_TOOL_NAME } from "./too
 import { toolGroupServerId, type ToolGroup } from "../../common/toolGroups.js";
 import { interpretToolEnvelope } from "./tool-envelope.js";
 import { isRecord } from "../../common/isRecord.js";
+import type { GuiCallRecorder } from "./gui-call-history.js";
+import { messageOf } from "../errors.js";
 
 // Shape of the dispatch route's response (POST /api/plugin/<tool>). `data` gates
 // whether a toolResult is published to the GUI; the rest is narration/metadata.
@@ -75,8 +77,15 @@ const SUBMIT_TRANSLATION_TOOL = {
  * @param opts.group  serve only one GROUP of the GUI tools (the `/api/mcp/<group>/<id>` URLs,
  *                    which exist so a directory can enable a subset through Claude Code's own
  *                    per-folder MCP config). Null/absent = every tool, the single view's URL.
+ * @param opts.history  feed each call into the tools pane's call history. Set only for agents
+ *                      that have no hooks of their own — see mcp/gui-call-history.ts, which
+ *                      also explains why claude must NOT get one.
  */
-export function buildGuiMcpServer(sessionId: string, baseUrl: string, opts: { submitTranslationTool?: boolean; group?: ToolGroup | null } = {}): Server {
+export function buildGuiMcpServer(
+  sessionId: string,
+  baseUrl: string,
+  opts: { submitTranslationTool?: boolean; group?: ToolGroup | null; history?: GuiCallRecorder | null } = {},
+): Server {
   const group = opts.group ?? null;
   // The advertised server name follows the id a group is expected to be registered under, so
   // what a user sees in `claude mcp list` matches what they wrote in their own config.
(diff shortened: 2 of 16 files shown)


snakajima:
Both findings were real and are fixed in ece240e9.

**Stale `/api/tools` reply** (codex + coderabbit, same issue) — correct, and the re-ask is what introduced it: two loads for the *same* session are now routine, and a session-id guard passes both. Fixed with a load generation, the same rule `useSessionFeed` already keeps for exactly this (#620). Regression test asserts the late pre-announcement reply does not put the empty list back; it fails with the generation check removed.

**`guiOnlyHistory: true` before the spawn** (codex, `tool-routes.ts:111`) — also correct, and worse than the report says in one way: for a single-view claude session the MCP URL carries no tool group, so no `tool-groups` event is ever published for it and the wrong label would never be re-asked away.

I did not fix it by publishing after the spawn, because the recording gate and the user-facing claim are not the same question. The gate runs when a tool call *arrives*, by which point the session is certainly spawned — it is fine as it is, and a claude session cannot call our MCP before `hookedSessions.add`, which happens synchronously with `ptys.set`. The *claim* is answered whenever the pane asks, which is early. So `historyIsGuiOnly` is now a stricter predicate than the gate: it additionally requires that we can see the session at all (a null agent means no pty and no tmux pane — not started, or gone). Nothing visible, nothing claimed.

The residual, stated plainly: a **single-view codex** session asked before its PTY exists shows no disclaimer until the pane is next re-rendered, since it too gets no `tool-groups` announcement. That is under-claiming rather than mislabelling, which is the direction this should fail in.

`yarn format` / `lint` / `typecheck` x3 clean, `yarn test` 6336 passing.

work in mulmoterminal

snakajima:
Fixed properly in b2f5ee61 — the verdict is right, and I should not have shipped this as an accepted residual.

The re-ask I added fixed the grid and left the single view exactly as broken, for the reason given: a single-view session connects on the all-tools URL, learns no tool group, and so published nothing on `tool-groups`. Its pane kept whatever the too-early `/api/tools` answered — an empty tool list, and for codex/agy no disclaimer — for the rest of the session.

**Now**: `mcp-routes.ts` announces on FIRST CONTACT from either route, once per session rather than per tool call.

The bare announcement deliberately carries **no `groups` field**, which is the one thing that makes this safe to broadcast. `TerminalGrid` reads that field to decide whether a cell has the Canvas MCP; an all-tools session genuinely has no groups to report, so sending `[]` would tell a cell that *can* draw that it cannot. So the grid now ignores a message without `groups`, while `ToolsPane` and `GuiPanel` read only the session id and act on both — `GuiPanel`'s tool hint had the identical too-early problem and gets the same fix for free.

New spec pins all four properties: the all-tools session is announced, the payload has no `groups` key, it fires once per session and not per call, and the group route still announces what it learned. It points `HOME` at a temp dir before importing (the group route persists under a `MULMOTERMINAL_HOME` resolved at import time, so without that the spec appends test uuids to the developer's own `~/.mulmoterminal`) and puts it back afterwards, since a vitest worker is shared between spec files.

`yarn format` / `lint` / `typecheck` x3 clean. `yarn test` 6340 passing, run three times over to confirm the env handling above left nothing flaky.

work in mulmoterminal

PR #1188, opened 2026-07-31
feat: make background workers findable, and fix ungrouped tools being hidden

snakajima:
Two independent changes, one commit each. Both come out of the same question — what happens to a truly background chat (`hidden: true`) once the single terminal view is gone — and the second was found by the owner while looking at the first.

They can be split into two PRs if preferred; they share only `registry.ts`.

---

## 1. `fix:` ungrouped tools were hidden from cells that had them

**Reported:** `spawnBackgroundChat` was missing from the tool list of a cell that could call it.

`narrowedTools` used *"is this a grid cell?"* as a stand-in for *"does it have only what its directory registered?"*. That held while a grid cell never received `--mcp-config`. It stopped holding twice over:

- **#1167** made an adopted chat report tool groups, and
- **#1187** gave a workspace cell the whole GUI MCP.

An ungrouped tool is reachable through **no group url**, so the filter dropped it from sessions that genuinely had it. Because every *other* tool now appears, it was the one conspicuous absence.

**The fix records the fact instead of inferring it.** Connecting to the all-tools url (`/api/mcp/:sessionId`) is proof of the whole MCP — `--mcp-config` is the only thing that hands that url out. That is now its own persisted mark, deliberately **separate from the group list**:

> "has all four groups" and "has the whole MCP" are different statements. A directory that registered all four group urls has the first and **not** the second.

A test pins exactly that distinction. It also shows the limit of the #1167 fix: recording the four groups was enough for the Canvas gate and could never be enough here, because a group list cannot express a tool that has no group.

**Audit of the whole map while here:** `spawnBackgroundChat` is the **only** ungrouped tool, and no map entry names a tool the server does not expose. It stays ungrouped on purpose — a plain grid cell has no business spawning sessions silently.

---

## 2. `feat:` background workers are findable, and a failed one says so

A hidden worker — `spawnBackgroundChat hidden:true`, a scheduled feed refresh — runs with no cell, no bold row, and nothing waiting to be clicked. That is right while it works. It is **wrong when it dies**: nothing pulls attention, so the failure is simply never learned.

It also has to be handled before the single view can go: the Background filter that lists these lives inside `App.vue`'s `!isGrid` block and would be deleted with it.

### Most of this already existed

Background workers are **already** in the grid launcher's "or resume here" list — `selectSessionRows` includes them in the cwd-scoped query with their own row cap — and every row **already** carries `hidden`. So adopting one into a cell needed no new mechanism: clicking the row already resumes it.

What was missing was being able to tell which rows they were.

### What was added

| | |
|---|---|
| **Label** | The launcher marks a background worker, and a failed one as `● failed`. Failed wins over the plain label — one badge, and it is the one that matters. |
| **Record** | A persisted set. Persistence is the point, not an optimisation: the value here is finding out **later**, and a live-only flag would be cleared by the very reap that discovers the failure. |
| **Outcome** | The `completion-hooks` seam that already exists for this ("a callback fired when a hidden worker finishes, so whoever dispatched it learns the outcome"). Its first-answer-wins contract already means what is needed: a finished turn reports success, and reaching teardown with no Stop means no turn ever completed. |
| **Signal** | A `worker-failed` notify kind. |

No clash with the feeds engine's own completion hook — that dispatches through its own spawner in `index.ts` and never this route.

`worker-failed` is checked **before** `closed` (the same teardown raises both) and deliberately needs **no prior state**. Every other kind compares against a baseline, and a worker never produces one — requiring it would drop the single notification that exists *because* nothing was watching.

### One decision worth stating

**The badge is always on; the sound is opt-in.** `notifyKinds.ts` is emphatic that a new kind must not start beeping at people who never asked, and it says so because that was violated once before. The durable half is what fixes "you would never find out"; the sound is what makes you look now. Easy to flip if the owner would rather it default on.

---

## On testing this

Each piece had a spec and none of them proved the pieces were **connected**. `worker-failure-wiring.spec.ts` now drives the real route and the real completion hook — the only place the decision "a hidden worker that never finished a turn has failed" actually happens — covering all four cases including the one that matters most: a worker whose turn finished first is **not** marked failed by the teardown that follows.

Two problems found and fixed in my own tests along the way, both worth naming:

- **The first three tests passed against the broken code.** They used `vi.waitFor`, which succeeds on the state *before* the merge runs. Switched to `flushPromises`; all three then failed, and passed with the fix.
- **Two specs set `process.env.HOME` to a temp dir**, which is process-wide and shared by every file in a vitest worker. Rewritten to mock `node:fs`, following `tool-group-reset.spec`'s existing pattern.

  **Correction to an earlier claim in this PR:** I first wrote that this was *causing* intermittent failures in unrelated specs. That was wrong, and it was concluded from one clean run against one failing run. Running `main` itself four times fails 2 of 4, each time on a different unrelated file (`files-browse`, `unusable-cwd-routes`, `mulmoscript`, `drop-route`). **The flakiness is pre-existing and unrelated to this PR** — and `vitest.config.ts` already documents the mechanism: files run in parallel across all cores, so "the victim is whichever test is running at peak load, not any one test". The `node:fs` rewrite still stands on its own merits (process-wide env mutation shared across a worker is a real hazard, and the repo has a precedent for avoiding it), but it fixed nothing.

Every new assertion was checked to fail without its fix.

**To see a real failure end to end** (the origin guard allows a same-machine caller with no `Origin`):

```
curl -s -X POST localhost:<port>/api/plugin/spawnBackgroundChat \
  -H 'content-type: application/json' \
  -d '{"message":"sleep for a while","hidden":true}'
# -> jsonData.chatId
tmux kill-session -t <chatId>
```

The PTY exit reaches `reap`, the hook fires with `didError: true`. An empty grid cell at `{workspace}` should then show the row as `● failed`.

## Verified

- `yarn test` — 7053 tests pass. Note the pre-existing parallel-load flakiness described above: a run may fail on one unrelated file, and `main` does the same.

**Not verified:** anything in a running app. The tool list and the badge are covered by tests, not by eyes.

---

## 3. `refactor:` the worker-status wire shape moved to `common/` (review follow-up)

CodeRabbit: `hidden` and `failed` are written by the server and read by the launcher, so they are a wire shape — and CLAUDE.md is explicit that a type both sides decide from belongs in `common/` rather than mirrored into `server/` and `src/`. It was right; they were mirrored.

The two sides genuinely differ, so this follows the pattern the rule prescribes (`common/sourceExtensions.ts`): share the core, keep each side's extras local, and **pin the asymmetry with a test**. The server always fills both fields; the client's are optional, because a page left open across an upgrade parses rows from a server that predates them.

The pin is split across the two test projects, and not for tidiness: a spec in the app project that imports a server type pulls node-pty's Node globals into that program, and `window.setTimeout` in an unrelated component then resolves to Node's overload and fails to compile. I hit exactly that while writing it.

**Not taken from the same review:** replacing `●` in the failed badge with a Material Symbol. CLAUDE.md allows it explicitly — *"Compact status notation stays text, not icons: `⎇ main ●3 ↑2`, `●` unsaved dots"* — and the same file already renders `● open` and a `●` dirty dot two lines away.

work in mulmoterminal

### DIFF
--- a/common/notifyKinds.ts
+++ b/common/notifyKinds.ts
@@ -9,10 +9,13 @@
 //   command-failed — a Run cell's command exited non-zero.
 //   session-exited — a session's PTY ended. Closing a cell yourself goes through the same
 //                    path, so this one fires on a deliberate close too.
+//   worker-failed  — a hidden background worker ended without ever completing a turn. The one
+//                    kind raised for a session with no cell on screen: a worker is invisible by
+//                    design, so nothing else would ever say so.
 //   pr-ci-failed   — a directory's PR phase became ci-failing. The phase poll runs only
 //                    while the roster is on screen, so a failure that lands while you are
 //                    in another view is not seen.
-export const NOTIFY_KINDS = ["finished", "waiting", "command-done", "command-failed", "session-exited", "pr-ci-failed"] as const;
+export const NOTIFY_KINDS = ["finished", "waiting", "command-done", "command-failed", "session-exited", "worker-failed", "pr-ci-failed"] as const;
 
 export type NotifyKind = (typeof NOTIFY_KINDS)[number];
 
--- a/common/workerStatus.ts
+++ b/common/workerStatus.ts
@@ -1,0 +1,35 @@
+// What a session row says about a BACKGROUND WORKER, as it travels from the server to the UI.
+//
+// Both sides decide from these two fields and they decide different things, so this is the shared
+// CORE rather than either side's whole row:
+//   - the server fills them per row from the persisted marks (session-reads.ts), and
+//   - the launcher's resume list renders them as the `background` / `● failed` labels
+//     (CellLaunchForm.vue), which is the only place a worker can be found once the single view
+//     is gone.
+// Each side keeps its own extras (title, mtime, working/waiting/event) next to its import.
+//
+// Shared rather than spelled twice because they are a WIRE shape: the server writes them, the
+// browser reads them, and two copies of that is the drift `common/` exists to prevent. The two
+// sides are deliberately not identical, though — see WorkerStatus vs PartialWorkerStatus below,
+// and the spec that pins the asymmetry.
+export interface WorkerStatus {
+  /** Spawned as a hidden background worker (spawnBackgroundChat hidden:true, a scheduled
+   *  refresh). It is still listed, but never bold and never unread — a helper finishing should
+   *  not pull the user's attention. */
+  hidden: boolean;
+  /** That worker ended without ever completing a turn. The counterpart to `hidden`: a worker is
+   *  quiet by design, so this is the one outcome the quiet is wrong for, and it rides on the row
+   *  so a picker can say WHICH worker failed rather than making the user open each one. */
+  failed: boolean;
+}
+
+/**
+ * The same two fields as the CLIENT may receive them.
+ *
+ * Optional on purpose, and the asymmetry is load-bearing rather than laziness: the server always
+ * fills both, but a browser also parses rows from an OLDER server (a page left open across an
+ * upgrade, the phone's cached list), where neither field exists. Declaring them required there
+ * would make the type lie about what can arrive, and the labels are absence-tolerant anyway —
+ * "no badge" is exactly right for a row that never said.
+ */
+export type PartialWorkerStatus = Partial<WorkerStatus>;
--- a/server/routes/app-routes.ts
+++ b/server/routes/app-routes.ts
@@ -130,7 +130,6 @@ export function mountAppRoutes(app: Express, deps: AppRouteDeps): void {
     spawnCodexPty: deps.spawnCodexPty,
     spawnAntigravityPty: deps.spawnAntigravityPty,
     registerBackgroundSession: deps.registerBackgroundSession,
-    publish: (c, d) => deps.publish(c, d),
   });
 
   // presentHtml View's source-editor dispatch (loadHtml/saveHtml) on
--- a/server/routes/app-routes.ts
+++ b/server/routes/app-routes.ts
@@ -44,7 +44,16 @@ import { mountNotificationRoutes } from "../backends/notifier.js";
 import { mountWhisperRoutes } from "../backends/whisper.js";
 import { mountSchedulerRoutes } from "../backends/scheduler.js";
 import { mountFilesRoutes } from "../backends/files.js";
-import { hookedSessions, ptys, sessionToolGroups, sessionToolGroupsHydrated, devTerminalSessions, devTerminalSessionsHydrated } from "../session/registry.js";
+import {
+  hookedSessions,
+  ptys,
+  sessionToolGroups,
+  sessionToolGroupsHydrated,
+  devTerminalSessions,
+  devTerminalSessionsHydrated,
+  hasAllGuiTools,
+  allToolsSessionsHydrated,
+} from "../session/registry.js";
 import { mountShortcutsRoutes } from "../backends/shortcuts.js";
 import { mountDecisionRoutes } from "./decision-routes.js";
 import { mountTranslationRoutes } from "../backends/translation.js";
@@ -121,6 +130,7 @@ export function mountAppRoutes(app: Express, deps: AppRouteDeps): void {
     spawnCodexPty: deps.spawnCodexPty,
     spawnAntigravityPty: deps.spawnAntigravityPty,
     registerBackgroundSession: deps.registerBackgroundSession,
+    publish: (c, d) => deps.publish(c, d),
   });
 
   // presentHtml View's source-editor dispatch (loadHtml/saveHtml) on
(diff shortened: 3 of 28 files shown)


snakajima:
Addressed in ebc4e4c1, and one correction to something I wrote in the PR body.

## Taken: the shared wire type

Right, and it is the project's own rule — CLAUDE.md says a type both sides decide from belongs in `common/` rather than mirrored into `server/` and `src/`, and `hidden` / `failed` were mirrored.

Followed the pattern the rule prescribes (`common/sourceExtensions.ts`): share the core, keep each side's extras local, and **pin the asymmetry with a test**. The server always fills both fields; the client's stay optional, because a page left open across an upgrade parses rows from a server that predates them — declaring them required there would make the type lie about what can arrive.

The pin is split across the two test projects, and that is load-bearing rather than tidiness: a spec in the app project that imports a server type pulls node-pty's Node globals into that program, and `window.setTimeout` in an unrelated component then resolves to Node's overload and fails to compile. I hit exactly that writing it, which is why both files name the same shared type instead of one file comparing them.

## Not taken: Material Symbol for `●`

Declining this one. CLAUDE.md permits it explicitly, in the same section as the no-emoji rule:

> Compact status **notation** stays text, not icons: `⎇ main ●3 ↑2`, `●` unsaved dots, `−12` diff counts. Icons there are bigger and slower to scan.

And `CellLaunchForm.vue` already renders `● open` (line 511) and a `●` dirty dot (line 461) — the new badge sits two lines from one of them. Swapping it for an icon would make this one status marker inconsistent with its neighbours.

## Correction to the PR body

I wrote that two of my specs were *causing* intermittent failures in unrelated specs by setting `process.env.HOME`. **That was wrong.** I concluded it from one clean run against one failing run, which cannot distinguish a cause from background noise.

Running `main` itself four times fails **2 of 4**, each time on a different unrelated file (`files-browse`, `unusable-cwd-routes`, `mulmoscript`, `drop-route`). The flakiness is pre-existing and has nothing to do with this PR — and `vitest.config.ts` already documents the mechanism: files run in parallel across all cores, so "the victim is whichever test is running at peak load, not any one test".

The `node:fs` rewrite stays, because process-wide env mutation shared across a worker is a real hazard and the repo has a precedent for avoiding it. But it fixed nothing, and the PR body now says so.

## Verified

- `yarn test` 7053 pass, `yarn typecheck` / `:test` / `:server` pass, `yarn lint` 0 errors, `yarn build` succeeds

snakajima:
Both findings were real. Fixed in 9d86162a.

## 1. `/api/sessions` did not await `failedWorkersHydrated`

Correct, and it was an inconsistency I introduced. The route already awaits the background set **deliberately** — its own comment says the background set is awaited for both queries because "it decides a flag on the row rather than whether the row is listed" — and `failed` is the same kind of flag, from the same kind of persisted log.

Without the await, the first list after a restart can serve a failed worker as `failed: false`. That is precisely the case the record exists for: the value of persisting it is finding out **later**, and the most likely "later" is the first look after a restart.

## 2. Two teardown messages could race

Also correct, and my code comment claiming `worker-failed` was "checked before closed" was misleading — that ordering holds *within* one message and says nothing about two.

**Now one message carries both facts:** `{ id, working: false, event: "closed", failed }`.

The event stays `"closed"` on purpose. I first tried replacing it with `"worker-failed"`, then found `sessionActivity.ts` keys on `"closed"` to drop a reaped session — swapping it would have left the activity entry stranded. So the outcome rides along as a field instead, and `notifyKindOf` lets the specific answer win. It still deliberately requires no prior state: a worker has no cell and never produces one.

**Consequence worth naming:** the hook now RECORDS only, and reap announces. That makes the recorder's synchrony a real contract — reap reads the flag on the very next line — so it is pinned by its own test rather than left implied:

> `it("records SYNCHRONOUSLY, so reap can read the flag on the next line")` — calls the hook **unawaited**, exactly as reap does.

An async recorder would otherwise publish `failed: false` and the failure would go unannounced, silently, which is the failure mode this whole feature exists to remove.

## Verified

- `yarn test` 7054 pass, `yarn typecheck` / `:test` / `:server` pass, `yarn lint` 0 errors, `yarn build` succeeds
- One run failed on `server/backends/mulmoscript.spec.ts`, which passes in isolation and is the pre-existing parallel-load flakiness described in the PR body (`main` does the same, 2 of 4 runs, different file each time)

snakajima:
Right, and this one is the worst defect in the PR. Fixed in ced1ef4d.

## What was wrong

The only success signal a PTY-hosted agent gives us is a finished turn reported by **Claude Code's Stop hook**. `hook-routes.ts` says so directly, and its comment even recorded the assumption I broke:

> No-op unless a hook is registered, **which only a hidden feeds worker ever does**.

codex and antigravity have no hook mechanism at all — they never post to `/api/hook`. So registering the recorder for every hidden spawn meant every **successful** hidden codex/antigravity worker reached teardown having reported nothing, and was marked **failed**.

Not a rare race: it is every run of those agents. The badge would have been wrong more often than right, and a signal like that is worse than the silence it replaced — it teaches you to ignore it.

## The fix

The recorder is registered for **claude only**, and that is a correctness limit rather than a scope decision. A non-claude hidden worker keeps today's behaviour: no failure signal. Giving it one needs a completion signal for those agents first, and none exists.

Pinned with a test per agent (`it.each(["codex", "antigravity"])`), verified to fail without the guard.

Also corrected the stale comment in `hook-routes.ts`: it now says a hidden claude `spawnBackgroundChat` registers a hook too, and why claude is the only agent that can ever report success through that endpoint.

## Verified

- `yarn test` 7056 pass, `yarn typecheck` / `:test` / `:server` pass, `yarn lint` 0 errors, `yarn build` succeeds

PR #1189, opened 2026-07-31
feat: adopt sessions the server spawned while nothing was open

snakajima:
PR3b of the single-view removal (design note: `plans/feat-remove-single-view.md`). Follows #1167, #1186, #1187, #1188.

This is the half PR3a could not cover.

## The gap

A chat started from the collections UI is placed by the browser that asked for it. That is the common case and **only** that case. Three others start a **visible** chat with no tab open at all:

- an agent calling `spawnBackgroundChat` from another session,
- a scheduled task firing at 3am,
- the phone.

Today those land in the chat sidebar. Once the single view is gone, they have nowhere to appear.

## The shape

The server remembers that a session is waiting, and the next grid to load adopts it.

**Marked** at the three server-side spawns that start a visible session: `spawnBackgroundChat` when NOT hidden, the phone's `startChat`, and the scheduler's. The translation worker is deliberately absent — and a spec pins that a hidden worker can *never* be marked, because "happens not to be marked" and "cannot be marked" read identically until someone adds a caller, and the difference is a background refresh silently opening a terminal.

**Cleared on ANY attach**, not just a grid one: *unplaced* means nobody is looking, and a viewer is a viewer. It lands at `ws-routes`' four attach points — the same choke points `markDevTerminalSession` already uses, covering new, resumed and reattached. This is what stops a browser-placed chat coming back as a duplicate cell on the next load.

**`GET /api/sessions/unplaced`** answers the marker directly, rather than having the grid diff "all sessions" against its own state. A diff would sweep up ordinary sessions and change what a reload does to a normal cell — the invariant this whole line of work is written around.

**The grid asks on ACTIVATE, not mount.** It is kept alive across route changes, so mount fires once per page load and would miss everything spawned while the user was elsewhere in the app.

## Two things the plan did not anticipate

**1. The marker needs two append-only logs, not one.**

`MULMOTERMINAL_HOME` is shared between server instances, so these logs cannot be rewritten — and hydration reads the unplaced log as it *was*, including ids that a later line has since answered. A single log would therefore re-adopt, on every load, **every session any grid has ever taken**. So `placed` is its own log and `unplacedSessionIds()` is the difference.

Pinned directly: `does not hand back a session that was placed BEFORE this process started`.

**2. No new pub/sub channel, which the plan opened with.**

The plan called for a `publishToOne` sibling of `LAUNCH_TERMINAL_CHANNEL`. It is not needed. Reconciliation on activate covers everything except a chat spawned while the user is *already sitting on* `/terminals` — and a channel for that window would be a second mechanism that has to be kept from double-placing what the browser just placed. Worth adding later if that window proves annoying; not worth it up front. Recorded in the plan.

## Duplicate cells

Guarded on **both** paths — the live placement and the sweep. Two cells for one session fight over its socket: the server supersedes the prior one, so the older cell goes dead while still looking live. The sweep's guard is not redundant with the server clearing the mark, because a tab may still hold a cell whose attach has not landed yet.

## Cap policy

Still unanswered, and this PR does not force it: a full grid simply leaves the session marked, so the next load with room adopts it. PR5 is where it has to be decided, since that is where the single-view fallback disappears.

## Verified

- `yarn test` — 7066 tests pass (+12)
- The grid-side adoption test was checked to fail with the `onActivated` hook removed

Note the pre-existing parallel-load flakiness in this suite (documented in `vitest.config.ts`, and `main` shows it too): a run may fail on one unrelated file.

**Not verified:** anything in a running app. The most convincing check is a real one — let a scheduled task fire, or `curl` a visible `spawnBackgroundChat` with no tab open, then open the grid and see the cell appear.

work in mulmoterminal

## Summary by CodeRabbit

* **New Features**
  * Server-created chat sessions can now appear in the grid automatically when it is activated.
  * Session details, including agent and working directory, are preserved during adoption.
* **Bug Fixes**
  * Prevented sessions from appearing in multiple grid cells.
  * Prevented already-attached sessions from being adopted again.
  * Grid activation retries adoption when the grid is temporarily full or loading fails.
* **Tests**
  * Added coverage for session persistence, adoption, deduplication, attachment, and restart scenarios.

### DIFF
--- a/plans/feat-remove-single-view.md
+++ b/plans/feat-remove-single-view.md
@@ -335,7 +335,33 @@ store — it is not client-only here.
    So a `presentCollection` result renders in a grid cell exactly as it does in the single view.
    Nothing to build here; the seeding work is all that is left.
 
-#### PR3b — the durable half
+#### PR3b — the durable half — **DONE**
+
+Implemented, with two departures from what is written below:
+
+1. **No new pub/sub channel.** The plan opened with a `publishToOne` sibling of
+   `LAUNCH_TERMINAL_CHANNEL`. It is not needed: reconciliation runs on **activate**, not mount, so
+   switching to the grid picks up everything spawned while the user was elsewhere — and if the grid
+   is not mounted there is nothing to deliver to anyway. A channel would only shorten the window
+   for a chat spawned while the user is *already sitting on* `/terminals`, at the cost of a second
+   mechanism that has to be kept from double-placing what the browser just placed. Worth adding
+   later if that window proves annoying; not worth it up front.
+2. **The marker needs a second log, not one.** `MULMOTERMINAL_HOME` is shared between server
+   instances so these logs are append-only, and hydration reads the unplaced log as it was —
+   including ids that a later line has since answered. A single log would therefore re-adopt, on
+   every load, every session any grid has ever taken. So "placed" is its own append-only log and
+   `unplacedSessionIds()` is the difference. A spec pins exactly that (`does not hand back a
+   session that was placed BEFORE this process started`).
+
+Also worth recording: the mark is cleared on **any** attach, not just a grid one — "unplaced" means
+nobody is looking, and a viewer is a viewer. That is what stops a browser-placed chat coming back
+as a duplicate on the next load, and it lands at `ws-routes`' four attach points, the same choke
+points `markDevTerminalSession` uses.
+
+The cap policy is still unanswered, and this PR does not force it: a full grid simply leaves the
+session marked, so the next load with room adopts it. PR5 is where it has to be decided.
+
+---
 
 The server-side unplaced marker + reconciliation on grid load, i.e. the case where no tab was
 open at all. Details below.
--- a/server/index.ts
+++ b/server/index.ts
@@ -66,6 +66,7 @@ import {
   sessionCwd,
   sessionMemos,
   sessionMemosHydrated,
+  markUnplacedSession,
 } from "./session/registry.js";
 import { hydrateClearedTranscripts } from "./session/cleared-transcripts.js";
 import { runWithHiddenMarker } from "./session/hiddenMarker.js";
@@ -560,6 +561,9 @@ initFeedsBackend({ workspace: CLAUDE_CWD, spawnWorker: feedsSpawnWorker });
 const remoteHostSpawnChat = (message: string) => {
   const sessionId = randomUUID();
   spawnClaudePty(sessionId, null, null, { initialPrompt: message });
+  // Started from the PHONE, so by definition no browser placed it. Marked so the next grid to
+  // load adopts it instead of leaving a live agent with nowhere to appear.
+  markUnplacedSession(sessionId);
   return { chatId: sessionId };
 };
 // The phone's remote terminal view (#435). Both accessors live here because the PTY table
@@ -774,6 +778,9 @@ function spawnScheduledChat(message: string): void {
   try {
     spawnClaudePty(sessionId, null, null, { initialPrompt: message });
     scheduledSessions.register(sessionId);
+    // The clearest case for the unplaced mark: a task firing at 3am has no browser to place its
+    // chat, and without this the only trace would be a row in a list nobody is looking at.
+    markUnplacedSession(sessionId);
   } catch (err) {
     console.error(`[scheduler] failed to spawn chat for a scheduled task: ${messageOf(err)}`);
   }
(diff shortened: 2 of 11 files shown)


snakajima:
Both findings were real. Fixed in 43faae5b.

## 1. The minted id — this one was the serious half

`resolveClaudeSession` (and codex/antigravity, all through `resolveReattachableId`) mints a **fresh** session id when the requested one can be served neither from a live pty nor from an on-disk transcript. A visible spawn that died before writing a transcript is exactly that — and exactly the kind of session that ends up unplaced.

So the id the grid **adopted** and the id the session ends up with differ, and clearing only the new one left the requested one marked forever. The consequence is not a stale flag: every activate would adopt it again and mint another session. The grid fills with cells nobody asked for, indefinitely.

Fixed for all four attach points through one helper, rather than the rule spelled four times — the launcher has a `requested` id too.

## 2. The agent

Correct. The endpoint answered `agent: null` once the `PtyEntry` was gone, and the grid reads that as claude, so an adopted codex or antigravity session would reconnect on the wrong endpoint.

The missing entry is precisely the case this marker exists for — the server restarted, or the pty was reaped, before anyone opened a tab — so the agent is now recorded **with** the id. The log becomes `<session id> <agent>` per line; a line with no agent is one written before the field existed, and those sessions were all claude. Both the round trip and that default are pinned.

The live `PtyEntry` still wins when there is one: it is the process actually running.

## Verified

- `yarn test` 7072 pass, `yarn typecheck` / `:test` / `:server` pass, `yarn lint` 0 errors, `yarn build` succeeds

snakajima:
Fixed in 8a3f828f. Real, and an inconsistency with every sibling in that file — `markUnplacedSession`, `markBackgroundSession`, `markDevTerminalSession` and `markFailedWorker` are all no-ops once the id is known; `markSessionPlaced` was not.

That matters more here than it would elsewhere, because this one runs at **all four** ws attach points. A long-lived session reconnecting — a reload, a network blip, a page switch in the grid — appended a line each time, so the log grew with attach count rather than session count, and `/api/sessions/unplaced` waits on hydrating it.

## The test found a second bug, in itself

Worth reporting rather than quietly fixing. The spec's helper matched the log file with `endsWith()`, and:

```
"unplaced-sessions.json".endsWith("placed-sessions.json") === true
```

So it read **both** logs as one, and every earlier assertion about the placed log had been silently counting the unplaced one too. Matched on the exact basename now. The idempotence test failed on first run because of this, not because of the fix — which is the only reason it surfaced.

Both the guard and the corrected helper were checked by removing the guard and watching the test fail.

## On the red ubuntu build

`lint-and-build (ubuntu-latest)` failed on 43faae5b in `yarn install --frozen-lockfile`, with no lockfile error in the log — an infrastructure hiccup rather than the code. macOS passed the same commit. This push re-runs it.

## Verified

- `yarn test` 7073 pass, `yarn typecheck` / `:test` / `:server` pass, `yarn lint` 0 errors, `yarn build` succeeds

snakajima:
Addressed in aaa5386e — one taken, one already fixed, and the nitpick taken with a change of location.

## Taken: the in-flight guard

Real. `onActivated` fires again when the user leaves the grid and comes straight back, and both runs read `cells` before either inserts — so the per-row duplicate guard cannot see the other, and the same unplaced session gets two cells fighting over one socket.

Pinned by a test that holds the fetch open across two activations and asserts the second sweep is refused. Verified to fail with the guard removed.

## Already fixed: `markSessionPlaced` appending unconditionally

That landed in **8a3f828f**, one commit before this review ran. Codex re-checked the same commit and confirmed it. No change needed.

## Taken, but relocated: the test reimplementing `markAttachedSessionPlaced`

Right that a copy of the rule in the test can keep passing while the real one drifts.

Importing it from `ws-routes` would not have fixed that, though, and this is the interesting part: the test calls `vi.resetModules()` to get a fresh registry, so a helper imported from a *different* module graph would mark a registry instance the assertions never read — the tests would go green while testing nothing. (It also drags in enough of `ws-routes` to need `existsSync` on the `node:fs` mock.)

So the helper moved into the registry instead, which is where it belonged: it is a rule about the **marks** — clear both ids, because the resolvers may have minted a fresh one — not about ws routing. Now both come from the same reset graph and the test drives the real rule.

## On the earlier red build

`lint-and-build (ubuntu-latest)` has since passed on 8a3f828f, confirming that failure was the transient `yarn install` and not the code.

## Verified

- `yarn test` 7074 pass, `yarn typecheck` / `:test` / `:server` pass, `yarn lint` 0 errors, `yarn build` succeeds

PR #1193, opened 2026-07-31
feat: let the grid be the view underneath a full-screen overlay

snakajima:
PR4 of the single-view removal (design note: `plans/feat-remove-single-view.md`). Follows #1167, #1186, #1187, #1188, #1189.

Everything the single view owns has to exist in the grid before PR5 can delete it. The full-screen overlays were the last piece: they could only appear by taking the grid off screen.

## The hoist alone does nothing

The plan said "hoist the overlays out of the `!isGrid` block". That is necessary and not sufficient.

Every overlay is **route-driven** — `/collections`, `/prs`, `/files`, `/wiki`, `/accounting`. Opening one leaves `/terminals`, and `isGrid` was `route.name === "terminals"` — so the grid came off screen and **the single view mounted behind the overlay**. Moving the overlays out only helps if the grid can also *be* the backdrop.

## `viewIsGrid` already answered this, and only the toolbar was asking

`overlayOrigin.ts` already exports exactly the right predicate — *is the view UNDERNEATH the current screen the grid?* — which follows the overlay's **origin** rather than its own route, so an overlay opened from the grid keeps the grid behind it.

Only `AppToolbar` read it. So for the entire time an overlay was open, the header and the shell behind it disagreed: the toolbar showed the grid's buttons while the single view's terminal was mounted behind. `App.vue` binds to the same value now, and the two agree.

`viewIsGrid` had **no tests**, and it decides which shell renders rather than which buttons show. Six cases added — grid, chat, opened-from-grid, opened-from-chat, no-origin, and an overlay→overlay hop — verified to fail against a broken predicate.

## The return-path fallback moves to the grid

An overlay with no recorded origin (a direct load, a link, a restored tab) used to fall back to chat. It falls back to the grid now: that is where `/` already sends a fresh load, and the fallback also decides what renders *behind* such an overlay — leaving it on chat would put the single view back on screen under an overlay opened from a link.

One existing regression test had to be rewritten rather than just re-pointed. It captured `/terminals` as the earlier origin and asserted the fallback was `chat`; with the fallback now `terminals`, the two answers coincide and it would pass whether the stale origin was ignored **or reused** — which is the whole thing it exists to catch. It captures `/chat` and expects `terminals` now, deliberately opposite values.

## Two things deliberately not done

**`AppSettingsModal` is not hoisted.** `GridView` renders its own, so hoisting this one would show two on `/terminals`. Five overlays moved, not six.

**The `AppToolbar` switch to `{name:"chat"}` stays.** Removing it here would strand the single view — unreachable but still present. That is PR5's job.

Also already true, so no work: **`ToolsPane` is already beside the zoomed cell** (`TerminalGrid.vue:867`). The plan listed it as an open decision; earlier work had answered it.

## One regression this introduced, and fixed

With the grid surviving under an overlay, a chat started from the collections browser is placed into a grid **the user cannot see**.

The queue-and-navigate path had been closing that overlay *by accident* — the grid was unmounted, so there was no handler, so the request queued and pushed to `/terminals`. The moment the grid stopped unmounting, that stopped happening and the chat went behind the overlay silently.

`placeSpawnedChat` now navigates whenever the user is not already on `/terminals`, whether or not a mounted grid took the request. Pinned by a test that places into a mounted grid from a `collections` route and asserts the push.

## Verified

- `yarn test` — 7086 tests pass (+12)
- Each new assertion checked to fail without its fix

Note the pre-existing parallel-load flakiness in this suite (documented in `vitest.config.ts`; `main` shows it too): a run may fail on one unrelated file.

**Not verified:** anything in a running app, and this PR is the most visual one yet. Worth opening a collection from the grid and confirming the grid is what sits behind it — and that starting a chat from that browser lands you on the grid with the new cell.

work in mulmoterminal

## Summary by CodeRabbit

* **New Features**
  * Full-screen overlays now appear over the grid while preserving the underlying view.
  * Overlay navigation without a saved origin now returns to the Terminals view.
  * Spawned chats now reliably navigate to Terminals when needed.

* **Bug Fixes**
  * Improved navigation when opening Files or overlays from different views.
  * Prevented stale navigation history from sending users back to an outdated location.

### DIFF
--- a/plans/feat-remove-single-view.md
+++ b/plans/feat-remove-single-view.md
@@ -393,7 +393,32 @@ open at all. Details below.
   background worker; the skill button's use of it is a workaround, not the same thing. Do not
   collapse the two while removing the workaround.
 
-### PR4 — make the grid self-sufficient
+### PR4 — make the grid self-sufficient — **DONE**
+
+Three of the six bullets below needed no work, and the one real change was bigger than "hoist":
+
+- **The hoist alone does nothing.** Every overlay is ROUTE-driven, so opening one leaves
+  `/terminals` — and `isGrid` was `route.name === "terminals"`, which meant the grid came off
+  screen and the single view mounted behind the overlay. Moving the overlays out of the `!isGrid`
+  block only helps if the grid can also BE the backdrop, so `App.vue` now binds to `viewIsGrid`
+  (the existing "which view is underneath" answer, which follows the overlay's origin). The two
+  halves had to travel together, exactly as the comment at `App.vue:39` warned.
+- **`viewIsGrid` had no tests**, and it now decides which shell renders rather than which buttons
+  show. Six cases added, verified to fail against a broken predicate.
+- **`AppSettingsModal` must NOT be hoisted** — `GridView` renders its own, so a hoist would show
+  two on `/terminals`. Five overlays moved, not six.
+- **`ToolsPane` is already beside the zoomed cell** (`TerminalGrid.vue:867`). The open question in
+  this section was already answered by earlier work.
+- **The AppToolbar `{name:"chat"}` switch STAYS**, deferred to PR5. Removing it here would strand
+  the single view — unreachable but still present — which is PR5's job, not this one's.
+
+**One regression this introduced, and fixed.** With the grid surviving under an overlay, a chat
+started from the collections browser was placed into a grid the user could not see: the
+queue-and-navigate path used to close the overlay *by accident*, and it stopped being taken the
+moment the grid stopped unmounting. `placeSpawnedChat` now navigates whenever the user is not
+already on `/terminals`, whether or not a mounted grid took the request.
+
+---
 
 Everything the single view currently owns has to exist in the grid *before* PR4.
 
--- a/src/App.vue
+++ b/src/App.vue
@@ -12,7 +12,6 @@ import WikiBrowseOverlay from "./components/WikiBrowseOverlay.vue";
 import PrsOverlay from "./components/PrsOverlay.vue";
 import FilesOverlay from "./components/FilesOverlay.vue";
 import GridView from "./components/GridView.vue";
-import { useRoute } from "vue-router";
 import AppSettingsModal from "./components/AppSettingsModal.vue";
 import AppToolbar from "./components/AppToolbar.vue";
 import { useSessions, type Filter } from "./composables/useSessions";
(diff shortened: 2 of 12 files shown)


snakajima:
All four findings were real, and they are all the same consequence of this PR. Fixed in 2f6b7f8e.

## The shared cause

The grid now stays mounted underneath a full-screen overlay. Every place that treated **"mounted"** as **"on screen"** became wrong at once — and three of the four were doing the right thing *by accident* before: the grid unmounted, so there was no handler, so the queue-and-navigate branch ran and brought the grid on screen. That accident stopped the moment the grid started surviving an overlay.

## 1. `placeSpawnedChat` overrode the full-grid fallback

Mine, and the worst of the four. A full grid falls back to showing the chat in the single view, and my push dragged the user straight off the view the session had just appeared in — a navigation added for overlays beating a fallback that predates it.

The handler now reports whether it took the request; a refusal navigates nowhere. Pinned exactly as suggested: handler mounted, grid full, non-terminals route.

## 2. The unplaced sweep lost its trigger (P1)

Right. It ran from `onActivated`, and the grid no longer deactivates when an overlay opens — so a scheduled or server-spawned session created while the user was in PRs or the wiki kept running without a cell until they switched to Chat and back, or reloaded.

Driven by the actual route now, which covers both the first mount and every return from an overlay.

## 3. Grid shortcuts reached the hidden grid (P2)

Right, and the CodeMirror detail is the sharp end of it: its editable surface is `contenteditable`, which `isEditableTarget` does not exclude, so **typing in an editor was reaching the shortcuts** — up to `terminal-close` closing the hidden grid's zoomed cell. Gated on the actual route.

## 4. Remote launches landed behind the overlay (P2)

Right, and the same shape as (1): `openTerminalAt` took the handler branch and returned without navigating, so the phone's request was reported as served while the terminal appeared where nobody could see it. Same rule applied.

## One implementation note

`GridView` reads the router **singleton** rather than `useRoute()`. Its own specs mount it without the router plugin, so `useRoute()` breaks 20-odd existing tests — `overlayOrigin` reads the singleton for the same reason. The specs that exercise shortcuts now push `/terminals` first, which is honest: those shortcuts only apply while the grid is what you are looking at.

## Verified

- `yarn test` 7088 pass, `yarn typecheck` / `:test` / `:server` pass, `yarn lint` 0 errors, `yarn build` succeeds

Issue #202, opened 2026-07-04
feat: add arbitrary MCP servers mid-conversation and run them in a sandbox, as MulmoClaude does


isamu:
## Background — how MulmoClaude does it today

MulmoClaude starts `claude -p` (or an interactive session) inside a Docker sandbox and calls host-side tool implementations through the mcp-config, getting both "sandbox isolation" and "full use of the host's capabilities". The details are written up in this new doc:

- https://github.com/receptron/mulmoclaude/blob/main/docs/claude-docker-boundary.md (PR #1966)

In summary:

| What | Where it runs |
|---|---|
| The `claude` CLI itself | container |
| Claude's built-in `Bash` / `Read` / `Write` / `Edit` / `Grep` / `Glob` | container |
| The mulmoclaude MCP stdio subprocess (bridge proxy) | container |
| The mulmoclaude MCP tool **implementations** | **host** (Express) |
| User HTTP MCP | **host** (reached via URL rewriting) |
| User stdio MCP (`hostExecInDocker: true`) | **host** (via the stdio<->HTTP gateway) |

The experience is: add a user MCP in the Settings UI, and Claude can call it straight away mid-conversation. Reference code:

- `server/agent/config.ts:116` `rewriteLocalhostForDocker` — URL rewriting for HTTP MCP
- `server/agent/config.ts:276` `buildMulmoclaudeServer` — starts the mulmoclaude broker MCP as a stdio subprocess inside the container and forwards to `MCP_HOST=host.docker.internal:PORT/api/*`
- `server/agent/config.ts:186` `startStdioHttpShim` — runs a user stdio MCP on the host behind a shim (opt-in `hostExecInDocker: true`)
- `server/agent/config.ts:588` `buildDockerSpawnArgs` — bind mounts, container env

## Goal

I want the same experience in MulmoTerminal:

1. Claude Code (or an alternative LLM CLI) runs inside a docker sandbox started by MulmoTerminal
2. Arbitrary MCP servers (HTTP or stdio) can be added mid-conversation
3. The added MCP tools can be called naturally by the Claude inside the sandbox
4. Tool implementation logic and state stay concentrated on the host (the MulmoTerminal process) and are never dragged into the sandbox's thin environment (`node:22-slim` and friends)

In short, I want MulmoTerminal to gain the equivalent of MulmoClaude's "MCP management in the Settings UI + automatic `--mcp-config` injection + URL rewriting / stdio shim".

## Proposed MVP

[...]

PR #1195, opened 2026-07-31
feat: remove the Docker sandbox

snakajima:
PR1 of the single-view removal (design note: `plans/feat-remove-single-view.md`) — the owner's call from the start of that work. Follows #1167, #1186, #1187, #1188, #1189, #1193.

Pure deletion. With the flag unset — the default, and how it shipped — every session already took the host path this now takes unconditionally.

## Why

It was **opt-in, macOS-only and single-view-only**, and the single view is being removed. Keeping it would have meant porting it to the grid, which is the opposite of what it was for: the sandbox existed to contain **one** interactive session, and the grid runs many.

## Deleted

`server/infra/sandbox.ts` (384 lines) and its spec, `Dockerfile.sandbox`, `scripts/ci-sandbox-image.ts`, and both `docker-sandbox` CI workflows.

## Unwound at each call site, not stubbed

- **`pty-spawn`** — `sandboxWouldRun` and `spawnSandboxEntry` gone; `ptySpawn` keeps the tmux path only.
- **`spawn-claude`** — the `sandbox` local goes, so `hookSettingsJson` / `mcpConfigJson` return to `localhost` / `127.0.0.1` unconditionally and the tool-group reset is decided by tmux alone.
- **`ws-routes`** — the Keychain-refresh branch and the per-reconnect credential re-sync.
- **`mcp-config`** — the `sandbox` parameter and the loopback rewrite, and with them the parameter on `SpawnDeps` and its binding in `index.ts`.
- **`provider-env`** — the refusal that existed only because a container inherits no environment. This is a capability **returning**: a provider session was refused outright under the flag.
- **`PtyEntry`** — the `sandbox` flag; and reap loses the container cleanup.
- The boot-time diagnostics and the CLI doctor's `docker` row.

## The documented env vars

`MULMOTERMINAL_SANDBOX`, `MULMOTERMINAL_SANDBOX_IMAGE`, `SANDBOX_MOUNT_CONFIGS` and `SANDBOX_SSH_AGENT_FORWARD` are swept out of the README, **both** language guides, and two skills (`-model`, `-bug-report`).

**Setting them is now silently ignored rather than an error**, so an existing `.env` keeps working — the session just runs on the host, which is what it already did whenever Docker was unavailable. There is a changelog entry saying so, because removing a documented variable is user-facing even when nothing breaks.

## Three specs lost their subject

Rewritten rather than deleted wholesale, since each still tests something real:

- **`add-dirs`** keeps the `--add-dir` half and drops the bind-mount half. Its pairing existed because "the flag without the mount reads fine and breaks only in the sandbox" — that sentence no longer has a second half.
- **`provider-env`** swapped the sandbox refusal in its every-reason loop for a missing-token one, so the loop still covers more than one reason.
- **`tool-group-reattach`** lost the fake socket it only carried in order to reach the sandbox branch.

## Verified

- `yarn test` — 7050 pass (the suite shrinks: 45 sandbox tests removed with their subject)
- `yarn typecheck` / `:test` / `:server` all pass, `yarn lint` 0 errors (4 pre-existing max-lines warnings), `yarn build` succeeds

**Correction to an earlier version of this description.** It said `typecheck:test` had three pre-existing failures in calendar specs and that the check was therefore red on main. **That was wrong.** My local `node_modules` held `@mulmoclaude/core@1.10.0` while `package.json` requires `^1.11.0` — a stale install, not a defect. CI installs with `--frozen-lockfile` and never saw it, which is exactly what its all-green run was telling me. After `yarn install`, all three typechecks pass locally too. Nothing on main is red.

**Not verified in a running app.** The meaningful check is that nothing changed: start a session with the flag unset and confirm it behaves exactly as before. If you have `MULMOTERMINAL_SANDBOX=1` in a `.env`, that now does nothing rather than failing.

work in mulmoterminal

## Summary by CodeRabbit

* **Changes**
  * Removed the experimental Docker sandbox; sessions now run directly on the host.
  * Removed Docker setup requirements, configuration options, and related documentation.
  * User-configured provider and MCP server connections now work without sandbox-specific restrictions or URL changes.
  * On startup, legacy sandbox credentials and containers are cleaned up automatically, with retries when cleanup cannot complete.
* **Documentation**
  * Updated English and Japanese guides and changelog to reflect the new host-based session behavior.


snakajima:
All three findings were real. Fixed in b4a5f1ad. Plus a correction to my own PR description.

## 1. Leftover sandbox credentials (P2, and the one that matters)

Right, and this is the sharpest thing in the PR. Those files are an **export of the Keychain OAuth credential** (mode 0600), and the only thing that ever deleted one was `cleanupSandbox` — which this PR removes. Anyone whose server was killed or upgraded mid-session was left holding a live credential in a directory nothing would ever touch again.

`~/.mulmoterminal/sandbox` is now removed at startup. Unconditional and idempotent rather than marker-guarded like the probe sweep: that directory is ours alone and nothing writes it any more, so there is no case where a second run could remove something a first run should have kept. Three tests, including "happy when there is nothing to remove" — it runs unguarded on every boot, and the usual case by far is that nobody ever turned the sandbox on.

## 2. `docs/index.md` still advertised `docker`

Right, and a plain miss: I swept the README, both language guides and two skills, and did not check the landing page. Removed from **both** language blocks.

## 3. Changelog wording

Fair. "the single view is being removed" reads as though this change removes it. It says now that the sandbox only ever wrapped the single-view session, which a *later* change removes — and that this one does not.

## Correction to the PR description

It claimed `typecheck:test` had three pre-existing failures in calendar specs and that the check was red on main. **That was wrong.** My local `node_modules` held `@mulmoclaude/core@1.10.0` while `package.json` requires `^1.11.0` — a stale install. CI installs with `--frozen-lockfile` and never saw it, which is what its all-green run was telling me and I read past. After `yarn install` all three typechecks pass locally. Nothing on main is red; the description is corrected.

## Verified

- `yarn test` 7053 pass, `yarn typecheck` / `:test` / `:server` all pass, `yarn lint` 0 errors, `yarn build` succeeds

snakajima:
Right, and fixed in d853c15f — along with a second thing the same search would have caught.

## The script

`scripts/sandbox-repro.sh` built from the deleted `Dockerfile.sandbox`, so it was a script that could only fail. It was the manual repro for the feature (#202) and goes with it.

## And one more, found by widening the search

`.github/workflows/ci.yml` justified the **macOS leg of the test matrix** by the sandbox being darwin-gated:

> macOS matters here beyond redundancy: the Docker sandbox is darwin-gated (`sandboxPlatformSupported`), so its code paths — Keychain credential export/refresh, mount building — only run for real on a macOS runner.

The leg stays — node-pty ships prebuilt binaries per platform, tmux differs, and the app's own platform branches only run for real on their own runner — but the comment now says *that*, instead of naming a feature that no longer exists. Left as-was it would have been the stated reason for a CI job, pointing at nothing.

## The actual lesson from this one

Both misses have the same cause: my sweep grepped `--include=*.ts --include=*.vue --include=*.md --include=*.json`, so **shell scripts and workflow YAML were never searched** — and that is exactly where both were. I redid it across every file type this time.

What still matches "sandbox" is unrelated and staying: the CSP `sandbox` header, iframe sandboxing, and Codex CLI's own `--sandbox` flag in the review workflow.

## Verified

- `yarn test` 7053 pass, `yarn typecheck` / `:test` / `:server` all pass, `yarn lint` 0 errors

snakajima:
Right. Fixed in 507fae17.

## The gap

`docker run --rm` cleans up when the **container** exits, not when its client dies — killing the client only detaches. That is precisely why `reap()` called `cleanupSandbox(id)` explicitly, and that call went with the feature. So a server killed or upgraded mid-session leaves a container running with the workspace and `~/.claude` mounted read-write, and nothing left in the app can reach it.

## The part worth explaining: what it is gated on

A container sweep means invoking `docker` at boot. The sandbox was **opt-in and macOS-only**, so nearly every install never turned it on — and paying for a subprocess on every start to clean up after a feature you never used is the wrong trade, especially when `docker` may be absent or a stopped daemon can make the call slow.

So the sweep is gated on the legacy credential **directory existing**, which is exactly the evidence that this machine ever ran the sandbox. If you never used it, nothing runs — not even a `docker` call.

That required `removeLegacySandboxDir` to answer *whether the directory was there* rather than whether the removal succeeded (`removeQuietly` returns true for a path that never existed, so the old return value was true for everyone). A test pins both answers, including that the second call is false — a wrong answer here is either a docker invocation on every boot, or a container nobody ever removes.

Best-effort and silent otherwise: docker absent, stopped or slow is not a reason to hold up a boot.

The changelog now says what cleanup happens and for whom.

## Verified

- `yarn test` 7054 pass, `yarn typecheck` / `:test` / `:server` all pass, `yarn lint` 0 errors

snakajima:
Right, and it is a contradiction between my own comment and my own code. Fixed in 4905a5f3.

The doc block said *"none of that is a reason to hold up a boot"* while both docker calls went through `spawnCapture` — `spawnSync`, **with no timeout**. A wedged daemon makes `docker ps` hang, so it would have blocked startup on exactly the machines the sweep exists to help. The comment described the intent; the implementation did the opposite.

Now `spawnCaptureAsync`, bounded at **5s**, and the caller does not await it. Shorter than `spawnCapture`'s own 15s default on purpose: nothing waits on this, and a container left behind by a feature that no longer exists is not worth minutes of a daemon that is not answering.

The `try/catch` went with it — `spawnCaptureAsync` already turns a non-zero exit *and* a timeout into a `status` rather than a throw, so catching around it was catching nothing.

## Verified

- `yarn test` 7054 pass, `yarn typecheck` / `:test` / `:server` all pass, `yarn lint` 0 errors, `yarn build` succeeds

snakajima:
Right — fixed in e65df885, which landed just as that verdict was being written (it describes 4905a5f3).

## The gap

The migration was one-shot. `~/.mulmoterminal/sandbox` was removed synchronously and the container sweep fired after it, so a first boot with Docker absent or not yet up lost the sweep **forever** — and an orphan container would outlive every later start. The gate that made the sweep cheap for everyone else was also what made it unrepeatable.

## The fix: split the two halves along what each is for

- **The credentials are the urgent part** and still go immediately, on the first boot, whatever Docker is doing. Nothing about a wedged daemon should keep an exported Keychain token on disk.
- **The directory is now what remembers that a sweep is owed.** It survives the credential deletion and is removed only once `docker ps` has actually answered. A non-zero status — absent, not up, timed out — leaves it, so the next boot asks again.

The gate stays self-limiting in the right direction: the directory is empty after the first boot, and the docker call stops happening as soon as one succeeds. A machine that never ran the sandbox still never invokes docker at all.

Five tests on the split, including the one that names the reason: *"keeps the directory, because the container sweep has not answered yet"*.

## Verified

- `yarn test` 7055 pass, `yarn typecheck` / `:test` / `:server` all pass, `yarn lint` 0 errors, `yarn build` succeeds

snakajima:
Right, and this one breaks the file's own premise. Fixed in d49e5b3e.

`fs-cleanup.ts` exists *because* cleanup must not throw — its header says a throw from it "can only turn a transient lock into a broken teardown". And I put a throwing call in it: `readdirSync` answers EACCES on an unreadable directory, ENOTDIR if the path is a file, and ENOENT if it goes away between the existence check and the read. Running at boot with nobody catching, that is a server that will not start — over litter from a feature that no longer exists.

Caught, and still reports `true`: the directory being there is the only question the caller asks. Nothing is lost by trying again — the sweep leaves the directory alone unless docker answers, and an unreadable one resists removal anyway, so the next boot retries both halves.

The test drives a real trigger rather than a mock: it creates the path as a **file**, which is what `readdirSync` answers ENOTDIR for.

## Where this PR has got to

Five rounds, and each fix opened the next question — which is what a deletion touching a credential path, a daemon and the boot sequence looks like:

1. credentials with no deleter →
2. containers with no deleter →
3. the container sweep blocking boot →
4. the gate making that sweep one-shot →
5. the credential removal able to throw at boot

All five were real. Every one of them lives in the migration I added to fix the first, not in the deletion itself — which is worth saying plainly: the sandbox removal was the easy part.

## Verified

- `yarn test` 7056 pass, `yarn typecheck` / `:test` / `:server` all pass, `yarn lint` 0 errors

PR #1196, opened 2026-07-31
feat: run a scheduled task's chat as a background worker

snakajima:
Owner's call, from noticing that 81 cells is a large number **as long as nothing accumulates in it**.

Hidden workers already can't: they are excluded from placement and from the unplaced marker, with a spec pinning that a hidden worker can never be marked. **Scheduled user tasks were the one path left that could fill the grid on its own** — a visible spawn is adopted as a cell the next time the grid loads, so an hourly task meant a cell per firing, indefinitely, reaching `MAX_TERMINALS` with nobody having asked for a terminal.

## The change

A task the scheduler runs is dispatched by a clock, not by someone at a keyboard — the definition of a background worker everywhere else here. It now gets that treatment: behind the **Background** filter, never bold, and **no grid cell**.

## It was already half of one

This is the argument for the change rather than a justification after the fact.

`scheduledSessions.register` has *always* put these on the background retention, whose stated reason is that *"nobody is waiting for it to finish and nothing else would ever end it"*. So the code already assumed nobody was watching — while the chat list still showed the session as one the user had started. The two halves now agree.

## A failed one still says so

Being quiet is right while it works and wrong when it dies, so the completion hook `spawnBackgroundChat` uses is registered here too. A scheduled task that reaches teardown without completing a turn is recorded and shows as `● failed` in the launcher's session list.

**Claude-only, by construction** — this path always spawns claude — and that limit matters: the Stop hook is the only success signal a PTY-hosted agent gives, so a recorder on an agent that cannot report success would mark every successful run as failed (the defect caught in #1188).

Without this, the change would have made scheduled tasks quiet **and** silent about failing, which is exactly the combination the failed-worker signal exists to prevent.

## Where the policy lives

Moved out of `index.ts` into `server/session/scheduled-chat.ts`, for the reason `background-chat.ts` was split from its route: these are rules, the caller is a uuid and a try/catch, and **policy nothing can assert is policy that drifts**.

Six tests, including two that are about the spawn rather than the classification: the order is spawn-then-retain (registering a session that failed to spawn would put a dead id on the retention), and a throwing spawn neither retains nor registers a hook (a hook for a session that never started would never fire or be cleared).

## What a user loses

A scheduled task's session no longer goes **bold/unread** when it finishes. If you were watching for that, it is under the Background filter, or in the launcher's list for the workspace. Said plainly in both the README and the changelog, since it is a visible behaviour change for anyone with the dev worklog on.

## Verified

- `yarn test` — 7062 pass (+6)
- `yarn typecheck` / `:test` / `:server` all pass, `yarn lint` 0 errors, `yarn build` succeeds
- The background-worker assertion was checked to fail with the hidden marker flipped off

**Not verified in a running app.** The check worth doing: with `worklogEnabled: true`, confirm the next firing shows under Background rather than among the chats, and that the grid gains no cell for it.

work in mulmoterminal

### DIFF
--- a/README.md
+++ b/README.md
@@ -630,7 +630,9 @@ disconnected, or with no device registered, the toggle is a no-op.
 **Dev worklog (cross-clone).** Set `worklogEnabled: true` in
 `~/.mulmoterminal/config.json` (and **restart** — the scheduler reads its tasks at boot)
 to register a built-in scheduled task. Every `worklogIntervalHours` (default 6) it spawns
-a Claude session that reviews the work you did across **all your saved working dirs**
+a Claude session — as a **background worker**: behind the Background filter, never bold,
+and it takes no grid cell, so an hourly task cannot fill the grid — that reviews the work
+you did across **all your saved working dirs**
 (`cwdPresets`) since it last ran, and writes it up as a short manager-style report.
 Multiple clones/worktrees of the same repo (e.g. `myapp`, `myapp2`) are **merged into one
 per-repository section**, each covering what problem was addressed, what got solved, what's
@@ -1490,7 +1492,8 @@ Key rules:
   exists) via the in-memory `knownSessions` registry + a `created` push; an
   unused one disappears when its PTY is reaped.
 - **Background workers get their own filter.** A session nobody started by hand —
-  a collection's scheduled refresh, or a plugin's `spawnBackgroundChat`
+  a collection's scheduled refresh, a **user scheduled task** (the dev worklog and
+  anything else the scheduler runs), or a plugin's `spawnBackgroundChat`
   `hidden: true` — is listed under the **Background** chip instead of among the
   chats, so a refresh schedule doesn't fill the history. It stays openable (a
   MulmoTerminal session is a live terminal, so a row you can't reach is a process
--- a/README.md
+++ b/README.md
@@ -630,10 +630,11 @@ disconnected, or with no device registered, the toggle is a no-op.
 **Dev worklog (cross-clone).** Set `worklogEnabled: true` in
 `~/.mulmoterminal/config.json` (and **restart** — the scheduler reads its tasks at boot)
 to register a built-in scheduled task. Every `worklogIntervalHours` (default 6) it spawns
-a Claude session — as a **background worker**: behind the Background filter, never bold,
-and it takes no grid cell, so an hourly task cannot fill the grid — that reviews the work
-you did across **all your saved working dirs**
+a Claude session that reviews the work you did across **all your saved working dirs**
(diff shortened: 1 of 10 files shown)


snakajima:
Real, and a consequence I did not consider. Fixed in a6a41c00.

Making these background sessions silenced their pushes: a background session's finished turn never reaches the phone, on the stated reasoning that it *"isn't a real user task"*. A task the user configured to run **while they are away** is exactly a real user task — and the phone is the only way they would ever hear about it. So the change had silenced the one case the setting exists for.

## What the finding exposed

`background` was doing two jobs that had never needed separating:

- **presentation** — out of the chat list, never bold, no grid cell;
- **identity** — not a real user task, never push.

A scheduled task wants the first and not the second. It now carries its own durable mark, and `shouldSuppressPush` takes a third answer.

A **translation worker is still refused whatever else is true** — an internal helper with no output a person reads, so there is nothing to tell them about. That one is not an exception waiting to be made, and the rule says so rather than leaving it to coincidence.

The mark is persisted like its siblings because a tmux session outlives a restart: the turn that raises the push can happen long after the process that spawned it is gone.

## Docs

README and the changelog now say push still fires — **being quiet means out of the way, not unreachable**. That is worth stating explicitly, because the rest of the entry reads as "this thing goes silent".

## Verified

- `yarn test` 7067 pass, `yarn typecheck` / `:test` / `:server` all pass, `yarn lint` 0 errors, `yarn build` succeeds

snakajima:
Right, and it is the same omission as the `failedWorkersHydrated` one on #1189: I added a durable set and read it without awaiting its hydration. Fixed in 60893f62.

## Why it matters more than it looks

**The hazard is the combination, not either answer.** Read separately during startup, `background` can hydrate before `userScheduled` — and `(background: true, userScheduled: false)` is precisely *"a background session that is not the user's task"*, so the push is suppressed for the one run that most needs it.

And this is not a corner case. The comment already sitting above that gate says why:

> Asked of the DURABLE predicate, not the live set: tmux keeps a worker's agent running across a server restart…

A scheduled session survives a restart in tmux, so **its turn finishing moments after boot is the ordinary case for this code**. The race window and the scenario the code exists for are the same window.

## The fix

One function answers both, awaiting both logs — rather than two awaits at the call site that a later caller could split apart again. The combination is the invariant, so it is held in one place.

The test drives the real ordering: the user-scheduled log is answered **late on purpose**, because an immediate answer would let a non-awaiting implementation pass. Verified to fail with the second await removed.

## Verified

- `yarn test` 7070 pass, `yarn typecheck` / `:test` / `:server` all pass, `yarn lint` 0 errors

PR #1201, opened 2026-08-01
feat: delete the single terminal view, and give the content surfaces a door

snakajima:
The last step of `plans/feat-remove-single-view.md`, and the door that had to exist first. Originally two PRs (#1202 was stacked on this one); folded together because they are one change — the door exists so the deletion is safe, and the deletion is why the door was needed. Merged, `main` never passes through a state where the content surfaces have both a door and a view that no longer needs one.

## 1. The blocker nothing had recorded

With the single view deleted, **how does anyone reach collections, wiki, accounting or files?**

They could not. Those buttons were gated `v-if="!inGrid"` — deliberately, with the reason written in:

> Single view only (#886): the content surfaces. The grid is for supervising agents, and every one of these replaces the whole screen anyway.

The plan's PR4 section listed the overlays and not the buttons that open them, which is how this stayed hidden until the owner asked directly.

**Collections now sits beside Grid** as a peer of the views, and reveals Feeds / Wiki / Accounting / Files once you are inside. One door rather than five: a terminal user's row does not grow by four buttons for surfaces they are not in.

- Collections joins the view-switch group (#941) — that group's rule is that it changes *which view you are in*, where everything right of the rule acts *within* the current one. The revealed siblings stay outside it.
- Membership is answered from the ROUTE (`CONTENT_ROUTES`, beside `OVERLAY_ROUTES` so the two cannot drift), so moving collections → wiki → files never blinks the row that got you there.
- **PRs is deliberately not content** — work under supervision belongs with the terminals.

## 2. Feeds was showing the collection list

The overlay rendered `CollectionsIndexView` for any index route, and that component lists collections and explicitly filters feeds **out** (`source !== "feed"`), with its open handler hardcoded to `gotoDetail("collection", …)`. The plugin ships `FeedsView` for exactly this and nothing here was using it. Feed *detail* was always fine — `CollectionView` asks the binding which kind it is showing — which is why only the index was wrong.

## 3. Pinned shortcuts moved, and lost their labels

Onto the row the overlay already had, replacing the Claude / Codex / Antigravity picker. They belong there — a favourite IS a collection or a feed — and they were about to be deleted with the toolbar that held them. Icon-only, with the name still reaching a screen reader (`aria-label`) and the pointer (`title`); the row hides itself when nothing is pinned.

What goes with the picker: `launchAgent` still decides which agent a collection action spawns, but it is now whatever was last chosen elsewhere rather than something you can change from here.

## 4. The single view is gone

`App.vue` is now the page-level concerns and the overlays. Everything the view owned has a home:

| The single view had | It lives in |
|---|---|
| the settings modal | GridView's own |
| Canvas / tools panes | beside a zoomed cell |
| the session list | the cockpit roster, and the launcher's resume list for sessions off the grid |
| collections, wiki, accounting | the Collections door above |
| one terminal filling the window | a zoomed cell |

**The 81-cell question answered itself.** Deferred through four PRs because the single view *was* the overflow destination — but with hidden workers and scheduled tasks both excluded from placement (#1196), nothing fills the grid on its own, so reaching the cap takes 81 terminals opened by hand. A full grid now parks the session: it stays marked unplaced, the next load with room adopts it, and the launcher's resume list shows it meanwhile.

**Two lifecycle consequences, neither of which fails loudly:**

- **`GridView`'s `onActivated` stops firing.** Mounted for the life of the page, there is no activate/deactivate cycle at all — the new-terminal opener and the spawned-chat placer would have *silently never registered*. Both moved to `onMounted`; the roster poll follows the route, since the grid still goes off screen under an overlay.
- **`viewIsGrid` became a constant** and is deleted. Every route is the grid or an overlay whose origin resolves to it.

Deleted with the view: the `/chat` route and button, `Sidebar`, `SessionTabBar`, `SessionFilters`, `sessionList`, and `useChatLauncher`'s `registerChatOpener` / `showSpawnedSession`. knip is clean afterwards.

## 5. The grid's own controls hide when the grid is covered

PRs, Worklog, New terminal, the ordering control and the rate-limit gauge show only on `/terminals` (owner's call). They act on cells nobody can see, and the gauge is status for a view that is not showing. The switch group never hides, so Grid view always brings the terminals back.

**One consequence, recorded rather than left to be found:** `/prs` now highlights **nothing** in the nav, because its own button hides with the rest. Inside the content section the Collections door stays lit. Say the word if the active overlay's button should be exempt.

## The specs

Rewritten rather than re-pointed wherever the rule they expressed is what changed — the two-view world. One could not survive as written: `useFilesView` pinned *"ignoring an earlier open's origin"* by capturing **chat** and expecting **the grid**, deliberately opposite values. With one view those coincide and it would pass whether the stale origin was ignored or reused. It now asserts the history **state**, which is where staleness would actually live.

## Verified

- `yarn test` — 7028 pass (the suite shrinks with the view: two component specs and one composable spec went with their subjects)
- `yarn typecheck` / `:test` / `:server` all pass, `yarn lint` 0 errors, `yarn build` succeeds, `knip` clean

**Not verified in a running app, and this deserves it most.** It deletes a whole view on top of six merged PRs that have not been clicked through either. Worth a `yarn build` + restart and a pass over: the grid, a zoomed cell with Canvas and tools, Collections and each sibling, a chat started from a collection, and the settings modal.

work in mulmoterminal


snakajima:
Right. Fixed in c2045f92.

`collectionsActive` / `feedsActive` required `mode === "index"`, so opening a collection or a feed put out the very button that got you there — while you are plainly still inside that section. The index/detail distinction belongs to the **view**, not to **which section you are in**.

**It matters more than it would have a day ago.** Since the grid's own controls now hide under a full-screen overlay, nothing else in the nav is lit either — so a collection detail page showed **no selected destination at all**. The two changes are individually reasonable and together produced a toolbar with nothing on.

Three tests, including that the two doors never light together — they are different sections sharing one component. Verified to fail with the old predicate.

## The related gap I flagged in the description is still open

`/prs` also shows no highlight, for the same underlying reason: its button hides with the grid controls. That one is deliberate — the owner asked for PRs to hide when the grid is not visible — and I have left it as asked. Worth deciding whether the button whose own overlay is open should be exempt; it would fix `/prs` the same way this fixes `/collections/:slug`.

## Verified

- `yarn test` 7031 pass, `yarn typecheck` / `:test` / `:server` all pass, `yarn lint` 0 errors

PR #1226, opened 2026-08-01
feat: resolve presentDocument / presentHtml relative paths against the session's cwd

snakajima:
## The bug

`presentDocument({ path })` / `presentHtml({ path })` resolve a **relative** path against one root — the workspace (`CLAUDE_CWD`, default `~/mulmoclaude`). MulmoClaude has only that directory, so workspace-relative is right there. MulmoTerminal runs every grid cell in its own project, so it almost never matches the directory the agent is looking at.

A session running in `~/git/ai/mulmoterminal` calling `presentDocument({ path: "README.md" })` opened `~/mulmoclaude/README.md`. When that file does not exist you get `No markdown file exists at README.md`. When it **does** exist, another project's file is opened silently and the user's edits in the view are written back to it. That second case is the real damage.

## The fix

Rewrite the relative `path` to an absolute one **at the boundary** — the `/api/plugin/<tool>` dispatch route — using the calling session's own directory. The MCP broker now sends the session id as an `X-MulmoTerminal-Session` header (a header, not the body: the args are the tool's own schema and every plugin sees them).

Not done as a per-request `rootFor`, and that is the load-bearing decision. Two later hops come from the **browser** and can name no session: the View's own dispatch (`loadDoc`/`saveDoc`/`loadHtml`/`saveHtml`) and the iframe's `/htmlfile/…` request. A per-request root fixes the tool call and then resolves those two against the workspace again — the view opens and the save silently goes to a different file, which is a worse failure than today's. An absolute path makes everything downstream session-independent, and a stored tool result keeps pointing at the file it actually opened.

Left alone deliberately, each for its own reason:

- **`artifacts/…`** — the workspace's own output area. The plugins route those values to `files.artifacts` (root fixed at `<workspace>/artifacts`), and it is where `saveNewDoc` / `executeHtml` put a document they just created. Re-rooting them would mean a tool could not re-open the document it had just saved.
- **already-absolute paths**, and **a View dispatch** (`kind` present — the browser replaying a path out of a tool result).
- **anything core's own gate would refuse** (wrong extension, `.`/`..`/empty segment, NUL). Rewriting those would either launder a bad path into a valid absolute one or duplicate a rule that belongs to `@mulmoclaude/core`. The existing validation still fails them.

Session cwd is three tiers: the live PTY's `cwd` (where the agent actually runs) → the persisted `dev-terminal-cwds.json` note → `CLAUDE_CWD`. **Any caller that names no session — the scheduler, feeds, a remote or direct POST — still gets `CLAUDE_CWD`, so no existing path changes behaviour.**

## Two decisions, both the owner's call

- **Always on; no config key.** A `presentPathRoot: "session" | "workspace"` switch was considered and dropped: with it, "which root resolved this" is not visible from the tool result, so the silent-wrong-file class of bug survives in a second form.
- **The tool prompts are not touched.** The "workspace-relative (`docs/design.md`)" wording lives in the shared definitions in `@mulmoclaude/markdown-plugin` / `@mulmoclaude/html-plugin`. Changing it here alone would make one tool mean two things across the two hosts. **Correcting it to a host-neutral phrasing in core remains open, and it is MulmoClaude's call** (per CLAUDE.md's reference-host rule).

Known cosmetic side effect, already documented in `backends/markdown.ts`: `publishFileChange` cannot stat an absolute path for its mtime, logs one `[file-change] stat failed` line and falls back to `Date.now()`. The channel name still matches, so live-refresh works; this change makes the line appear more often. The real fix is teaching core's publisher about absolute paths — MulmoClaude's side.

## Verification

- New `test/server/backends/presentPathRoot.spec.ts` — the pure rewrite (relative / absolute / `artifacts` / `kind` / refused values / no usable `path`) and the route middleware (session header, no header, a header that is not a session id, another tool's `path`, and a tool named `constructor`).
- New `test/server/session/session-cwd.spec.ts` — the three-tier pick, including live PTY winning over the remembered note.
- `test/server/backends/openPath.spec.ts` — one end-to-end case: with the path absolutized against another project, `loadDoc`/`saveDoc` hit **that** project's `README.md` and the workspace's copy is untouched.
- `test/server/backends/html.spec.ts` already pins the absolute `/htmlfile/abs/…` round trip; unchanged and still green.
- `yarn format` / `lint` / `typecheck` / `typecheck:server` / `typecheck:test` / `build` / `test` (7294 passed).

Not yet done: the manual check in a live browser session (call `presentDocument({ path: "README.md" })` from a cell in this repo, edit in the view, confirm the write lands on this repo's README).

Plan: `plans/feat-present-path-session-cwd.md`.

work in mulmoterminal

## Summary by CodeRabbit

- **New Features**
  - Relative document and HTML paths now resolve within the active session’s working directory.
  - Session paths use the live terminal directory, remembered session directory, or workspace directory as a fallback.
  - Absolute paths and workspace artifact paths continue to work unchanged.

- **Bug Fixes**
  - Opening or saving similarly named files from different sessions now targets the correct session workspace.
  - Unrelated tools and invalid path requests are left unaffected.


snakajima:
Both Codex findings were real and are fixed in 48b1435.

**P1 — relative images.** Confirmed the mechanism: `rewriteMarkdownImageRefs` joins a ref onto the document's directory and emits `/api/files/raw?path=…` with no `cwd`, which `mountFilesRoutes` resolves under the workspace root. An absolute `docPath` therefore asks for `<workspace>/home/…`. The fix is to leave a session whose directory IS the workspace **byte-identical** rather than rewrite it to the same file's absolute path — so nothing that works today stops working, and the no-session path (scheduler, feeds, direct POST) becomes a literal no-op, which is what this PR claimed and the first commit did not quite do. A document in *another* project does still lose its relative image refs; there is no way to name a cwd in that URL, and today those same refs resolve against a different project entirely, so this is not a regression. Teaching the raw-file URL about a source cwd is the real fix and it belongs to the shared plugin.

**P2 — dot-prefixed directories.** Confirmed, and it is worse than the example given: MulmoTerminal's own managed worktrees live under `~/.mulmoterminal/worktrees/<repo>-<hash>/<task>`, so this is where real sessions run. `isPresentableHtmlPath` is right to refuse the segment — the `/htmlfile` mount refuses it too (`resolveHtmlFileRequestPath` returns null), so accepting the call would report success for a page that can never render. The middleware now asks the plugin's own gate and answers 400 with the reason, instead of letting `invalid path: /Users/…/report.html` read like a typo in an argument that was correct. `presentDocument` does not go through that mount (`isDocumentPath` accepts dotfile segments) and is unaffected — pinned by a test.

Full suite green: 7295 passed, plus `format` / `lint` / `typecheck` / `typecheck:server` / `typecheck:test` / `build`.

work in mulmoterminal

PR #1227, opened 2026-08-01
feat: open the Canvas by itself when the agent draws on the cell you are looking at

snakajima:
## What

When the agent calls `presentDocument` (or any other tool that draws a card) on the cell you currently have enlarged, the Canvas pane now opens by itself.

## Why

The drawing IS the agent's answer to what was asked. With the pane closed, that answer left no trace but a count on the cell's unread chip — which reads as a notification rather than as the reply, and finding it meant knowing a pane exists and which button opens it.

## The limits, which are the design

- **The enlarged cell only.** A background cell drawing must not take the screen away from what is being done in another one, so that case keeps exactly what it has today: `TerminalCell`'s unread chip, which enlarges and opens on click. (Owner's call.)
- **Only a result that will actually RENDER** — gated on `getPlugin(toolName)`, the same question `GuiPanel` asks before drawing a card, so the two cannot disagree about what "drew something" means. Every GUI tool result publishes on the session channel, including ones with no view of their own (`manageCollection`, `google`); switching the pane for those is a takeover with nothing behind it to explain itself.
- **It re-opens after the user closed the pane by hand.** Closing dismisses the drawing in front of you; it is not a standing preference against the next one.

Routed through the existing `openCanvasFor` rather than `setRightPane`, so the files pane still flushes its buffer on the way out and may refuse to go — a buffer that can be saved neither way keeps the pane it is visible in.

## Verification

New spec `test/src/components/canvasAutoOpen.spec.ts` (7 cases): opens on the enlarged cell; ignores a result no plugin renders; leaves the pane alone when another cell draws; does nothing while no cell is enlarged; re-opens after a manual close; flushes the files pane and yields to one that refuses; follows the zoom to another cell.

`yarn format`, `yarn lint`, `yarn typecheck`, `yarn typecheck:server`, `yarn typecheck:test` and `yarn test` (7286 passed) all clean.

Not done: live browser check — this instance serves a built `dist/`, so it needs a `yarn build` on the box that runs it.

work in mulmoterminal

### DIFF
--- a/src/components/TerminalGrid.vue
+++ b/src/components/TerminalGrid.vue
@@ -37,6 +37,9 @@ import {
 import { setFilesPaneOpener } from "../composables/filesPaneOpener";
 import { paneCanShowClick } from "./paneClickTarget";
 import { onToolGroupsAnnounced } from "../composables/useToolGroupsAnnounce";
+import { usePubSub } from "../composables/usePubSub";
+import { getPlugin } from "../plugins-registry";
+import { isRecord } from "../../common/isRecord";
 import { hasCanvasGroup } from "../../common/toolGroups";
 import type { RightPane } from "./gridCell";
 import { parsePaneStore, rememberPane, recallPane } from "./filesPaneStore";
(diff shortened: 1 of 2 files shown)


snakajima:
Fixed in 12f7e93e — good catch, the race was real.

`openCanvasFor` awaits the files pane's save, and the zoom can move while a dirty buffer is going down; on the way back it emitted `toggle-expand` and dragged the drawing cell back, taking the screen from the cell the user had just moved to.

The two callers genuinely disagree about that, so it is now an argument rather than a re-check: the unread chip on a tiled cell still means "bring that cell here" whatever the zoom did in the meantime, while the auto-open path gives up instead. Kept as one function so the buffer flush cannot drift between them.

Pinned by a new case that holds the flush open across a zoom change and asserts no `toggle-expand` is emitted — verified it fails without the fix.

Issue #1267, opened 2026-08-02
fix(windows): presentPathRoot.spec builds its expected path with a POSIX separator, so Windows (daily) is red since #1226


isamu:
## Symptom

`Windows (daily)` has failed on every run since #1226 landed (first red:
[run 30709308166](https://github.com/receptron/mulmoterminal/actions/runs/30709308166), latest:
[run 30724035812](https://github.com/receptron/mulmoterminal/actions/runs/30724035812)). Last green
was [30702245163](https://github.com/receptron/mulmoterminal/actions/runs/30702245163).

One assertion, on both 22.x and 24.x, in `test/server/backends/presentPathRoot.spec.ts`:

```
FAIL  absolutizePresentPath > resolves a relative path against the session cwd
-   "path": "/repos/mulmoterminal/README.md",
+   "path": "D:\\repos\\mulmoterminal\\README.md",
```

Linux and macOS are green, so #1226 merged clean.

## Root cause

The spec's fixture is `const CWD = "/repos/mulmoterminal"` and the expectation is written as the
template `` `${CWD}/README.md` `` — a hardcoded POSIX join. The implementation
(`server/backends/presentPathRoot.ts`) uses `path.resolve(cwd, value)`, which on Windows produces a
drive-qualified, `\`-separated path.

The product code is correct — a session cwd is a native path on whatever platform the server runs
on, and every other expectation in the same file already goes through `path.resolve`. Only this one
test hardcodes the separator.

## Fix

Make the fixture a platform-native absolute path (`path.resolve("/repos/mulmoterminal")`, the idiom
already used in `test/server/config/cwd-presets.spec.ts` and `test/server/files/pathContainment.spec.ts`)
and build the expectation with `path.join` rather than string concatenation.

Verification is a `workflow_dispatch` of `Windows (daily)` at the fix branch — the suite passing on
macOS proves only that the change is not a regression.

isamu:
Working on this in `mulmoterminal`.

isamu:
Fixed by #1269, merged as 313388ce2d986add8d0dc52632f474082d5fba64.

Verified before merge by a `workflow_dispatch` of `Windows (daily)` at the fix branch — [run 30724980050](https://github.com/receptron/mulmoterminal/actions/runs/30724980050), success on both 22.x and 24.x. The run triggered by the merge to `main` is the confirmation on the mainline.

PR #1269, opened 2026-08-02
fix(windows): spell the present-path spec's session cwd the way the platform does

isamu:
## Summary

`Windows (daily)` has been red on **every** run since #1226 landed. One assertion, on both 22.x and
24.x, in `test/server/backends/presentPathRoot.spec.ts`:

```
FAIL  absolutizePresentPath > resolves a relative path against the session cwd
-   "path": "/repos/mulmoterminal/README.md",
+   "path": "D:\\repos\\mulmoterminal\\README.md",
```

The spec's fixture is `const CWD = "/repos/mulmoterminal"` and its expectation is the template
`` `${CWD}/README.md` `` — a hardcoded POSIX join. `absolutizePresentPath` resolves with
`path.resolve(cwd, value)`, which on Windows picks up the current drive and joins with `\`.

The product code is correct; only this expectation is POSIX-only. Every other expectation in the
same file already goes through `path.resolve`, which is why the middleware tests covering the same
rewrite passed and only the unit one failed.

Fix: resolve the absolute fixtures (`CWD`, `WORKSPACE`, `DOT_CWD`) and build the expected value with
`path.join`. Test-only change — no product code touched.

refs #1267

## Items to Confirm / Review

- **`path.join`, not `path.resolve`, in the expectation.** `path.resolve(CWD, value)` is the exact
  call the implementation makes, so it would assert nothing beyond "the function is itself". With
  `CWD` already absolute the two agree, and `join` is a separate function reaching the same answer.
- **Resolving `WORKSPACE` / `DOT_CWD` too** is not needed to make the job green — both already pass
  because their expectations run through `path.resolve`. They are resolved for coherence: a session
  cwd that is drive-relative on Windows is not one any server would report. Say so if you would
  rather keep that diff to the one failing constant.
- **Scope.** The Windows job runs the whole suite and this was its only failure, so there is no
  second site with the same POSIX-only spelling to sweep.

## User Prompt

> fix https://github.com/receptron/mulmoterminal/actions/runs/30724035812

## Root cause and history

| run | commit | result |
| --- | --- | --- |
| [30702245163](https://github.com/receptron/mulmoterminal/actions/runs/30702245163) | before #1226 | green (last) |
| [30709308166](https://github.com/receptron/mulmoterminal/actions/runs/30709308166) | #1226 | **red** (first) |
| [30724035812](https://github.com/receptron/mulmoterminal/actions/runs/30724035812) | #1260 merge | red (the reported run) |

Linux and macOS are green throughout, which is why #1226 merged clean — `Windows (daily)` is kept
out of `pull_request` on purpose.

Plan: [`plans/fix-1267-windows-present-path-spec.md`](plans/fix-1267-windows-present-path-spec.md).

## Verification

- `yarn format` / `lint` / `typecheck` / `typecheck:server` / `typecheck:test` / `build`: clean.
- `yarn test` on macOS: 512 files, 7429 passed — which proves only that this is not a regression;
  the bug is invisible on macOS by construction.
- Ground truth: a `workflow_dispatch` of `Windows (daily)` at this branch. Linked below once green.

work in mulmoterminal

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed Windows-specific failures when resolving and displaying present-path information.
  * Updated path handling to use platform-compatible resolution, including session working directories.

* **Documentation**
  * Added a plan documenting the cause, proposed solution, and Windows verification steps for the path-related issue.

### DIFF
--- a/plans/fix-1267-windows-present-path-spec.md
+++ b/plans/fix-1267-windows-present-path-spec.md
@@ -1,0 +1,51 @@
+# fix #1267 — the present-path spec must spell a session cwd the way the platform does
+
+## The failure
+
+`Windows (daily)` has been red on every run since #1226 landed (2026-08-01 16:58 UTC); the first is
+[run 30709308166](https://github.com/receptron/mulmoterminal/actions/runs/30709308166) and the most
+recent is [run 30724035812](https://github.com/receptron/mulmoterminal/actions/runs/30724035812).
+Last green: [30702245163](https://github.com/receptron/mulmoterminal/actions/runs/30702245163).
+
+One assertion, on both 22.x and 24.x, in `test/server/backends/presentPathRoot.spec.ts`:
+
+```
+FAIL  absolutizePresentPath > resolves a relative path against the session cwd
+-   "path": "/repos/mulmoterminal/README.md",
++   "path": "D:\\repos\\mulmoterminal\\README.md",
+```
+
+Linux and macOS are green, which is why #1226 merged clean.
+
+## Root cause
+
+The spec's fixture is `const CWD = "/repos/mulmoterminal"` and the expectation is the template
+`` `${CWD}/README.md` `` — a hardcoded POSIX join. `absolutizePresentPath` resolves with
+`path.resolve(cwd, value)`, which on Windows both picks up the current drive and joins with `\`.
+
+The product code is right: a session cwd is whatever the platform spells, and the two hops
+downstream (the View's dispatch, the `/htmlfile` mount) receive that same native path. Every other
+expectation in the file already goes through `path.resolve` — which is why only this one test
+failed while the middleware tests covering the same rewrite passed.
+
+## Fix
+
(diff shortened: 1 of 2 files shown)


isamu:
## Bot review triage

- **Codex** — `LGTM`, no findings.
- **CodeRabbit** — one Minor nitpick: add a `text` language identifier to the fenced failure-output
  block in `plans/fix-1267-windows-present-path-spec.md` (MD040). **Deliberately skipped.** Bare
  fences are the convention in this directory (207 bare vs 5 `text` across `plans/*.md`), markdownlint
  is not part of CI, and `.prettierignore` excludes Markdown — so the change would buy nothing and
  cost a commit that re-runs every check on a branch already verified on Windows.
- **Socket Security** — both checks pass.

## Windows verification (the point of this PR)

`Windows (daily)` dispatched at this branch:
[run 30724980050](https://github.com/receptron/mulmoterminal/actions/runs/30724980050) — **success**
on `14ac0ebe`, both matrix jobs (`lint_test_windows (22.x)` and `(24.x)`). That is the job that has
been red on every run since #1226, and it is green on this commit.

PR #1371, opened 2026-08-03
feat: reveal a drawing on the tiled grid, and let one terminal enlarge

snakajima:
## What

Two changes, the second found by the first.

**1. A drawing that lands while nothing is enlarged reveals itself.** An agent calling `presentDocument` / `presentChart` IS its answer to what was asked. On the tiled grid that answer left no trace but a count on a chip — the Canvas pane exists only beside an enlarged cell, so reading it meant knowing the pane exists and which button opens it. Now the drawing cell enlarges itself and the Canvas opens beside it: the same reveal the unread-canvas chip performs on click, through the same `openCanvasFor` (files-buffer flush included).

**2. `toggleExpand` no longer refuses to zoom with fewer than two running cells (#374).**

## Where each half lives, and why

- The new case is in `GridView.vue`: un-zoomed, `TerminalGrid` is handed one PAGE of cells, and a cell on another page can draw too (`toggleExpand` turns the page for it).
- `TerminalGrid.vue` keeps the case it already owned — a drawing on the cell ALREADY enlarged — and still refuses to enlarge a background cell while another one is being worked in. That one keeps the chip.
- Nothing happens while the grid sits behind a full-screen overlay: the user is reading something else, and rearranging the grid they left would greet them, on the way back, with a zoom they never asked for.
- `isDrawnResult` moved to `src/utils/drawnResult.ts` so both sides ask "will this actually render a card?" the same way — tools that publish without a view of their own (`manageCollection`, `google`) reveal nothing.

## Why #374 had to go (please don't put it back)

Trying the above on a one-terminal grid did **nothing at all**, with no error on either side: the reveal asks for the zoom, the state politely declines, and the feature simply appears not to work.

#374's rule reads as sound — zooming is "one big, the rest as a filmstrip", so with nothing to switch to it swaps a working layout for an empty filmstrip and squeezes the terminal's status bar off the bottom. What it missed is that the zoomed row is **also the only place the Canvas / Tools / Files panes exist**. On a single-terminal grid the rule did not merely decline a layout; it locked those panes away entirely: the unread-canvas chip did nothing when clicked, and a drawing had nowhere to be shown.

The owner also wants a single terminal to enlarge from the manual UI on its own terms, so the rule is gone from BOTH entry points (`toggleExpand` and `zoomAt`, which the keyboard zoom shares). Collapsing was never refused and still isn't.

Because the issue still reads as correct, the reasoning is written down in three places rather than left to be re-derived: at `toggleExpand`, at zoom invariant 5, and in `plans/feat-canvas-auto-reveal.md`.

**Knock-on:** a grid of nothing but launch cells can now be enlarged too (a bigger launch form). An expand button that works on some cells and silently not on others seemed worse; pinned by a test.

## Tests

- `test/src/components/gridCanvasAutoExpand.spec.ts` (new) — the reveal, the single-terminal grid, a result no plugin renders, a cell already enlarged, another route. Verified it fails when the enlargement is routed back through the old refusal.
- `canvasAutoOpen.spec.ts` — unchanged behaviour for the already-enlarged cell.
- `gridTabs.spec.ts` — the #374 cases inverted, plus zoom invariant 5.

`yarn test` 7796 passed, lint (0 errors), typecheck and build all green. Live-verified by the owner.

work in mulmoterminal

## Summary by CodeRabbit

* **New Features**
  * Canvas now automatically opens and expands the cell producing a drawing result.
  * Automatic Canvas opening works for cells on other pages and single-cell grids.
  * Any existing cell can now be expanded or zoomed, including single-cell and launch-only grids.
* **Bug Fixes**
  * Drawing events are ignored when unsupported, invalid, or blocked by an active overlay.
  * Canvas expansion and zoom behavior now remain consistent across grid states.
* **Tests**
  * Added coverage for automatic Canvas expansion, single-cell grids, inactive routes, and invalid drawing results.


snakajima:
Fixed in a1176bdd — good catch, and the premise is right: the zoom row is hidden rather than unmounted when the grid is tiled, so a dirty `FilesPane` really can be sitting there with nothing enlarged.

`openCanvasFor` now takes a `stillWanted` predicate that is re-asked **after** the flush. GridView passes its own two conditions (`expandedUid === null && onTerminalsRoute()`), so a reveal that was overtaken during the save is dropped and the unread-canvas chip reports the drawing instead. A click passes nothing — "bring that cell here" survives whatever moved, which is the existing #1227 behaviour.

Pinned by two tests in `canvasAutoOpen.spec.ts` ("drops an un-clicked reveal whose conditions stopped holding during the flush" / "still enlarges for a click, which passes no condition to re-check"); verified the first fails without the guard.

PR #1376, opened 2026-08-03
feat(github): keep one work comment per issue and edit it as the work moves

isamu:
## Summary

The issue work comment (`issueWorkComments`, off by default) said two things and then went quiet:
**work started**, and **it merged**. Between them the issue said nothing, and each state was a new
comment.

Now a clone keeps **one comment per issue and edits it** as the work moves.

```
Working on this in `1234-fix-login`.

- started — 2026-08-04 14:20 UTC

posted by MulmoTerminal
```

…becomes, without a second comment appearing:

```
Merged in #1240. Work done in `1234-fix-login`.

- started — 2026-08-04 14:20 UTC
- PR #1240 — 2026-08-04 15:05 UTC
- merged in #1240 — 2026-08-04 16:40 UTC

posted by MulmoTerminal
```

- **Three milestones, no more**: started, PR opened, merged. **CI is deliberately absent** — it is
  on the PR already, and it flaps; an issue reporting every turn stops being readable.
- **Times are UTC**, and they are the point: a claim posted three weeks ago and never updated reads
  differently from one that moved this morning.
- The comment **says it came from MulmoTerminal**. These land on issues other people filed.
- The directory stays the **folder name only** (unchanged) — which is what tells two terminals, two
  worktrees, or two people that the work is already taken.
- Editing sends **no notification**, on purpose: the first line is news, the rest is status.

`refs #1369`. `issueWorkComments` stays **off by default** and stays a global setting.

## Items to Confirm / Review

1. **The lock moved from the milestone to the comment.** Recording a milestone is a
   read-modify-write — the body is the only place the earlier milestones live — so two arriving
   together could post two comments, or drop one another's line. `server/git/work-comment.ts` now
   serializes per `(repo, issue, dir)`. The serializer is **extracted and shared** with
   `issue-work.ts`'s identical one-worktree-per-issue rule (`server/infra/serialize-per-key.ts`) —
   please check that refactor, it touches a path that spawns sessions.
2. **The `start` marker is reused as the comment's anchor**, so no migration is needed and comments
   already on issues get adopted and edited. A comment written by an older build carries no
   milestone lines, so its **`createdAt` becomes the `started` time**. An older build's separate
   `merged` comment is still recognised, so an issue upgraded mid-flight is not told twice.
3. **The PR milestone is only reported when this session watched it appear** (`before.pr !==
   now.pr`), the same rule `merged` already used. A reload finding a month-old PR adds no line —
   this side would only know when it noticed, not when it happened. That means the PR line is
[...]

PR #1383, opened 2026-08-04
fix: bound the PR number in the work comment's own patterns

isamu:
## Summary

Follow-up to #1376. The milestone lines in an issue work comment read a PR number back out of a
body **anyone on the issue can edit**, and the pattern accepted a digit run of any length.

#1376 already caught what that produces — `Number("9".repeat(20))` is 1e20, which the next edit
would have written back as `- PR #1e+20` — but it fixed it with a check *after* the match. The
bound belongs in the pattern:

```diff
-const PR_LINE = /^- PR #([1-9]\d*) — (.+)$/;
-const MERGED_LINE = /^- merged(?: in #([1-9]\d*))? — (.+)$/;
+const PR_DIGITS = "[1-9]\\d{0,9}";
+const PR_LINE = new RegExp(`^- PR #(${PR_DIGITS}) — (.+)$`);
+const MERGED_LINE = new RegExp(`^- merged(?: in #(${PR_DIGITS}))? — (.+)$`);
```

Ten digits is far more than any forge issues and still exact as a `Number`, so an oversized line
now **fails to match at all** and is dropped like every other line that is not exactly what render
wrote. The post-match check goes with it — it could no longer fire.

It also settles the `security/detect-unsafe-regex` warning the merged pattern raised (12 lint
warnings → 11).

## Items to Confirm / Review

1. **The lint warning was a false positive, and I am not fixing it because a linter said so.**
   `\d*` inside an optional group is star-*height*, not ambiguity; measured, the old pattern was
   linear (5000-digit and 5000-space inputs both returned in 0 ms). The reason to bound is the
   `1e+20` value, which is real. Had the bound not been independently worth it, the right response
   would have been to say so rather than reshape the code — please push back if you disagree.
2. **Ten digits is a judgement call.** GitHub's own numbers are seven; the widest safe choice would
   be fifteen (still under `Number.MAX_SAFE_INTEGER`). Ten leaves plenty of room and keeps the
   pattern readable. A spec pins that `#1234567890` still parses.
3. **Behaviour is unchanged** for every input that was already handled: the regression test from
   #1376 (`- PR #99999999999999999999 — …` is dropped) still passes — it now fails at the pattern
   instead of at the check.

## User Prompt

Not a user request. Found while running `/gh-review-loop`: CI on #1373 surfaced a
`security/detect-unsafe-regex` annotation pointing at `common/workComment.ts`, which #1376 (merged
minutes earlier) had introduced. Cleaning up after my own merge rather than leaving a new warning
in the repo.

## Verification

`yarn format` / `lint` (0 errors, 11 warnings — one fewer than before) / `typecheck` / `build` /
`test` (7867 passed) all green, from a `--frozen-lockfile` install after merging current `main`
(which now carries #1373's dependency bumps).

work in mulmoterminal2


isamu:
Codex was right, and the finding is fixed in `93a9842`.

Bounding the number in the pattern made the **parser stricter than the writer**: the route still accepts any safe positive integer, so an eleven-digit PR would have been rendered as a line the parser then dropped. That is worse than it sounds — a dropped line does not merely fail to display, it disappears from the comment on the **next** edit, because the body is where the earlier milestones are stored. The render/parse inverse is the invariant the whole design rests on, and I broke it while tightening one half.

One bound now, `MAX_LINE_PR`, on both sides:

- `withWorkEvent` refuses a PR milestone whose number a line could not carry, instead of recording it as if it had been written.
- `eventLine` refuses to write one. A **merge** keeps its milestone and loses only the number — the merge still happened.
- The headline drops the number on the same rule, or the comment would name a number the line beneath it does not.

Specs pin the agreement at the boundary: `9999999999` round-trips; `10000000000` is refused rather than written.

I did **not** tighten the route's validation to match. `positiveInt` there is `isIssueNumber`, the shared rule the work-item chip also applies, and the bound that matters here belongs next to the pattern that defines it rather than duplicated at the entry.

PR #1388, opened 2026-08-04
feat: open a MulmoScript in the Canvas from the Files pane (#1374)

isamu:
## Summary

The Canvas button added in #1380 now also opens **MulmoScript stories**. Pick `artifacts/stories/<name>.json` in the Files pane, press Canvas, and the storyboard opens — no agent, no tool call.

This is **step 2 of #1374**; step 1 (`presentDocument` + `presentHtml`) shipped in #1380. The issue stays open until the whole thing has shipped.

mulmoScript breaks **both** assumptions the first two tools were built on, and that is most of this diff:

|  | markdown / html (#1380) | mulmoScript (here) |
| --- | --- | --- |
| Path the plugin takes | absolute is fine | **absolute is refused** — only `stories/x.json` |
| What the card carries | a path; the View self-fetches | **the parsed script** |
| So the gate asks | "can you render this file?" | **"is this file in the workspace's story directory?"** |

## Items to Confirm / Review

1. **The gate is rooted at the workspace, deliberately.** A project cell may have an `artifacts/stories/` of its own; those stories are not the ones the plugin opens. Matching on shape alone would offer a button that opens the wrong file or nothing at all. Please sanity-check that this is the intended rule and not too strict — a story genuinely outside the workspace currently has no way in, which I believe is correct because the plugin has no way to address it.
2. **The card comes from the plugin's reopen route, not from reading the file here.** `POST /api/plugin/presentMulmoScript { filePath }` — the same call the agent's tool makes. It costs a round trip before the Canvas opens. The alternative was parsing and validating the script in the browser, which duplicates the route's logic (and its schema defaults — the route fills in `canvasSize`, `speechParams` and friends, which a hand-built card would not have).
3. **`canOpenInCanvas` gained a second parameter with a default of `null`.** A caller that forgets it silently loses stories while markdown and html keep working. It defaults rather than being required because the FilesOverlay mount has no workspace to pass. Worth a second opinion.
4. **Failure is a field, not a status.** The route narrates a missing or refused file as a **200 with no `data`**, pinned by `server/backends/mulmoscript.spec.ts`. This reads `data`, not `res.status`.
5. **One flaky test, not from this change.** `test/server/session/tool-store.spec.ts > drops a placeholder the agent's real card has already beaten` failed once in a full run, then passed in 3 further full runs on this branch and 2 on clean `main`. This diff contains no server code, and that spec's fixture uses a per-test `mkdtemp`. Reporting rather than diagnosing — I did not chase it further.

## User Prompt

> 次いこう!!

Following the plan agreed on #1374 and recorded when #1380 merged: step 1 was `presentDocument` + `presentHtml`, step 2 is `presentMulmoScript`.

## Implementation

`plans/feat-1374-canvas-mulmoscript.md` has the design record. Three additions to `src/composables/canvasOpenFile.ts`:

- **`storyWirePath(absolutePath, workspace)`** — pure. Is this file under `<workspace>/artifacts/stories/` and a `.json`? Answers with `stories/<rest>`.

  Built on `dirPathKey` (`common/dirPathKey.ts`), the key the workspace chip already compares with: both separators fold, so the mixed path `absoluteUnder` produces on Windows still matches; `..` folds away, so a traversal stops matching the prefix rather than needing to be spotted; drive and UNC roots survive. Lexical, and **nothing rests on it** — a browser cannot resolve a symlink, and the reopen re-runs the plugin's own guard plus a realpath check server-side, so a path this lets through still yields no card.

- **`reopenStory(wirePath)`** — the round trip. Returns the card, or null when the route reports no `data`.

- **`buildCanvasCard(absolutePath, workspace)`** — the async entry point. markdown and html are still decided in memory; only a story costs a request.

`FilesPane` takes a `workspace` prop and `TerminalGrid` passes `defaultCwd` (CLAUDE_CWD, what `launchChips` labels WORKSPACE). **The button and the card builder are called with the same arguments** — #1380 shipped with those two disagreeing, and the symptom was a button that did nothing when pressed.

### On MulmoClaude

Per CLAUDE.md I checked the reference host first. MulmoClaude has **no counterpart** — nothing there opens a file into the canvas without the agent; its `src/utils/collections/presentSeed.ts` seeds a *client-only* placeholder and reconciles it away, where mulmoterminal deliberately persists (that divergence predates this PR, from #1768). Its route namespace also differs (`/api/mulmoScript/save` vs. this repo's `/api/plugin/presentMulmoScript`), which likewise predates this PR; I used the route this repo already mounts and tests.

## Verification

**Real browser, real server** (throwaway `HOME`, so no personal directories appear):

- A workspace story opens and renders in full — title, both beats, Movie / PDF, Edit Script Source — **on a shell cell whose Canvas button read "No render MCP for this directory"**. That is the case the feature exists for.
- An identically-shaped `artifacts/stories/tale.json` under a **non-workspace** directory offers **no button**.
- Zero console errors in both.
- The route itself checked directly: `{"filePath":"stories/tale.json"}` returns `data` with the schema-completed script; `stories/nope.json` returns `{"message":"File not found: …"}` with no `data`.

**Tests** — 16 new. Every one was re-run against the code with its own fix removed, and five mutations were each caught by the intended test:

| mutation | caught by |
| --- | --- |
| root at any directory, not the workspace | the wire-path and Windows tests |
| match the story shape anywhere | "refuses an identically-shaped path under another directory" |
| send the absolute path to the route | "reopens a story … and carries back what it returned" |
| drop the `.json` requirement | "refuses a file in the story directory that is not a script" |
| treat any 200 as a card | "has no card when the route narrates the file as missing" |

`yarn format` / `lint` (0 errors) / `typecheck` / `build` / `test` (7937 passed).

refs #1374

work in mulmoterminal3

## Summary by CodeRabbit

* **New Features**
  * Added support for opening eligible MulmoScript story files in Canvas.
  * Story files are recognized only within the workspace’s designated stories directory.
  * Canvas cards are generated from reopened story data while retaining existing Markdown and HTML support.
  * Added safeguards against invalid paths, traversal attempts, unsupported extensions, and project artifact conflicts.

* **Bug Fixes**
  * Improved Canvas file handling for missing or unsuccessful story responses.

* **Tests**
  * Added coverage for workspace scoping, path validation, Windows separators, asynchronous loading, and error scenarios.

### DIFF
--- a/plans/feat-1374-canvas-mulmoscript.md
+++ b/plans/feat-1374-canvas-mulmoscript.md
@@ -1,0 +1,85 @@
+# feat: MulmoScript を Canvas に開く (#1374) — 第2段
+
+issue #1374 の**実装順 2**。第1段（`presentDocument` + `presentHtml`、#1380）で入った
+Files ペインの Canvas ボタンに `presentMulmoScript` を足す。
+
+## 第1段と決定的に違う 2 点
+
+第1段の md / html は「パスだけ載せた合成カードを書けば View が自己 fetch する」で済んだ。
+mulmoScript は**どちらも成り立たない**。
+
+### 1. 絶対パスが使えない
+
+`normalizeStoryPath`（`@mulmoclaude/mulmoscript-plugin`）は **絶対パスとバックスラッシュを拒否**する。
+受けるのは `stories/foo.json` / `foo.json` / `artifacts/stories/foo.json` の 3 つだけ。
+
+第1段は「ペインはセルの cwd、プラグインはワークスペース」の食い違いを**絶対パスに結合して**
+解決した（md / html のガードは絶対パスを受ける）。mulmoScript ではその手が使えないので、
+**ワークスペース相対の `stories/<name>.json` を組み立てる**しかない。
+
+つまりゲートは「拡張子が合うか」ではなく **「このファイルはワークスペースの
+`artifacts/stories/` の中にあるか」** になる。プロジェクトセルが自分の `artifacts/stories/` を
+持っていても、それはプラグインが開くファイルではないので**出してはいけない**。
+
+ワークスペースは UI 側に既にある — `defaultCwd`（CLAUDE_CWD、`/api/config` 由来。
+`launchChips` が WORKSPACE チップに使っているのと同じ値）。
+
+比較は `common/dirPathKey.ts` の `dirPathKey` に載せる。セパレータ両方を畳み、`.` / `..` を
+解決し、Windows のドライブルート / UNC を落とさない。**`..` が畳まれる結果、traversal は
+プレフィックス照合に失敗して自然に落ちる**。
+
+なおこの照合は**字句的**（ブラウザに symlink は解決できない）。第1段と同じく、
+**間違ってもフェイルセーフ**である点が効く — 下の reopen がサーバ側で `normalizeStoryPath` と
+realpath ガードを通すので、UI が通してもカードは書かれない。
+
+### 2. カードにスクリプト本体が要る
+
+`MulmoScriptData` は `{ script: MulmoScript; filePath: string }`。md の `docPath` や html の
+`filePath` と違い、**`script` を欠いたカードは描けない**。
+
+自前で読んで検証するのは、プラグインのロジックを二重化することになる（第1段で「判定は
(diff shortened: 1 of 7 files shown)


isamu:
@coderabbitai review

PR #1391, opened 2026-08-04
fix: bound every request in canvasOpenFile, and pin the plugin ordering

isamu:
## Summary

Two findings from reviewing #1388 that **missed the merge**. #1388 was merged at its first commit while the review loop was still running, so this carries the loop's result forward. No behaviour from #1388 changes.

## Items to Confirm / Review

1. **The timeout value.** 10s, matching the shape of the repo's other bounded callers. The reopen can legitimately take a moment on a large script — if 10s is too tight for a real story, say so and I will raise it.
2. **Three call sites, one helper.** `seedCanvasCard` and `hasStoredCard` shipped unbounded in #1380; they are fixed here too rather than left inconsistent beside the new one. Tell me if you would rather keep this PR to the reopen alone.

## 1. No request here was bounded

CLAUDE.md requires `AbortController` timeouts on network requests, and this repo already does it in nine places with a named constant (`useCost`, `useHandoff`, `useLaunchOptions`, …). None of the three fetches in `canvasOpenFile.ts` had one.

It matters most on the **reopen**, because that call blocks the Canvas from opening: a server that never answers leaves the button pressed and nothing happening, which a user cannot tell apart from "this file cannot be shown". The seed and the stored-card read have the same shape, so all three go through one `fetchBounded` rather than three copies.

## 2. The ordering in `buildCanvasCard` was load-bearing and unpinned

markdown and html are asked before the story branch. If either ever accepted `.json`, **every story in the workspace would quietly open as that other thing** — no error anywhere.

Pinned against the plugins themselves rather than against my reading of them, since a package upgrade is exactly how it would change. The converse is pinned too: a `.md` inside `artifacts/stories/` is still a document, not a story.

## Verification

Each new test was re-run against the code with its own fix removed:

| mutation | result |
| --- | --- |
| `fetchBounded` → plain `fetch` | all three timeout tests hang to the 15s test timeout |
| drop `storyWirePath`'s `.json` requirement | the ordering test fails |

`yarn format` / `lint` (0 errors) / `typecheck` / `build` / `test` (7959 passed) on top of current `main`.

refs #1374

work in mulmoterminal3

### DIFF
--- a/src/composables/canvasOpenFile.ts
+++ b/src/composables/canvasOpenFile.ts
@@ -109,6 +109,26 @@ export const canOpenInCanvas = (path: string | null, workspace: string | null = 
   path !== null && (canvasCardForFile(path) !== null || storyWirePath(path, workspace) !== null);
 
 /**
+ * Every request here bounded, per the repo's other API callers (useCost, useHandoff, …).
+ *
+ * Not a formality on this path: the reopen BLOCKS the Canvas from opening, so a server that never
+ * answers leaves the button pressed and nothing happening, with no way for the user to tell that
+ * from "this file cannot be shown". A rejection at least reaches the caller, which gives up
+ * visibly.
+ */
+const REQUEST_TIMEOUT_MS = 10_000;
+
+async function fetchBounded(url: string, init?: RequestInit): Promise<Response> {
+  const abort = new AbortController();
+  const timer = setTimeout(() => abort.abort(), REQUEST_TIMEOUT_MS);
+  try {
+    return await fetch(url, { ...init, signal: abort.signal });
+  } finally {
+    clearTimeout(timer);
+  }
+}
+
+/**
  * The mulmoScript card for `wirePath`, built by the plugin's own reopen rather than here.
  *
  * `MulmoScriptData` requires the parsed `script`, not just a path — unlike markdown and html,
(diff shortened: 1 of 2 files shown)


isamu:
## Review loop — iteration 1

**Codex:** LGTM. **CodeRabbit:** "Review limit reached" again — no review happened, so its SUCCESS check is not a second opinion. **CI:** 11/11 green.

My own reading of the diff turned up two things worth stating rather than silently leaving.

### Deliberately deferred — adjacent code with the same gap (now #1393)

Bounding the three requests in `canvasOpenFile.ts` immediately raises "what about the ones next door", and the answer is that they are unbounded too:

- `src/composables/seedCollectionCanvas.ts:20` — `/api/collections/list`
- `src/composables/seedCollectionCanvas.ts:54` — `/api/agent/toolResult` (the same route this PR bounds)
- `src/components/GuiPanel.vue:117` — `/api/agent/toolResult`

It is not three stragglers. Across `src/` there are **77 `await fetch(` calls and 10 with a signal**, and those 10 inline the same `AbortController` + `setTimeout` in **10 separate files** — there is no shared helper, so `fetchBounded` here is the tenth copy of a pattern rather than a new one.

Fixing that is a repo-wide sweep behind one shared helper, which is a different change from "carry #1388's review forward". Filed as **#1393** with the call sites and the proposal, rather than widened into this PR.

### Accepted consequence of the timeout

A seed that times out may still be applied server-side — the client gives up while the request completes. The user then sees nothing happen, presses again, and a second card is written. `collapseByIdentity` collapses the two by `filePath`, so the Canvas shows one card either way; the store keeps both, inside its 50-entry cap.

That is strictly better than the unbounded version, where the same failure is indistinguishable from a hang with no way out, and it is the same trade the repo's other bounded callers make. Recording it rather than hiding it.

work in mulmoterminal3

PR #1394, opened 2026-08-04
fix: refuse a harness-injected prompt on the hook path too (#1384)

### LINKED ISSUE #1384: fix: ハーネスが注入したプロンプト（task-notification / system-reminder）がセルの row 1 と AI タイトルに出る
## What happened

セルの row 1 に、ハーネスが注入した XML がそのまま出る:

```
<task-notification> <task-id>boen82mbm</task-id>
```

バックグラウンドタスク（Monitor / サブエージェント）を持つセッションで、その通知が届くたびにヘッダーのテキストが通知本文に置き換わる。ユーザーが実際に打った依頼は見えなくなる。

## What I expected

ヘッダーには、ユーザーが打った直近の意味のあるプロンプトが出続けること。ハーネスが注入したものは、スラッシュコマンドと同じで「打ったプロンプト」ではない。

## なぜ起きるか — 同じ判定が2経路にあり、片方にしか無い

row 1 のテキストは `memo → aiTitle → lastPrompt` の優先順（`common/sessionMemo.ts:42` の `sessionDisplayName`）。今出ているのは `lastPrompt`。

`lastPrompt` を書いているのは `UserPromptSubmit` フック:

```
server/routes/hook-routes.ts:98-107   trackPromptForHeader()
  → lastPrompts.set(sessionId, preferredHeaderPrompt(...))
```

`preferredHeaderPrompt`（`server/session/transcript.ts:225`）が弾くのは `isTrivialPrompt` — 「ok」のような短い相槌だけ。注入された長い XML はむしろ「意味のあるプロンプト」と判定され、以後そこに居座る。

**一方、transcript を読む経路には正しいフィルタが既にある。** `server/session/transcript.ts:14-21`:

```ts
if (typeof text === "string" && text.trim() && !/^\s*<(local-command|command-|bash-|task-notification)/.test(text)) {
  return text.trim();
}
```

コメントまで付いている:

> `task-notification` is in the list for the same reason as the rest: the harness writes it into the user channel when a background task finishes, and it is no more a typed prompt than a slash command is.

つまり **「これはユーザーのプロンプトではない」という同じ判断が2箇所にあり、transcript 側だけが実装されていて、ライブのフック側には無い**。

`<system-reminder>` はどちらのリストにも入っていない。

## 影響範囲

`applyHeaderHooks`（`server/routes/hook-routes.ts:135-142`）は `kind: "prompt"` で2つのことをする:

```ts
await trackPromptForHeader(sessionId, effect.text, cwd);
deps.noteTitleTurn(sessionId, effect.text);
```

なので壊れるのはヘッダーのテキストだけでなく、**その注入テキストから AI タイトルが生成される**経路も同じ。

`handleActivityHook` は別なので、working/waiting の追跡には影響しない（注入プロンプトも実際にターンを始めるので、そこは今のままで正しい）。

## Steps to reproduce

1. セルの claude セッションでバックグラウンドタスクを起動する（Monitor など）
2. そのタスクが通知を出す
3. セルの row 1 が `<task-notification> <task-id>…` に変わる

※ セッションノートを書いていると、ノートが `lastPrompt` より優先されるので**症状が隠れる**（`sessionDisplayName` の優先順）。再現時はノート無しのセルで。

## 直し方の案

判定を1つに寄せて両方から呼ぶ。CLAUDE.md の「両側が判断する値は共有し、2箇所に写さない」に沿う形:

1. `server/session/transcript.ts` の `userPromptText` にインラインで書かれている正規表現を、名前付きの述語（例 `isInjectedPrompt(text): boolean`）として切り出して export する
2. `userPromptText` はそれを呼ぶ形に変える（挙動は変わらない）
3. **`<system-reminder>` をそのリストに足す** — 現在どちらのリストにも無い
4. `headerHookEffect`（`server/session/header-hook.ts:27-31`）が `UserPromptSubmit` に対して、注入プロンプトなら `null` を返すようにする

**4 を `headerHookEffect` に置くのが良い理由**: あの関数は「このフックはヘッダーについて何も変えない」を `null` で表す純粋関数で、既に空プロンプトを同じ理由で弾いている（`transcript.ts` ではなく `header-hook.ts` の話）。ここで止めれば `trackPromptForHeader` と `noteTitleTurn` の**両方**が一度に守られる。呼び出し側に `if` を2つ足すより、判断が1箇所で済む。

`null` を返すのは「何もしない」であって「消す」ではないので、直前の本物のプロンプトがそのまま残る — これが欲しい挙動。

### テスト

- `header-hook` の spec に、`<task-notification>` / `<system-reminder>` / `<local-command…` の `UserPromptSubmit` が `null` になることを追加
- 述語の spec で、普通のプロンプトが弾かれないこと（本文の途中に `<` があるだけのものなど）を固定
- `userPromptText` の既存の挙動が変わらないことを確認

## Environment

- mulmoterminal 4.3.1 / Node v22.18.0 / macOS (Darwin arm64) / zsh
- tmux 3.7b / Claude Code 2.1.221
- ローカル開発（ビルド済み `dist/` を配信）
- `~/.mulmoterminal/config.json`: `chips: null`（未設定＝既定）

isamu:
## Summary

A cell whose session runs a background task showed the harness's own XML on row 1 —
`<task-notification> <task-id>boen82mbm</task-id>` — and kept showing it, with the AI title then
generated from that XML.

The judgment "a harness-injected block is not a typed prompt" already existed. It was added in
`e1e19c66` (2026-07-22), a two-file commit that touched **the path that reads a transcript** and
nothing else. The **live `UserPromptSubmit` hook** — the only unfiltered writer of `lastPrompts`, and
the one that actually puts text on the header — was never taught it.

This lifts the rule into one exported `isInjectedPrompt` and calls it from both paths, refusing an
injected prompt in `headerHookEffect` before it can become an effect.

## Items to Confirm / Review

1. **`<system-reminder>` is added on zero evidence of the harness ever producing it at the start of a
   user line, and I want that called out rather than buried.** The issue asks for it. Scanning every
   transcript on this machine (9,902 files, 26,465 user lines carrying text), `<system-reminder>`
   appears 9 times and is *embedded* in all 9 — never leading. Running the new predicate over all of
   those lines refuses exactly the same **4,832** as the old one, so nothing real changes hands. It
   is defence against a shape not currently observed. The one cost: a person whose prompt literally
   begins with `<system-reminder>` gets no header update — a trade already accepted for the other
   four markers.
2. **`header-hook.ts` previously imported nothing at all**, deliberately ("pure decision"). It now
   imports `isInjectedPrompt` from `transcript.ts`. That file imports only `isRecord` / `readString`
   and touches no filesystem, so the module stays pure — but if you'd rather the predicate lived in
   its own module than in the transcript reader, that's a one-file move and I'd take the redirect.
   The issue proposed `transcript.ts`, which is what I followed.
3. **Whether refusing is right for the AI title as well as the header.** One `kind: "prompt"` drives
   both, so refusing at `headerHookEffect` means a turn started by an injected prompt contributes
   nothing to the title either. I believe that's wanted — the title should describe the user's task —
   but it is a behaviour change beyond row 1, so it deserves a look.
4. **`docs/ChangeLog.md` gets an `Unreleased` entry.** Say if fix-only entries are meant to wait for
   the release commit instead.

## User Prompt

- https://github.com/receptron/mulmoterminal/issues/1384 の原因はわかるか。
- 原因が判明したら、その「強固な方法」（判定を1つに寄せて両方から呼び、テストで固定する）で進めて、PR 作成からレビューまでやること。

## Why it happened

`lastPrompts` — the map row 1 reads — has exactly two writers:

| writer | filtered? |
| --- | --- |
| `hook-routes.ts:103`, seeding from the transcript (`latestUserPrompt` → `collectPrompts` → `userPromptText`) | yes, by `e1e19c66` |
| `hook-routes.ts:106`, the `UserPromptSubmit` hook (`headerHookEffect` → `trackPromptForHeader`) | **no** |

The transcript *fallback* in `sessionDetailView` is filtered too (`readSessionSummary` →
`createSummaryScan().finish()` → `latestMeaningfulUserPromptFromParsed` → `userPromptText`). So the
live hook was the only way the XML could reach the screen — which is why the symptom needs a *live*
background task to appear, and why writing a session memo hides it (`sessionDisplayName` ranks the
memo first).

`preferredHeaderPrompt` did not save it either. Its only guard is `isTrivialPrompt`, which drops
short acks like "ok"; a 200-character XML block is the opposite of trivial, so it was taken as the
session's most meaningful prompt and stayed.

### What it reached

`applyHeaderHooks` turns one `kind: "prompt"` into two writes, and `lastPrompts` is read from five
places — so this was wider than row 1:

| consumer | what the user saw |
| --- | --- |
| `session-routes.ts:91` → `sessionDetailView` | cell row 1 (the reported symptom) |
| `hook-routes.ts:140` `noteTitleTurn` | the AI title generated **from the XML** |
| `lifecycle.ts:199` | the session-list payload |
| `task-push.ts:54` → `buildPushDetail` (`reply \|\| lastPrompt \|\| aiTitle`) | a Web Push whose body is the XML, whenever the reply read comes back empty |
| `index.ts:707` `promptOf` → `remoteHost/terminalScreen.ts:240` | the remote host's terminal screen |

`handleActivityHook` is a separate path, so working/waiting is untouched — correctly: an injected
prompt does start a real turn.

## The change

1. `server/session/transcript.ts` — the inline regex becomes an exported
   `isInjectedPrompt(text: string): boolean`, with `system-reminder` added to the marker list.
   `userPromptText` calls it; behaviour for the four existing markers is unchanged.
2. `server/session/header-hook.ts` — `headerHookEffect` returns `null` for a `UserPromptSubmit`
   whose prompt is injected.

`null` is the function's existing meaning of "this hook changes nothing about the header", already
used for a blank prompt. It is **not** "clear it" — the real prompt from before the background task
stays on screen, which is the wanted behaviour. Deciding it there covers `trackPromptForHeader` and
`noteTitleTurn` at once, and through them all five consumers above, instead of two `if`s at the call
site.

Design note in `plans/fix-1384-injected-prompt-header.md`.

## Ground truth

Two checks against real harness output rather than against fixtures I wrote.

**What the harness actually writes** — every user line of every transcript on this machine, joined
the way `userPromptText` joins content blocks:

| marker | at start | embedded |
| --- | --- | --- |
| `<task-notification` | **3,643** | 591 |
| `<command-` | 669 | 0 |
| `<local-command` | 361 | 0 |
| `<bash-` | 158 | 0 |
| `<system-reminder` | **0** | 9 |

The `^\s*` anchor is load-bearing: those 591 embedded mentions are real prompts (the `/loop` skill's
own documentation among them), and an unanchored match would delete them.

**The shipped predicate over that same corpus** — importing `isInjectedPrompt` itself, not a copy:

```
user lines with text : 26465
refused BEFORE       : 4832
refused NOW          : 4832
newly refused        : 0
```

The same run asserts `isInjectedPrompt` and `userPromptText` never disagree on any real line.

**Against a running server** — `PORT=39847 node --import tsx server/index.ts`, posting real-shaped
`UserPromptSubmit` payloads to `/api/hook` and reading `lastPrompt` back from
`/api/session/<id>` after each:

| # | posted | `lastPrompt` after |
| --- | --- | --- |
| 1 | `fix the login bug in the auth flow` | `fix the login bug in the auth flow` |
| 2 | `<task-notification><task-id>boen82mbm</task-id>…` | **unchanged** — `fix the login bug in the auth flow` |
| 3 | `<system-reminder>background context</system-reminder>` | **unchanged** |
| 4 | `what is a <task-notification> block?` | `what is a <task-notification> block?` |
| 5 | 250 chars of real prompt | accepted, capped to 200 |

Rows 4 and 5 are the negative controls: the guard refuses the injection without becoming a blanket
"reject anything with a tag in it", and the existing cap still applies.

## Tests

The failure mode being fixed is **drift between two call sites**, so the tests pin the agreement,
not just each side:

- `test/support/injectedPrompts.ts` — one fixture list of injected samples (shapes copied from real
  transcripts) and of prompts that merely mention a marker.
- `transcript.spec.ts` and `header-hook.spec.ts` both iterate **that same list**. Adding a marker to
  one path and forgetting the other fails the other's spec.
- A direct `isInjectedPrompt` spec for what the anchor decides: leading whitespace/newlines count as
  leading, mid-text mentions do not.
- A case pinning that a reminder appended as its own content block leaves the prompt in front of it
  intact — the shape every `<system-reminder>` on disk actually has.

**Proven to catch the bug**: with the new branch disabled, `header-hook.spec.ts` goes red on 7 cases
(all injected shapes); restored, it is green.

`yarn format` / `yarn lint` (0 errors) / `yarn typecheck` / `yarn build` / `yarn test` — **547 files,
7,989 tests passed**.

Closes #1384

work in mulmoterminal


isamu:
## Bot review triage

| reviewer | outcome |
| --- | --- |
| Codex | `CODEX VERDICT: LGTM` on `9aecfa24`, re-run clean on `c1e5af84`. No findings to address. |
| CodeRabbit | **Never reviewed** — hit the per-developer rate limit on the first push and posted only the limit notice. Proceeding without it, by the author's decision. |
| Socket Security | Project Report + PR Alerts pass. No dependency change in this PR. |
| knip / jscpd | pass — the new `test/support/injectedPrompts.ts` is neither dead nor a duplicate. |

Nothing from a bot needed fixing. The three commits after the first one all come from my own
re-reading, and none of them was flagged by any reviewer:

1. **`01e862d7`** — a comment claimed "for a year only one of them did". `e1e19c66` is dated
   2026-07-22 and this PR is 2026-08-04, so the real gap is two weeks. A comment a maintainer
   cannot trust is worse than no comment.
2. **`24909f6b`** — the comments claimed the shared fixture list catches "a marker added to one path
   and forgotten on the other". One shared predicate makes that state impossible, so the sentence
   was describing the fix as if it were the test. Reworded to what the list actually buys: both call
   sites are held to one specification, so a path that *stops consulting* the predicate — an early
   return, a re-inlined copy — fails its own spec.
3. **`a22cb2a9`** — the same overclaim in the plan file.

## Answering my own review item 3 (the AI title)

I raised whether refusing an injected prompt should also stop it counting toward the AI title.
Checking the two counters says yes, and that the change makes them agree:

- The transcript-side count, `countUserTurnsFromParsed`, goes through `conversationTurnsFromParsed`
  → `userPromptText`, so it has **always** skipped injected lines.
- The live counter, `titleTurnCounts` (via `noteTitleTurn`), counted them.

So before this PR the two disagreed, and a background task reporting could push a session past
`TITLE_REGEN_EVERY_TURNS` on the live side while the transcript side saw no new turn at all. After
it, both count only what a person typed. The one visible consequence: on a session with no title
yet, an injected prompt no longer forces a generation, so the title is produced at the next real
prompt's `Stop` instead — later, from the right text.

## Deliberately not changed

- **The codex rollout path.** `last-turn.ts` reads codex sessions through a different parser and
  those rollouts carry no `<task-notification>` block, so the predicate has no call site there. Out
  of scope for this issue rather than overlooked.
- **`handleActivityHook`.** An injected prompt does start a real turn, so working/waiting must keep
  tracking it. Untouched.

PR #1397, opened 2026-08-04
feat(grid): give each cell its own right pane

### LINKED ISSUE #1378: feat(grid): 右ペインの開閉をセルごとに持てるようにする（今はグリッド全体で 1 値）
右ペインの開閉をセルごとに持てるようにできないでしょうか。バグではなく要望ですが、経緯が分かる方が判断しやすいと思い、bug report の書式に沿って書きました。

## What happened

グリッド表示（どのセルにもズームしていない状態）でエージェントに `presentHtml` を出させたところ、Canvas は開きませんでした。

これは仕様だと分かりました。`test/src/components/canvasAutoOpen.spec.ts` に

```js
// Nothing is enlarged, so there is no slot to put it in.
it("does nothing while no cell is enlarged", …)
```

があり、冒頭にも `The two limits pinned here are the whole design.` と書かれています。「背景のセルが画面を奪わない」という判断は納得しましたし、`unseenCanvas` チップという受け皿が用意されていることも確認しました。**ここは変更を求めていません。**

そこから「では、このセルだけ Canvas を開けっ放しにしておけばよいのでは」と考えたところ、それができないことに気づいた、というのが本 issue です。

## What I expected

セルごとに「このセルでは右ペインを開いておく／開かない」を持てること。

## Steps to reproduce

1. セルを 2 つ開く（A, B）
2. A にズームして Canvas ペインを開く
3. B にズームを移す

**実際**: B でも Canvas が開いた状態で始まる。B で閉じると、A に戻ったときも閉じている。
**期待**: B は B が自分で開いたことのある状態で始まる。B で閉じても A の状態は残る。

## Environment

- mulmoterminal 4.3.0 / Node v24.18.1 / macOS (Darwin arm64) / zsh
- tmux 3.7b / gh 2.92.0 / claude 2.1.220 / codex-cli 0.146.0
- 起動方法: ローカル開発（`server/index.ts` を tsx 経由 + vite）
- ブラウザ: Google Chrome 150.0.7871.187

## 調べたこと

ペインが「**何を映すか**」は既にセル単位になっていました。`GuiPanel` / `ToolsPane` は `:session-id="expandedSessionId"` でズーム中セルのものに張り替わりますし、files ペインに至っては開いていたファイルと展開ツリーをセルごとに覚えています（#958 / #910）。

```ts
// TerminalGrid.vue
const paneStateByUid = new Map<number, FilesPaneState>();  // :449
const claimPaneState = (uid, cwd) => { … }                 // :467
const paneUid = ref<number | null>(null);                  // :483
```

`claimPaneState` のコメント（「同じリポジトリに端末 2 枚は普通の使い方」）を読んで、per-cell が意識的に作り込まれた部分だと理解しました。

一方で「**映すかどうか**」だけがグリッド共有の 1 値でした。

```ts
const PANE_OPEN_KEY = "files_pane_open";   // :155
const rightPane = ref<RightPane | null>(restoredPane(stored(PANE_OPEN_KEY))); // :186
```

ズームが動いても誰もこれをリセットしないため（`watch(zoomed, …)` が触るのは `paneExpanded` だけ、:202-206）、開閉だけがセルをまたいで引き継がれます。キー名が `files_pane_open` のままなのも、files ペイン時代から素直に育った結果だとお見受けしました。

## 判断が要りそうな点

**a. 復元キーは uid / session / cwd のどれか。**
files は uid + cwd の 2 層です。ただし Canvas の中身は cwd ではなく session に紐づく（`:session-id="expandedSessionId"`、履歴も `/api/agent/toolResults/:sessionId`）ので、Canvas は session をキーにする方が自然かもしれません。files と割れることを許すかどうかは、設計を持っている方の判断だと思います。

**b. `files_pane_open` の移行。**
現在 `"files" | "canvas" | "tools" | ""` を持ち、`restoredPane` が旧 `"1"` からの移行も担っています（:181-185）。per-cell 化で形が変わるので、もう一段の移行が要ります。

**c. 未使用セルの既定値。**
「1 度開けばどのセルでも開いている」現状に慣れていると、全セル閉じ始まりは後退に感じられる可能性があります。

**d. `paneWidth` / `paneExpanded` は共有のままでよいか。**
`paneExpanded` は "the takeover is asked for, never inherited" として意図的に記憶しない設計（:190-196）なので、そのままが筋だと思っています。幅も共有で違和感はありません。開閉だけを per-cell にする、という切り分けで合っているでしょうか。

## 関連

- #1237 — 単セルでズーム自体が拒否され Canvas に到達できない（別問題ですが `openCanvasFor` を共有）
- #1374 — present 系をエージェント抜きで開く
- #958 / #910 — files ペインの per-cell / 復元まわりの前例

isamu:
## Summary

The right pane already knew everything about a cell except whether to be there. Which file the
tree had open and which directories were expanded were **per-cell** (#958, #910); the canvas is
read **per session**; but *"is a pane open, and which"* was **one value for the whole grid**. So
opening the canvas on one terminal opened it on the next, and closing it there closed it on the
first.

Each cell now answers for itself.

- Walking the zoom shows **that cell's** pane — files, canvas, tools or nothing.
- A cell that never asked for one **arrives with none**. Closing is an answer too, so it stays closed.
- A reload restores by **session** (a uid is a different number next time), capped at 40, LRU.
- A pane can be asked for on a **tiled** cell — which is where #1378 starts, since the canvas cannot
  open with nothing enlarged. The press records what that terminal should have; enlarging it opens it.
- Each cell's header buttons now show **its own** state.
- Width and the full-width takeover stay shared, as the issue proposed.

`Closes #1378`. One file of source (`TerminalGrid.vue`) plus specs.

## Items to Confirm / Review

Three traps, none of them visible from the feature itself, and each one a regression I hit while
building it:

1. **Collapsing the zoom only HIDES the row** — the pane stays mounted with its editor and buffer.
   Deriving the shown pane from `expandedUid` would close it on every collapse and flush an editor
   the user is coming straight back to. It derives from `paneUid`, which already trailed the zoom
   for exactly this reason. Pinned by the existing "keeps the pane and its buffer mounted while the
   zoom is collapsed" spec.
2. **A button pressed on a TILED cell must not move the pane there** — that would unmount an editor
   merely hidden behind the grid, with a buffer nobody asked to close. The press records; the zoom
   watcher moves it, flush and all, when that cell is actually enlarged.
3. **Reopening files on the cell it was closed on** changes nothing the old watcher looked at, so
   its remembered tree never came back. `rightPane` is a watcher dependency now — please check that
   this cannot loop (it re-enters once and early-returns on `paneCwd === expandedCwd`).

Also worth a look:

4. **`paneByCell` distinguishes absent from null.** Absent means "never asked" and is what lets a
   reload restore from the session store; an explicit `null` is "the user closed it" and must
   survive. A `Map<number, RightPane | null>` carries both, which is easy to flatten by accident.
5. **The old `files_pane_open` key is dropped, not migrated.** Since a cell that has never asked
   starts closed, there is nothing for a single grid-wide value to migrate INTO. An existing browser
   comes back with every cell closed once, then remembers per cell from then on.
6. **The decisions came from the issue's own list (a–d).** Key = uid in memory / session across
   reloads (a); new storage key (b); untouched cells start closed (c); width and `paneExpanded` stay
   shared (d).

## User Prompt

> https://github.com/receptron/mulmoterminal/issues/1378 これってできる？右ペインで mcp の present 系と
> ファイルエクスプローラーがある。それぞれが状態を持つ必要あるからね。編集中とかの切り替え、今どうなって
> いるんだ？まずはできるか検討して。
>
> あーー、良く読んだらセルごとに開閉をもっていて、zoom を閉じて開いても状態を持ってほしいし、grid で開く、、
> では開かないから grid で開くになったときに expand すると開いてほしい、、、なのでセルごとに右の開いている
> 状態＋何が開いているかを持っておきたいってことだね。それなら、やっぱりセルごとに presentHTML なのか
> ファイルエクスプローラーなのかまで含めて持っておいて、セル切り替えると変化してほしいよね。そのときに
> ファイル開いている状態も持って。

The investigation that preceded this is in [`plans/feat-1378-per-cell-right-pane.md`](plans/feat-1378-per-cell-right-pane.md),
including the state table (what was already per-cell) and what happens to an unsaved buffer on
every switch path.

## What was already there

Worth stating so the diff is not read as doing more than it does: the **file the pane had open and
its expanded tree** were already per-cell and already restored across reloads, and the **canvas
contents** were already per-session and served by the API. This change adds the open/close half
only — which is why it is one file.

## Verification

`yarn format` / `lint` (0 errors) / `typecheck` / `build` / `test` — **7951 passed**, from a clean
`--frozen-lockfile` tree.

Five new specs pin the behaviour: walking the zoom between two cells with different panes, a cell
that never asked starting closed, a pane asked for while tiled opening on enlarge, a closed cell
staying closed across a reload, and restore-by-session.

**Not done: a live click-through.** This is a UI state machine and the specs mock the panes, so a
real two-cell walk in a browser would be worth doing before merge.

work in mulmoterminal2

### DIFF
--- a/plans/feat-1378-per-cell-right-pane.md
+++ b/plans/feat-1378-per-cell-right-pane.md
@@ -1,0 +1,81 @@
+# feat(grid): 右ペインの開閉をセルごとに持つ (#1378)
+
+## 何を変えるか
+
+グリッドの右ペインについて、**「開いているか」と「何が開いているか」をセルごとに持つ**。
+セルを切り替えると、そのセルの状態に切り替わる。
+
+いまグリッド全体で 1 値なのは、この 2 つだけ（`TerminalGrid.vue:155,186`）。
+「何を映すか」は既にセル単位になっている:
+
+| 何 | 今の紐づけ | この変更後 |
+| --- | --- | --- |
+| 開いているか / どのペインか | **グリッドで 1 値**（localStorage `files_pane_open`） | **セルごと** |
+| files が開いていたファイル・展開ツリー | セル(uid) ごと + cwd ごと（#958） | 変更なし |
+| files の編集バッファ | 保持しない（出る前にディスク保存、無理なら backup） | 変更なし |
+| canvas の中身 | セッションごと・サーバ（`/api/agent/toolResults/:id`） | 変更なし |
+| 幅 / 全画面 | グリッドで 1 値 / 記憶しない | 変更なし |
+
+## 決めたこと
+
+| 論点 | 決定 |
+| --- | --- |
+| まだ開いたことのないセル | **閉じた状態で始める**。自分で開いたセルだけが開く |
+| リロード後 | **sessionId をキーに localStorage から復元**（uid はリロードで変わるため） |
+| 幅 `paneWidth` / 全画面 `paneExpanded` | **共有のまま**（`paneExpanded` は「記憶しない」が既存の設計判断） |
+| セル移動時に files の保存が失敗したら | **既存の「ペインがズームに追従せず前のセルに留まる」挙動をそのまま使う**。新しい概念を足さない |
+
+## 踏んではいけない罠
+
+**ズームを閉じてもペインは unmount していない。** `zoom-row` は `hidden` になるだけ
+（`TerminalGrid.vue:983`）で、FilesPane は mount されたまま — だからいまはズームを閉じて開き直しても
+編集中のバッファごと残る。
+
+ここで「表示するペイン = `props.expandedUid` のセルの状態」と素直に書くと、**ズームを閉じた瞬間に
+uid が null → ペインが閉じる → FilesPane が unmount → 毎回 flush** になり、今より悪くなる。
+
+逃げ道は既にある。**`paneUid`（`:532`）が「ペインが今どのセルに乗っているか」**という意味の ref で、
+ズームに追従しつつ、ズームを閉じている間は最後のセルに留まる。これを files 専用から
+**ペイン全体の identity に一般化**する。新しい ref を増やさない。
+
+## 実装
+
+### `TerminalGrid.vue`（実質このファイルだけ。`rightPane` の参照は 24 箇所、代入は 1 箇所）
+
+1. `rightPane` を ref から **computed** にする。実体はセル別 Map:
+
+   ```ts
+   const paneByCell = ref(new Map<number, RightPane>()); // 不在 = 閉じている
+   const rightPane = computed(() => (paneUid.value === null ? null : (paneByCell.value.get(paneUid.value) ?? null)));
+   ```
+
+2. `setRightPane(pane, uid)` が Map を書く。`openCanvasFor` は uid を明示して呼ぶ
+   （`emit("toggle-expand")` の直後は `props.expandedUid` がまだ更新されていない）。
+
+3. `paneUid` を一般化する:
+   - 「files を離れるとき null にする」（`:240`）を **やめる** — ペインの identity になったので消さない。
+     `paneCwd` / `paneState` のリセットは今のまま
+   - 再ルート watcher（`:538`）を **files が開いているときだけ動く**から **どのペインでも動く**へ。
+     files 固有の処理（flush / snapshot / reload）は、files が絡むときだけ実行する
+
+4. リロード復元: `localStorage` に `Map<sessionId, RightPane>` を持つ。セルが初めて表示されたとき、
+   メモリに無ければ sessionId で引く。files の cwd 別レイヤ（#958）と同じ発想。
+   旧キー `files_pane_open` は**捨てる**（新規セルは閉じ、が決定なので移行先が無い）。
+
+5. `gridCellProps` の `rightPane` / `filesOpen` を**そのセルの値**にする。
+   各セルのヘッダのボタンが、自分の状態を示すようになる。
+
+### テスト
+
+- ズーム A→B でペイン種別が入れ替わる / A に戻ると A の状態
+- **ズームを閉じて開き直してもペインが閉じない**（罠の回帰テスト。FilesPane が unmount しないこと）
+- 初めてのセルは閉じている
+- リロード相当（新しい uid + 同じ sessionId）で復元される
+- files が dirty で flush 失敗のとき、ペインが前のセルに留まる
+- 既存の canvas 自動オープン（ズーム中セルのみ）が壊れていないこと
+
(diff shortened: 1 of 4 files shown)


isamu:
Codex is right, and this was already being fixed here when the verdict landed — `02331720` has it, with a test.

`paneExpanded.value = false` ran on every `setRightPane`, including the new path where the button belongs to a **tiled** cell. That path exists to record what a terminal should have when it is enlarged and to change nothing on screen, so collapsing the full-width Canvas the user is reading was a second cell's button rearranging the one in front of them. It is now gated on the pane being the one shown (`paneUid.value === uid`).

The spec added with it (`paneExpand.spec.ts` → "survives a pane asked for on another, tiled cell") fails on the previous commit.

isamu:
Right again — `aa78a360`.

The watcher only filed the pane's tree when it moved to **another cell**, but a terminal that changes directory is leaving its tree behind just as surely, and that snapshot is what the directory layer (#958) restores from. So cd-ing away and coming back later restored whatever was there before, or nothing.

It was unconditional before this PR. The guard I added to keep a **first showing** from snapshotting a pane that has not mounted yet took the cd case with it; the first showing is now recognised by having no root yet (`paneCwd === null`), which is what it actually is.

A spec covers it: a cell whose `cwd` prop changes files its tree under the directory it is leaving.

isamu:
Review triage for this PR, so a human reviewer does not have to re-walk the threads.

**Codex** — two rounds of `CHANGES REQUESTED`, both real, both fixed with a targeted test; third pass `LGTM`.

| # | Finding | Fix |
|---|---|---|
| 1 | A pane button on a **tiled** cell reset `paneExpanded`, collapsing the full-width Canvas the user was reading | `02331720` — gated on the pane being the one shown. (Found here in parallel; the fix was written before the verdict landed.) |
| 2 | The zoom watcher snapshotted the files tree only when the pane moved to **another cell**, so a terminal that **cd'd** lost its tree for the directory it left | `aa78a360` — snapshot on every re-root; "first showing" is now recognised by having no root yet, which is what it actually is |

**CodeRabbit** — reviewed, no actionable comments.

**Nothing deliberately skipped.** All checks green (10/10).

**Still outstanding, and not something a bot can cover:** no live click-through. This is a UI state machine and the specs mock both panes, so a real two-cell walk in a browser — open files on one, canvas on the other, walk the zoom, collapse and re-enlarge, reload — is worth doing before merge.

PR #1403, opened 2026-08-04
feat(github): say why a work comment could not be written

isamu:
## Summary

`issueWorkComments` failed **silently whichever way it failed**. A `gh` logged in without write
access did nothing, forever, with no explanation anywhere — indistinguishable from leaving the
setting off. This names the cause and shows it once.

Three layers were dropping it: `ranOk` discarded the CLI's stderr, `ensureWorkComment` collapsed
every cause into one `gh-failed`, and the client's `postWorkComment` ignored the response inside a
`catch {}`.

Now: the stderr is classified into `cli-missing` / `auth` / `permission` / `unknown`, returned as
`failure` on the response, warned once per (repo, cause) in the server log, and shown in the cell
header as a dismissible `issue not updated — …` notice naming the fix. **The work itself is
unaffected, exactly as before** — the comment is skipped and the next milestone retries.

This is the last unaddressed item from #1369's checklist. The other two are deliberate: CI results
were decided against in #1376, and the "edits send no notification" question was parked pending
real use.

## Items to Confirm / Review

1. **Classifying on the HTTP status, not the English.** The three spellings agree on nothing else
   (`gh: … (HTTP 403)` vs `HTTP 401: …` vs `glab: … (HTTP 401)`). Every string in the spec is real
   captured output, not invented to match the code.
2. **404 is deliberately NOT `permission`.** The issue was read successfully moments earlier, so a
   404 on the write is as likely to be a deleted comment — naming the wrong fix seemed worse than
   naming none. Say if you'd rather it were treated as no-access.
3. **`permission` (403) is the one case not reproduced live.** No under-scoped token here, and
   probing it against someone else's repository would mean actually attempting a write. Instead a
   genuine 403 was captured from a read that requires push access, and used as the spec's input.
4. **The notice is dismissed globally, not per cell.** One login produces one cause and nine cells
   hit it; nine chips for one problem seemed worse. Dismissal is keyed by cause, so silencing
   "log in" does not also silence "no write access".
5. **The notice sits outside the configurable chip list**, like `CellTidyPrompt` — a user who
   removed the `work` chip still needs to hear that a setting they enabled is doing nothing.
6. **Wording** of the four labels/hovers (`src/composables/workCommentNotice.ts`) — these land in a
   terminal header, so they are terse.

## User Prompt

> https://github.com/receptron/mulmoterminal/issues/1369 続きできる？
>
> これを実装したらissueはcloseしよう

Asked which of #1369's remaining items to take; chose **the `gh` write-permission one** — the only
item the issue records as still untouched.

## Verification

Unit tests are argued against fakes, so the CLI behaviour was checked against the real thing.

**Real `gh`, end to end** (`ensureWorkComment` against `receptron/mulmoterminal#1369`; all three
fail at the read, so **nothing was ever posted to the issue**):

| Condition | Result |
| --- | --- |
| `GH_TOKEN=bogus_xxx` (401) | `{"posted":false,"reason":"gh-failed","failure":"auth"}` |
| Logged out — empty token + empty `GH_CONFIG_DIR` (**no HTTP status at all**) | `failure:"auth"` |
| `gh` removed from `PATH` | `failure:"cli-missing"` |

The logged-out row is why the classifier does not read the status alone: `gh` declines to call the
API and prints only an invitation to log in, so the commonest cause would have filed as `unknown`.

**Icon renders** — `comments_disabled` measured in a real browser: it collapses to one glyph (48px)
exactly like the already-shipping `task_alt`, where a nonexistent name stays 1152px. A missing
ligature renders as literal text, which no unit test can see.

**Tailwind utilities** `border-border` / `bg-elevated` / `text-muted` confirmed present in the built
CSS — a mistyped class still builds, it just renders unstyled.

`yarn format` / `lint` / `typecheck` / `build` green. Full suite (8103 tests) green on two
consecutive runs.

## Docs

`docs/guide/{en,ja}/config.md` and the README row now say write access is required, and what the
cell shows when it is missing.

work in mulmoterminal4


isamu:
Addressed in f6bcd2c9.

**Codex — stale `postWorkComment` response overwriting `commentFailure`**: accepted and fixed. Gated on the same `my === req` token `item` already uses. This one matters beyond lateness: `permission` is a per-repository answer, so a cell that moved on would have been shown a verdict about the repository it left.

Added `ignores a cause for a directory the cell has already left` — verified it fails with the guard removed and passes with it, so it is not a vacuous test.

CodeRabbit was rate-limited on the first push and posted no review; I will re-check once its window resets before merging.

Independent review of the rest of the diff found nothing further. Full suite green (8104 tests), format/lint/typecheck/build clean.

isamu:
Addressed in b7c01f1e — **accepted in part, and I want to flag the part I did not take.**

**Accepted:** a cause did survive the cell losing its `cwd`, and surviving a move to a branch with no issue. Both leave a notice about work the user cannot act on from there. Cleared on both.

**Not accepted: clearing on `kind === null`.** A cause is recorded on a *milestone*, and every poll between milestones reports nothing — `workCommentToPost` returns null on all of them. Clearing there would take the notice down on the next 30s tick, which is the same silence this PR exists to end.

I verified rather than argued: applying the suggestion literally (`next.issue === null || kind === null`) makes `keeps the cause through the ordinary polls that report nothing` fail. Each of the three new tests was checked for vacuity by reverting its guard individually.

Moving to a **different** issue needs no case — that is a `start` milestone, so the attempt overwrites the cause anyway.

Full suite green (8107 tests); format/lint/typecheck/build clean. CodeRabbit was rate-limited on both pushes so far; I will re-check before merging.

isamu:
Addressed in 3bec5f06 — accepted, and it is the sharpest of the three.

Switching `issueWorkComments` off is precisely what a user reaches for **because** the notice told them their login cannot write. The notice then stayed up, arguing with the switch they had just used.

`commentFailure` is now cleared whenever there is nothing left for a comment to be about: the setting is off, or the branch carries no issue. Still not cleared on an ordinary no-milestone poll, for the reason given on the previous commit — both behaviours have a test, and each fails when its own guard is reverted.

The lifecycle is now closed on every path I can find: no directory, no issue, setting off, a later milestone succeeding, a stale in-flight response, and dismissal. The one deliberate survivor is a cause between milestones on the same issue, which is the feature.

Full suite green (8108 tests); format/lint/typecheck/build clean.

PR #1441, opened 2026-08-05
feat: Grok as a fourth first-class agent in the Agent Picker

snakajima:
Pick **Grok** in an empty cell and get a real agent session: it runs, it resumes across a reload and a server restart, it carries the GUI MCP tools its directory registered, and it wears a badge in the sidebar and the tab bar. The phone can ask the grid to start one.

Plan and reasoning: `plans/feat-grok-agent.md`.

## The design in one table

grok is **claude-shaped on resume** and **antigravity-shaped on MCP**, which is why it is a copy of neither.

| axis | claude | codex | antigravity | **grok** |
| --- | --- | --- | --- | --- |
| session id | we mint (`--session-id`) | it mints, we watch | it mints, we watch | **we mint** |
| resume | `--resume <id>` | `resume <id>` | `--conversation <id>` | **`--resume <id>`** |
| GUI MCP | `--mcp-config` per spawn | `-c mcp_servers.…` per spawn | `.agents/mcp_config.json` | **`.grok/config.toml`** |
| full GUI MCP in the workspace | yes | yes | no | **no** |

Because we mint the id, none of the codex/antigravity apparatus is needed — no watcher, no `claimed…` set, no `remember…` map, no `agentResumeId`. And because grok takes no MCP flag, it stays out of `FULL_GUI_MCP_AGENTS` for exactly the reason agy is: there is no per-spawn URL to hand it. The launcher form therefore offers grok the four per-group toggles rather than "All of them, automatically" — pinned by a spec, since promising tools it cannot deliver is what #1423 was.

**The two id flags are never sent together.** With `--resume`, a `--session-id` is only legal alongside `--fork-session`, where it *names the fork* — so sending both would branch the conversation instead of resuming it, leaving the user's history behind under the old id.

## Why grok's MCP registration is driven through its own CLI

`.grok/config.toml` is TOML and it is the user's file. Unlike agy's JSON it cannot be round-tripped without a TOML library, and adding one to append two tables is a large dependency for a small write. So `grok mcp add -s project` / `grok mcp remove -s project` does the writing, only for the ids we wrote, and a directory already in the desired state spawns no subprocess at all (the diff is a cheap text scan). It is the slower of the two designs and the only one that cannot corrupt a file someone hand-wrote.

## Verified against a real grok 0.2.118, not only in tests

- A cell runs `grok --session-id <minted uuid> --permission-mode auto`; TUI up, status line `Grok 4.5 (high) · auto`.
- grok persisted the conversation at `~/.grok/sessions/<url-encoded cwd>/<the id we minted>/`, so the id we chose **is** grok's id. The resume probe finds it, rejects the same id under another cwd, and rejects an id grok never minted.
- The sync registers the bridge, removes an entry when its group is switched off, removes all of them when none is on, and leaves a user's own `playwright` server untouched throughout.

**The cwd encoding was measured, not assumed** — `encodeURIComponent` exactly, checked against a directory containing a space, a hyphen and non-ASCII. Guessing it fails silently as "resume declined", which is the worst failure this integration has.

## The bug the live check caught, which every green signal had missed

`grok mcp add` ends in `<command> -- <args…>`, and everything after the `--` belongs to the server being registered. The first implementation appended `-s project` at the **end**, so the flag went to the *bridge*, grok never saw it, and the add silently fell back to **user scope** — writing four entries into the developer's own `~/.grok/config.toml`, where every directory on the machine would have picked them up. Exit code 0, project file untouched, typecheck and 8337 tests green.

Only running it against a real directory and reading the file back found it. The leaked entries were removed; the argv is now built by `grokMcpAddArgs` / `grokMcpRemoveArgs` with a spec asserting the scope flag precedes the `--`.

## Incidental cleanups, each because a third copy was the alternative

- `server/agents/gui-mcp-bridge.ts` — the stdio bridge command and the set of server ids that are ours to rewrite, now shared with the antigravity path.
- `server/agents/git-exclude.ts` — same, for the `.git/info/exclude` rule.
- `BadgedAgent` (`Exclude<TerminalAgent, "claude">`) replaces two hand-written `"codex" | "antigravity"` unions.
- `AgentMark.vue` gains an explicit codex branch: its `v-else` **was** codex's mark, so a newly added agent silently wore someone else's logo.

## Out of scope, deliberately

Rate limits (grok exposes neither a statusline nor a rollout), draft injection (needs a measured TUI marker), activity tracking, `customAgents` (`CUSTOM_AGENT_KINDS` stays `["claude"]`), and **launcher chips** — a chip whose command is `grok` still runs verbatim and records `agent: "shell"`. `/api/grok/sessions` and the sidebar restore path are the remaining piece of PR#3.

`yarn format` / `lint` / `typecheck` / `test` green — 8340 passing.

work in mulmoterminal


snakajima:
Good catch — this is real, and fixed in 5b69e6c2.

Confirmed the mechanism rather than taking it on trust: `ptyWouldReattach` is `persistent && tmuxAvailable() && tmuxHasSession(id)`, and after a server restart `ptys` is empty while the tmux session survives, so `spawnGrokPty` really is reached for a reattach. The grok in that pane read `.grok/config.toml` once at its own start and nothing re-reads it, so the rewrite cannot affect the process being reattached — it can only speak for the *other* grok sessions in the directory, which is exactly what that file's header says this path must never do.

The `[]` path you flagged is the sharp end: `ws-routes` resolves the groups with `.catch(() => [])`, so a transient failure reading Claude Code's config arrives here as "no groups" and the unconditional sync would deregister the servers every live grok in that directory is using.

Now gated on the same `ptyWouldReattach(sessionId, true)` the codex spawn already consults. The new spec (`test/server/session/spawn-grok-mcp-sync.spec.ts`) was run against the *unfixed* source first — it fails there and passes with the guard, so it is not a test that would have agreed either way.

One thing worth recording for whoever picks it up: **`spawn-antigravity.ts` has the same shape and therefore the same hole** — `syncAntigravityMcpConfig` is called unconditionally, and a reattach can rewrite `.agents/mcp_config.json` for sessions it does not speak for. That is pre-existing and I have deliberately left it alone rather than change agy's behaviour under a Grok heading; it deserves its own PR and its own review.

work in mulmoterminal

PR #1455, opened 2026-08-05
fix: let the new-terminal form use the cell's width

isamu:
## Summary

Every row of `CellLaunchForm` was capped at `max-w-[360px]`, so a wide cell rendered a narrow
centred column with the rest of the width unused. The chips are what actually suffers: **25 chips
wrapped into 20 rows inside 360px while the cell was 1535px wide.**

The cap is removed in the three components that shared it — `CellLaunchForm.vue` (6 rows + the chip
row), `LaunchChipList.vue`, `ModelPicker.vue`. Rows now span the cell and take their margin from the
form's own `p-4`.

One exception: the **agent picker** keeps its content width (`max-w-full` rather than `w-full`). It
is a segmented control, so stretching it would only widen the pill's background — the cap there
existed to decide *where it wraps*, which `max-w-full` still does in a narrow cell.

Measured against a running server at a 1600px viewport: chip row **360px → 1503px**, the same 25
chips fall into **4 rows instead of 20**.

## Items to Confirm / Review

Two controls now read oddly at extreme width, and I'd rather you judge them than pick for you:

- **The MCP toggle rows** are `justify-between`, so at 1535px the checkbox sits ~1500px from the
  label it belongs to. Functional, but a long way for the eye to travel.
- **The model `<select>`** now spans the full width, which is unusual for a dropdown.

If those bother you, the alternative is: chips and chip lists go edge-to-edge as they do here, and
the *input* controls (directory, model, MCP, worktree, resume) take a generous cap instead — much
wider than 360 but not full-bleed. Say the word and I'll push it.

Also worth noting: **no test pinned the 360px cap**, so nothing failed when it changed. The
verification below is a live measurement, not a spec.

## User Prompt

> 新規ターミナルの画面、横幅制限しているのでこんな感じ。paddingくらいだけとるようにして画面幅をもっと使いたい

(with a screenshot of the chips wrapping into a narrow centred column)

## Verification

- `yarn format` / `yarn lint` (0 errors) / `yarn typecheck` / `yarn test` — 8479 passed. One failure
  in `test/server/backends/calendarPush.spec.ts` ("socket hang up"), which **passes on its own** and
  is a server-route spec untouched by this diff.
- Live check against `yarn dev`, on the **Vite** port rather than the Express one — the Express port
  serves a stale `dist/` and showed the old `max-w-[360px]` in the DOM after the source had already
  changed, which is exactly how a UI fix gets called broken.

work in mulmoterminal4

## Summary by CodeRabbit

* **UI Improvements**
  * Launch form controls now use the full available width for a more flexible layout.
  * Chip lists and model selection no longer have a narrow fixed-width limit.
  * Agent selection preserves content-sized controls while wrapping as needed.
  * Worktree, resume, MCP, and working-directory sections now expand to fit available space.

* **Tests**
  * Added coverage to verify responsive, full-width launch form behavior.

### DIFF
--- a/src/components/CellLaunchForm.vue
+++ b/src/components/CellLaunchForm.vue
@@ -441,7 +441,7 @@ async function removeWorktree(w: Worktree): Promise<void> {
     >
       <span class="material-symbols-outlined" aria-hidden="true">close</span>
     </button>
-    <div v-if="chips.length" class="flex max-w-[360px] flex-wrap justify-center gap-1.5">
+    <div v-if="chips.length" class="flex w-full flex-wrap justify-center gap-1.5">
       <span
         v-for="p in chips"
         :key="p.label + p.path"
@@ -514,7 +514,7 @@ async function removeWorktree(w: Worktree): Promise<void> {
          wrapped row is centred too rather than hanging off the left. -->
     <div
       data-testid="agent-picker"
-      class="inline-flex max-w-[360px] flex-wrap justify-center gap-0.5 rounded-[7px] border border-border bg-deep p-0.5"
+      class="inline-flex max-w-full flex-wrap justify-center gap-0.5 rounded-[7px] border border-border bg-deep p-0.5"
       role="radiogroup"
       aria-label="Agent picker — what this terminal runs"
     >
@@ -533,7 +533,7 @@ async function removeWorktree(w: Worktree): Promise<void> {
         {{ option.label }}
       </button>
     </div>
-    <label class="flex w-full max-w-[360px] flex-col items-center gap-1.5">
+    <label class="flex w-full flex-col items-center gap-1.5">
       <span class="font-sans text-[11px] uppercase tracking-[0.05em] text-dim">Working directory</span>
       <span class="flex w-full items-stretch gap-1.5">
         <input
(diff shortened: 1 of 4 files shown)


isamu:
### Review triage — iteration 2

**Codex: LGTM** (both iterations). **CodeRabbit: 2 nitpicks**, both on the spec I added in iteration 1.

| finding | verdict |
|---|---|
| Spec asserts before the fetches land, so `cell-worktrees` / `cell-resume` are absent | **fixed** — real hole |
| Assert the model `<select>`'s width, not just the wrapper | **kept as-is, test renamed** |

**Fixed.** CodeRabbit was right and it mattered: the guard's second test walks the form's direct
children and claims to cover "every row", but `cell-worktrees` and `cell-resume` render only once
`/api/worktrees` and `/api/sessions` resolve — and those are **two of the rows this PR widened**. The
spec was checking a DOM that did not contain them. Now it awaits the fetches, the fixtures produce a
worktree and a session, and it asserts both rows are *present* before the loop — so a fixture that
stops producing them fails here rather than quietly shrinking what gets walked.

**Kept.** The `<select>` takes its width from `SELECT_CONTROL` (`src/components/selectClasses.ts`),
which every select in the app shares. Pinning it from a launch-form test would tie this file to a
decision made for the whole app and would fail for reasons that have nothing to do with the launch
form. The wrapper — the row the form stacks — is what this PR changed, and a pixel cap anywhere in
the picker's markup, `<select>` included, is already caught by the html-wide scan in the first test.
The test is renamed to "spans the width in the model picker's **row**" so it no longer claims more
than it checks, which was CodeRabbit's own suggested alternative.

**Found in my own review, not flagged by either bot** (iteration 1):

- Nothing pinned the 360px cap, so nothing failed when it was removed and nothing would fail if it
  came back — hence the guard, written against *any* pixel cap rather than the one number. Verified
  it actually bites by re-adding the cap to the chip row and to `cell-resume`; both go red.
- The agent picker's comment said it was "centred like everything else in this column", which this
  diff inverts — it is now the only content-sized row. It also still said "four options" from before
  Grok was added.

PR #1461, opened 2026-08-05
fix: never offer a provider with no models in the launch picker (#1432)

isamu:
## Summary

A custom provider appeared in the launch form's MODEL dropdown and could not be selected by a
click or by the arrow keys (#1432). It was not an option at all: a provider whose `models` is
empty renders as `<optgroup label="DeepSeek"></optgroup>` — a **group header with nothing under
it**, which a browser draws as a shaded row that a click and the arrow keys both skip.

The picker now offers only providers that are reachable **and** have a model to pick, so an empty
group cannot be rendered; the ones it leaves out are explained instead of vanishing.

## Items to Confirm / Review

1. **The help link changes wording** — it now reads **"Needs attention"** whenever a configured
   backend is not offered (previously "How this works" / "Use another model…" only). A provider
   with a missing key used to say "Use another model…"; it now says "Needs attention".
2. **`sanitizeProviders` gained a `console.warn`** naming the model ids the schema dropped. It is
   said **once per process** because this config is re-read on paths that run per session spawn
   (`getAppendSystemPrompt`), which would otherwise repeat the line at every launch. That makes
   the module stateful across calls — deliberate, but worth a look.
3. **Settings now shows "not in the picker"** (in the error colour) for a provider that is
   reachable with zero models. `ready` itself is unchanged and still means "a session CAN run on
   it" — a directory that pins a model of its own launches fine, which is why this is a separate
   state rather than "not ready".
4. **Skill / docs wording reversed for non-OpenRouter ids.** `mulmoterminal-model` said *"Do not
   write a `models` array … registering the provider is enough"*, which is true **only** for the
   id `openrouter`. Under any other id it produced exactly the broken entry in this issue.
5. **Not changed, noticed while here:** `useLaunchOptions`'s wire guard drops a **whole provider**
   when one model row fails `isModelPreset`, while the config schema drops **per model**. Not
   reachable today (the server builds those rows itself), so it is left alone — but the two layers
   disagree on the same principle.
6. **Not reproduced in a real Chrome** — this checkout has no Playwright browser installed. What
   was verified is that the empty `<optgroup>` is no longer in the DOM (see below); the painting of
   an empty group as an unreachable row is the reporter's own observation.

## User Prompt

> 調査、修正、テスト強化できる？？
> https://github.com/receptron/mulmoterminal/issues/1432

## Root cause

`ModelPicker.vue` rendered one `<optgroup>` per **ready** provider and one `<option>` per model.
Ready says only that this server can reach the backend — it says nothing about there being
anything to pick. Two silent paths lead to an empty list, and #1432 hit at least one of them:

1. **The built-in presets are OpenRouter's alone.** Every entry in `common/modelPresets.ts` carries
   `provider: "openrouter"`, and `presetsForProvider` matches on that id — so a backend registered
   as `deepseek`, `moonshot` or a company gateway starts with **no models**, and must list its own.
2. **A malformed `models` entry is dropped without a word.** `providerSchema.models` filters
   anything that is not a usable model-id string and `.catch([])`s a non-array, so
   `models: [{"id": "deepseek-chat"}]` becomes `[]` — the user reads a config file that lists
   models while the UI has none.

A session cannot start on such a provider anyway (`resolveProvider` refuses a named provider
without a model), so it is not a choice — it is a setup problem, and it now reads as one.

## What changed

| File | Change |
|---|---|
| `src/components/launchOffer.ts` (new) | One rule: `isOfferable = ready && models.length > 0`, plus `notOfferedReason()` — the sentence to show about a provider the picker leaves out |
| `src/components/ModelPicker.vue` | Offers only offerable providers (an empty group can no longer be rendered); shows the select only when something is offerable; the help link says "Needs attention" when a backend is not offered |
| `src/components/ModelSetupHelp.vue` | "Needs attention" now covers the no-models case; step 1 says presets come with the id `openrouter` and no other, and what a model id may contain |
| `src/components/settings/ModelsSection.vue` | A reachable provider with no models reads "not in the picker" rather than "ready" |
| `server/config/app-config.ts` | Warns, once per process, naming every model id it dropped (and a `models` that is not an array) |
| `server/skills/mulmoterminal-model/SKILL.md` | `models` is REQUIRED under any id other than `openrouter`; the id shape rule; two new troubleshooting rows |
| `docs/guide/{en,ja}/providers.md` | Same correction, plus troubleshooting rows for "the backend is missing from the MODEL list" |

## Verification

Against a real server, with a scratch `HOME` holding the reported config (`DEEPSEEK_API_KEY` set,
`models` written as objects — one of the two silent paths):

```
[providers] 'deepseek': dropped 1 unusable model id(s) — {"id":"deepseek-chat"}. A model id is letters, digits and . _ : / - ~.

GET /api/launch-options
{"providers":[{"id":"deepseek","label":"DeepSeek","ready":true,"tokenEnv":"DEEPSEEK_API_KEY","models":[]}],"anyReady":true}
```

That is the reporter's exact server state — `ready`, and nothing to pick. With the same config
corrected to `"models": ["deepseek-chat", "deepseek-reasoner"]`, the route answers two unmeasured
models and the log is silent, so the fix does not over-filter.

The DOM before the fix, captured from the mounted component:

```html
<option value="">This directory's default</option>
<optgroup label="DeepSeek"></optgroup>   <!-- the row that could not be clicked -->
```

The four new picker tests were run against the **unfixed** component and all four failed, then
pass with it.

## Tests

- `test/src/components/launchOffer.spec.ts` (new) — the rule, and both sentences
- `ModelPicker.spec.ts` — **no `<optgroup>` is ever empty**, written against the shape rather than
  one payload; a models-less provider is not offered; the select hides when it is the only one;
  the help link and the help text
- `launch-options.spec.ts` — a provider whose id is not `openrouter` gets no presets (the rule that
  made this reachable); it offers exactly what it lists; and four cases pinning what the config
  loader says when it drops a model id
- `globalSettingSections.spec.ts` — the Settings row says "not in the picker", not "ready"

`yarn format` / `lint` (0 errors) / `typecheck` / `test` (8503 passed) / `build` all green.

## The other two console errors in the report

- `GET /api/dir-sound?… → 404` is **by design**: the route 404s when the directory has no custom
  sound, and the client falls back to the global sound, then the built-in chime.
- `426 Upgrade Required` is emitted **nowhere in this repo** — every `WebSocketServer` runs in
  `noServer` mode, and `ws` only sends 426 from its own standalone `port` server. It also cannot
  disable a `<select>` option. Left open, with a request for the failing URL, in the issue.

refs #1432

work in mulmoterminal3

### DIFF
--- a/docs/guide/en/providers.md
+++ b/docs/guide/en/providers.md
@@ -129,9 +129,15 @@ In any directory, tell a Claude session `add OpenRouter to my mulmoterminal conf
 `mulmoterminal-config` skill keeps your existing settings and writes a valid entry — **without your
 key in it**.
 
-**Do not list models here.** Registering a provider with this `id` is enough — the
-[27 measured models](#verified) appear in the picker on their own. You only list models to add ones
-that are **not** in that list (→ [Adding models](#add-models)).
+**Under the id `openrouter`, do not list models here.** Registering a provider with this `id` is
+enough — the [27 measured models](#verified) appear in the picker on their own. You only list models
+to add ones that are **not** in that list (→ [Adding models](#add-models)).
+
+**Under any other `id`, `models` is required.** The measured list is OpenRouter's alone, matched by
+that exact id, so a backend registered as `deepseek`, `moonshot` or an in-house gateway starts with
+none — and **a backend with no models is not offered in the picker**, because a session named on a
+provider without a model refuses to start. List the ids you want to run
+(→ [Adding models](#add-models)).
 
 | Key | Meaning |
 |---|---|
@@ -140,7 +146,7 @@ that are **not** in that list (→ [Adding models](#add-models)).
 | `baseUrl` | **no trailing `/v1`** — see below |
 | `tokenEnv` | the **name** of the environment variable holding the key, never the key |
 | `maxOutputTokens` | optional; defaults to 16000 |
-| `models` | optional; only to add models **not** in the built-in list (→ [Adding models](#add-models)) |
+| `models` | **required unless `id` is `openrouter`** — that is the only id with presets. Under `openrouter`, only to add models **not** in the built-in list (→ [Adding models](#add-models)) |
 
 ### Never end `baseUrl` in `/v1`
 {: .no_toc }
@@ -330,7 +336,9 @@ If you are unsure where to start, try **Kimi K2.7 Code** (fast, coding-oriented)
 | `404 No endpoints available …` | that model is excluded by your [privacy settings](https://openrouter.ai/settings/privacy) |
 | Empty replies, looks hung | `maxOutputTokens` too low — keep it at 16000+ |
 | Talks but never uses tools | the model's own limitation — check the table or `model-trials.ts` |
-| No MODEL select in the launch form | no provider is registered (step 2) or its key is missing; "Use another model…" in the form names what's missing |
+| No MODEL select in the launch form | no provider is registered (step 2), its key is missing, or it has no models to pick; the link beside the MODEL label — "Needs attention" / "Use another model…" — names what's missing |
+| A registered backend is missing from the MODEL list | it has no models: presets come with the id `openrouter` and no other, so [list its `models`](#add-models) |
+| `models` is in the config and the picker still offers none | every id in it was refused — a model id is letters, digits and `. _ : / - ~`, so an object like `{"id": "…"}` or a value with a space is dropped. The server's log names what it dropped |
 | No model in the header | `ctx` is not in your `chips` |
 
 ---
--- a/docs/guide/ja/providers.md
+++ b/docs/guide/ja/providers.md
@@ -127,8 +127,15 @@ console.log("added openrouter to " + file);
 どのディレクトリでもいいので、Claude のセッションに `mulmoterminal の設定に OpenRouter を追加して` と
 頼めば、同梱の `mulmoterminal-config` スキルが既存の設定を保ったまま書いてくれます（**鍵は書きません**）。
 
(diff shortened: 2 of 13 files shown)


isamu:
## Review loop — iteration 1

**Bot signals**

| Reviewer | This iteration |
|---|---|
| Codex (`CODEX VERDICT`) | **LGTM** on both `4076ace5` and `27a639a4` — no findings |
| CodeRabbit | **rate limited**, posted no review. Not waited on (project policy: never block a merge on CodeRabbit) |
| knip / jscpd / Socket / lint-and-build (mac+ubuntu) / package-smoke | all green |

**Applied this iteration** (found in my own read of the diff, not flagged by any bot):

- `server/config/app-config.ts` — the dropped-model warning echoed the rejected value **whole**.
  A rejected value is whatever the user's JSON held, and "longer than `MODEL_ID_MAX_LENGTH`" is
  itself one of the rejection reasons, so a 50 KB string became a 50 KB log line. It is now quoted
  to 80 characters, with a test asserting the line stays bounded.
- Merged `origin/main` (#1460 moved the picker's row width into `LAUNCH_ROW`). The two changes
  touch different lines of `ModelPicker.vue` and merged clean; the full suite was re-run after the
  merge.

**Deliberately not changed** — noted in the PR description as well:

- `useLaunchOptions`'s wire guard drops a **whole provider** when one model row fails
  `isModelPreset`, while the config schema drops **per model**. The two layers disagree on the same
  principle, but it is not reachable today (the server builds those rows itself), so widening this
  PR to chase it would be speculative.

Local: `yarn format` / `lint` (0 errors) / `typecheck` / `test` (8505 passed) / `build` all green.

PR #1474, opened 2026-08-05
feat: 起動画面の history で「まだ動いている」を出し、そこで止められるようにする (#1467)

isamu:
## Summary

The launcher's `or resume here` rows now say when a conversation's session is **still running with
nobody attached** — what a server restart leaves behind — and offer a **stop** button for exactly
those rows. This is the manual half of #1467: it puts the state where the user already looks, and
lets them end it without `tmux kill-session` by hand.

The row already answered "is anyone holding this" (`attached`). That answer comes from
`tmux list-clients`, which reports **only sessions that have a client** — so the sessions that
accumulate, running with nobody attached, were structurally invisible to it. One
`tmux list-sessions` per list is what was missing.

## Items to Confirm / Review

1. **`runningKey` is a KEY, not the row's id.** A codex/agy conversation started from a grid cell
   runs under a key MulmoTerminal minted, and only that agent's conversation log connects the two —
   the same reason `attached` consults it. Stopping the row's id there would kill nothing and report
   success. Pinned on both sides (`runningKeyOf` server-side, and a UI test that fails if the button
   posts the conversation id).
2. **Only unattached rows get the button** (agreed with @isamu). A row marked `● open` belongs to
   the terminal that has it; ending it from another cell's launch form would pull a session out from
   under a tab the user cannot see from here.
3. **It asks first.** Stopping kills a running claude, so the in-flight turn is lost. The
   conversation is not — the transcript stays and the row can be resumed afterwards, which is what
   the confirmation says.
4. **New wire type rather than a field on `SessionOccupancy`.** That type is documented as the one
   fact a second terminal needs before offering to open a session, and the worktree row depends on
   it; "what is running, and under which key" is the launcher list's own question, so it went into
   `common/sessionRunning.ts` beside it.
5. **Cost**: one extra `tmux list-sessions` per conversation-list request, next to the
   `list-clients` the route already ran.

## What this does NOT do

**#1467 stays open.** The report asks for automatic cleanup, and this is not that:

- The list is **per directory and per agent**, so it does not reach a project you no longer open —
  which is where the reporter's sessions are.
- Wiring `POST /api/tmux/cleanup-orphans` at boot (the issue's own proposal) would not fix it
  either. It spares everything **resumable** — live pty, persisted grid session, or a transcript on
  disk — and we launch claude with `--session-id <session id>`, so the transcript's filename *is*
  the tmux session name: every conversation ever prompted reads as resumable, forever.

Measured while investigating, on one machine with 21 surviving sessions:

| | count | pane |
|---|---|---|
| attached (held by a terminal) | 12 | claude |
| **unattached, transcript on disk** | **5** | **claude still running**, idle 10–11 h |
| unattached, no transcript | 4 | `zsh` / `node` (launcher / run cells) |
| claude exited, session left behind | 0 | — |

So what accumulates is **live claude processes nobody is attached to**, and `cleanup-orphans` would
reap 4 of 21 — none of them those. (Nothing sets `remain-on-exit`, so a session whose command exits
ends with it; the issue's "the claude process has already exited" is not what survives.) The root
fix is making `isResumableTmuxSession` less conservative — a transcript on disk means the
conversation can be resumed **without** the tmux session, which is an argument for reaping it, not
against. That is a separate decision and a separate PR.

## User Prompt

> 2 [＝ resumable 判定が過剰、が本丸] だけと、ターミナル起動画面の history のところで、tmux が
> 生きているかどうか判定して表示できないかな？そこで stop できれば、ひとまず自分でコントロール
> できないかな。

Two decisions taken with the user before implementing: the stop button appears only on unattached
rows, and it asks for confirmation first.

## Changes

| file | |
|---|---|
| `common/sessionRunning.ts` (new) | the `runningKey` wire field, and why it is not part of `SessionOccupancy` |
| `server/session/dir-session.ts` | `runningSessionKeys()` (surviving tmux ∪ live ptys) and `runningKeyOf()` |
| `server/routes/session-routes.ts` | fill it in the claude route and in `withAttached` (all four agents) |
| `src/composables/useSessionStop.ts` (new) | confirm → terminate → reload, with an in-flight guard |
| `src/components/CellLaunchForm.vue` | the `● running` marker and the stop button |
| `docs/guide/{en,ja}/basics.md` | what `● running` means and what stop does to the conversation |

## Verification

Local: `yarn format` / `lint` (0 errors) / `typecheck` / `test` (8669 passed) / `build`.

End-to-end against a **real server and a real browser**, with a genuinely surviving session (a
transcript on disk plus a live `mt-<id>` tmux session, under a scratch `HOME` so no real config was
touched):

```
GET /api/sessions?cwd=…  →  "attached": false, "runningKey": "aaaaaaaa-1467-…"

BEFORE: { rows: 1, running: 1, stop: 1,
          runningTitle: "Still running with nobody attached — resume it here, or stop it" }
CONFIRM ASKED: Stop the session running "tidy the parser"? Its conversation is kept — you can
               resume it later — but anything it is doing right now is lost.
AFTER:  { rows: 1, running: 0, stop: 0 }

tmux session is gone — the stop button really killed it
transcript kept — the row can still be resumed
```

refs #1467

work in mulmoterminal3

## Summary by CodeRabbit

* **New Features**
  * Running sessions now remain available after a server restart, even without an attached terminal.
  * Resume listings show which sessions are still running.
  * Added a confirmation-based stop option for eligible running sessions while preserving conversation history.
  * Session lists refresh automatically after stopping or failed stop attempts.

* **Documentation**
  * Updated English and Japanese guides with running-session behavior, resume options, and stop restrictions.


isamu:
Thanks — two of the three are in as of 6bea808a, and I am declining the third with a reason.

**Applied**

- **Disable resume while its stop is in flight** (Major). Real, and the failure is nasty in a quiet
  way: if the resume wins the race, the cell attaches to a session the terminate then kills — a
  terminal that dies on arrival with nothing on screen saying why. Guarded on the button *and* in
  `resume()`, since the row is reachable by keyboard too, with a test that fails without the guard.
- **The plan file's implementation map** named `common/sessionOccupancy.ts` (Minor). That was the
  first design; it moved to its own `common/sessionRunning.ts` during implementation, because
  `SessionOccupancy` is what the worktree row also depends on. The map now says so.

**Not applied — the `●` marker**

`●` (U+25CF BLACK CIRCLE) is a geometric shape, not an emoji, and this repository's rule names it
explicitly as the sanctioned form for exactly this:

> Compact status **notation** stays text, not icons: `⎇ main ●3 ↑2`, `●` unsaved dots, `−12` diff
> counts. Icons there are bigger and slower to scan.
> — [CLAUDE.md](https://github.com/receptron/mulmoterminal/blob/main/CLAUDE.md), the same "No emojis" section the finding cites

It is also the established notation in the two rows immediately above the new one, in this same
list: `● open` (line 801) and `● failed` (line 791). Replacing the marker on one of the three
statuses is what would make the list inconsistent, and the guides describe all three the same way.

Happy to revisit if the maintainers want the whole notation changed — that would be one PR across
every status marker, not this one.

Issue #1456, opened 2026-08-05
feat: automate conversations between cells — have them join a chat "room"


isamu:
## What I want

**I want conversations between cells to be automated.** Two participants, and ideally three or more.

Cross-terminal talk in `#550` stops at "a human presses a button, one round trip". `#1454` proposes "an MCP tool that reads another session's last turn". Beyond that, I want to get to **agents autonomously continuing a conversation**.

## Three things that block the naive MCP approach

1. **Being able to start a conversation unbidden is a problem** — it means handing an agent the power to wake the other side's turn
2. **How do you find the other party** — enumerating all sessions exposes unrelated work on the machine
3. **How do you switch turns** — an agent can only act inside its own turn

### A room makes 1 and 2 disappear by design

**1 -> a post wakes nobody.** Writing to the room only appends to a log; the other side's turn does not run. It is read only when the other side comes to read it during its own turn.

> Make it a bulletin board, not a telephone.

Handing an agent something like `insertText` (injecting into the other side) makes it **a telephone**, and "starts talking on its own" becomes possible.

**Principle: participants may post. Only the host (MulmoTerminal) may start a turn.**

**2 -> membership becomes the ACL.** Not "every session on the machine" but "whoever is in this room". And **a human does the joining**. If agents are not given the right to join, they can neither see outside the room nor add members.

### 3 does not disappear — this is the heart of it

Something has to "wake the next participant's turn". **MCP cannot solve this** (trying to solve it takes you straight back to 1). A **runner on the host side** drives it.

## Decisions made

- **Participants are agents + humans + shells.** Make the room an **HTTP API** with a CLI equivalent to `mt room post`, and make the MCP tool a thin wrapper over it. Shell cells, CI and humans can all take part, and it can be tested with curl. Making it MCP-only pins participation to agents alone
- **Turns are driven round-robin by the host's runner.** It injects "it's your turn, read the room and respond" into the current speaker's cell, waits for completion, then moves on
- **Joining is done by a human** (agents are not given the right to join)

## Parts that already exist (80% of it)

| Part | Location | What it solves |
|---|---|---|
| Getting another cell's utterance | `sessionLastTurn` in `server/session/session-reads.ts` | Reading |
| The "this is a record, not an instruction" frame | `formatHandoff` in `server/session/handoff-text.ts` | Minimal defence against injection |
[...]

PR #1483, opened 2026-08-05
feat: 部屋を見せる、部屋に入る、部屋を選ぶ (#1456 v3)

isamu:
## Summary

v2 (#1476) gave the conversation rooms a file, an API and a CLI, and left the app unable to show
them. A round table finished with `agreed · 4 turns` and the conversation itself — the only thing
the feature produces — was reachable only by knowing the room id and running the CLI. This is the
half that makes a room worth having: **see it, speak in it, and choose which one a table uses.**

- **`RoomsOverlay`** — rooms on the left, the conversation on the right, a post box below. Same
  shape as the accounting / PRs / files overlays; the open room is the URL (`/rooms/:room`), so a
  conversation can be linked to and reloaded.
- **The post box is not a convenience.** Agents are typed into by the runner and can call nothing,
  so a person joins a conversation *from outside*, exactly as the CLI and a CI job do. This is that
  door for the person already looking at the screen.
- **A room box in the round-table picker** — empty mints a new room (what every table did before),
  a name uses that room, and an existing name **continues that conversation**: the members read it
  back before they speak. Reuse and naming are one control because the answer to "which room" is a
  name, and whether it exists yet is not the user's problem.
- **`DELETE /api/rooms/:room`** — every table mints a room, so a listing nobody can tidy is what
  makes people stop opening it.
- **`listRooms` now matches its own doc comment** — it said "newest activity first" and sorted
  alphabetically, and a table's id starts with its creation time, so the *oldest* conversation sat
  at the top of the list.

## Items to Confirm / Review

1. **Where the "could not read" line is drawn.** #1476 taught the server to answer **500** rather
   than an empty conversation. This PR extends that decision rather than restating it: the runner
   still degrades to the previous turn (a dropped request must not kill a live table), and the
   **overlay shows an error** (a person cannot tell an unreadable room from an empty one). That is
   why `useRooms` now has two forms of each call — `loadRoom`/`sendRoomMessage` say whether it
   worked, `fetchRoom`/`postRoomMessage` throw that answer away for the runner. One HTTP call site,
   two policies, neither hidden inside the other. **Please check that split is where you want it.**
2. **The picker refuses an unusable room name instead of falling back to a new room.** Deliberate:
   a table that quietly ran somewhere the user never named is a conversation they will look for
   under the name they typed. The alternative (mint one silently) is friendlier and, I think, worse.
3. **The ask menu now stays OPEN when a table starts** (it used to close). Both things you want next
   live in it — stop, and read the conversation. It still closes on the next click outside.
4. **The overlay polls the whole room every 2s rather than using `?since=`.** Two posts can share a
   millisecond, so a tail keyed on the newest timestamp can step over one; the room is small and on
   loopback. `since` stays for followers that cannot afford the whole room. If a room is ever
   expected to get long, this is the line to revisit.
5. **`from` is a display name and nothing authenticates it** — unchanged from v2, restated because
   the UI now lets a person type one.

## User Prompt

> 続けよう

Continuing #1456 after v1 (#1458, the runner) and v2 (#1476, the server-side room). The remaining
items were listed on the issue when v2 merged; this PR takes **room-viewing UI / posting from the
UI** and **reusing and naming rooms**.

## Implementation

Plan: `plans/feat-1456-rooms-ui.md`.

**Server**

- `deleteRoom()` + `DELETE /api/rooms/:room`, same-origin guarded like the post — this one destroys
  a record rather than adding to one.
- `listRooms()` orders by mtime, newest first, tie-broken by name. Return shape unchanged
  (`rooms: string[]`) because the CLI reads it.
- `requestedRoom(req, res)` — the room-id check was written out at each route and a third one was
  about to be added. The id becomes a filename, so this is the check no route may skip.

**UI**

- `useRoomsView` (route seam), routes `/rooms` + `/rooms/:room`, both added to `OVERLAY_ROUTES` but
  **not** `CONTENT_ROUTES` — a room is the record of what the terminals said to each other, not
  workspace content, so it sits with PRs.
- `newRoomId` moved from `useRoundTable.ts` to `roundTableRules.ts`: it is a pure decision, and the
  picker needs it without pulling in the runner's sockets.
- `roomForTable(typed, mintNew)` — the whole room-choice rule, pure and tested.

## Verified against a running app, not just the build

The last two features each had a bug that only a live run found, so this one was driven end to end
with a real server (`MULMOTERMINAL_HOME` and `HOME` pointed at a scratch dir, so the live config was
never touched) and a real browser:

- listing order is by activity — checked with two rooms whose alphabetical order is the reverse
- `DELETE`: 403 cross-origin, 400 on a bad id, and the file actually gone
- a CLI post appears in the overlay; a UI post reaches the room; an outside post appears **without a
  reload** (the poll)
- **the log follows new messages only when you are already at the end** — this is the one the live
  run found. jsdom lays nothing out, so it could not have come from a unit test; the two tests added
  for it each fail when their own half of the rule is broken (checked both directions).
- no console errors on `/rooms`, `/rooms/:room` or `/terminals`

`yarn format` / `lint` / `typecheck` / `build` / `test` (8860 passing) all green.

## Still open on #1456

Moving the runner to the server (closing the tab still stops a table), roles / @mention /
asynchronous rooms, searching or exporting a room.

refs #1456

work in mulmoterminal5


isamu:
## bot レビュー対応（iter-1 → 09ed3eb4）

### Codex

**CHANGES REQUESTED → 修正 → LGTM。**

`RoomsOverlay.vue` の watcher が poll を張る前に await するので、部屋の切り替えが速いと
古い run が中断されたまま残る。新しい run が interval を張った**後に**古い run が再開して
`pollTimer` を上書きし、新しい方のハンドルが誰からも届かなくなる — **interval が2本回り、
片方は永久に止められない**（オーバーレイを閉じても、unmount しても残る）。

世代トークンで、自分が最新の run のときだけ張るようにした。テストは、読み込みが in flight の
うちに部屋を変えて1ティック進め、`loadRoom` が1回だけ呼ばれること・unmount 後は0回になることを
見る。**ガードを外すと `expected 2 to be 1` で落ちる**ことを確認済み。

### CodeRabbit

このPRではレビューが走っていない（アカウントのレート制限）。プロジェクトの方針として
CodeRabbit は待たないので、後から指摘が来て**内容が正しければその時に直す**。

### CI

全 green（lint-and-build macOS/Ubuntu, package-smoke, knip, jscpd, Socket ×2, codex-review）。
ローカルも format / lint / typecheck / build / 8861 tests green。

PR #1479, opened 2026-08-05
feat: 再起動を生き延びたセッションを Settings に一覧し、そこで終了できるようにする (#1478)

isamu:
## Summary

Settings gains **Sessions that survived a restart** — every terminal still running from an earlier
server, **across all directories**, with a stop button. It is the only place in the app where a
session in a project you no longer open, or a plain shell, can be seen at all.

Closes #1478. It is the manual half of #1467, which stays open (see below).

| the row says | from |
|---|---|
| working directory | the live pty's own cwd, else the remembered one (#1021) |
| what it is | the live pty's agent, else claude (transcript) / codex (rollout) / **`shell or command`** |
| last active | tmux `#{session_activity}`, one call for the whole list |
| held by a terminal | the same `attached` fact the launcher rows use → `● open`, no stop |
| **not resumable** | nothing on disk can bring it back, so the session is the only copy |

Ordered **what can be cleared first, longest idle at the top**; held rows sink to the bottom.

## Items to Confirm / Review

1. **One row at a time, no "stop all".** A single click ending many live agents is automatic
   cleanup wearing a button, and that decision belongs to #1467.
2. **Stop is offered only for unattached rows**, the rule set in #1474 — a held row belongs to the
   terminal holding it. The confirmation and the terminate call are `useSessionStop` from that PR,
   reused unchanged.
3. **`GET /api/tmux/sessions` is exempt from the origin guard**, like `google.ts`'s `GET /status`
   (#1094): safe methods are not judged by origin because a cross-site `` and a legitimate
   local fetch both send no Origin header. What keeps the list private is that a cross-origin
   caller cannot READ the reply. The guard is still asked, exactly as its neighbours ask it, and a
   test pins both halves (the GET answers, the POST beside it 403s).
4. **The list is deliberately the opposite selection from the phone's** (`buildSessionList` keeps
   only resumable AND grid sessions) — that filter is why a shell left behind by a restart appears
   nowhere today.
5. **Cost**: two tmux calls per open of the Settings modal (`list-sessions` twice — names, and
   activity), plus the `list-clients` already used. Fetched on mount, not polled.

## User Prompt

Across several turns, after I broke #1467 down into four problems and the user picked the root
cause (#2) as the one to think about separately:

> 2 [＝ resumable 判定が過剰、が本丸] だけと、ターミナル起動画面の history のところで、tmux が
> 生きているかどうか判定して表示できないかな？そこで stop できれば、ひとまず自分でコントロール
> できないかな。 — shipped as #1474
[...]

PR #1486, opened 2026-08-06
fix: 使われていないセッションを起動時に終了する — 1 つの述語が 3 つの問いに答えていたのをやめる (#1467)

### LINKED ISSUE #1467: bug: cleanup-orphans が自動呼び出しされず、再起動後のゾンビ tmux セッションが蓄積する
## What happened

サーバー再起動のたびに、内部プロセス（claude）が既に終了した tmux セッションがゾンビとして残り続ける。`cleanup-orphans` エンドポイントは存在するが、起動時にも定期的にも呼び出されないため、孤立セッションは `tmux -L mulmoterminal kill-session` で手動削除するまで永続する。

## What I expected

サーバー起動後、再接続されなかった（または内部プロセスが終了済みの）tmux セッションが自動的にクリーンアップされること。

## Steps to reproduce

典型的なシナリオ: バージョンアップ時の停止→更新→起動

1. 複数プロジェクトでセッションを開く
2. 一部のセッションで Claude の作業が完了している（claude プロセスが終了した状態）
3. 新バージョンが来たのでサーバーを停止する（pm2 stop / Ctrl+C / kill など）
4. `pnpm install` 等でパッケージを更新する
5. サーバーを再起動する
6. `tmux -L mulmoterminal list-sessions` で確認 → 停止前に claude プロセスが終了していたセッションがゾンビとして残っている
7. UI にはそのセッションのクローズボタンが表示されない（別プロジェクトの cwd のため）
8. 手動で `tmux -L mulmoterminal kill-session -t mt-<id>` しない限り残り続ける

毎回再現する。バージョンアップに限らず、クラッシュからの自動復帰や手動再起動でも同じ。

## Analysis

- #366 で `POST /api/tmux/cleanup-orphans` が追加されたが、フロントエンドにもサーバーブートにも呼び出し元がない（`grep -r "cleanup.orphans" dist/` = 0件）
- `server/index.ts` のブート処理（lines 890–917）は surviving セッション数のログ出力と settings/drops ファイルのプルーニングのみ行い、tmux セッション自体のクリーンアップは行わない
- クローズボタン（WS `terminate` + `POST /api/session/:id/terminate`）はライブセッションに対しては正常に動作する（#366 で修正済み）
- 問題は「再起動後に in-memory ptys に入らなかったセッション」で、UI からも到達できない

## Environment

- mulmoterminal 4.5.1
- Node v24.18.1
- Linux x86_64 (7.1.5-1-cachyos)
- tmux 3.7b
- Claude Code 2.1.222
- 起動方法: pm2 経由 (ecosystem.config.cjs → bin/mulmoterminal.js → server/index.ts)
- pm2 固有の問題ではなく、どの再起動方法でも同じ（in-memory ptys がリセットされるため）

### ecosystem.config.cjs

```js
module.exports = {
  apps: [
    {
      name: "mulmoterminal",
      script: "./node_modules/mulmoterminal/bin/mulmoterminal.js",
      args: "--port 34567",
      interpreter: "node",
      env: {
        MULMOTERMINAL_HOST: "0.0.0.0",
        MULMOTERMINAL_ALLOWED_ORIGINS: "http://<IPアドレス>:34567",
      },
    },
  ],
};
```

isamu:
## Summary

`POST /api/tmux/cleanup-orphans` has had no caller since #367 — that is what #1467 reports. **Calling
it would not have fixed it.** The predicate it consulted is made of permanent records, so on the
machine this was measured on a sweep would have ended **2 of 22** sessions while 15 sat idle for
hours with nobody attached.

`isResumableTmuxSession` = `live ∨ grid log ∨ claude transcript ∨ codex rollout`, and three callers
were asking it three different questions. This splits it:

| question | rule | caller |
|---|---|---|
| can a user usefully **open** this? | unchanged | the phone's session picker |
| does ending it **lose the conversation**? | **on-disk evidence only** — `isRestorableSession` | the Settings list |
| may we **end** it unasked? | **nothing about the past** — `reapableTmuxSession` | the sweep |

Then the sweep runs **once at boot**, through that rule, and says what it did.

Closes #1467.

## Items to Confirm / Review

1. **What ending a session costs, stated plainly:** the process and its scrollback. The conversation
   is on disk and resumes without the tmux session — which is exactly why "it has a transcript" was
   the wrong reason to keep one alive. The Settings list marks a row `not resumable` when that is
   not true of it.
2. **Every unknown answers NO.** An attach count tmux would not give (#747's case) and an unknown
   age both spare the session. Reaping on a guess is worse than the pile-up.
3. **`sessionIdleReapDays`, default 7, `0` = off.** A corrupt value falls back to **7, not 0** — a
   sweep silently disabled by junk in the config looks exactly like the bug this fixes.
4. **#1479's `resumable` column was wrong and is corrected here** (it counted `live ∨ grid log`,
   so 20 of 22 rows claimed to be restorable when 15 were).
5. **The route lost its decision.** `cleanup-orphans` now just calls the same sweep, so the manual
   and automatic paths cannot drift. `orphanReapable` is gone; its tests moved to the sweep.
6. **Not done:** pruning `dev-terminal-sessions.json` (append-only, 439 entries here). Nothing
   decides a KILL from it any more — it is the chat sidebar's exclusion list, where permanence is
   what makes it correct.

## User Prompt

> 残りの課題は？ … 2

— where "2" is the second of the four problems #1467 was broken into: *resumable 判定が過剰*（本丸）.
Problems 3 and 4 shipped in #1474 and #1479; problem 1 (the route with no caller) is fixed here too,
since the sweep is that caller. Two decisions were taken with the user before implementing: the
threshold defaults to **7 days and is configurable**, and the sweep runs **once at server start**.

## Verification

Local: `yarn format` / `lint` (0 errors) / `typecheck` / `test` (**8847 passed**) / `build`.

**Against real tmux**, both directions — because this feature ends processes:

```
# 1. boot with the default threshold, against this machine's 26 live sessions
[tmux] persistence on — 26 session(s) survived; reattach on connect
[tmux] kept 26: 12 in use, 14 active within 7 day(s)
sessions before/after: 26 / 26          ← nothing was touched

# 2. the kill path, on ONE throwaway session, aged by a 30-day-future clock
result: {"reaped":["bbbbbbbb-1467-…"],"heldBack":0,"recent":0}
still there? 0 — reaped
everything else intact: 26
```

The full suite was also checked for the obvious hazard — a spec that imports the server entry would
run the sweep against the developer's own tmux. None does, and the session count is unchanged across
a full `yarn test`.

## Tests

- `reapableTmuxSession`: attached / attach count unknown / a pty of ours / age unknown / inside the
  threshold — each spared; the one case that is ended; and zero meaning off even at a year idle.
- `isRestorableSession`: a grid-logged shell is **not** restorable (the #1479 column's bug).
- the sweep: what it ends, and the counts it reports for what it kept and why; a `null` attach map
  reads as "everything is held"; a zero threshold does nothing at all.
- the log lines: both halves, and "off" said out loud rather than silence.
- `sanitizeReapIdleDays`: rounding, range, `0` kept, junk → 7.
- the section: the `ends at next start` marker, and the stepper writing its own config field.

work in mulmoterminal3

## Summary by CodeRabbit

* **New Features**
  * Added automatic cleanup of unused sessions when the server starts.
  * Added configurable idle-session thresholds, including an option to disable cleanup.
  * Added Settings controls showing sessions scheduled for cleanup and allowing threshold changes.
  * Preserved on-disk transcripts while ending eligible idle sessions.

* **Documentation**
  * Documented configuration, behavior, limits, defaults, and cleanup effects in English and Japanese guides.


isamu:
Thanks — six of the eight are in as of `ca65aad2`, including the two that were genuinely dangerous.

**Applied**

- **A kill that failed was still recorded as a kill** (Major). `tmuxKillSession` discarded tmux's
  status, so a refused kill went into `reaped` — and `server/index.ts` then removes those ids from
  the surviving set and deletes their session settings files, which can hold a provider's API token.
  It returns a boolean now, and only a confirmed kill is recorded.
- **Unknown state was reported as a fact** (Major). An unreadable attach count was counted as
  "in use" and an unknown age as "active within N days". Both now land in their own `unclear`
  bucket, and the log says `… , 2 tmux would not report on`. The whole point of that log is to
  replace a confident wrong answer with a true one.
- **The row marks went stale after changing the threshold** (Major). `reapable` is the server's
  answer against the old number, so raising it left rows promising `ends at next start` about a
  start that will now spare them. The section re-reads after a successful save.
- **`Math.round(0.4)` is `0`, which is the off switch** (Minor). A fractional value silently
  disabled the sweep — the exact failure the fallback in that function exists to prevent, and my own
  comment there says so. Fractions are refused now, not rounded.
- The fence language in the plan file.

Each has a test that fails without the fix.

**Not applied — moving `sessionIdleReapDays` out of `mulmoterminal-config`** (both the SKILL.md and
the settings-coverage comments)

The rule being quoted is real, but it is about not documenting a key in the router *in addition to*
its owning writing skill. There is no other owning skill here: the writing skills are `-dirs`,
`-theme`, `-header`, `-keys`, `-model` and `-notify`, and an idle-session sweep belongs to none of
them. The repository already treats `mulmoterminal-config` as the owner for exactly this case —
**ten other keys** are registered to it in the same map:

```
prRepos, gitlabHosts, worklogEnabled, worklogIntervalHours,
decisionDigest, issueWorkComments, prWorkdirFooter, appendSystemPrompt, cockpitLines, …
```

`worklogEnabled` / `worklogIntervalHours` is the closest sibling — a background behaviour with a
number and a Settings control — and it lives there. Following the finding would mean inventing a
skill for one key, or filing it under an unrelated one; either is worse than the convention already
in place. Happy to revisit if the maintainers want a `mulmoterminal-sessions` writing skill, but
that is a repository-wide move, not this PR's.

PR #1497, opened 2026-08-06
fix: gate REST item create/update on the schema before writing

### LINKED ISSUE #1489: REST item create/update skip schema validation (unlike view-data and manageCollection putItems)
## Summary

`POST /api/collections/:slug/items` and `PUT /api/collections/:slug/items/:itemId` write the request body through `store.write` without `validateRecordObject`. View-data PUT, mutate actions, and `manageCollection` putItems all validate.

## Why it matters

Any client (UI bug, curl, future remote writer) can persist invalid enums, missing required fields, or bad types. Detail later surfaces `issues` and a Repair affordance, so bad data becomes first-class rather than rejected at the door.

Note: current MulmoClaude REST create/update has the same gap — fixing here can lead or stay in lockstep, but defense-in-depth is still worth it for both hosts over the shared workspace.

## Evidence

- `itemCreateHandler` / `itemUpdateHandler` in `server/backends/collections.ts` call `target.write(...)` with no schema gate.
- `writeViewItem` (view-data PUT) calls `validateRecordObject` before write.
- Core `putOneItem` (manageCollection putItems) validates before write.

## Proposed fix

Prefer one shared write path so REST, views, and the agent cannot diverge:

1. Best: route REST create/update through the same validation as putItems (or call putItems-equivalent helpers).
2. Minimal: call `validateRecordObject(recordWithId, itemId, collection.schema)` and return 400 with the problem before `write`.

Also consider the computed-key checks putItems runs (`computedKeyProblem`) so host-computed fields cannot be forged via REST.

## Acceptance

- [ ] Invalid create/update bodies return 4xx with an actionable problem; nothing is written.
- [ ] Valid CRUD still works (existing collections.spec CRUD cases pass).
- [ ] Behavior aligns with view-data / putItems on the same schema cases (bad enum, missing required, etc.).

## Source

From the collection-mechanism architecture review (priority 2).

snakajima:
Fixes #1489.

## The bug

`POST /api/collections/:slug/items` and `PUT /api/collections/:slug/items/:itemId` handed the request body to `store.write` with no schema gate. The other two write paths over the same records already validate — view-data PUT (`writeViewItem`) and the agent's `manageCollection` putItems (core's `putOneItem`).

So any client (a UI bug, a curl, a future remote writer) could persist a missing required field or an off-enum value. The record then came back through the detail route as an `issue` with a Repair affordance — bad data as a first-class citizen rather than a rejected write.

## The fix

One shared helper, `rejectedInvalidRecord`, called by both handlers: `validateRecordObject(record, itemId, schema)` → 400 with the problem, nothing written. It sits next to `writeViewItem`, which makes the identical call, so the three write paths now agree.

**Tier is core's default `"enforced"`** — primaryKey↔id identity, required fields non-empty, enum membership. Deliberately *not* the `"strict"` tier: core documents that one as report-only, so a legacy record whose `number` field holds a string stays editable rather than becoming unsaveable. That is also why this cannot lock users out of existing data in general — the enforced tier is exactly the historical three checks that `validateCollectionRecords` has always reported.

## What I did NOT do, and why

The issue says "also consider the computed-key checks putItems runs (`computedKeyProblem`) so host-computed fields cannot be forged via REST". I left that out on purpose:

1. **`computedKeyProblem` is not exported from core.** It is a private function in `manageTool.ts`. Adding it here means a MulmoTerminal-local reimplementation of shared semantics — the kind of divergence CLAUDE.md warns about, and MulmoClaude's REST has no such check either.
2. **It can reject writes the UI legitimately makes.** Inline table editing goes through core's `buildUpdatedRecord`, which spreads the row object as-is. A record that already carries a stale computed key on disk (written by an agent before any gate existed, or a persisted `toggle` projection) would start 400-ing on every inline edit, telling the user to remove a field the form does not show. `draftToRecord` skips computed types, so the edit panel is safe, but the inline path is not.

If we want it, the right shape is exporting `computedKeyProblem` from `@mulmoclaude/core` and adopting it in **both** hosts at once, after checking real workspaces for records that already carry computed keys. Happy to file that as a follow-up.

## Tests

Four cases added to `record CRUD` in `test/server/backends/collections.spec.ts`, each **verified to fail without the gate** (reverted the handler change, watched all four go red, restored it):

- create with an off-enum value → 400, and the item never appears in detail
- update with an off-enum value → 400, and the stored record still reads `name: "Foo"`
- create missing a required field → 400 naming the field, and the same create with the field succeeds
- update that empties a required field → 400, stored value untouched

The off-enum assertions check for the same `"not one of"` problem string the existing view-data test asserts on `rejected[].problem`, which is the parity acceptance item.

A new `reqcol` fixture provides a required non-primary field; `testcol` could not, since its view-data tests write partial records on purpose.

## Verification

`yarn format`, `yarn lint` (0 errors; 13 pre-existing warnings, none in touched files), `yarn typecheck`, `yarn test` — 8914 passed / 45 skipped. Existing CRUD cases pass unchanged.

work in mulmoterminal

### DIFF
--- a/server/backends/collections.ts
+++ b/server/backends/collections.ts
@@ -194,6 +194,26 @@ function resolveWriteTarget(res: Response, collection: ResolvedCollection, body:
   return { write, record };
 }
 
+/** Gate a record on its schema before it reaches the store, answering 400 with the
+ *  problem when it fails. True means the request is already answered — stop.
+ *
+ *  The same `validateRecordObject` call the other two write paths make: view-data
+ *  PUT (writeViewItem, below) and the agent's manageCollection putItems (core's
+ *  putOneItem). REST create/update wrote straight through, so a client — a UI bug,
+ *  a curl, a future remote writer — could persist a missing required field or an
+ *  off-enum value that the detail route then reports back as `issues` with a Repair
+ *  button, rather than the write being refused at the door (#1489).
+ *
+ *  The tier is core's default `"enforced"`: primaryKey↔id identity, required fields
+ *  non-empty, enum membership. Deliberately NOT the `"strict"` tier, which exists to
+ *  REPORT (a legacy record whose `number` field holds a string stays editable). */
+function rejectedInvalidRecord(res: Response, collection: ResolvedCollection, record: CollectionItem, itemId: string): boolean {
+  const problem = validateRecordObject(record, itemId, collection.schema);
+  if (!problem) return false;
+  res.status(400).json({ error: problem });
+  return true;
+}
+
 /** The action's skill template, or null after sending a 500. Shared by the per-record
  *  and collection-level action routes. */
 async function readActionTemplateOr500(res: Response, collection: ResolvedCollection, action: { id: string; template: string }): Promise<string | null> {
@@ -384,6 +404,7 @@ const itemCreateHandler: RequestHandler<{ slug: string }> = async (req, res) => 
   const { collection, target } = resolved;
   const itemId = resolveCreateItemId(collection.schema, target.record) ?? generateItemId();
   const recordWithId: CollectionItem = { ...target.record, [collection.schema.primaryKey]: itemId };
+  if (rejectedInvalidRecord(res, collection, recordWithId, itemId)) return;
   const result = await target.write(itemId, recordWithId, { refuseOverwrite: true });
   if (isCommonStoreFailure(result)) {
     respondForStoreFailure(res, collection.slug, result);
@@ -408,6 +429,7 @@ const itemUpdateHandler: RequestHandler<{ slug: string; itemId: string }> = asyn
     return;
   }
   const recordWithId: CollectionItem = { ...target.record, [primaryKey]: req.params.itemId };
+  if (rejectedInvalidRecord(res, collection, recordWithId, req.params.itemId)) return;
   const result = await target.write(req.params.itemId, recordWithId);
   if (isCommonStoreFailure(result)) {
     respondForStoreFailure(res, collection.slug, result);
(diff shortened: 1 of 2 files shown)


snakajima:
Good catch on the `when` predicate — confirmed and fixed in 2b77c56.

The claim checks out: core's `validateOneField` opens with `if (!fieldVisible(field, record)) return null;` ("a `when`-hidden field has no input the user can fill — never missing"), while `validateRecordObject` runs the `required` check unconditionally. A field that is `required` behind a `when` would therefore have made the editor's own valid save come back 400.

**Fix:** the gate now runs core's `recordFieldProblem` — the exact per-field check `validateRecordObject` itself runs, so the problem strings a REST client sees stay identical to view-data's and putItems' — over the visible, non-computed fields only. `COMPUTED_TYPES` are skipped for the same reason core's compiled validator skips them. The primaryKey↔id identity check is not repeated: both callers build the record as `{ ...body, [primaryKey]: itemId }` from a string id, so it holds by construction.

I did not reach for the alternative you offered (fix the shared validator first) because that is a `@mulmoclaude/core` change and a publish cycle; the comment on the helper says so and says this should become a call to a visibility-aware core validator if one lands.

**Test:** the `reqcol` fixture grows `rating` — `required`, `when: { field: "visited", in: ["true"] }`. One test pins both directions: a create with `visited: false` and no rating is 200, a create with `visited: true` and no rating is 400 naming `rating`. Verified it fails without the fix (removed the `fieldVisible` skip, watched it and the required-field test go red).

Local: `yarn format`, `yarn lint` (0 errors), `yarn typecheck`, `yarn test` — 8915 passed / 45 skipped.

PR #1498, opened 2026-08-06
feat: catch the custom-view host surface up to MulmoClaude (#1490)

### LINKED ISSUE #1490: Catch up custom-view host surface: view-data/image, view-data/actions, view-i18n (or document scope)
## Summary

MulmoClaude exposes several custom-view host routes that MulmoTerminal either stubs or omits. If custom views are a supported surface on MT, these should ship; if not, the supported scope should be explicit in docs.

## Gaps vs MulmoClaude (`API_ROUTES.collections`)

| Feature | MulmoClaude | MulmoTerminal | Impact |
|---|---|---|---|
| `GET …/view-data/image` | yes | missing | Desktop custom views cannot safely thumbnail workspace `image` fields (mobile path inlines via remote-view items) |
| `POST …/view-data/actions/:actionId` | yes | missing | Custom views cannot run `kind: "mutate"` under the view token |
| `GET …/view-i18n` | real dict | stub `{ locale: "", dict: {} }` in `collectionUi.ts` | Custom-view i18n falls back to raw keys |

Also related (lower priority for this issue): registry is list+import only (no preview/export). Can be a follow-up.

## Proposed approach

Either:

**A. Implement** the three endpoints + UI bindings, matching MulmoClaude paths/status bodies (see `../mulmoclaude/src/config/apiRoutes.ts` and its collections routes), including:
- image path authorization against current record image-field values (no arbitrary file reads)
- mutate-only under view token (never chat/agent)
- i18n dict load via core `readCustomViewI18n` when present

**B. Document** that on MulmoTerminal, custom views are read-mostly (no image resolve, no view-token mutate actions, no i18n files), and point authors at mobile remote views / desktop table for those capabilities.

## Acceptance

- [ ] Choice A or B is explicit in code comments + guide/changelog as appropriate.
- [ ] If A: routes, binding hooks, and tests exist; parity notes vs MulmoClaude paths.
- [ ] If B: `docs/mulmoclaude-parity.md` (or collection docs) states the scope so the next host binding does not re-open this as accidental drift.

## Source

From the collection-mechanism architecture review (priority 3).

snakajima:
Closes #1490 — option **A** (implement), not B (document the gap).

## Why A

The gap was not a scope choice we had made; it was drift nobody could see from inside this repo. Core's bundled custom-view docs — which MulmoTerminal itself serves to the agent through `manageCollection`'s `schemaDocs` — document `<dataUrl>/image`, `<dataUrl>/actions/<id>` and `"i18n": "views/<name>.i18n.json"` as things a view author can use. On this host, a view following those instructions rendered raw keys, showed broken images, or 404'd, with nothing reporting a problem anywhere.

## What ships

New `server/backends/customViewRoutes.ts`, all three running the shared `@mulmoclaude/core` engines:

| Route | Notes |
| --- | --- |
| `GET …/view-i18n` | Parent-side (not the view token) — core's `readCustomViewI18n`, one locale picked host-side. `{ locale: "", dict: {} }` when a view declares no `i18n`. |
| `GET …/view-data/image` | Scoped token. Resolves an image field's stored path to a downscaled `data:` thumbnail via the same `resolveThumbnail` the phone path uses. |
| `POST …/view-data/actions/:actionId` | Scoped **write** token. **Mutate kind only** — 403 otherwise. |

Two rules worth keeping when this code is touched again:

- The image route's **authorization is the record scan**. `isAuthorizedImagePath` accepts only a **current** value of an `image` field, so a scoped token can never be turned into an arbitrary workspace file read.
- The action route is **mutate-only** on purpose. A view token must never be able to start LLM work, so `chat` / `agent` actions stay on the parent-side route. It otherwise runs the identical pipeline the UI button does: `require` re-checked against the record, params validated, write gate, atomic write.

`server/backends/viewRateLimit.ts` ports MulmoClaude's per-(IP, slug) budget. Images get their own roomier bucket (300/min) because a gallery's first paint is legitimately dozens of fetches; actions share 60/min with `/query`, which had no limiter here before.

**UI side.** `fetchViewI18n` was a stub returning the empty dict; it is now a real fetch. More importantly, `customViewSrcdoc.ts` was missing the i18n half of the bridge *entirely* — no `dict`, no `locale`, and `__MC_VIEW.t` simply `undefined`, so the issue's "falls back to raw keys" was optimistic. Both boot fields and the vue-i18n-compatible `t(key, named)` helper are now there, matching MulmoClaude's string verbatim. `@mulmoclaude/collection-plugin` already called `fetchViewI18n` and forwarded both fields into `buildViewSrcdoc`, so nothing in the plugin needed changing.

## Structure

The handlers sit in their own module behind a deps seam (`resolveCollection`, `resolveView`, `respondForMutateAction`, `visibilityGate`, `guarded`, the CORS and concurrency middlewares handed in at mount time) rather than in `collections.ts`, which is at its 600-line lint budget. Passing the helpers in also keeps the two modules free of an import cycle.

## Parity notes

`docs/mulmoclaude-parity.md` gains a section recording this surface as shared, and states the two differences left open deliberately so the next host binding does not re-open them as accidental drift:

- `GET …/view-data` ignores `ids` / `fields`. MulmoClaude routes that through `manageCollection`'s `getItems`; MT returns every enriched record, so a view passing `?fields=…` (as the core doc's examples do) gets a working superset — it just transfers more than it asked for.
- The registry is list + import only — no preview or export route.

Route paths, status codes and refusal wording were checked against `../mulmoclaude/server/api/routes/collections.ts` and `src/config/apiRoutes.ts` rather than guessed.

## Tests

`yarn typecheck`, `yarn lint` (0 errors), `yarn test` — 8939 passed, 45 skipped.

- 12 route tests over real fixtures: a real PNG through `sharp` for the image route (plus the negative case — a path held by a non-image field 404s), a `require`-gated mutate action for the 200 / 403 / 409 / 400 / 404 mapping, a two-locale i18n file for locale pick, `en` fallback and the no-`i18n` shape.
- 3 unit tests for `isAuthorizedImagePath`, 3 for the rate limiter (budget, per-key isolation, window reset).
- 3 srcdoc tests that **execute** the bootstrap against a stand-in `window` and assert what `t()` returns, rather than string-matching the bundle.

Not verified in a live browser yet — that needs a `yarn build` and a restart against a collection with an image field and a custom view.

work in mulmoterminal

## Summary by CodeRabbit

* **New Features**
  * Added custom-view support for locale-specific translations with fallback text and named placeholders.
  * Added secure image resolution for current view data, including downscaled previews.
  * Added declared custom-view actions for updating eligible records.
  * Added authorization, validation, CORS handling, and separate rate limits for image and action operations.
* **Documentation**
  * Documented the custom-view host surface, endpoint behavior, access rules, and known parity gaps.

### DIFF
--- a/docs/ChangeLog.md
+++ b/docs/ChangeLog.md
@@ -8,6 +8,22 @@ This file records **what changed and why**. For **how to actually use** a new fe
 
 Entries here are folded into the next release's heading when it ships.
 
+- **Custom views catch up to MulmoClaude's host surface** (#1490): a collection's custom view can
+  now read its translations, resolve a stored image, and press a declared mutate action — the three
+  things core's own authoring docs (which MulmoTerminal serves to the agent through
+  `manageCollection`'s `schemaDocs`) already told view authors it could.
+  `GET …/view-i18n` returns the active locale's dict, and the iframe bootstrap gained the
+  `__MC_VIEW.dict` / `__MC_VIEW.t(key, named)` half of the bridge it had been missing entirely — an
+  i18n-declaring view used to render raw keys, with nothing reporting a problem. `GET
+  …/view-data/image` turns an image field's stored path into a downscaled `data:` thumbnail, which
+  is how a desktop view shows a picture at all (the sandboxed iframe cannot put its token on an
+  `<img>`); the authorization rule is the record scan itself, so only a **current** value of an
+  image field resolves and never an arbitrary workspace file. `POST
+  …/view-data/actions/:actionId` lets a write-capable view press a declared **mutate** action
+  instead of hand-rolling the same transition as a raw write — mutate only, so a view token still
+  cannot start LLM work. All three run the shared `@mulmoclaude/core` engines, and the two scoped
+  ones sit behind per-minute budgets with images in a roomier bucket than actions.
+
 ## mulmoterminal@4.6.0 — 2026-08-06
 
 > **Setup guide:** [Cells that talk to each other, and a history that knows whose it is](https://receptron.github.io/mulmoterminal/guide/en/v4.6.0.html) — written at release time. ([日本語](https://receptron.github.io/mulmoterminal/guide/ja/v4.6.0.html))
--- a/docs/mulmoclaude-parity.md
+++ b/docs/mulmoclaude-parity.md
@@ -6,7 +6,7 @@ the two hosts can't drift. This doc records where that effort stands: which
 subsystems are shared today, which are deliberately **not**, and what picking
 up each remaining item would involve.
 
-Status date: **2026-07-18** (core `0.23.0`, collection-plugin `0.11.1`).
+Status date: **2026-08-05** (core `2.1.0`, collection-plugin `2.0.0`).
 
 ---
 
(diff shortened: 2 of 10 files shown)


snakajima:
Thanks — three of the four applied in af12471c, one skipped:

- **Rate-limiter sweep** (`viewRateLimit.ts`): applied. The point about the limiter running ahead of `requireViewToken` is what makes it worth fixing — unauthenticated requests create entries, so the pathological case is reachable without a token. Now sweeps at most once per window, with a comment recording that this diverges from MulmoClaude's size-only gate and why.
- **Parity wording**: applied. It now claims *endpoint and authorization* parity and points at the response-body difference the section goes on to list, instead of claiming parity and then contradicting itself.
- **`guarded` on the query route**: the omission is deliberate, so I added the comment saying so. The handler catches its own throw in order to answer a **fixed** message — a raw DuckDB error can carry host paths, and a scoped view is not a trusted audience for those. `guarded` would put exactly that in the body.
- **Setup-guide link under `## Unreleased`**: skipped. In this repo the `> **Setup guide:**` line is written at release time and points at `docs/guide/{en,ja}/v<version>.md`, a dated page that doesn't exist until a version is cut. `## Unreleased` is a holding area whose entries are folded into the next release's heading, which is where that link is added.

snakajima:
On the image route's concurrency cap — I'm keeping the shared `queryConcurrency` (4 in flight per slug), and I've added a comment at the mount site (f5849d5a) saying why, so it doesn't read as an oversight.

The cap is the **documented contract**, not an accident of copying. Core's `custom-view.md` — the doc this app itself serves to the agent through `manageCollection`'s `schemaDocs` — says, under *Displaying images*:

> **Never fire one fetch per image in parallel** (`Promise.all` over every record) — the host caps in-flight `/image` + `/query` requests at **4 per collection** and answers the rest **429**, so a first paint of a dozen images half-fails. Resolve through a small worker pool instead:

…followed by a `resolveImages(paths, onResolved, workers = 3)` snippet with a 429 retry. MulmoClaude's route mounts the identical `viewQueryConcurrency` guard.

So the failure mode described — a gallery firing N parallel fetches and getting 429s — happens on **both hosts, by design**, and the view authors are told how to avoid it before they write the fetch. Raising the cap only here would do the more damaging thing this PR exists to prevent: make a view that works on MulmoTerminal and breaks on MulmoClaude look correct while it is being written.

The two axes are doing different jobs: in-flight concurrency bounds simultaneous decode cost (each request is a record scan plus a thumbnail decode), and the roomier per-minute image budget (300 vs 60) exists precisely so a throttled pool of 3 isn't starved partway through a full gallery.

If we do want a higher ceiling for images, the right move is to raise it in core's doc + MulmoClaude first and follow here — happy to open that as a separate issue.

PR #1499, opened 2026-08-06
feat: accent collection records that have a pending completion bell (#1491)

### LINKED ISSUE #1491: Wire collectionUi notifiedSeverities to the shared notifier (or drop the affordance)
## Summary

`collectionUi.ts` stubs `notifiedSeverities: () => new Map()`, so the collection plugin never shows severity chrome for completion bells. Meanwhile `server/backends/collectionWatchers.ts` already publishes into the shared notifier file (`<ws>/data/notifier/active.json`) with MulmoClaude-compatible adapter shape.

## Why it matters

Watchers and MulmoClaude can publish collection completion bells, but the collection UI in MulmoTerminal cannot surface which records are pending / done via the plugin's severity map. Users get incomplete parity for a feature whose backend half already runs.

## Evidence

- Stub: `src/composables/collectionUi.ts` (`// No notifier in MulmoTerminal.` / empty map).
- Watchers: `server/backends/collectionWatchers.ts` + `collectionNotifierAdapter.ts` (byte-identical pluginPkg / legacy markers with MulmoClaude).
- Integration doc still lists the stub under the original gap table.

## Proposed fix

1. Expose a read of active collection notifications to the browser (existing notifier API if any, or a thin `/api/…` over the same active.json).
2. Map entries into `Map<string, CollectionNotifySeverity>` the plugin expects (same keying MulmoClaude uses).
3. Keep the cross-app invariant: one bell per record regardless of which app published it (`readEntry` already recognises legacy markers).

Alternatively, if MT will not show severity in the collection chrome, remove or hide the plugin affordance that depends on a non-empty map and document that completion bells are delivered only via MT's own notifier UI (if any).

## Acceptance

- [ ] Either live severities update when watchers publish/clear, or the UI no longer implies that map is wired.
- [ ] No double bells when switching between MulmoClaude and MulmoTerminal on the same workspace.
- [ ] Spec or fixture for map keying if implementing the live path.

## Source

From the collection-mechanism architecture review (priority 4).

snakajima:
Closes #1491 — the "wire it" branch, not the "drop the affordance" one.

## The premise had gone stale, which makes this small

The stub carried the comment `// No notifier in MulmoTerminal.` That stopped being true a while ago. `src/composables/useNotifications.ts` keeps a live `active` list fed by `/api/notifications` plus the `notifications` pubsub channel, and `NotificationBell.vue` renders it — so step 1 of the issue's proposed fix ("expose a read of active notifications to the browser, or a thin `/api/…` over `active.json`") needed nothing at all.

Both hard halves were already done, in fact:

- `server/backends/collectionWatchers.ts` publishes completion bells into the shared notifier file, and `collectionNotifierAdapter.ts` writes MulmoClaude's exact `pluginData.action.target = { view: "collections", slug, itemId }`.
- The browser already holds those entries live.

What was missing was the mapper between them. So the Kanban view — which colours a card's left border from `notifiedSeverities` — never accented anything, on either app's bells.

## What ships

- **`common/collectionNotifyTarget.ts`** — the `view: "collections"` marker plus the defensive narrowing of `pluginData.action.target`. In `common/` deliberately: the server writes this target and the UI reads it, so per CLAUDE.md it is a value both sides decide from. A second copy of the literal in the reader is precisely how that pair drifts without a symptom — the bell keeps arriving, only the accent quietly stops. The adapter now builds from the same constant.
- **`src/utils/collectionNotified.ts`** — a port of MulmoClaude's `collectionNotifiedSeverities` (`src/utils/collections/notifiedItems.ts`). That file is MulmoClaude **app source**, not `@mulmoclaude/core`, so it cannot be imported — same situation as `isAuthorizedImagePath` in #1490. `itemId → worst severity`; entries carrying a slug but no concrete `itemId` are skipped, since a collection-level bell has no single card to accent.
- **The binding** reads `useNotifications().active` *inside* the call. That is what makes it live: the plugin wraps the binding in a `computed`, so reading the ref during evaluation registers the dependency and the accents follow a watcher publishing or clearing. `useNotifications()` is a module singleton with an idempotent init, and `usePubSub` has no lifecycle hooks, so calling it per invocation is safe outside `setup`.

## Against the issue's acceptance criteria

- **Live severities update when watchers publish/clear** — yes, via the reactive read above.
- **No double bells across apps** — unaffected, and it could not be otherwise: this path only reads. Dedupe stays where it already was, on the shared `legacyId` server-side. The upside of sharing the wire shape is the inverse — a bell **MulmoClaude** published accents the same card here.
- **Spec for the map keying** — 7 tests, including the one that matters: a `pluginData` built by this app's own publisher (`buildPluginData`) is read back by the mapper. That fails if either end stops going through the shared module.

## Docs

`docs/collection-plugin-integration.md` listed `notifiedSeverities → empty map` in its Tier-1 stub table, which the issue flagged. I annotated the bullet rather than rewriting it — every binding in that list is real now, and the section is a record of how the increment was scoped, so editing it in place would rewrite history rather than update it.

## Tests

`yarn typecheck`, `yarn lint` (0 errors), `yarn test` — 8946 passed, 45 skipped.

Not verified in a live browser yet: that needs `yarn build` + a restart, a collection with a `kanban` group field, and a watcher-published bell against one of its records.

work in mulmoterminal

## Summary by CodeRabbit

* **New Features**
  * Collection boards now display active notification accents for records with pending completion reminders.
  * Notifications support informational, nudge, and urgent severity levels.
  * Multiple notifications for the same record are represented using the highest active severity.

* **Documentation**
  * Updated release and integration documentation to describe live notification severity mapping and board behavior.

* **Tests**
  * Added coverage for filtering, severity precedence, malformed data, and unknown severity handling.

### DIFF
--- a/common/collectionNotifyTarget.ts
+++ b/common/collectionNotifyTarget.ts
@@ -1,0 +1,37 @@
+// The navigate target a collection-completion bell carries, and the one narrowing
+// that reads it back. Shared because BOTH sides decide from it: the server writes
+// the target when it publishes a bell (server/backends/collectionNotifierAdapter.ts)
+// and the UI reads it to accent the record's card (src/utils/collectionNotified.ts).
+// A second copy of the literal in the reader is exactly how a writer/reader pair
+// drifts silently — the bell keeps arriving, only the accent stops appearing.
+//
+// CROSS-APP: the shape is MulmoClaude's, which spells the view constant
+// `NOTIFICATION_VIEWS.collections` in its own app source. The VALUE is what
+// crosses the workspace, so keeping it in one place here (rather than importing a
+// constant that isn't published) changes nothing on the wire — entries from either
+// app narrow identically.
+import { isRecord } from "./isRecord.js";
+
+/** `pluginData.action.target.view` for a bell pointing at a collection record. */
+export const COLLECTION_NOTIFY_VIEW = "collections";
+
+export interface CollectionNotifyTarget {
+  slug: string;
+  /** Absent on a collection-level bell — those can't accent a specific record. */
+  itemId?: string | undefined;
+}
+
+/** Narrow a notification's opaque `pluginData` to its collection navigate target,
+ *  or null when it isn't one. `pluginData` reaches the browser as `unknown` (it is
+ *  whatever the publishing app put there, and the other app's version of this
+ *  repo may be older), so every level is checked rather than trusted. */
+export function collectionNotifyTargetOf(pluginData: unknown): CollectionNotifyTarget | null {
+  if (!isRecord(pluginData)) return null;
+  const { action } = pluginData;
+  if (!isRecord(action)) return null;
+  const { target } = action;
+  if (!isRecord(target)) return null;
+  const { view, slug, itemId } = target;
+  if (view !== COLLECTION_NOTIFY_VIEW || typeof slug !== "string" || slug.length === 0) return null;
+  return { slug, itemId: typeof itemId === "string" && itemId.length > 0 ? itemId : undefined };
+}
--- a/docs/ChangeLog.md
+++ b/docs/ChangeLog.md
@@ -8,6 +8,15 @@ This file records **what changed and why**. For **how to actually use** a new fe
 
 Entries here are folded into the next release's heading when it ships.
 
+- **A record with a pending completion bell is now visible in the board** (#1491): the collection
+  view's Kanban cards accent (red for urgent, amber for a nudge) when a completion watcher has an
+  open bell for that record. The publishing half already ran — `collectionWatchers` writes into the
+  shared notifier file, and MulmoTerminal's bell has shown those entries for some time — but the
+  binding the plugin asks for the per-record severities returned an empty map, so nothing in the
+  board reflected it. It now reads the live bell, which means a bell **MulmoClaude** published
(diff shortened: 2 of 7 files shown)


snakajima:
Good catch on the contradiction — fixed in 7ff8b021, though in the opposite direction to the suggestion, because the note is the accurate half.

I checked all 17 bindings named in that Tier-1 stub list against `src/composables/collectionUi.ts`: `createItem` `updateItem` `deleteItem` `deleteCollection` `deleteFeed` `runItemAction` `runCollectionAction` `refreshCollection` `deleteView` `mintViewToken` `fetchViewHtml` `buildViewSrcdoc` `listFeeds` `reconcileShortcuts` `unpin` `notifiedSeverities` — and `startChat`, which is `startCollectionChat(prompt, { hidden: false })` at line 252. Every one is wired.

So the stale statement is the *other* one: the Tier 2 section's "Still open: `startChat` is a no-op … and Tier 1 write routes". Both shipped (the write routes are mounted in `server/backends/collections.ts`). Weakening the new note to say only `notifiedSeverities` is done would have made the doc agree with itself by adopting the wrong fact.

The correction is appended rather than overwriting that sentence — the section is titled *Tier 2 status*, so what it originally said is the record of where Tier 2 stopped, and the addendum states where things stand now.

PR #1505, opened 2026-08-06
feat: accept Claude Code's [1m] extended-context suffix in a model id (#1503)

### LINKED ISSUE #1503: feat: model id のバリデーションが Claude Code の拡張コンテキスト構文 [1m] を拒否する
## What happened

`.mulmoterminal.json` の `model` に Claude Code の公式な拡張コンテキスト構文である `claude-opus-5[1m]`（末尾 `[1m]`）を指定すると、セッション起動時に以下のエラーで拒否される。

```
[unusable model id "claude-opus-5[1m]" — letters, digits and . _ : / - ~ only]
```

## What I expected

`[1m]` サフィックスは Claude Code CLI が公式にサポートする「拡張コンテキストウィンドウ（1M）を明示指定する」記法（`code.claude.com` の `model-config.md` "Extended context" セクションに記載）であり、Claude Code 自身がこのサフィックスを剥がしてからプロバイダに送るため、`--model claude-opus-5[1m]` は Claude Code 上で正当な入力のはず。MulmoTerminal 側のバリデーションでこれを弾かず通してほしい。

## Steps to reproduce

1. `.mulmoterminal.json` に `"model": "claude-opus-5[1m]"` と書く
2. 該当ディレクトリで Claude セッションを起動する
3. `[unusable model id "claude-opus-5[1m]" — letters, digits and . _ : / - ~ only]` で起動が拒否される

毎回再現する。

## Analysis

- `common/modelIds.ts` の `MODEL_ID_RE = /^[A-Za-z0-9~][A-Za-z0-9._:/~-]*$/` が `[` `]` を許可文字に含めておらず、コンフィグスキーマ・起動ピッカー・ws query の3箇所すべてで共有されているため、`.mulmoterminal.json` の `model` / `provider` 経由のどちらからも `[1m]` サフィックスを渡す手段がない。
- `[1m]` は現状 Opus 5 / Sonnet 5 のような最新世代モデルでは Anthropic API 上デフォルトで1Mコンテキストになるため実害は小さいが、Opus 4.6 や Sonnet 4.6 など「`[1m]` が実際に1Mへのアップグレードを制御する」モデルでは、この記法が使えないと明示的な1Mコンテキスト指定ができない。
- 影響範囲: `.mulmoterminal.json` の `model`、グローバル設定の `providers[].models`、起動ピッカーでの手動入力、すべて同じ正規表現を通るため同様に拒否される。

## Suggested fix

`MODEL_ID_RE` に、末尾のみ許可する形で `\[1m\]` のような拡張コンテキストサフィックスを追加する（例: `/^[A-Za-z0-9~][A-Za-z0-9._:/~-]*(\[1m\])?$/`）か、もしくは別枠のオプション（例 `contextWindow: "1m"`）として分離して `--model` 構築時に付加する。

## Environment

- mulmoterminal 4.6.0
- Node v24.18.1
- Linux x86_64 (7.1.5-1-cachyos)
- Claude Code 2.1.223

isamu:
## Summary

`.mulmoterminal.json` holding `{"model": "claude-opus-5[1m]"}` refused to launch with
`[unusable model id "claude-opus-5[1m]" — letters, digits and . _ : / - ~ only]`.

`[1m]` is Claude Code's own syntax for asking a model for the 1M context window. The CLI
strips the suffix before the id reaches the provider, so it never leaves Claude Code — but
MulmoTerminal's id shape rejected `[` and `]`, and the config schema, the ws query and the
spawn-time check all read that one shape, so there was no way to write it at all.

This widens the shape by **exactly** the suffix the docs define, and passes the value to
`claude --model` verbatim. Interpreting or stripping it here would mean this repo owning a
rule it does not own.

A model id and a provider id stop sharing one predicate, because the suffix is a model-NAME
suffix and means nothing on a `providers[]` key the user invents:

| Call site | Predicate |
| --- | --- |
| `providerSchema.id`, ws query `?provider=` | `isUsableProviderId` (unchanged shape) |
| `providerSchema.models[]`, ws query `?model=`, `resolveProvider` | `isUsableModelId` (suffix allowed) |

Both keep reading the same `BASE_ID_RE`, so the part that keeps a value safe in argv — no
whitespace, no leading dash, no `|`, no control characters — cannot drift between them.

Closes #1503

## Items to Confirm / Review

1. **The suffix is passed through untouched on the third-party provider path too**, so a
   directory pointing at a gateway gets `ANTHROPIC_MODEL=…[1m]` and
   `ANTHROPIC_SMALL_FAST_MODEL=…[1m]`. That is deliberate: Claude Code reads those variables
   and strips the suffix itself, and the docs describe selecting the 1M window behind a
   gateway in exactly this way: *"when `ANTHROPIC_BASE_URL` points at a gateway, Claude Code
   can't verify 1M support. To use the full window, select Sonnet 5 (1M context) in the model
   picker, which maps to `sonnet[1m]`."* Since the CLI strips the suffix before the request
   goes out, the gateway sees the same model id it saw before — the only difference is the
   1M beta header, which is the thing being asked for. Not measured against a live gateway
   though, so please sanity-check if you run one.
2. **`[1m]` is matched as a literal, not as a `[\d+[a-z]]` family.** Checked the whole
   model-config page: 19 occurrences of `[1m]` and no other bracketed form; the effort level
   is a separate `effort` setting, not a name suffix. So `[2m]`, `[200k]`, `[1M]` are all
   still refused. Flagging in case anyone knows of a suffix the docs do not list.
3. **Splitting one predicate into two** is the reviewable design call. The alternative was to
   let a provider `id` carry `[1m]` as well and keep a single function. I chose the split
   because a provider id reaches no CLI, so the suffix there could only ever be a typo that
   the schema and the ws query would then both silently honour.
4. **The refusal wording changed**, since it is generated from the rule rather than written
   twice: `unusable model id "x" — allowed: letters, digits and . _ : / - ~, plus an
   optional [1m] suffix`.

## User Prompt

- https://github.com/receptron/mulmoterminal/issues/1503 調査と対策
- これって 1m 以外はない? 公式 doc を確認してね。そのうえで url があればソースにコメント
- （方針の選択）正規表現に `[1m]` を許可する案で
- はい review -> merge まですすめて

## What the official docs say

<https://code.claude.com/docs/en/model-config#extended-context> — recorded as a comment in
`common/modelIds.ts` beside the constant, and in the guide.

- Valid on an alias (`opus[1m]`, `sonnet[1m]`, `opusplan[1m]`) and on a full model name
  (`claude-opus-4-8[1m]`).
- Accepted by `/model`, by the `model` setting, and by `ANTHROPIC_DEFAULT_*_MODEL`.
- "Claude Code strips the suffix before sending the model ID to your provider."
- It is the only bracketed suffix that exists.

## Verification

Against an external ground truth rather than against our own output:

- **The CLI itself.** `claude --model 'claude-opus-5[1m]' -p 'Reply with exactly: OK'` on
  Claude Code 2.1.223 answers `OK` and exits 0 — so `--model` takes the suffix as readily as
  `/model` does, which the docs only show for the latter.
- **The real load path**, not a hand-called predicate. `dir-model-choice.spec.ts` writes a
  `.mulmoterminal.json`, runs `loadDirConfig` → `effectiveChoice` → `resolveProvider` →
  `buildClaudeArgs`, and asserts the suffix arrives in argv unchanged — so it fails if any
  link in that chain starts rewriting the id.
- `yarn format` / `lint` (0 errors; the 14 warnings are pre-existing, in files this PR does
  not touch) / `typecheck` / `build` / `test` — **621 files, 8977 tests, all passing**.

Not done: launching a session through a running server by hand. The change is a validation
predicate, and the two checks above cover both ends of it.

## Tests

- `test/common/modelIds.spec.ts` (new) — the shape itself: every documented form accepted;
  the suffix refused in the middle, doubled, unterminated, or with nothing in front of it;
  invented suffixes refused; the suffix counted against the length limit; the provider/model
  asymmetry; and a block pinning that the base shape did **not** loosen.
- `test/server/config/dir-model-choice.spec.ts` — the suffix through to `claude --model` and
  through to `ANTHROPIC_MODEL` on the provider branch.
- `test/server/session/launch-choice.spec.ts` — a `?model=` carrying the suffix travels; a
  `?provider=` carrying it drops the whole pair.

## Docs

- `docs/guide/{en,ja}/providers.md` — how to write it, that it goes to the CLI untouched,
  that it is the only bracketed suffix, and that a provider `id` may not carry it.
- `server/skills/mulmoterminal-model/SKILL.md` — the skill that owns model settings, incl.
  when *not* to suggest it (on Opus 5 / Sonnet 5 the Anthropic API already runs 1M, so it
  changes nothing there).
- Dated release pages are snapshots and were left alone.

## Not in scope

- A `[1m]` entry in the launch picker. `MODEL_PRESETS` is measured pass-rate data for
  third-party backends, and the picker has no Anthropic-model section to add it to.
- Codex / Grok / Antigravity model overrides — they come from `CODEX_MODEL` and friends in
  the server's environment and never pass through this shape.

work in mulmoterminal5


isamu:
## Bot review triage (iteration 1)

| Bot | Signal | Action |
|---|---|---|
| Codex (`github-actions[bot]`) | `CODEX VERDICT: LGTM` | nothing to address |
| CodeRabbit | rate limited — no review produced | nothing posted to triage |
| Sourcery | did not run on this repo | — |
| Socket Security (project + PR alerts) | pass | — |

No inline comments from any bot.

## One thing I chased myself, and why it is NOT in the diff

Reviewing beyond the bots, the badge looked like a real gap. `src/components/modelBadge.ts`
picks a context window by ordered substring match, and `opus[1m]` / `sonnet[1m]` match none of
the dated entries — they would fall through to the bare-family `200k` and report ~5x the
context actually in use, which is exactly #985's failure mode. I wrote the fix (`[1m]` → 1M,
ranked below a window the agent itself reported) and its tests, and they passed.

Then I checked the assumption underneath it against a real transcript instead of against my
own reasoning — **and it is false.** The badge's model id comes from the transcript, not from
what we passed to `--model`, and Claude Code records the *resolved, stripped* id:

```
$ grep -o '"model":"[^"]*"' ~/.claude/projects/<this session>.jsonl | sort | uniq -c
 186 "model":"claude-opus-5"
```

That session is running `claude-opus-5[1m]`. The suffix never reaches the transcript, so it
never reaches the badge — the fix was unreachable code defending against a case that cannot
occur. Reverted rather than shipped.

The same reasoning clears `server/session/cost.ts`, which matches model ids by **prefix**: it
also reads the transcript's stripped id, and a trailing suffix could not have broken a prefix
match in any case.

So the suffix's blast radius really is what the PR body claims — argv and the provider `env`
block, both of which hand it to Claude Code, which is the thing that understands it.

PR #1513, opened 2026-08-06
fix(muse): resume keeps the workspace, the context badge stops reading a high-water mark, and the 33 MB log is folded

snakajima:
Review of the recently added **grok** and **muse** agents, concentrating on the three areas asked about — chat history, resuming from it, and MCP. grok came out of it in good shape; everything below is muse, which arrived later and had not yet been brought in line with the shape the other three agents share.

Measured against the real CLIs on this machine (`muse` 2026-08-06, whose `--help` and `resume --help` are the source for the argv claims) and against a real 33 MB session log.

## Bugs

**A resumed muse session lost its workspace tools.** `buildMuseArgs` returned `["--yolo", "resume", <id>]` and nothing else, dropping `--workspace <cwd>` — which is what registers the policy-gated workspace tools. muse takes root options on either side of the subcommand (`muse resume --help` says so, and `muse --yolo --workspace /tmp resume <id>` was run to confirm it parses), so the flag now rides both paths. A seed prompt is still not passed on a resume, because `muse resume <id> <prompt>` is not a command line muse accepts — that is now stated in the type rather than being a silent drop.

**The context badge read a high-water mark.** `museBadgesFromLog` took the LARGEST `input_tokens` it had ever seen, and `input_tokens` is the context that call ran with, not an increment — so the reading never came down after a compaction. On the session measured here that is `ctx 397k` against a live 266k: a badge telling the user to `/compact` when they need not. It now takes the last completed call's value, which is what grok's and codex's badges mean.

## Performance

**The badge poll re-read the whole session log every time.** 33 MB after a day's work here, `JSON.parse` per line, once per cell per minute plus every `/api/session/:id`. It now folds through `transcript-fold` exactly as grok's `updates.jsonl` does (#1377) — measured on that same log: **105 ms cold, 2 ms warm**, and the totals are byte-for-byte what the old reader produced.

**Every sqlite question was a full-table scan** (`SELECT … FROM sessions`, filtered in JS) — including the spawn watcher, which asks twice a second for 15 s. All queries are now `WHERE session_id = ?` / `WHERE workspace_root = ?`, opened and closed through one helper instead of three copies of the open/close/close-in-a-finally dance.

**A claude cell paid for muse on every poll.** The self-healing fallback in `sessionDetail` fired whenever a claude read came back empty — which is the normal state of every claude cell before its first turn — costing a sqlite query and a log read to answer for a session muse has never heard of. It is now gated on the in-memory conversation map, which knows every session this server started as muse (including across a restart, since it hydrates from the log). Same self-heal, no cost elsewhere.

**muse looked up MCP groups it cannot use.** It has no MCP of any kind, so `handleDirectoryMcpAgentConnection` was reading Claude Code's per-directory config on every muse spawn for an answer nothing acts on. `DirectoryMcpWsAgent` now carries `readsDirectoryMcpConfig`, false for muse and true for the two agents that really do read a file in the directory.

## Redundancy

- `museSessionExists` and `museModelFromSessionLog` were dead — deleted.
- `resolveMuseSession` had two resume paths, three dynamic `import()`s inside the function, and a `conversationExists: () => false` beside a real existence probe. One path now: ask the map, then the probe, in the order `agent-resume.ts` gives the reason for.
- `spawn-muse.ts` called `carriesFullGuiMcp` and voided the result behind three eslint disables, for guideline parity. The guideline exists so a new spawn path cannot silently differ from the cell beside it — a discarded call does not enforce that, so the answer is declared where both sides read it (`common/guiMcpAgents.ts`) and pinned by a spec.
- `muse-session.ts` lost its four-rule `eslint-disable` header by being split rather than suppressed.
- `modelBadge.ts` had `muse-spark` and `muse` as separate rows with the same window; the first could never be reached.
- Three comments that enumerate the agents nothing can map back to a session key ("agy or grok") now name muse too.

## Not changed, on purpose

- **`--yolo` stays.** It disables approval and the sandbox, which is heavier than grok's `--permission-mode auto`, but a grid cell has no way to answer a modal approval prompt in a TUI nobody is watching — so it is the same decision the other cells make, and changing it is a product call rather than a review finding.
- **Muse sessions surviving a restart still read "shell or unknown"** in Settings' surviving-sessions list. `agentOfKey` consults only claude's transcripts and codex's rollouts, so agy, grok and muse have all had this blind spot since they landed; fixing it is one change for the three of them, not part of this.

## Verification

`yarn typecheck`, `yarn lint` (0 errors), `yarn test` — **9042 passed, 45 skipped**, including 34 new assertions across `muse-args.spec.ts`, `muse-usage.spec.ts`, `muse-gui-mcp.spec.ts` and the muse half of `directory-mcp-ws-agent.spec.ts`. The badge path was additionally run against the real index and log, and the argv against the real CLI.

## Review follow-ups in this PR

- **Codex (P2), badge fallback:** gating the claude-cell self-heal on the conversation map alone dropped the case it was written for — one of muse's own ids polled as claude with no mapping yet (a history row opened before its spawn recorded one, or a session started from a plain terminal). Both gates now, map first, then one indexed lookup.
- **Duplication scan:** the token-count guard grok's and muse's folds each carried is now shared as `usageCount`. `jscpd` is back to the two clones that were there before this branch, and `knip` reports no new dead code.
- **Not taken:** the follow-up review's swallowed `.catch`, unbounded `claimed…` set and snapshot race are shapes muse inherited from codex's and agy's spawners, which have them identically — worth fixing for the three together rather than making muse the odd one out.

work in mulmoterminal

### DIFF
--- a/server/agents/grok-usage.ts
+++ b/server/agents/grok-usage.ts
@@ -19,6 +19,7 @@
 import { isRecord } from "../../common/isRecord.js";
 import type { SessionContextInfo } from "../../common/sessionContext.js";
 import type { SessionUsage } from "../session/transcript.js";
+import { usageCount as num } from "./usage-count.js";
 
 export const emptyGrokUsage = (): SessionUsage => ({ inputTokens: 0, outputTokens: 0, cacheReadTokens: 0, cacheCreationTokens: 0 });
 
@@ -28,11 +29,6 @@ export const isGrokUsage = (value: unknown): value is SessionUsage =>
   typeof value.outputTokens === "number" &&
   typeof value.cacheReadTokens === "number" &&
   typeof value.cacheCreationTokens === "number";
-
-const num = (record: Record<string, unknown>, key: string): number => {
-  const value = record[key];
-  return typeof value === "number" && Number.isFinite(value) && value > 0 ? value : 0;
-};
 
 /**
  * The context reading in a conversation's `signals.json`, or nulls/zeroes when it says nothing.
--- a/server/agents/muse-args.ts
+++ b/server/agents/muse-args.ts
@@ -1,39 +1,52 @@
 // Builds the argv for spawning `muse` as a first-class session.
 //
-// `muse` has no --session-id; it mints its own id (sqlite session-index.db).
-// Fresh: `muse --workspace <cwd>` (and optional --model/--reasoning-effort).
-// Resume: `muse resume <id>`.
+// `muse` is codex-shaped on the id axis: there is no `--session-id`, it mints its own (a row in
+// `session-index.db`), so a fresh spawn is watched for the new row and a resume names it —
+// `muse resume <uuid>`.
+//
+// Everything else is a ROOT option, and muse accepts those on either side of the subcommand
+// (`muse resume --help` says so). That is what makes the resume path below carry `--workspace`
+// rather than dropping it: `--workspace <PATH>` is what registers the policy-gated workspace tools,
+// so a resumed session started without it comes back with the conversation and without the tools
+// it had — file edits and shell in the project it was working on. The flags are emitted BEFORE the
+// subcommand, which is the form the CLI documents first.
+//
+// `--yolo` disables approval prompts and the sandbox for the run. It is the counterpart of the
+// `--permission-mode auto` grok is given and the permission mode claude cells run under: a cell in
+// the grid has no way to answer a modal approval prompt in a TUI it is not being watched.
 
 export interface MuseArgsInput {
+  /** A muse session id to resume, or null to start fresh in `workspace`. */
   resume?: string | null;
+  /** The directory whose tools the session is given (--workspace). Passed on BOTH paths. */
   workspace?: string | null;
+  /** Model override (--model), or null to use muse's own configured default. */
   model?: string | null;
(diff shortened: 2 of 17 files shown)


snakajima:
Follow-up Code Review Results: A comprehensive code review identified critical issues with error handling, memory leaks, and race conditions. See issue tracking for details. Critical fixes needed for: (1) Silent error swallowing in capture, (2) Stale claims memory leak, (3) Race condition in snapshot. All tests passing; PR ready for merge with follow-up PRs for critical issues.

snakajima:
@codex Taken — that window is real. The mapping is recorded synchronously for a resume, and a fresh session's key is a Mulmo-minted id that names no muse log, so the map-only gate held for those; what it dropped is exactly the case you name — one of muse's own ids polled as claude before (or without) a mapping, which includes a session started from a plain terminal. Both gates now, map first, then one indexed `session_log_path` lookup (57577477).

Two more things from this round, for the record:

- The duplication scan flagged the token-count guard grok's and muse's folds both carried; it is now shared as `usageCount` (7d8df37e).
- The follow-up review's other three findings — the swallowed `.catch` in the capture, the unbounded `claimed…` set, and the snapshot race — are all shapes muse INHERITED from codex's and agy's spawners, which have them identically. They are worth fixing for the three of them together rather than making muse the odd one out here.

PR #1514, opened 2026-08-06
feat(muse): GUI MCP for Muse cells, through a plugin and a process-tree session resolve

snakajima:
Muse cells reach the GUI tools their directory switched on. This is the piece the last PR said was impossible, and the reason it looked impossible was a wrong premise: Muse does have MCP, behind its own experimental flag, through a mechanism neither of the other two agents uses.

## Verified end to end

Against a real server and a real cell (`muse-bin 0.1.0-R708.1`, 2026-08-06), in a directory with only `render` registered — a cell asked what tools it had and answered:

```
mcp__plugin_mulmoterminal_render__presentChart
mcp__plugin_mulmoterminal_render__presentDocument
mcp__plugin_mulmoterminal_render__presentForm
mcp__plugin_mulmoterminal_render__presentHtml
```

…and only those: `data`, `media` and `external` stood down, as the directory's switches say they should. A `presentChart` call then wrote a real `artifacts/charts/2026/08/museworks-*.chart.json` into that session's own workspace, with the server logging all four bridges resolving to the right session id.

## Why Muse is a third shape, not a variant

**Registration is per machine.** `muse plugins install` copies the bundle into a content-addressed cache and records it for the user; `--scope project` writes nothing into the project (measured — installing from one directory left it untouched and the plugin listed from another). So the registration cannot express "this directory gets render". One `mulmoterminal` plugin therefore declares all four group servers, and the narrowing happens per session.

**Nothing reaches a plugin's MCP server but its command line.** It is started with a curated environment — measured at exactly 16 variables, all of Muse's own (`MUSE_PLUGIN_ROOT`, `PATH`, `HOME`, `TMPDIR`, …). Neither the muse process's environment nor an `env` block in the manifest arrives; both were tried, and both fail *silently*, which is the whole difficulty. So the group and the port go in argv, and the session is **asked for**.

**The session is resolved from the process tree.** `POST /api/mcp-resolve` takes the bridge's own pid, walks its ancestors, and matches them against `tmux list-panes` — whose session name is the id this server minted. That is exact: it tells two Muse cells in ONE directory apart, which nothing else could. The cwd is a fallback for a session not under tmux, and it is trusted only when it names exactly one live muse session; otherwise the answer is `null` and the bridge serves nothing. A wrong session would draw one cell's chart in another, so refusing is the better failure.

**A group the session is not entitled to serves an EMPTY toolset**, not an error. Muse starts all four servers for every session; erroring would show three broken servers in a cell that switched one group on, which reads as a broken app.

## The bug that hid all of this

`POST /api/mcp/resolve` is one segment, so `/api/mcp/:sessionId` — registered first — swallowed it and answered `400 invalid sessionId`. The bridge treats an unreachable resolve as "no session" and stands down, so the whole feature failed by serving zero tools **with nothing logged anywhere**. It survived every green test run and was caught only by running the bridge by hand against a live server. The route is now `/api/mcp-resolve`, and it logs both outcomes — an unresolved bridge is otherwise indistinguishable from an agent that was never given any tools.

## What else changed

- `server/agents/muse-mcp.ts` — generates the bundle, installs and approves it through muse's own CLI (never writing into muse's store), and stamps what it installed so an unchanged manifest spawns no subprocess at all. A changed bridge path or port rewrites the manifest, which re-installs and re-approves — the trust is keyed to the definition hash.
- `server/mcp/bridge.mjs` — `--group` / `--port` in argv, resolve-once-per-process (the promise is cached, not the result: three concurrent `tools/list` otherwise asked three times, caught by the spec), and the stand-down replies.
- `server/session/bridge-session.ts` — the pure resolution rule, plus the per-session entitlement recorded at spawn and dropped when the session is reaped.
- `server/infra/process-tree.ts`, `tmuxPanePids()` — the walk and the pane map, both bounded and both spec'd.
- The launcher form now offers Muse the same four per-group toggles as Grok and Antigravity, instead of "Muse has no GUI MCP bridge".
- README's "MCP server ids" table gains the third route; CLAUDE.md gains the section, including the two things not to re-try (per-directory install, an env var for the bridge).

## Checks

`yarn typecheck`, `yarn lint` (0 errors), `yarn test` — **9088 passed, 45 skipped**, with 40 new assertions across `muse-mcp.spec.ts`, `bridge-session.spec.ts`, `spawn-muse-mcp.spec.ts`, the muse half of `directory-mcp-ws-agent.spec.ts`, and the bridge and tmux specs. `knip` and `jscpd` report nothing new.

## Two things to know before merging

- **The flag is experimental.** `muse plugins` answers "plugins are not available in this build" without `MUSE_EXPERIMENTAL_PLUGINS=1`, and any update can rename or drop it. Every failure path here degrades to "no GUI tools" with one warning, never to a session that will not start.
- **The plugin is machine-wide.** It appears in `muse plugins list` for every Muse session on the machine, including ones MulmoTerminal did not start — those resolve to no session and serve nothing. Removing it is `muse plugins remove mulmoterminal`; the next Muse cell re-registers it.

work in mulmoterminal

## Summary by CodeRabbit

* **New Features**
  * Added Muse support for GUI MCP tools through automatic plugin setup.
  * Muse sessions now receive directory-based tool groups with per-group access controls.
  * Added automatic session discovery for reliable tool availability across workspaces and projects.
  * Muse now displays the same MCP group toggles as supported agents.
  * Sessions without authorized tools continue operating normally with an empty tool list.
  * Plugin configuration updates are applied automatically when Muse sessions start or reconnect.

* **Documentation**
  * Updated MCP setup guidance, Muse configuration requirements, tool naming, and agent comparisons.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -110,10 +110,11 @@ had no rule until now.
 
 Deliberate divergence is fine — say so in a comment with the reason, and flag it in the PR.
 
-## The GUI MCP has two server ids, and they are not meant to match
+## The GUI MCP has three server-id shapes, and they are not meant to match
 
-The same tool is called `mcp__mt__presentChart` in a workspace cell and
-`mcp__mulmoterminal-render__presentChart` in a project cell. Both are current. The branch is
+The same tool is called `mcp__mt__presentChart` in a workspace cell,
+`mcp__mulmoterminal-render__presentChart` in a project cell, and
+`mcp__plugin_mulmoterminal_render__presentChart` in a muse cell. All three are current. The branch is
 `carriesFullGuiMcp()` in `server/session/mcp-config.ts`: the workspace / single view / cell-less chat
 gets a **generated** `--mcp-config` carrying every tool under `GUI_SERVER_ID`; a project cell is
 handed **no `--mcp-config` at all** and reaches the tools through the user's own `.mcp.json` under the
@@ -150,11 +151,32 @@ So do not re-add a recogniser, however narrow, and do not "restore" the worktree
 that hole is known and accepted (a `codex` chip can occupy a worktree twice). A chip that wants GUI
 tools asks for them in the flags the user writes. Agent behaviour belongs to the Agent Picker.
 
+**Muse is the third shape and the one that breaks the pattern.** It reads neither a flag nor a file
+in the directory: MCP servers are declared by an installed PLUGIN (`server/agents/muse-mcp.ts`), and
+`muse plugins install` records one PER MACHINE — `--scope project` writes nothing into the project.
+So the registration cannot express "this directory gets render", and two things follow that nothing
+else here does:
+
+- a plugin MCP server is started with a **curated environment** — measured at 16 variables, all of
+  muse's own — so `guiMcpEnv` does not reach it and neither does an `env` block in the manifest.
+  The group and the port are argv; the SESSION is asked for, by walking the bridge's process tree
+  back to a tmux pane whose name is the session id (`server/session/bridge-session.ts`).
+- the four servers are registered for every session, and the ones a session is not entitled to
+  serve an EMPTY toolset rather than failing. Erroring would show three broken servers in a cell
+  that switched one group on.
+
+Do not "fix" that by trying to install per directory, and do not add an env var for the bridge —
+both were tried, and both fail silently by serving zero tools.
+
 The ids differ in **who owns them**, which is what decides whether a rename is free:
 
 - `GUI_SERVER_ID` (`mt`) — regenerated on every spawn, written to no file a user keeps. Ours. It is
   short because the client repeats it on **every tool name** (`mcp__<id>__`, or codex's
   `mcp-<id>-` with `-` rewritten to `_`), so the id is paid per tool, per listing, per session.
+- muse's per-group ids (`render`, `data`, …) — **ours, inside a manifest we generate**, and short
+  for `GUI_SERVER_ID`'s reason: muse builds the tool name out of the plugin id AND the capability
+  id, so `mulmoterminal-render` inside a `mulmoterminal` plugin would say our name twice in every
+  tool name. Free to rename, and a rename re-registers itself on the next spawn.
 - `toolGroupServerId()` (`mulmoterminal-render`, …) — **keys in config files users wrote**, read
   back by the launcher's per-group switch, documented in the setup guide. Renaming these breaks
   working setups with no error anywhere; it needs a migration over existing per-folder configs.
--- a/README.md
+++ b/README.md
@@ -1205,7 +1205,7 @@ it came from — `mcp__<id>__presentChart` in Claude Code, `mcp-<id>-presentChar
 also rewrites `-` in the id to `_`). So the id you register under is repeated on every tool, in
 every listing, for the life of the session.
 
-MulmoTerminal delivers the GUI MCP by **two different routes**, and they do not share an id:
+MulmoTerminal delivers the GUI MCP by **three different routes**, and they do not share an id:
 
 | | Workspace cell / single view | Project-directory grid cell |
 |---|---|---|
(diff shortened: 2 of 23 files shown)


snakajima:
Fixed a real bug this uncovered, pushed as 116e639b.

**Symptom:** MCPs enabled for a directory, Muse started, no Canvas and no MCP servers in the cell at all.

**Cause:** the sync skipped the install. It compared the manifest against a digest it had recorded when it last installed one — a record of what *we* did, not of what muse *has*. Anything that removes the plugin on muse's side (`muse plugins remove`, a cleared cache, a muse reinstall) left the two disagreeing permanently: our note said "installed", muse had nothing, every muse cell came up with zero tools, and nothing was logged because the sync returned on its first line.

**Fix:** the stamp is gone. Every spawn reads `muse plugins inspect --json` once (~100ms) and re-registers unless muse's own answer matches on all three counts — `active`, every capability `trusted_enabled` (trust is keyed to the definition hash, so a changed manifest leaves capabilities listed and never started), and the declared commands equal to ours (the bridge path or the port may have moved).

Verified against the reporter's own directory: a cell in `../mag2` with all four groups registered lists all twelve tools, and `/api/tools` reports `[data, external, media, render]` — which is what the Canvas button and the Tools pane read. Removing the plugin by hand and opening a cell reinstalls it.

snakajima:
Second report ("restarted the server, same result") — found and fixed, 0890445c.

**Measured on the machine:** the muse process the cell was attached to started at **09:01**; the plugin was written at **16:08**. The server restart replaced the server, not the session — MulmoTerminal's sessions live in tmux and outlive it, so the cell was reattached to a muse that had been running since long before the plugin existed.

Two things were wrong:

1. **The registration was gated on `!ptyWouldReattach`**, copied from agy and grok. That rule exists to stop a reattach from rewriting a config file *shared with other sessions in the directory* — muse has no such file. Its registration is machine-wide and inert on its own, so the gate protected nothing, and it broke the first-run case completely: right after this lands every muse cell IS a reattach, so nothing ever registered. Now every path registers, throttled to one `muse plugins inspect` per minute per process (one browser reload reattaches every cell at once). A failed check never starts that clock.

2. **muse reads its plugins when its own process starts**, so an already-running session cannot be given them retroactively. That is now said in the log line and in the README, because "restart the server" is the reasonable thing to try and it is not enough — the CELL has to be started again.

Verified against the reporter's own server (34567) and directory: a new muse cell in `../mag2` starts four bridges, lists all twelve tools, and `/api/tools` returns `[data, external, media, render]`.

snakajima:
Review round addressed in 89ce7ec1 — thanks, the first one was a real hole.

**[Codex P1] the cwd fallback → removed.** "The single live muse session in this directory" was treated as evidence and it is not: the plugin is machine-wide, so a muse the *user* starts in a normal terminal has the bridge too, matches no pane of ours, and would have been handed a neighbouring cell's session id and groups — enough to draw into that Canvas and write artifacts under that session. The rule is now **descent only**: the bridge's ancestors must contain that session's tmux pane, or (persistence off) the pty this server spawned for it. `cwd` is read for the log line and nothing else. New spec case pins the stranger-in-the-same-directory rejection.

**[Codex P2] background muse chats.** `registeredGuiMcpGroups` was resolved only for agy and grok, so `spawnBackgroundChat` on muse got an empty entitlement — no GUI tools in a workspace that had them registered. Muse is in that condition now.

**[Codex P2] the synchronous CLI.** `spawnCapture` had no timeout and runs on the `/ws/muse` path, so a hung `muse plugins` would block the event loop — every request and every open terminal, not just the spawn. It now takes and defaults a timeout; the muse calls pass 20s against a measured ~0.3s.

**[Codex P2] verify the plugin exists rather than trusting the stamp.** Already fixed in 116e639b, before this review ran — the stamp is gone and every spawn reads `muse plugins inspect`. It was found the hard way: the reporter's cells came up with no tools because of exactly this.

**[CodeRabbit] two stale comments.** `guiMcpAgents.ts` still said the session environment carries the groups (it cannot — that is the whole reason the resolve exists), and `bridge.mjs` named `/api/mcp/resolve`, the path that 400s. Both corrected.

Not re-run live: this round only *removes* a resolution path, and the pane path it leaves is the one already verified end to end on the reporter's own server and directory. `yarn typecheck`, `lint` (0 errors), `yarn test` — 9094 passed.

snakajima:
CODEX VERDICT addressed in 3f829439 — the reattach path was genuinely wrong.

A tmux-only reattach (a reconnect after a server restart) has no live pty either, so it took the same branch as a spawn and resolved `registeredGuiMcpGroups` from the **request's** cwd — and such a reconnect often carries no `?cwd=`, which resolves to the default workspace. For agy and grok that only produced a file write suppressed further down; muse records the result as the session's entitlement, so another directory's switches could have become this session's.

- The lookup now runs against where the session actually **is**: a live pty's cwd, else the recorded `sessionCwd`, and only then the request's.
- muse re-records nothing for a session it is merely reattaching — that muse is already running with what it read at its own start, so a new entitlement could only move a session that cannot act on it. The one exception is a session this process has no record of at all, which is precisely the server-restart case: the record is then the only answer the bridge can be given.

Both pinned by specs (`directory-mcp-ws-agent.spec.ts` for the directory the groups are read against, `spawn-muse-mcp.spec.ts` for the two reattach cases). `yarn typecheck`, `lint` (0 errors), `yarn test` — 9097 passed.

PR #1532, opened 2026-08-07
feat(roster): the picker's agent mark on every row, and none on a row that runs no agent

snakajima:
## What

The cockpit roster row wore the agent's **name** in a bordered pill (`codex`, `agy`, `grok`, `mu`) and only for a non-Claude agent. It now wears the same **drawn mark** the Agent Picker and the rate-limit gauge use (`AgentMark.vue`), so one agent looks the same wherever it is named; a shell row gets the picker's `terminal` symbol.

## Why the roster had to learn what a row is NOT

Marking every row makes the default visible, which is where the trap was. `Cell.agent` is absent for Claude **and** for every cell that is not an agent session at all — a launcher chip, a run-command cell, a cell nobody has launched in yet — so the roster's `c.agent ?? "claude"` would have put Anthropic's burst on all of them: a `yarn dev` chip claiming to be a Claude session.

`rosterAgent()` in `GridView.vue` asks the cell's **kind** first:

- a launcher or a command cell is `"shell"` — whatever command line it names, which is the same answer the server records (`spawnLauncherPty`), and nothing here parses that text;
- a session cell keeps its agent (`c.agent ?? "claude"`);
- a cell running nothing answers `null`, and the header marks it with nothing at all — the status dot and the directory already identify the row.

`CockpitRow["agent"]` is `SessionAgent | null` rather than `string`, so the "runs nothing" case is a value the type carries instead of a default someone re-adds.

## Also

- The mark is named for assistive tech (`role="img"` + `aria-label`), which cannot read a drawn glyph; the hover title says the same thing.
- The Material Symbol's ligature text is hidden the way the picker hides its own, so the row's text does not read "terminal".

A custom agent still shows Claude's burst — `TerminalCell` resolves a custom pick to `agent: "claude"` on purpose (a wrapper decides which command line starts Claude Code, not what the session is), so that is the existing model rather than something this PR changes.

## Test plan

- `test/src/components/CockpitHeader.spec.ts` — a mark for every built-in agent and for shell, none when the row runs nothing, the aria name, and no agent NAME printed beside it.
- `test/src/components/GridView.spec.ts` — a launcher row says `shell`, a codex row keeps `codex`, a plain session row says `claude`, driven from persisted state the way a reloaded grid gets it.
- `yarn typecheck`, `yarn lint` (no new warnings), `yarn test` — 9175 passed.

work in mulmoterminal

## Summary by CodeRabbit

* **New Features**
  * Added accessible icons for active agents and terminal sessions, with labels for assistive technologies.
  * Added support for sessions without an assigned agent.
  * Improved grid views to distinguish agent sessions, terminal cells, and inactive cells.

* **Bug Fixes**
  * Corrected roster information so launcher and session entries display the appropriate agent details.

### DIFF
--- a/src/components/CockpitHeader.vue
+++ b/src/components/CockpitHeader.vue
@@ -13,12 +13,15 @@ import { HOVER_TIP_ID, useHoverTipAnchor } from "../composables/useHoverTip";
 import { phaseDisplay, WORK_WORD, type PrPhase, type WorkPhase } from "./rosterPhase";
 import { textTip } from "./tipContent";
 import type { AttentionStatus } from "./attentionStatus";
-import { agentBadge } from "../../common/sessionAgent";
+import AgentMark from "./AgentMark.vue";
+import { isTerminalAgent, type TerminalAgent } from "../../common/sessionAgent";
 
 const props = withDefaults(
   defineProps<{
     status: AttentionStatus;
-    agent: string;
+    // What the row is running, or null when nothing does yet (an empty cell) — see rosterAgent()
+    // in GridView.vue for how a cell answers this.
+    agent: string | null;
     cwd: string | null;
     home: string | null;
     headerColor: string | null;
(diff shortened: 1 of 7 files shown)


snakajima:
> Add regression coverage for `RunCommand` cells: the new classifier's command branch must produce the shell mark, but the current test covers only launchers.

Fair — and the GridView test could not have reached it: a run-command cell is ephemeral, so `parseGridState` (which keeps only cells with a session) cannot express one, and driving it from persisted state was never going to cover that branch.

The classifier now lives in `src/components/rosterAgent.ts` and is pinned directly over all five kinds a cell can be, in `test/src/components/rosterAgent.spec.ts`:

- a session cell with no `agent` → `claude` (the absent case that really is Claude);
- a session cell with one → `codex` / `grok`;
- a launcher, both shapes (`{shell:true}` and a configured `{index,label}`) → `shell`;
- a run-command cell, both sources → `shell`. The `button` source carries an `agent` of its own — the session context the button was resolved against, not what the cell runs — so it is asserted to stay `shell` even when that field says `codex`;
- a cell that has launched nothing → `null`, no mark.

The GridView test stays as it was: it pins the wiring from cell to roster row, not the rule.

Issue #1533, opened 2026-08-07
bug: terminal view reattaches to the wrong backend terminal process


snakajima:
## What happened

While working across the last few PRs, a terminal cell in the web UI ended up attached to the **wrong backend terminal process** — the view displayed and drove a different PTY session than the one it should. This was observed **several times**, so it is recurring, not a one-off.

It is not yet known whether the bug is specific to non-Claude agents (Codex / Antigravity / Grok / Shell / custom agents) or affects Claude cells too.

## What I expected

A terminal view, once bound to a session, stays bound to that exact backend PTY across reconnects, reloads, paging, zoom/roster transitions, and server restarts — or fails visibly rather than silently binding to a different session.

## Steps to reproduce

Not yet isolated — intermittent. Observed during normal multi-session development work (many concurrent cells) over the last few PRs. A read-only investigation of the attach/reattach paths (server session registry, client binding, recent PR diffs, resume/tmux paths) is in progress; findings and a concrete repro/mechanism will be posted here as follow-up comments.

Candidate areas under investigation:

- server-side session lookup on WebSocket (re)attach — any non-unique keying (cwd/agent) or id reuse after reap
- client-side binding of a cell to a session by index/order rather than stable id (grid paging, zoom teleport, cockpit roster)
- resume/restore after server restart — tmux pane-name → session-id mapping collisions
- per-agent asymmetries in session tracking (#1449 per-agent conversation lists, `attached` flag)

## Related (not duplicates)

- #1207 — second terminal in a worktree connected to the in-progress session (closed; the chip-side limit was later deliberately removed)
- #1443 — antigravity tmux reattach rewrote shared config for sessions it does not speak for (closed)
- #957 / #1496 — reload/reattach display issues (different symptom: blank/missing, not wrong session)

## Environment

- mulmoterminal 4.7.1 (git install, commit f5062654), 52 live sessions at time of filing
- Node v22.18.0 / Darwin arm64 (macOS) / zsh
- tmux 3.7b / Claude Code 2.1.224 / codex-cli 0.147.0
- Browser: Chrome (desktop)

## Attachments

None yet — no screenshot was captured at the moments it happened. Investigation results to follow in comments.

snakajima:
## Investigation synthesis

[...]

PR #1534, opened 2026-08-07
fix(session): resume a running conversation under its surviving key; occupancy sees tmux survivors (#1533)

snakajima:
Fixes the two top-priority mechanisms confirmed on a live instance in #1533: a conversation whose backend was still running in tmux read as free, and resuming it started a **second backend on the same conversation** — after which a cold reconnect could come back on either process, i.e. "the terminal view reattached to a wrong backend terminal process".

## What was happening

The probe on #1533 found the double-session state in all three non-Claude agents' conversation logs (codex ×2, antigravity ×2, muse ×1, Aug 5–7), and the first backend of the freshest pair still running, detached and invisible to the registry — along with 19 more such sessions. Two gaps produced it:

1. **The launcher's resume row emitted the row's own id** — the agent's conversation id — even when the server had said (via `runningKey`, #1467) that the conversation was still running under a different, minted key. "Resume it here" then spawned `codex resume <conv>` in a fresh `mt-<convId>` tmux session while `mt-<mintedKey>` kept running the same rollout.
2. **`dirSession` could not see a backend that survived its pty** (#1496 keeps the tmux session but drops the `ptys` entry; a server restart leaves the same state). For codex/agy/muse there is no transcript pass to fall back on, so the worktree read as unoccupied: the one-session-per-worktree refusal (#1207) stopped firing, and the row offered the newest transcript instead of the running session.

## The fix

- `CellLaunchForm.resume()` reattaches `s.runningKey ?? s.id` — picking a running conversation up means reattaching the key it runs under. For Claude and grok the two are the same string, so nothing changes there.
- `dirSession` gains a **survivor pass** (`survivorCandidates`, pure and pinned by spec): sessions in `runningSessionKeys()` that `ptys` no longer knows, tied to their directory through the codex/agy/muse conversation logs, ranked live. Claude transcripts whose session survives in tmux are likewise upgraded to live, so the running conversation outranks a merely newer transcript.
- `runningSessionKeys()` is passed into `dirSession` alongside `tmuxCounts` — one `list-sessions` call per list, not per worktree row.

Behavioural consequence, intended: a worktree with a surviving codex/agy/muse backend now **refuses a fresh session and offers resume** (restoring #1207's rule for the post-#1496 state), and the resume reattaches the surviving process.

## Second commit: the client slot layer (a live sighting during this work)

While this PR was being written the bug struck again with a different mechanism: one Claude cell open, the server updated under a dev (HMR) page, and the cell came back showing an earlier **shell** session. Slot keys are positional (`cell-<uid>`, renumbered from array position on every grid parse) while the slot map outlives components — the reloaded GridView handed the Claude cell the ghost slot a **closed** shell cell had left behind, and `attach()` reused it without comparing sessions, then persisted the wrong pairing via the `onSession` replay.

- `attach()` on an existing slot now **reconnects** when the view names a session other than the slot's, instead of silently reusing the socket and buffer (with a console warning naming both ids). A view with no opinion (`sessionId: null`) keeps the reuse — that replay is the deliberate repair path for a parent that never learned its id.
- `GridView.onClose` tears the slot down with the cell: the `terminal-close` shortcut and a launcher cell's close used to drop the cell while its slot kept an open socket — a PTY that read as attached forever, never reaped, and the ghost the key collision hands out. `TerminalCell`'s own close button already tears down first; the new `slotLive` check keeps that path a no-op.

## Third commit: refuse invalid session ids, and reap the unreachable (`mt-undefined`)

Probing the live instance also turned up a tmux session literally named **`mt-undefined`** — a spawn had reached tmux with a session id that was never set. Such a name is a shared bucket (every spawn with the same defect attaches the same pane) and it is unreachable by every route: attach, resume and terminate all validate the id, so nothing could ever open it — or clean it up.

- `ptySpawn` now refuses an id `SESSION_ID_RE` rejects, at the one choke point every spawner passes — the defect becomes a failed launch with a message instead of a silently shared terminal. Probe ids keep passing (UUID-shaped by construction).
- `reapableTmuxSession` gains a `validId` fact: an unparseable id reaps at once — no recency can make it reachable, and the sweep is the only thing that can ever end it. Still spared while a terminal is attached, and the `sessionIdleReapDays: 0` off switch still holds. The sweep takes the fact injected (like `liveHere`) so its decision stays pure; the Settings list's `reapable` column feeds it too, so the row now honestly says the server will end it.

## Fourth wave: the rest of the synthesis hardening

- **Watcher guards** — muse's spawn watcher attributes only an unambiguous sole row (it took the FIRST new row, and its `before` snapshot can come back empty from a busy sqlite); codex's sole-rollout branch now requires the rollout's own cwd to agree (compared canonically), closing the cross-directory claim; all three watchers claim synchronously with the selection instead of in the caller's `.then`; and the registry seeds the muse/agy claimed sets from their persisted logs at hydration.
- **Per-session-id connect serialization** — every agent endpoint's admission (claude, launch, codex, antigravity/grok/muse) runs in a per-id turn (`infra/serialize-per-key`), and `settledEntry` re-reads the live entry AFTER the admission awaits: the resolve-time snapshot can be a corpse the idle reap killed mid-admission (reattaching it wired the browser to a dead pty while the next connect spawned a fresh one under the same id). That case closes the socket plainly so the client reconnects and resolves afresh.
- **Agent-kind assertion** — `wrongEndpointReason` refuses a live PTY whose recorded agent does not match the endpoint, loudly (an error frame, so the client stops retrying and the user is told to open the session from its own agent's cell). `ptys` is one table for every agent and nothing compared the two, while `asTerminalAgent()` coerces any unrecognised persisted agent to "claude".

Note for the reviewer: the last three commits interleave two sessions' work on the same tree; the FINAL state is the thing to review, and it is typecheck-clean with the full suite green.

## Not in this PR (tracked in #1533)

Lower-priority residue from the synthesis: `parseGridState` still keys slots positionally (guarded now by the attach mismatch reconnect) and does not dedupe by session id; `dirSession`'s mtime ranking for issue-work's `resumed` outcome; positional launcher/script identity. 

## Verified

- `yarn typecheck`, `yarn lint` (0 errors), `yarn build`, full `yarn test` — 9217 passed.
- New specs: `survivorCandidates` selection + rank interplay (6 cases); "resumes under the surviving key, not the row's own id" for the launcher row; the slot-inheritance guard (3 cases, including the no-opinion repair path); the close-path cleanup (3 cases, including the shortcut and the already-torn-down no-op); the ptySpawn id guard (refusal + probe-id pass); and the unreachable-id reap (immediate, attach-spared, off-switch-respected).

Addresses the confirmed mechanism in #1533 — deliberately not a closing keyword, so the issue stays open for the remaining hardening.

work in mulmoterminal

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved resuming of active sessions across supported agents.
  * Prevented stale or mismatched terminal connections when reusing slots.
  * Closing a live terminal now reliably terminates its session and avoids duplicate cleanup.
  * Reduced duplicate session attachments during simultaneous launches.
  * Improved detection and cleanup of invalid or abandoned sessions.
  * Added safeguards against attaching to sessions owned by another endpoint.
* **Reliability**
  * Session status now reflects running conversations more accurately, including sessions restored from conversation history.

### DIFF
--- a/server/agents/antigravity-session.ts
+++ b/server/agents/antigravity-session.ts
@@ -61,7 +61,7 @@ const delay = (ms: number): Promise<void> => new Promise((resolve) => setTimeout
 export async function watchForAntigravitySession(
   root: string,
   before: ReadonlySet<string>,
-  opts: { pollMs?: number; maxWaitMs?: number; isCancelled?: () => boolean; claimed?: ReadonlySet<string> } = {},
+  opts: { pollMs?: number; maxWaitMs?: number; isCancelled?: () => boolean; claimed?: Set<string> } = {},
 ): Promise<string | null> {
   const pollMs = opts.pollMs ?? WATCH_POLL_MS;
   const deadline = Date.now() + (opts.maxWaitMs ?? WATCH_MAX_WAIT_MS);
@@ -71,5 +71,9 @@ export async function watchForAntigravitySession(
     await delay(pollMs);
     result = pickFreshAntigravitySession(root, before, opts.claimed);
   }
+  // Claimed here, synchronously with the selection, not only in the caller's `.then`: two watchers
+  // awaiting the same poll interval would otherwise both pick the same conversation before either
+  // claim landed (#1533). The caller's own add stays and is idempotent.
+  if (result) opts.claimed?.add(result);
   return result;
 }
--- a/server/agents/codex-session.ts
+++ b/server/agents/codex-session.ts
@@ -2,6 +2,7 @@ import { existsSync, readdirSync, readFileSync } from "node:fs";
 import path from "node:path";
 import os from "node:os";
 import { isRecord } from "../../common/isRecord.js";
+import { canonicalPath } from "../infra/canonical-path.js";
 
 const ONE_DAY_MS = 24 * 60 * 60 * 1000;
 const ROLLOUT_RE = /^rollout-.*\.jsonl$/;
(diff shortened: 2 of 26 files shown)


snakajima:
All three review findings addressed in `c154eceb`:

1. **Grok survivors** — a `grokSurvivorCandidates` pass in `dirSession`: grok keeps no conversation log (its session key IS its conversation id, store partitioned by directory), so a surviving key is tied to a directory by probing the store, both spellings like the transcript pass. Pinned by 4 new spec cases.
2. **Cross-process log staleness** — `refreshAgentConversations()` re-reads the three conversation logs before survivor matching: whole-file (they are a few KB, append-only) and idempotent via the existing `hydrateAgentConversationInto` fold, with own-process writes still protected by `writtenIds`.
3. **Unlearned-slot reuse** — the attach guard now treats `knownSessionId === null` as a mismatch too when the view names a concrete session (that state belongs to a fresh launch whose socket may still be connecting); only a view with no opinion keeps the reuse. Pinned by a new spec case.

Verified: lint 0 errors, typecheck clean, full suite 9222 green.

PR #1550, opened 2026-08-08
fix(#1549): hold the worktree buttons while they work, and say why they failed

### LINKED ISSUE #1549: fix(worktree): 「+ New worktree」に多重実行ガードも進行表示も無く、連打した回数だけ worktree が作られる（同じフォームの #1527 / #1447 が未適用）
## What happened

新規セルの起動フォーム（worktree セクション）で **「+ New worktree」を押しても、見た目が一切変わりません**。
スピナーも出ず、ボタンが無効にもならず、押した直後と何も変わらないので**作成が始まったのかどうか分かりません**。

そのため反応が無いと思ってもう一度押せてしまい、**押した回数だけ worktree が作られました**。
`agent/<slug>`、`agent/<slug>-2`、`agent/<slug>-3` … と、内容の同じ作業ツリーが並びます。
当方は約 33,000 ファイルのモノレポなので、1 本あたりフルチェックアウトで **6 秒前後**かかります。
その間ずっとボタンは押せる状態のままです。

（重複してできた分は報告者が削除済みのため、いま `git worktree list` には残っていません）

## What I expected

- 作成中はボタンが無効になる、またはスピナー等で「処理中」と分かる
- 処理が終わる前の 2 回目以降のクリックは無視される

## Steps to reproduce

1. グリッドの空きセルで新規セルの起動フォームを開く
2. 既定の作業ディレクトリ以外のディレクトリを選び、worktree セクションを出す
3. タスク名を入力して「+ New worktree」を押す
4. 見た目が変わらないので、続けて 2〜3 回押す
5. → `agent/<slug>`, `agent/<slug>-2`, `agent/<slug>-3` が作られる

大きいリポジトリほど 1 本の作成に時間がかかるため、押し増す余地が広くなります。

## 原因

作成ハンドラに **in-flight 判定がありません**（`dist` のバンドルより、名前は最小化されたもの）:

```js
async function createWorktree() {
  const repoDir = selectedDir.value, task = taskName.value.trim();
  if (!repoDir || !task) return;
  try {
    const res = await fetchJson(`/api/worktrees/create`, { method: `POST`, ... });
    if (!res.ok) return;
    const body = await readJson(res);
    if (typeof body.path === `string`) { taskName.value = ``; await selectDir(body.path); emit(`start`, body.path); }
  } catch {}
}
```

ボタン側の `disabled` も入力欄が空かどうかだけを見ています:

```js
{ "data-testid": `wt-start`, class: `... cursor-pointer ... hover:bg-hover hover:text-fg`,
  disabled: !taskName.value.trim(), onClick: createWorktree }
```

入力欄はクリアされるのが**成功レスポンスを受け取ったあと**なので、往復している数秒間はテキストが残ったまま
＝ボタンは有効なままです。

サーバ側も重複を弾きません。`server/git/worktrees.ts` の `serializeCreate()` は**直列化するだけ**で、
`uniqueBranch()` が空いている `-2` / `-3` を採るため、連打分がすべて別の worktree として成立します。

## 同じ起動フォームの隣のボタンは、すでに対策済みです

同フォームのフォルダ選択ボタンが呼ぶヘルパーには、**多重実行ガードもエラー表示も入っています**
（#1527 と #1447 の修正）:

```js
const pending = ref(false);
async function pickPaths(opts = {}) {
  if (pending.value) return { paths: [], error: null };     // ← 多重実行ガード（#1527）
  pending.value = true;
  try {
    const res = await fetch(`/api/pick-file`, { ... });
    const body = await readJson(res);
    return res.ok ? { paths: ..., error: null }
                  : { paths: [], error: errorText(body, res.status) };  // ← サーバの理由を返す（#1447）
  } catch (e) { return { paths: [], error: `Could not ...` } }          // ← 例外も返す
}
```

issue 起点の「これをやる」も同様で、#1219 の本文に
「連打では起きません（デスクトップは `starting`、スマホ側 UI も送信中は無効化する想定）」とあります。

**つまり worktree 作成の 3 つの入口のうち、ランチャーでタスク名を打つ経路だけが、
ガードも進行表示もエラー表示も持っていません。**

## 副次的に: 失敗も完全に無音です

上の `if (!res.ok) return;` と `catch {}` により、サーバが 500 を返しても画面には何も出ません。

実際、当方は**ローカルに base ブランチ `main` が無い**状態で
`git worktree add -b agent/x <dir> main` が `fatal: Not a valid object name: 'main'.` (exit 255) で落ち、
ルートが 500 を返していましたが、UI 上は「押しても何も起きない」としか見えませんでした。
原因にたどり着くまで `dist` のバンドル解析が必要で、**この無音と、上の「連打できる」は同じ一行から来ています**。

（なおその `main` 不在は、`main` を checkout していた worktree を「ブランチも削除」付きで消すと起きます。
`removeWorktree()` が `git branch -D <その worktree のブランチ>` を実行するためです。
`git branch main origin/main` で復旧させたところ、同じ作成が 6.2 秒で成功しました）

## ついでに: disabled のスタイルが当たっていません

上記のとおり `disabled` は付きますが、class は `cursor-pointer` と `hover:bg-hover hover:text-fg` を
含んだままです。タスク名が空のときボタンは**押せるように見えて無反応**で、これも「壊れている」の
印象に寄与しました。

## 提案

- 作成ハンドラに pending フラグを入れ、処理中はボタンを `disabled` にする（`pickPaths` と同じ形）
- 処理中であることを見た目に出す（スピナー / ラベル変更）
- 失敗時はサーバが返している `error` を表示する（#1447 と同じ扱い）
- `disabled` 時に `cursor-not-allowed` / 減光を当て、hover を無効化する

## Environment

- mulmoterminal 4.7.2（npx 起動 / `/api/update-status`: `{"ready":true,"install":"npm","version":"4.7.2","latest":null}`）
- Node v24.16.0
- Ubuntu 22.04.2 LTS / Linux aarch64（ヘッドレスのリモート機に SSH 接続、ブラウザは手元の別マシン）
- Google Chrome 150.0.7871.182（公式ビルド, arm64）
- bash / tmux 3.2a / git 2.34.1 / gh 2.31.0
- claude 2.1.224 / codex-cli 0.144.1
- 対象: 約 33,000 ファイルのモノレポ。管理 worktree を 6〜8 本併用

## Attachments

なし（症状が「見た目が変わらない」ため、静止画に写るものがありません）

isamu:
## Summary

`+ New worktree` in the launcher POSTed `/api/worktrees/create` and changed **nothing** on screen for the whole round trip. Its only `disabled` test was "is the task field empty", and the field is cleared once the **answer** lands — so on a large repository (the reporter's is ~33,000 files, ~6s per checkout) the form was identical to the one before the click and the button stayed live. Each extra press succeeded, because `uniqueBranch` takes the next free suffix: `agent/<task>`, `-2`, `-3`.

The same line hid the failure. `if (!res.ok) return;` plus `catch {}` meant a 500 showed nothing at all — the reporter's actual cause (`git worktree add … main` with no local `main`) was only findable by reading the shipped `dist` bundle.

This adds **one guard for the whole worktree section** and gives every control in it a visible working state and a readable failure.

| where | what a second click used to do |
|---|---|
| `+ New worktree` | **a second worktree** for the same task (the issue) |
| a worktree row | a second `emit("start")` — the row waits on up to four `claude mcp add` calls first |
| the delete beside a row | a second `git worktree remove` at the same path |
| the running cell's `Discard & remove` | terminated the pty again and removed twice |

## Items to Confirm / Review

1. **The guard is exclusive across the section, not per button.** Creating holds the delete buttons and the rows too. Reasoning: they all shell out to git in one repository, where two commands contend on the index lock; and "the form is busy" is a state a user can read, where "these two are live and that one is not" is not. Same shape as `useIssueStart`'s `starting` (#1219) and `useSessionStop`'s `stopping`.
2. **Disabled cursor.** The issue suggested `cursor-not-allowed`; I used `disabled:cursor-default disabled:opacity-40` to match the two buttons immediately beside it in the same form (`cell-dir-go`, `cell-dir-pick` from #1527). Happy to switch all three to `cursor-not-allowed` instead if the "refused" reading is preferred — but they should agree.
3. **The server's 500 message is unchanged and still generic**: `could not create the worktree (is this a git repo?)`, which is misleading in the reporter's case — it *was* a repo, the base branch was missing. Surfacing git's real stderr means changing `git()` (which drains and discards stderr deliberately) and `createWorktree`'s `null` return, read by every caller and a dozen specs. Left as a self-contained follow-up; the UI now shows whatever the server does say.
4. **Server-side de-duplication is deliberately not added.** `agent/<task>` + `agent/<task>-2` from two intentional creates is documented behaviour with a spec pinning it.
5. `TerminalCell`'s `Cancel` is now also held during a removal. The removal has already terminated the pty by then, so cancelling would leave the confirmation dismissed over a half-removed worktree.

## User Prompt

> https://github.com/receptron/mulmoterminal/issues/1549
> たしかにそうだ。ボタンロックしてloading spinner みたいなのをまわしたほうがよいかも。ここ以外でも必要そうなのは直してね

## What ships

**`src/composables/useBusyAction.ts` (new)** — the rule in one place. `busy` holds the *key* of the action in flight, so the pressed control can spin while the others are merely disabled; `run` drops any call arriving while it is set, and releases in `finally` so a failure does not leave the form dead.

**`CellLaunchForm.vue`**
- create / open-row / delete all go through `runWorktreeAction`
- the acting control shows `progress_activity` (the spinner the `cell-dir-loading` row already uses); the create button's label becomes `Creating…`
- new `wt-error` line carries the server's own sentence — `{ error }` from create, `{ ok:false, reason }` from remove — and is dropped when the field moves to another repository, since the failure was about the previous one
- `wt-start` / `wt-del` get `enabled:hover:` + `disabled:` styling, so an empty task name no longer leaves a button that looks pressable and does nothing
- the delete confirmation is not raised at all while another action runs — a dialog answered "yes" for work that is then dropped is worse than a button that does not respond

**`TerminalCell.vue`** — `Discard & remove` runs through the same composable and reads `Removing…`; it, `Cancel` and the error-state `Retry` are held.

**`cellChromeRules.ts`** — `worktreeRequestFailure(body, status)` reads the routes' two refusal shapes; `dirty` and `not-managed` join the reason table (remove answers them; nothing rendered them before).

## Verification

`yarn format` / `lint` / `typecheck` / `build` / `test` (9314 passed) all green.

Beyond the specs, driven in a **real browser against a real dev server** (scratch `HOME` + `MULMOTERMINAL_HOME`, a throwaway git repo, the create/remove routes delayed 3.5–4s to stand in for the 33k-file monorepo):

- **create**: button `disabled=true`, label `progress_activity / Creating…`; four further clicks plus three Enter presses in the (never-disabled) task field over the 4s window → **1 request**, and `git worktree list` afterwards shows exactly one `agent/fix-login`, no `-2`
- **failure**: with the route returning `{"error":"fatal: Not a valid object name: 'main'."}`, that exact sentence appears under the section, the button is released, and the task name is kept for the retry
- **remove**: delete icon becomes the spinner, the row and the create button are held, four clicks → **1 request**, row gone afterwards
- no `pageerror` / console errors throughout

Closes #1549

work in mulmoterminal3

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented duplicate worktree creation and removal requests during slow operations.
  * Disabled conflicting controls and added progress indicators while actions are running.
  * Improved worktree error messages, including server-provided refusal details and network failures.
  * Cleared outdated errors when switching repositories or directories.
  * Preserved task input after unsuccessful worktree creation.
* **Improvements**
  * Resume actions now retain the selected agent and active session where available.
* **Tests**
  * Added coverage for concurrency protection, loading states, validation, retries, and error handling.

### DIFF
--- a/plans/fix-1549-worktree-button-guard.md
+++ b/plans/fix-1549-worktree-button-guard.md
@@ -1,0 +1,82 @@
+# fix #1549 — a slow button that never says it is working gets pressed again
+
+## What happens
+
+`+ New worktree` in the launcher POSTs `/api/worktrees/create` and changes **nothing** on screen
+while `git worktree add` checks the tree out. On the reporter's ~33,000-file monorepo that is ~6
+seconds of a form identical to the one before the click, with the button still enabled — because
+its only `disabled` test is "is the task field empty", and the field is cleared **after** the
+response lands. Three presses produced `agent/<slug>`, `agent/<slug>-2`, `agent/<slug>-3`:
+`uniqueBranch` takes the free suffix, so every press succeeds.
+
+The same line hides the failure. `if (!res.ok) return;` plus `catch {}` means a 500 shows nothing
+at all — the reporter's real cause (`git worktree add … main` with no local `main`, exit 255) was
+only found by reading the `dist` bundle.
+
+The neighbouring controls in this very form already have the rule: `pickPaths` refuses a second
+call while one is in flight (#1527) and returns the server's sentence (#1447); the `/prs` issue row
+holds itself with `starting` for exactly this reason (#1219). Worktree creation is the entry point
+that got neither.
+
+## The rule, and everywhere it was missing
+
+**A click that shells out must hold itself until it settles, and must say why it failed.** Three
+handlers in the launcher and one in the running cell were missing it — all of them slow (`git
+worktree add` / `remove`, up to four `claude mcp add` calls), all of them silent:
+
+| where | what a second click did |
+|---|---|
+| `CellLaunchForm` `+ New worktree` | **a second worktree** for the same task (#1549) |
+| `CellLaunchForm` worktree row | a second `emit("start")` — the launcher awaits `syncMcpGroupsInto` first |
+| `CellLaunchForm` delete (`wt-del`) | a second `worktree remove` at the same path |
+| `TerminalCell` `Discard & remove` | terminates the pty twice, removes twice |
+
+## What ships
+
+**`src/composables/useBusyAction.ts` (new)** — the rule in one place. `busy` holds the KEY of the
+action in flight (so the pressed control can spin while the others are merely disabled) and `run`
+drops any call arriving while it is set. Exclusive across keys, not per key: these all shell out to
+git in one repository, and `useIssueStart` / `useSessionStop` already hold themselves the same way.
+
+**`CellLaunchForm.vue`**
+
+- all three handlers go through `runWorktreeAction`; `wt-start` and `wt-del` are `disabled` while
+  any of them runs, and the acting control shows `progress_activity` (the spinner the
+  `cell-dir-loading` row already uses) with the label changing to `Creating…`
+- `wt-error` under the section carries the server's own sentence — `{ error }` from create,
+  `{ ok:false, reason }` from remove — the treatment `cell-dir-pick-error` got in #1447
+- `wt-start` gets `enabled:hover:` / `disabled:cursor-not-allowed disabled:opacity-40`, so an
+  empty task name no longer leaves a button that looks pressable and does nothing
+- the delete confirm is not raised at all while another action runs, or the user would answer a
+  dialog for work that is then dropped
+
+**`TerminalCell.vue`** — `Discard & remove` runs through the same composable; the button reads
+`Removing…` and both it and the retry button are held.
+
+**`cellChromeRules.ts`** — `worktreeRequestFailure(body, status)` reads the route's two refusal
+shapes; `dirty` and `not-managed` join the reason table (remove answers them; nothing rendered
+them before).
+
+## Deliberately not done
+
+The route's 500 body is `could not create the worktree (is this a git repo?)` whatever went
+wrong — misleading in the reporter's case, where it **was** a repo and the base branch was
+missing. Surfacing git's own stderr means changing `git()` (which drains and discards stderr on
+purpose) and `createWorktree`'s `null` return, which every caller and a dozen specs read. Out of
+scope here; the UI now shows whatever the server says, so improving that message is a
+self-contained follow-up.
+
+Server-side de-duplication of concurrent creates is also left alone: `agent/<task>` and
+`agent/<task>-2` from two deliberate creates is the documented behaviour, pinned by a spec.
+
+## Files
+
+| file | |
+|---|---|
+| `src/composables/useBusyAction.ts` (new) | one-at-a-time guard, keyed |
+| `src/components/CellLaunchForm.vue` | three handlers guarded, spinner, `wt-error`, disabled styling |
+| `src/components/TerminalCell.vue` | `removeAndClose` guarded + `Removing…` |
+| `src/components/cellChromeRules.ts` | `worktreeRequestFailure`, two more reasons |
+| `test/src/composables/useBusyAction.spec.ts` (new) | second call dropped, released after a throw |
+| `test/src/components/CellLaunchForm.spec.ts` | one POST per press, spinner, error text |
+| `test/src/components/cellChromeRules.spec.ts` | the two refusal shapes |
--- a/src/components/CellLaunchForm.vue
+++ b/src/components/CellLaunchForm.vue
@@ -469,10 +469,12 @@ const reportWorktreeFailure = (repoDir: string | null, message: string): void =>
 // worktree may well appear a moment later. Saying otherwise sends the user to look at their network.
 // Reachable — SLOW_COMMAND_TIMEOUT_MS is 60s, and a checkout large enough to make this bug worth
 // fixing is a checkout that can exceed it.
+const errorText = (e: unknown): string => (e instanceof Error ? e.message : String(e));
+
 const requestFailureText = (e: unknown): string =>
   e instanceof DOMException && e.name === "AbortError"
     ? "Timed out waiting for git — it may still finish. Re-select this directory to see."
-    : `Could not reach the server: ${e instanceof Error ? e.message : String(e)}`;
+    : `Could not reach the server: ${errorText(e)}`;
 
 // Create a fresh worktree for the typed task and start the selected agent in it.
 //
--- a/src/components/CellLaunchForm.vue
+++ b/src/components/CellLaunchForm.vue
@@ -456,7 +456,23 @@ const removeKey = (w: Worktree): string => `remove:${w.path}`;
 // looked identical — and the difference was only findable by reading the shipped bundle.
 const worktreeError = ref<string | null>(null);
 
-const unreachable = (e: unknown): string => `Could not reach the server: ${e instanceof Error ? e.message : String(e)}`;
+// Only if the form is STILL pointed at the repository the action was for. The directory field stays
+// editable for the whole round trip, so a failure reported under whatever the user typed meanwhile
+// is a sentence about somewhere else — #1372's rule, applied to the error line rather than the rows.
+// A successful create deliberately does not ask: the worktree was made because it was asked for, and
+// launching in it is the click being honoured, not a stale answer.
+const reportWorktreeFailure = (repoDir: string | null, message: string): void => {
+  if (isSameDirPath(targetDir.value, repoDir)) worktreeError.value = message;
+};
+
+// A timeout is NOT "could not reach the server": the request landed and git is still working, so the
+// worktree may well appear a moment later. Saying otherwise sends the user to look at their network.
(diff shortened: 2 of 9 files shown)


isamu:
## Review triage — iteration 1

**Applied** (`23d6729b`) — Codex P1 and CodeRabbit's Major are the same real hole, and both are right: `requestRemove()` terminates the pty *before* it calls the route, so every exit from the confirmation was a promise the cell could not keep.

- guard in `cancelClose()` rather than on the button, because **Escape reaches it too**
- `:disabled` on `ccx-keep` and on the error-state `Close cell`
- re-entry guard on `close()` itself — the header's own close button is **not** under the `inset-0` overlay, and re-entering would have cleared the error the removal was about to write (not flagged by either bot; found while checking their paths)
- a spec covering all three dismissal routes, plus that the failure survives and the dialog becomes dismissible again once the request settles. It is red without the fix.

**Not applied**, with reasons:

1. **CodeRabbit — "duplicate `const gate` declarations, TypeScript rejects this file"** (`test/src/components/TerminalCell.spec.ts:1734`, marked Critical). False positive: the two `const gate` are at lines 1708 and 1734, inside two **different** `it()` callbacks, so they are separate function scopes. `yarn typecheck` (`vue-tsc -b` over the whole repo, test projects included) and the full suite both pass — 9315 tests green.

2. **CodeRabbit — "Out of Scope Changes: MCP synchronization and resume event payloads"**. False positive: `git diff origin/main...HEAD` only **moves** the existing `syncMcpGroupsInto` calls inside the new guard, and does not touch the resume payload at all (`runningKey` does not appear in the diff — that is #1533, already on main).

3. **CodeRabbit — "Docstring Coverage 66.67% < 80%"**. Against this repo's convention: `CLAUDE.md` says to default to no comments, forbids comments that restate what the code does, and caps them at one line except for public-API JSDoc on a published package.

CI was green on the previous commit across all 11 checks.

isamu:
## Iteration 2 — Codex LGTM, plus two of my own

**Codex: `CODEX VERDICT: LGTM`** on `23d6729b` — the close-confirmation interleaving is closed and covered.

Pushed `7dd4d98f` anyway, for two things I found reviewing the diff against the bot threads that **neither bot flagged**:

1. **A refusal could be reported under the wrong repository.** The directory field stays editable for the whole round trip, so a 500 can *land* after the user has typed their way somewhere else — and the error line would then say "fatal: Not a valid object name: 'main'" under a repo that has one. Clearing on the change (already in the PR) is not enough; the answer has to know which repository it was for. Same rule as #1372, applied to the error line rather than the rows. A **successful** create deliberately still launches — the worktree was made because it was asked for, and honouring the click is not a stale answer.

2. **A timeout was reported as an unreachable server.** `fetchWithTimeout` gives up at `SLOW_COMMAND_TIMEOUT_MS` (60s), which a checkout large enough to make #1549 worth fixing can exceed — and git carries on regardless, so the worktree appears a moment later. "Could not reach the server" sent the user to look at their network. It now says the request timed out and may still finish.

Both have specs; local `format` / `lint` / `typecheck` / `build` / `test` (9317) green.

CodeRabbit had not re-reviewed `23d6729b` when this went up — its findings on the previous head were triaged in the comment above.

isamu:
## Iteration 3 — one Codex finding, one CI failure of my own

**Codex on `7dd4d98f`: CHANGES REQUESTED** — `CellLaunchForm.vue:501`, a timeout while consuming the response body is swallowed by `jsonBody()` and shown as a missing worktree path. **Valid, and applied** in `b154cea6`.

`fetchWithTimeout` keeps its deadline armed across the **body** (its own comment says why), and `jsonBody` absorbs an unreadable body into `{}` — so an aborted read after a `200` came out as *"the server answered without a worktree path"* for a worktree that exists. That is precisely the trap `jsonBody`'s doc names from #1300; this was a fourth call site of it.

Split by status rather than dropping `jsonBody`:
- **a refusal** still goes through it — a 403 from the origin guard is not JSON at all, and the status is already the answer
- **a 200** is read directly and left to throw, landing on the timeout message

Both halves have a spec.

**CI on `7dd4d98f` was red** and that one was mine: `sonarjs/no-nested-conditional` on the ternary I added the commit before. Fixed in `7dc9a109`. Root cause worth recording — I had been reading `yarn lint` through `grep "  error"` instead of its exit code, and ANSI colouring between the spaces and the word made one error match zero times. Every check since is verified by exit status (lint 0, typecheck 0, build 0, test 0 / 9319 passing).

CodeRabbit is rate-limited; not waiting on it per the project's rule.

PR #1552, opened 2026-08-08
feat(#1551): grey the whole cell and spin over it while its worktree is removed

### LINKED ISSUE #1551: worktree セルを閉じるときの削除中、UI が何も変わらないので閉じない理由が分からない
## What happens

worktree セルの close は、確認ダイアログで **Remove worktree / Discard & remove** を押すと `git worktree remove` を実行する。大きなリポジトリではこれが数秒かかるが、その間 **セルの見た目はほぼ変わらない**。

#1550 でボタン自体は `Removing…` に変わり無効化されるようになったが、それはダイアログ内の 12px のラベル1つだけで、**セル本体・ヘッダは通常どおり生きているように見えたまま**。結果として「押したのに閉じない」「固まった？」としか分からない。

## What I expected

- 削除が走っている間は **そのセル全体（ヘッダ含む）がグレイアウト** して、もう触るものではないと分かる
- 中央でスピナーが回り、「いま削除している」ことが一目で分かる

## Notes

- `.cell-inner` には parked セル用の全体フェード（`SUNK_CELL` = `opacity-40`、ヘッダも含めて掛かる）が既にある。同じ語彙を再利用できる。
- 確認ダイアログのオーバーレイは `.cell-inner` の内側にあるため、ヘッダには掛からない（#1550 で `close()` に再入ガードを足したのはそのため）。全体を覆うには `.cell-inner` の兄弟としてオーバーレイを置く必要がある。
- ボタンが全部無効になったダイアログを薄く残しても情報量はないので、削除中はダイアログをスピナー表示に差し替えるのが自然。
- 削除が失敗したときは、これまでどおり確認ダイアログがエラー付きで戻ること。

refs #1549, #1550

isamu:
## Summary

#1550 made the removal button hold itself and read `Removing…`. That is a 12px label inside a dialog — meanwhile the **cell** (header, chips, terminal) went on looking completely live for the several seconds `git worktree remove` takes on a large repository. So the screen still said "nothing happened", which is #1549's complaint one layer out.

While the removal runs:

- **the whole cell fades**, header included, through the same one class parked cells already use (`SUNK_CELL` = `opacity-40` on `.cell-inner`, which wraps every child)
- **one spinner sits over it**, centred, naming what is going: `Removing ⎇ demo-repo (fix-login)…`
- **the confirmation dialog is replaced**, not faded

## Items to Confirm / Review

1. **The spinner is a SIBLING of `.cell-inner`, not a child.** Two reasons, both load-bearing: a busy indicator inside the layer it is dimming would be dimmed by it (40%), and only a sibling at the cell root reaches the **header** — the close-confirm overlay is inside `.cell-inner` and never did, which is why #1550 needed a re-entry guard on `close()`.
2. **One class for the fade, not two.** `sunkClass` became `fadedClass`, yielding `SUNK_CELL` for *either* reason (parked, or removing). Adding a second opacity utility on the same element would be settled by Tailwind's output order rather than by intent — the rule `SUNK_CELL`'s own comment sets, now noted there.
3. **Replacing the dialog rather than fading it** is a judgement call. Every control in it is already disabled during the removal, so a dialog of dead buttons at 40% is noise where a spinner is an answer. It comes **back**, with its reason, if the removal fails — the dismissal guards from #1550 are what keep that true and are unchanged. The #1550 spec that asserted the dialog stays visible mid-removal has been rewritten against the new shape, and still proves what it was written for: Escape and the header close button must not clear `closeConfirm` underneath, or the failure would return to a confirmation that is no longer rendered.
4. **Not in scope:** the launcher's `wt-del` row already shows its spinner in place (#1550) and is a row in a list — there is no cell to grey out.

## User Prompt

> つづけてworktreeをとじるときに削除するけど、それもUI変化しないでなかなか閉じないので分かりづらい。そのセル全体をグレイアウトしてスピナーを回そう

## Verification

`format` / `lint` / `typecheck` / `build` / `test` all exit 0 (9319 passing).

Driven in a **real browser against a real dev server** (scratch `HOME` + `MULMOTERMINAL_HOME`, a throwaway git repo, the remove route delayed 6s to stand in for a large repository) — the actual flow, worktree created from the launcher and the cell closed from its own header:

- `.cell-inner` computed `opacity` = **0.4**; the overlay's own = **1**
- overlay bounding box `1082×744` against the cell's `1084×746` — the whole cell, border in
- text: `Removing ⎇ demo-repo (fix-login)…`, `role="status"`
- the confirmation is gone while it runs, and the launcher is back once it finishes
- no `pageerror`, no console errors

Screenshot: the header, the `connecting` chip and the terminal are all visibly dimmed behind a single crisp spinner.

Closes #1551

work in mulmoterminal3

## Summary by CodeRabbit

* **New Features**
  * Added a clear in-progress state when removing a worktree, including a full-cell spinner and faded appearance.
  * Disabled cell interactions and prevented repeated removal attempts while removal is underway.
* **Bug Fixes**
  * Restored the close confirmation dialog when worktree removal fails, allowing the action to be retried.
* **Tests**
  * Expanded coverage for removal progress, disabled states, completion, and failure recovery.

### DIFF
--- a/plans/feat-1551-cell-removing-overlay.md
+++ b/plans/feat-1551-cell-removing-overlay.md
@@ -1,0 +1,43 @@
+# feat #1551 — a cell being removed should look like a cell being removed
+
+## The gap #1550 left
+
+#1550 made the removal button hold itself and read `Removing…`. That is a 12px label inside a
+dialog. Meanwhile the **cell** — its header, its directory chip, its terminal — carries on looking
+exactly like a live cell for the several seconds `git worktree remove` takes on a large repository.
+So the whole screen still says "nothing happened", which is the complaint #1549 was about, one
+layer out.
+
+## What ships
+
+**While `closeBusy === REMOVE_KEY`:**
+
+1. **The whole cell fades**, header included, through the mechanism parked cells already use
+   (`SUNK_CELL` = `opacity-40` on `.cell-inner`, which wraps every child — see `cellParked.ts`).
+   One computed yields the class for **either** reason, because two opacity utilities on one
+   element are settled by Tailwind's output order rather than by intent — the rule `SUNK_CELL`'s
+   own comment sets.
+2. **One spinner over the whole cell**, as a **sibling** of `.cell-inner` rather than a child: a
+   busy indicator inside the layer it is dimming would be dimmed by it. `absolute inset-0` on the
+   cell root therefore covers the header too, which the close-confirm overlay never did (that is
+   why #1550 needed a re-entry guard on `close()`).
+3. **The confirmation dialog is replaced, not faded.** Every one of its controls is already
+   disabled during the removal, and a dialog of dead buttons at 40% is noise where a spinner is an
+   answer. It comes **back** — with the failure — if the removal fails, which is the behaviour
+   #1550 shipped and this must not lose.
+
+Nothing else changes: the dismissal guards from #1550 (`cancelClose`, `close`, the disabled
+controls) all stay, because they are what makes the confirmation return intact on a failure.
+
+## Not in scope
+
+The launcher's own `wt-del` row already shows its spinner in place (#1550) and is a row in a list,
+not a cell — there is no cell to grey out. Left alone.
+
+## Files
+
+| file | |
+|---|---|
+| `src/components/cellParked.ts` | `SUNK_CELL` gains a second caller; comment says so |
+| `src/components/TerminalCell.vue` | the dim computed, the busy overlay, the dialog's `v-if` |
+| `test/src/components/TerminalCell.spec.ts` | the overlay appears and covers, the dialog returns with its error on failure |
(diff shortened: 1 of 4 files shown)


isamu:
## Iteration 1 — Codex CHANGES REQUESTED, applied

**Valid.** The overlay veils the cell but stops the **mouse and nothing else**: Tab still walked into the header buttons and xterm's own textarea behind it, and a screen reader still read a cell that is being deleted.

Fixed in `e23d4ea0` with `inert` on `.cell-inner` — which is the answer this repo already reached for the same problem on `zoom-main` (Codex, #1333), **including its trap**: `inert` is Booleanish to Vue, so a plain `false` reaches the DOM as `inert="false"`, and an element carrying that *is* inert. `|| undefined` is what keeps a live cell live, and the spec pins the absent case as well as the present one.

Verified in a **real browser**, since jsdom does not implement `inert` behaviour and the unit test can only see the attribute:

| | |
|---|---|
| focusable controls in the cell body, before | **13** |
| …during the removal | **0** |
| `Tab` presses (25) landing inside the removing cell | **0** |
| attribute value | `""` — not `"false"` |
| cleared once the removal finishes | yes |

CI was green across all 11 checks on the previous head; local `format` / `lint` / `typecheck` / `build` / `test` all exit 0.

PR #1565, opened 2026-08-08
feat(settings): group the Settings modal into categorised tabs (#1563)

isamu:
## Summary

The Settings modal was 24 headings in one flat 560px scroll — no grouping, no search, no table of
contents. This gives it a **left sidebar of grouped tabs**: 9 groups, one tab per section, one pane
on screen at a time, following MulmoClaude's own settings sidebar (#1333 there).

- **`src/components/settings/settingsTabs.ts`** (new) — the whole table: group → label → tabs. The
  label lives here and **each section loses its own `<h3>`**, so the sidebar entry and the pane
  heading are one string rather than two that can drift (#1097's lesson). It is also exactly the
  shape the i18n follow-up needs.
- **Panes are `v-if`.** Opening Settings no longer mounts 23 sections and fires their GETs. The one
  section that hid *itself* — Voice input, behind a server capability probe — would have left an
  empty pane behind a button, so **the probe moved up to the modal** and now decides whether that
  tab is offered at all.
- **`ShortcutsSection` splits** into `TerminalKeysSection` + `KeyboardShortcutsSection`. It was the
  one file with two headings, and two sidebar entries are two chances to find the thing.
- Dialog is `w-[min(780px,94vw)]` at a fixed `h-[85vh]` so it doesn't resize under the pointer on
  every click.

Design notes: [`plans/feat-1563-settings-categories.md`](plans/feat-1563-settings-categories.md).

## Items to Confirm / Review

1. **Two deliberate divergences from MulmoClaude** (which uses a plain `<nav>` of buttons with
   `aria-current`, and has no narrow-screen form). Both because 24 entries is not 15 — please sanity
   check the reasoning:
   - **`role="tablist"` + roving tabindex + arrow keys.** 24 plain buttons are **24 Tab stops
     standing between the dialog and the setting you opened it for** — more keystrokes than the flat
     scroll being replaced. Selection follows focus (automatic activation), as `ThemeSection`'s
     radiogroup already does in this repo; the cost is that arrowing past a pane mounts it.
   - **Below `sm`, the sidebar becomes an `<optgroup>`'d `<select>` above the pane.** Found by
     capturing at 390px: a 160px sidebar left ~190px of pane and the sound rows lost their own
     labels off the left edge — worse than what this replaces. First use of a Tailwind breakpoint in
     this app.
2. **The grouping itself** — 9 groups, and which section sits in which. Group order and within-group
   order are meant to be expected access frequency. This is the part worth arguing about.
3. **`settings.png` is deliberately NOT updated.** Two dated release pages (`v2.1.0`, `v2.2.0`)
   reference it, and a dated page is a snapshot of the screen it describes. README points at a new
   `settings-tabs.png` instead; the living guide's `config-settings-modal.png` IS replaced in place.
4. **The landing tab is Theme**, and the tab resets on every open (the modal is `v-if`'d by its
   shells). No "remember the last tab" — say if that is wanted.
5. `docs/guide/{en,ja}/config.md` said "in this order" and counted **twenty-one** sections; there
   were already twenty-four. Corrected, and the table reordered to the sidebar's order **by script
   off `SETTINGS_GROUPS`** so no row was retyped.

## User Prompt

> https://github.com/receptron/mulmoterminal/issues/1563
> カテゴリーわけするのと、i18nの見直しだね

Agreed up front, with the answers that shaped this PR:

- **Layout** — left sidebar + grouped tabs (MulmoClaude's shape), rather than collapsible groups in
  the existing single scroll or top tabs.
- **i18n scope** — introduce vue-i18n for the Settings modal only (`en`/`ja`), other screens later.
  MulmoTerminal has **no host i18n today**: `vue-i18n` is in `package.json` but there is no
  `createI18n` / `useI18n` anywhere, and every UI string is hardcoded English.
- **Language switching** — browser language by default, overridable from Settings.
- **PR split** — two PRs. **This is PR 1 (categorisation only; wording stays English).** i18n gets
  its own issue and PR, built on `settingsTabs.ts`.

## Verification

Against a **real running server** with a scratch `HOME` (so no real paths leak into shots), driven
with Playwright:

- **Tab stops measured from the open dialog**: `Theme tab → Midnight → Create a theme… → Close →
  ✕ (wraps)`. The whole 24-entry sidebar is **one** stop, and the next stop is inside the pane.
- **6 × ArrowDown from Theme** lands on `Directory settings` (the 7th tab) and the pane heading
  follows. Wrapping at both ends is pinned by a spec.
- **Zero console errors** across the tabs exercised (theme / sounds / shortcuts / dirSettings /
  help / mcp), at 1440×1000, 1280×860 and 390×844.
- `yarn lint` 0 errors, `yarn typecheck`, `yarn build`, `yarn test` — **9389 passed**, 45 skipped.

`SettingsModal.spec.ts` was rewritten around the tabs (each test opens the tab it is about) and
gained: every tab in the table renders a pane and none is unreachable; the Voice input tab appears
iff the probe says capable, including when the probe throws or 404s; one Tab stop with arrows
moving and wrapping; the narrow-screen picker offering the same sections, grouped, and switching
the pane.

## Screenshots

| Sidebar + a full pane | Narrow (390px) |
| --- | --- |
| Settings with the sidebar | the sidebar becomes a section picker above the pane |

work in mulmoterminal2

## Summary by CodeRabbit

* **New Features**
  * Settings are now organized into grouped tabs for easier navigation.
  * Added responsive section selection on smaller screens and keyboard-accessible navigation.
  * Added Terminal keys settings for copy-on-select and submit-key preferences.
  * Voice settings appear only when voice input is available.

* **Documentation**
  * Updated English and Japanese Settings guides with the new layout, sections, and screenshots.
  * Refreshed the documentation image inventory.


isamu:
## Review triage

**Codex [P1] — unsaved draft lost on tab change: valid, fixed.** `v-if` unmounted the open pane on every tab change, and `TerminalFontFamilySection` keeps a typed stack in a local `draft` ref precisely so a failed POST doesn't throw it away — so a single arrow key discarded it. Panes are now **created on first visit and hidden with `v-show` after that**. Closing the modal still resets everything (the shells `v-if` the modal itself), and a never-opened tab still doesn't mount, so the "don't fire 23 GETs on open" half of the change survives.

Each pane is wrapped in a `<div>` rather than taking `v-show` directly: most sections are **fragment-root** components, and `v-show` needs one element to set `display` on.

Two regression tests, and the first one **was confirmed to fail** against the old `v-if` before being kept:
- a typed-but-unapplied font stack survives a trip to another tab and back
- a tab never opened has not mounted

Also verified in a real browser (Playwright, real server): with Notification sounds open the font pane reports `visible: false`, and after clicking back the field still reads `'Cica'`. No console errors.

**CodeRabbit — screenshot description says nine groups: valid, fixed.** The image is scrolled to **Appearance → Sessions**; Help is below the fold. `docs/guide/images/README.md` and the guide's alt text now say eight and why.

Full suite after the fix: **9391 passed**, 45 skipped. lint / typecheck / build clean.

work in mulmoterminal2

PR #1568, opened 2026-08-08
feat(settings): say what a skill button does, and let it be cancelled (#1564)

isamu:
## Summary

A Settings skill button was an icon and a label. Pressing it **closed Settings, spawned an agent
session in a new grid cell, and sent the first turn** — with nothing on screen saying that would
happen, and nothing saying how to undo it. #1564 is someone who decided they did not need the
setting after all and could not find the way back.

Both halves of what was asked for:

- **A one-line hint under every skill button** — "An agent does this for you — it asks a few
  questions and edits the config file itself." It lives in `SkillLaunchButton`, so it is one
  sentence rather than seven.
- **A confirmation on press**, with **Cancel / Start**. It names the agent, says one new terminal
  will open, says closing that terminal ends the session, and that Settings reopens from the
  toolbar. Cancel is focused and comes first — the safe answer should not be under the pointer.

Design notes: [`plans/feat-1564-skill-launch-confirm.md`](plans/feat-1564-skill-launch-confirm.md).

The confirmation

## Items to Confirm / Review

1. **The confirmation is owned by `SettingsModal`, not by the button — and that is the load-bearing
   decision.** `useModalKeyboard` binds Escape on `document`; a dialog with its own binding would
   answer one press by closing the confirm **and** the Settings modal behind it. With the state in
   the modal, its existing `onClose` becomes "dismiss the confirm if one is open, else close" — one
   listener, layered — and `Tab` is moved into the confirm through a `computed` trap element
   (`useModalKeyboard` takes the trapped element as a `{ value }`, so a computed satisfies it).
   Verified in a real browser, not just jsdom.
2. **`launch-skill` now fires only on Start**, so `GridView` is untouched and Cancel leaves Settings
   exactly where it was (today `closeSettings()` ran first, which is why there was nothing to go
   back to).
3. **The agent name comes from `launchAgent`**, the picker's saved choice — not a hard-coded
   `claude`. The seed is rewritten server-side for codex, so a dialog claiming "claude" while codex
   starts would be worse than saying nothing.
4. **The wording**, in both languages. One thing I deliberately did NOT claim: an earlier draft said
   "nothing is written until you answer". I could not verify that for all seven skills, so it says
   only what #1564 established — closing the terminal ends the session.
5. **Not done, deliberately**: no "don't show this again" (starting a session is worth confirming
   every time, and the checkbox would need its own way back), and no banner on the started cell (the
   dialog says it before the cell exists, which is where the question is actually asked).

## User Prompt

> https://github.com/receptron/mulmoterminal/issues/1564
> skillのボタンの前に説明を追加するか、もしくはボタンを押したときに、skillでai が設定していくとして、ok / 戻るを追加するか、などして、初心者にもわかるようにして

Agreed up front: **both** — the hint before the button *and* the Cancel/Start dialog; the way to
stop is carried by the dialog's wording rather than by a banner on the cell; and no "don't show
again" checkbox.

## Verification

Against a **real running server** with a scratch `HOME`, driven with Playwright, in **both**
locales (`en-US` and `ja-JP`):

- pressing a skill button focuses **Cancel**, and `terminals started = 0` — nothing spawns
- **Escape closes the confirm only**: `confirm = 0`, `settings still open = 1`; a **second** Escape
  closes Settings
- **Start works end to end**: `settings closed = true`, `terminals added = 1`
- zero console errors in either language

`yarn lint` 0 errors, `yarn typecheck`, `yarn build`, `yarn test` — **9406 passed**, 45 skipped.

New specs in `SettingsModal.spec.ts`: it asks before starting anything; it names the picker's agent
rather than claude; Cancel emits nothing and leaves Settings open; Escape answers the confirm first
and only then the modal; and **every skill button in the modal routes through it**, enumerated from
what renders rather than from a list typed in the test.

The seven existing "button X launches skill Y" cases now assert `launch-skill` is **undefined** on
the press and only fires after Start — the regression this PR is about, pinned in the tests that
would otherwise have kept passing.

work in mulmoterminal2

## Summary by CodeRabbit

- **New Features**
  - Added a confirmation dialog before launching skills from Settings.
  - Shows the selected agent, new terminal behavior, stop instructions, and localized Start/Cancel actions.
  - Keeps Settings open when a launch is canceled.
  - Added English and Japanese translations and documentation.

- **Bug Fixes**
  - Improved keyboard focus and Escape-key handling for layered dialogs.

- **Tests**
  - Added coverage for confirmation, cancellation, agent naming, keyboard behavior, and all skill-launch buttons.


isamu:
## Review triage

**Codex + CodeRabbit — focus is not restored after Cancel / Escape: valid, fixed.** Focus moves *into* the confirmation, and dismissing it unmounted the focused element without putting focus back, so a keyboard user was dropped onto `<body>` — out of the modal they were part way through.

One deviation from the suggested fix, and it matters: the button is found by **the skill it launches** (a `data-skill` attribute on `SkillLaunchButton`) rather than by remembering `document.activeElement` in `askBeforeLaunch`. **A browser does not reliably focus a `<button>` when it is clicked — Safari does not** — so what was active when the press arrived is not a dependable handle on the trigger. Writing it the suggested way first is also what surfaced this: the new test failed with focus on the ✕ close button, because Vue Test Utils' `trigger("click")` does not focus either.

Verified in a real browser (Playwright, real server), for **both** dismissals:

```
Cancel: focus inside dialog = BUTTON[data-skill=null] "Cancel"
Cancel: focus after dismiss = BUTTON[data-skill=mulmoterminal-notify] "notifications_active Con"
Escape: focus inside dialog = BUTTON[data-skill=null] "Cancel"
Escape: focus after dismiss = BUTTON[data-skill=mulmoterminal-notify] "notifications_active Con"
errors = []
```

Covered by a new `it.each` over Cancel and Escape, mounted with `attachTo: document.body` so `document.activeElement` is real rather than a value jsdom will happily report without the element being in the document.

**CodeRabbit — `launchAgent` leaks `"codex"` when an assertion throws: valid, fixed.** Restored in `afterEach` rather than at the end of the test body, so a failure cannot make the next test fail for the wrong reason.

Full suite after the fix: **9408 passed**, 45 skipped. lint 0 errors / typecheck / build clean.

work in mulmoterminal2

isamu:
## Review loop — converged (2 iterations)

**Iteration 1** — Codex `LGTM`, CodeRabbit no new findings. One issue **I** found that neither bot flagged: the confirmation block had landed between the ARIA tablist comment and the `navEl` / `onTabKey` it explains, so a reader met "arrows move within the list" followed with no blank line by "a skill button starts a live agent session". Relocated — a pure move, the diff pairs line for line.

**Iteration 2** — Codex `LGTM` again, CodeRabbit rate-limited, **all 11 checks green**, zero inline comments.

### One thing deliberately not changed, so it is not silent

Both this dialog and the Settings modal behind it carry `aria-modal="true"`. Strictly, only the topmost dialog should — the one underneath ought to be `inert` while the confirmation is open. I left it because:

- the practical failure it guards against is already handled: `Tab` is trapped **inside the confirmation** (the trap element is a `computed` that returns the confirm while it is open), and `Escape` is layered so it answers the confirmation first;
- making the background `inert` is a change to how the whole modal renders, not to this feature, and it would want its own tests across all 25 panes.

Worth doing as a follow-up if the modal grows more nested dialogs — say the word and I'll open an issue.

### Verified against a real browser, not just jsdom

Both locales, real server: pressing a skill button starts **nothing** (`terminals started = 0`) and focuses Cancel; Escape closes the confirmation only (Settings stays open) and a second Escape closes Settings; Cancel and Escape both return focus to the exact button (`data-skill=mulmoterminal-notify`); Start closes Settings and adds one terminal. No console errors.

Local: lint 0 errors, typecheck, build, **9408 tests passing**.

work in mulmoterminal2

PR #1571, opened 2026-08-09
feat(collections): resolve a root per request and leave the engine none

snakajima:
Phase 2+3 of `plans/feat-collections-project-root.md`. **Behaviour is unchanged** — every request still resolves to the shared workspace — but the collection engine no longer has an ambient root to fall back to, which is what makes serving a second one safe.

## What changed

`configureCollectionHost` now binds `workspaceRoot: null` (core 3.0.0's explicit-root mode), so `getWorkspaceRoot()` **throws** instead of guessing. Every engine call names its root through the new `server/infra/project-root.ts` — one choke point, shared with accounting / wiki / resources later, so phase 5 changes one function rather than the ~30 call sites this touched.

## Why the ceremony for a no-op

With one root, a forgotten `opts.workspaceRoot` resolves correctly by accident. With one root **per project** it is not a crash — it reads or writes another project's data: tests green, types green, nothing logged. Turning the fallback into a throw is the only way to know the threading is complete.

It immediately earned that: two call sites this change had missed showed up as test failures rather than as silence.

- the custom-view **i18n** read
- the **thumbnail resolver's containment check** — which is a security boundary, so it must check against the root the request resolved, never whatever happened to be bound

## Shapes worth a look in review

- **Injected-dep modules are bound per scope**, not given a scope parameter: `buildRemoteViewFor(scope)`, `mutateRemoteViewFor(scope)`, `remoteViewItemsFor(scope)`, `thumbnailResolverFor(scope)`. The root ends up next to the engine call, and no dep interface — or its specs — had to change.
- **`manageCollection`'s binding takes the root as a getter.** The tool is built at module scope, before boot binds anything, so a captured value would be `undefined` — which under the new binding is a throw, not a fallback.
- **The completion watchers are started with an explicit root.** Without one they would throw `COLLECTION_ROOT_REQUIRED` on first discovery, and since the caller is fire-and-forget with a `.catch`, that would not stop the server — it would leave every collection bell silently dead.
- **`resolveProjectRoot` refuses a `?project=` parameter rather than ignoring it.** Ignoring it would serve one project while the client asked for another, with nothing saying so — the same silent-wrong-root failure this PR removes, moved onto the wire. Serving it is phase 5.

## The regression pin

`test/server/infra/projectRoot.spec.ts` asserts that after `initCollectionsBackend` the engine throws on `getWorkspaceRoot()`. Binding a string root again would restore the silent fallback and **every other test would still pass** — this is the only assertion that would notice.

## Also lands

The §6.5 note written before this work: skill staging (`data/skills` → `.claude/skills`) is a **workspace** mechanism and must not follow into project roots — the engine reads staging *first* for `source: "project"` collections, so a stray tree there would shadow the committed skill. That needs an upstream change (U5) and is not in this PR.

## Gate

`yarn format`, `yarn lint` (0 errors), `yarn typecheck`, `yarn build`, `yarn test` — **9412 passed / 45 skipped**.

work in mulmoterminal

### DIFF
--- a/plans/feat-collections-project-root.md
+++ b/plans/feat-collections-project-root.md
@@ -100,6 +100,14 @@ we are not re-binding, we are passing arguments.
 Zero callers, and it is precisely the helper a future contributor would reach for — at which
 point containment is silently checked against the wrong root. Delete it, or make it take a root.
 
+### U5. `skillsStagingDir` must be able to say "there is no staging here" *(required — see §6.5)*
+
+`CollectionHost.paths.skillsStagingDir: (root) => string` cannot express "this root has no
+staging tree", and the engine reads staging **first** for `source === "project"` collections
+(`skillAssets.ts:55`). Widen it to `(root) => string | null` and skip the staging base when null.
+
+MulmoClaude returns a string for its one workspace and is unaffected.
+
 ### Not upstream
 
 - **Route shapes** — the collection HTTP routes are host-owned (MulmoTerminal mounts its own in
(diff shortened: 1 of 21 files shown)


snakajima:
On the remaining blocker — **"Pin a core release that supports explicit-root mode"** — this one is incorrect, and I'd rather show the evidence than just assert it.

The claim is that the locked `@mulmoclaude/core@3.0.0` has `workspaceRoot` as a required `string`, no options on `listRegistry`, and a non-throwing `getWorkspaceRoot()`. Checked against the **published tarball, pulled from the registry** (not this checkout's `node_modules`):

```
$ curl -sL "$(npm view @mulmoclaude/core@3.0.0 dist.tarball)" | tar -xz
$ grep -n "workspaceRoot: string | null" package/dist/collection/server/host.d.ts
21:    workspaceRoot: string | null;
$ grep -n "COLLECTION_ROOT_REQUIRED" package/dist/collection/server/host.d.ts
9:export declare const COLLECTION_ROOT_REQUIRED = "COLLECTION_ROOT_REQUIRED";
$ grep -n "declare function listRegistry" package/dist/collection/registry/server/index.d.ts
17:export declare function listRegistry(scope?: RegistryScope): Promise<RegistryListResponse>;
```

All three APIs are in the published 3.0.0. Explicit-root mode **is** the 3.0.0 release (receptron/mulmoclaude#2838) — the major bump was taken for it, so there is no newer release to pin to.

The empirical check agrees: `lint-and-build` runs a clean install and `vue-tsc`, and it passed on the previous commit. If the locked package lacked these types, that job is the one that would have failed first.

The other three findings were real and are fixed in `1cbbf5cd`:

- **calendar push (P1)** — genuinely broken. `liveDeps` bound the ambient `loadCollection` / `getWorkspaceRoot`, so every push would have 500'd. The deps now carry one root (`scope: (req) => ProjectScope`).
  Worth flagging: **CI passed on the broken commit.** Every test for that route injects stub deps, so the production wiring had no coverage — the status-vs-field split was pinned precisely while the route could not have served a request. Added the missing live-deps test.
- **mobile-view i18n (P2)** — real. `readCustomViewI18n` was passed through unbound, so a view declaring `i18n` would have failed to build while an untranslated one worked.
- **repeated `project` parameters (P2)** — real. `?project=a&project=b` is an array and `?project[x]=y` an object; the `typeof === "string"` guard read both as absent. Now keyed on presence.

PR #1580, opened 2026-08-09
Collections: per-root watchers, project-aware completion bells, and the dead pane button

snakajima:
Picks up `plans/HANDOFF-collections-projects.md` §3.1 and §3.2, and fixes a bug the first hand-press of the feature found.

## 1. Per-root watchers (§3.1)

The completion watcher mounted ONE generation, for the workspace, so a collection in a project folder got **no completion bells and no live refresh** when an agent wrote a record directly — which is the canonical authoring path. The collection created in `~/git/ai/mag2` declares `completionField`, i.e. it asks for a feature that could not fire.

core 3.1.0 already supports this (a generation per root, concurrently; `WATCHER_ROOT_CONFLICT` is no longer thrown; `stopCollectionWatchers({ workspaceRoot })` releases exactly one). This wires it: one generation per root the server serves — the workspace plus every saved `cwdPresets` directory.

- **"Open" means "served by this server", not "on screen".** A completion bell is the notification that the user is *not* looking, so scoping the watchers to the visible Collections pane would let a bell ring only while its collection is already up. Bounded by discovery: a root with no collections mounts no watcher.
- Reconciled by a **60s poll**, because `listProjectRoots` reads a live thunk over `cwdPresets` and nothing emits when it changes. `syncCollectionWatcherRoots` is exported so a caller that knows it changed can pull the pass forward.
- A root that fails does not stop the roots after it, warns **once**, carries its error `code`, and retries next pass.

## 2. A bell opens the project it came from (§3.2)

`buildNavigateTarget` produced `/collections/<slug>` with no project, so a bell from a project's `tasks` opened the workspace's `tasks` — same row, same title, wrong records. The bell now carries its project as the **opaque id**, in the URL and on `action.target.project`.

- Root→id resolution lives in `collectionWatchers.ts`, **not** the adapter: a `test/src` spec imports the adapter, and reaching `node:crypto` through it puts Node's globals into a DOM-typed project, where `window.setTimeout` stops returning a number and an unrelated component fails to typecheck.
- **Cross-app parity holds where it is observable**: the id is omitted for the workspace, so the only link MulmoClaude can produce or receive is byte-for-byte what it was. Concatenated rather than built with `URLSearchParams`, which writes a space as `+` where MulmoClaude's builder writes `%20`.
- Client: the project rides in the route query, browse pushes carry the open project (a ref hop stays in its project) while a bell passes its own explicitly, the overlay's surface reads it through a getter, and the plugin views are keyed on it so a same-path project switch refetches.
- `collectionNotifiedSeverities` matches project too — with a `primaryKey` schema the record id is drawn from the data, so the same record in two projects has the **same** id and an unscoped reader accents the wrong card reliably, not rarely.

## 3. The "Show this folder's collections" button never worked

Reported from the running app. It has done nothing since it shipped in #1573.

`CellChromeButtons` emits `toggle-collections`, but `cellChromeBinding`'s `chromeEvents` — the single object every cell binds with `v-on` — had no key for it, and `GridCellEmits` did not declare it. The emit was dropped inside the cell; `TerminalGrid`'s handler waited for something nothing ever sent.

Two things let it survive a green suite: an unlistened Vue emit is legal, so `typecheck` was clean; and the existing "forwards every chrome event" test listed the events **by hand** from the same wrong set of four, so it agreed with the bug. The replacement derives the expectation from `CellChromeButtons`' own runtime `emits`, so the next button is covered the day it is added.

## Verification

- `yarn format / lint / typecheck / build / test` — 9481 passing, 0 lint errors.
- Removing the button fix fails the three new tests (checked, not assumed).
- Server side confirmed against the running instance: `/api/collection-projects` lists mag2, and `/api/collections/list?project=<id>` returns its `newsletters` collection.
- **Not verified in a browser.** The pane's height, the plugin views inside the shadow root and the nav containment are exactly what the suite does not see — which is how §3 got here in the first place. `plans/HANDOFF-collections-projects.md` §3.7 now says so with this as the example.

## Left for later

Nothing calls `syncCollectionWatcherRoots` at the moment a directory is recorded, so a first-time project waits up to a minute for its watcher; the config writers are several and the poll covers them all. And MT publishes *rooted* bell ids while MulmoClaude publishes root-less ones for the same workspace — core's sweep handles that asymmetrically, so alternating between the apps can strand a bell MulmoClaude will `skip` rather than clear. Pre-existing since #1571, by upstream design, and recorded in the handoff to confirm with MulmoClaude rather than "fix" here.

work in mulmoterminal

## Summary by CodeRabbit

* **New Features**
  * Collection monitoring now supports multiple project roots concurrently, with automatic synchronization and isolated recovery.
  * Collection links and notifications preserve project context, including encoded deep links and project-specific navigation.
  * Collection browsing and plugin views update correctly when switching projects.
  * Added support for toggling Collections from cell controls.
* **Bug Fixes**
  * Notification severity indicators are now scoped to the active project, preventing cross-project highlights.
* **Documentation**
  * Updated project-root watcher and notification handoff documentation.
* **Tests**
  * Expanded coverage for project navigation, watcher reconciliation, notification links, and Collections toggle events.

### DIFF
--- a/common/collectionNotifyTarget.ts
+++ b/common/collectionNotifyTarget.ts
@@ -19,6 +19,12 @@ export interface CollectionNotifyTarget {
   slug: string;
   /** Absent on a collection-level bell — those can't accent a specific record. */
   itemId?: string | undefined;
+  /** The record's project, as the opaque id the API takes. ABSENT MEANS THE WORKSPACE — both
+   *  because that is what a workspace bell has always written and because it is the only thing
+   *  MulmoClaude, which has one root, can write. A slug is unique within a root and nowhere
+   *  else, so a reader that ignores this accents the same-named record of whichever project is
+   *  on screen. */
+  project?: string | undefined;
 }
 
 /** Narrow a notification's opaque `pluginData` to its collection navigate target,
@@ -31,7 +37,11 @@ export function collectionNotifyTargetOf(pluginData: unknown): CollectionNotifyT
   if (!isRecord(action)) return null;
   const { target } = action;
   if (!isRecord(target)) return null;
-  const { view, slug, itemId } = target;
+  const { view, slug, itemId, project } = target;
   if (view !== COLLECTION_NOTIFY_VIEW || typeof slug !== "string" || slug.length === 0) return null;
-  return { slug, itemId: typeof itemId === "string" && itemId.length > 0 ? itemId : undefined };
+  return {
+    slug,
+    itemId: typeof itemId === "string" && itemId.length > 0 ? itemId : undefined,
+    project: typeof project === "string" && project.length > 0 ? project : undefined,
+  };
 }
--- a/plans/HANDOFF-collections-projects.md
+++ b/plans/HANDOFF-collections-projects.md
@@ -35,8 +35,8 @@ uses. Confirmed present in the installed package:
   `stagedSkillAuthoring` — **both wired**
 - **`startCollectionWatchers` mounts one generation PER ROOT, concurrently**, each stamping its
   root on payloads and driving bells whose ids carry it; `stopCollectionWatchers({ workspaceRoot })`
-  stops exactly one — **not wired**
-- `buildNavigateTarget(slug, itemId, root?)` — the adapter receives the root — **not used**
+  stops exactly one — **wired** (§3.1)
+- `buildNavigateTarget(slug, itemId, root?)` — the adapter receives the root — **used** (§3.2)
 - `withCardScope` for per-card project — **blocked, see §3.3**
 - `feedRefreshTaskDef({ workspaceRoot })` with a per-root task id — MT registers one, for the
   workspace only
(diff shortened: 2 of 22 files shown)


snakajima:
## Review findings applied

Applied three fixes from the spike review:

1. **collectionWatchers**: A scoped stop that fails leaves a generation alive. Separate `failedStops` from `failedRoots` so retries continue when `stopCollectionWatchers()` throws (the package deletes its generation on the last line, so the root must stay tracked or the next pass has no handle).

2. **useCollectionBrowse record modal**: The modal survived navigation via path alone, across projects with the same route. Now keyed on path+projectId so a project switch closes it.

3. **useCollectionBrowse watcher**: The watch was on `router.path`, not the computed `pageKey` (which includes project), so project changes were invisible. The watcher now reads `currentPageKey()` each time.

Commit: 6656066f

PR #1582, opened 2026-08-09
Collections: a clone-portability check, and the pane button only where the tools are

snakajima:
Picks up `plans/HANDOFF-collections-projects.md` §3.6, plus one thing asked for after using the feature. **§3.5 was attempted and reverted in this branch** — see below.

## 1. "Would this collection survive a clone?" (§3.6)

`GET /api/collections/:slug/self-containment`, project-scoped like every other read. Five rules, each named for what breaks on the **other** machine:

| code | severity | what goes wrong there |
|---|---|---|
| `user-scope` | blocker | the skill is in `~/.claude/skills`; the clone gets whatever that machine has |
| `data-ignored` | blocker | schema committed, records not — reads as an empty collection |
| `sqlite-store` | blocker | one binary file; git cannot merge it |
| `csv-runtime` | warning | the file travels, the DuckDB runtime must be there too |
| `no-primary-key` | warning | 4-byte random ids, so two machines can mint the same one |
| `not-a-repo` | info | nothing to clone yet; the git-dependent checks did not run |

Three decisions worth reviewing:

- **It asks git; it does not parse `.gitignore`.** The rule that matters may come from a parent directory, `.git/info/exclude`, or the global ignore file. `--no-index` is passed so the answer is about the **rule** rather than current tracking — a data dir that is already tracked reads as "not ignored" without it, which is true of the files already committed and misleading about every record written from now on.
- **It asks about the path the RECORDS live at** (`recordPathOf`), not the data directory: a `dataSource` collection's records are the rows of the external file and a `storage` collection's are in its backend file. (Codex P2 — the first version asked about `dataDir`, so a `*.csv` ignore line went unreported with `portable: true`.)
- **Unknown is not clean.** `dataDirIgnored` is `boolean | null`; null (no repo, no git) reports nothing rather than clearing the collection.

The module mounts its own route because `collections.ts` is at its `max-lines` budget. Nothing **surfaces** it yet — no button, no agent tool, no creation-time hook; the route is the seam a UI would use and the rules are a pure function anything can call.

## 2. No Collections button where the directory has no collection tools

Asked for after the button was first pressed. A directory that never registered the `data` MCP group has no `manageCollection`, so the pane is a door onto a room the agent beside it cannot enter.

**Hidden rather than disabled** — the opposite of the Canvas button, and the difference is what each state can say. A disabled Canvas button explains a pane that would open empty and names the fix; a directory with no collection tools is simply not a place where collections are a thing.

Answered from the **same** `/api/tools` reply as the canvas question (`hasCollectionsGroup` beside `hasCanvasGroup`), so the two buttons cannot disagree about what a session has. Two consequences, both deliberate:

- The pane has no close control of its own, so this button is also its only way out — it stays visible while the pane is open regardless of the groups. Remove that clause and an open pane is stranded for the life of the cell.
- A cell with **no session** gets no button (no MCP client, no groups). Honest, but it also hides the pane on a launcher or command cell whose directory does have collections.

## 3. Per-project feed refresh (§3.5) — attempted, reverted, blocked upstream

The commit is in this branch's history (`26ca37ae`) and reverted by `07b0c086`, so the net diff carries none of it.

Codex was right (P1). A collection with `ingest.kind: "agent"` refreshes by dispatching a hidden worker, and the seed prompt core builds carries `dataPath` **straight from the schema** — root-relative (`promptPathsFor`). `feedsSpawnWorker` calls `spawnClaudePty` with no cwd, so the worker stands in `CLAUDE_CWD`. Scheduling a project root therefore resolves `data/collections/<slug>/items` against the **workspace** and writes into the workspace's same-named collection — a silent cross-project write, which is exactly what this plan exists to remove.

**The host cannot fix it alone**: `AgentWorkerRunner` receives `{ message, roleId, hidden, onComplete }` and no root, so only core knows which root a refresh is for. The upstream change is additive — add `workspaceRoot` to that argument object and pass it from `refreshViaAgent`; MulmoTerminal then forwards it as the spawn's cwd, and this branch's commit re-lands as written.

Declarative RSS/JSON feeds were never at risk, but there is no way to register only those: the refresh walks a root's collections internally. Held rather than shipped with a caveat, because the failure is silent and lands in another project's data.

## Verification

- `yarn format / lint / typecheck / build / test` — 9511 passing, 0 lint errors. (One full run had a single failure that did not reproduce on re-run, consistent with the known flake this repo has under parallel load.)
- The portability check is pinned two ways: the rules against **facts** (the judgement a later edit is likely to get wrong), and the git half against **real repos in a tmpdir**, since its whole point is that it asks git instead of guessing — including a repo whose `*.csv` line ignores the external data file.
- **Not verified in a browser** — the standing gap recorded as §3.7 in the plan.

work in mulmoterminal

## Summary by CodeRabbit

- **New Features**
  - Added collection self-containment checks to identify portability concerns, including ignored files, unsupported storage, missing keys, and non-repository projects.
  - Added an endpoint for retrieving collection portability reports.
  - Collections controls now appear only when supported by the current session and remain accessible while the pane is open.

- **Documentation**
  - Updated project handoff guidance for collection portability checks and conditional Collections controls.

- **Tests**
  - Added coverage for portability reports, route responses, tool availability, and Collections button behavior.

### DIFF
--- a/common/toolGroups.ts
+++ b/common/toolGroups.ts
@@ -31,6 +31,20 @@ export const isToolGroup = (value: unknown): value is ToolGroup => TOOL_GROUPS.s
 // front of the spend (same reasoning as presentDocument below).
 export const CANVAS_TOOL_GROUPS: readonly ToolGroup[] = ["render", "media"];
 
+// The groups the COLLECTIONS pane is made of. `data` is the group manageCollection and
+// presentCollection belong to, and the pane is a window onto exactly that store — a directory
+// that never registered the group has no collection tools, so the pane is a door onto a room the
+// agent beside it cannot enter.
+//
+// Named here rather than written as `"data"` at the call site for the same reason
+// CANVAS_TOOL_GROUPS is: a rename must not leave a literal pointing at a group that is gone.
+export const COLLECTIONS_TOOL_GROUPS: readonly ToolGroup[] = ["data"];
+
+/** Does this session/directory have the collection tools? Same shape as `hasCanvasGroup`, and
+ *  validated the same way — an unknown group name from a newer server must not count. */
+export const hasCollectionsGroup = (groups: unknown): boolean =>
+  Array.isArray(groups) && groups.some((group) => isToolGroup(group) && COLLECTIONS_TOOL_GROUPS.includes(group));
+
 // Does a session/directory reach the Canvas at all? Asked of the group list the server reports,
 // whose members arrive as plain strings — validated rather than cast, since an unknown name from
 // a newer server must not count as a canvas group.
--- a/plans/HANDOFF-collections-projects.md
+++ b/plans/HANDOFF-collections-projects.md
@@ -152,22 +152,36 @@ collections do not exist on the phone. Preparation is specified in
 params defaulting to the host root (so adding the parameter later changes no handler), the scope
 must be an opaque id, and the artifact stays host-built.
 
-### 3.5 Per-root scheduled feed refresh — DONE
+### 3.5 Per-root scheduled feed refresh — BLOCKED UPSTREAM (attempted, reverted)
 
-`buildSystemTasks` now registers ONE `feedRefreshTaskDef` per root — the workspace plus every
-saved project directory, passed in from `server/index.ts` as `feedRoots`. A task id is the
-scheduler's primary key and core builds it from the root, so N roots register N tasks instead of
-overwriting each other down to the last one. Roots are deduped by RESOLVED path: core canonicalises
-the root into the id, so two spellings of one directory would otherwise produce one id and silently
-drop a registration.
+Written, reviewed, and taken back out in #1582. The mechanism is sound and the blocker is one
+missing argument in core.
 
-- **Read at BOOT**, because the scheduler registers once. A directory saved later starts refreshing
-  after the next restart; its feeds still update on demand meanwhile. The alternative is
-  re-architecting the scheduler around a live task list, which the user tasks (also read once, from
-  disk) would need too — a bigger change than this item.
-- **The calendar stays workspace-only, deliberately.** A Google grant is user-scope and its sync
-  marker is workspace state, so a per-project sync would need a per-project answer to "which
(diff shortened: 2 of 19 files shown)


snakajima:
## Both Codex findings verified and acted on

**P1 — "Run project feed workers from their project root" (`server/index.ts:879`): correct, and it cannot be fixed here. The feature is reverted.**

Traced it through: `refreshViaAgent` builds the seed prompt with `promptPathsFor`, which emits `dataPath` **straight from the schema** — root-relative. `feedsSpawnWorker` calls `spawnClaudePty` with no cwd, so the worker stands in `CLAUDE_CWD`. Scheduling a project root therefore resolves `data/collections/<slug>/items` against the workspace and writes into the workspace's same-named collection. Silent, and into another project's data — the exact failure this plan exists to remove.

The host has no way to do better: `AgentWorkerRunner` receives `{ message, roleId, hidden, onComplete }` and no root, so only core knows which root a refresh is for. Declarative RSS/JSON feeds were never at risk, but there is no way to register only those — the refresh walks a root's collections internally.

So `26ca37ae` is reverted by `07b0c086` rather than shipped with a caveat. The upstream change is additive: add `workspaceRoot` to the `AgentWorkerRunner` argument object and pass it from `refreshViaAgent`; MulmoTerminal then forwards it as the spawn's cwd and the reverted commit re-lands as written. Recorded in `plans/HANDOFF-collections-projects.md` §3.5.

**P2 — "Check the file that actually stores CSV records": correct, fixed in `07b0c086`.**

`recordPathOf` now picks the path the records actually live at — `dataSourceFile` for a `dataSource` collection, `storageFile` for a `storage` one, else `dataDir`. It was the right rule aimed one directory to the left, and it also covered sqlite (`*.db`), not only csv. Pinned by a spec that ignores the external file with `*.csv` while leaving the data directory perfectly committable, and asserts the collection now reports `data-ignored` and `portable: false`.

Thanks — the first one was worth the revert.

PR #1583, opened 2026-08-09
fix: catch scheduled tasks up after the server was off (#1581)

### LINKED ISSUE #1581: bug: 開発作業ログ(worklog)がUTC境界発火＋キャッチアップ無しで黙って永久にスキップされる
## What happened
`worklogEnabled: true` / `worklogIntervalHours: 6`（開発作業ログ機能）を有効にしてからしばらく経つが、`data/wiki/pages/dev-log-*.md` も `data/wiki/pages/worklog.md` も `config/scheduler/worklog-state.json` も、`CLAUDE_CWD` 配下に一度も作られない。4.7.5 → 4.7.6 へのアップデート後にサーバーを再起動し、6 時間以上待っても変化なし。エラー表示も一切ない。

対照として、同じく「定期実行」系である decisionDigest（`~/.mulmoterminal/decisions/<project>.md`）は正常に動作しており、こちらはサーバー起動時刻と一致するタイムスタンプで実際に再生成されていた。

## What I expected
`worklogEnabled` が true の間、`worklogIntervalHours` ごとに開発作業ログのバッチが起動し、週次 Wiki ページ（`data/wiki/pages/dev-log-YYYY-www.md`）とハブページ（`worklog.md`）が作成・更新される。

## Root cause（ソースを読んで特定）
`decisionDigest`（`server/index.ts` の `refreshDecisionDigests`）は起動時に即実行＋その後 `setInterval` で 6 時間ごとに実行される単純な実装で、これは正しく動く。

一方 `worklog`（`server/backends/worklog.ts` → `system-tasks.ts` → `server/backends/scheduler.ts` の `initUserTaskScheduler()`）は `@mulmoclaude/core/scheduler` の `createTaskManager()` に `taskManager.registerTask()` で直接登録されている。`initUserTaskScheduler` 自身のコメントに明記されている:

> Tasks are registered directly on the task-manager (matching MulmoClaude's user tasks) — they fire forward on schedule, with no system-task persistence/catch-up.

`@mulmoclaude/core` のタスクマネージャ（`task-manager.ts` の `isDue()`）は、interval スケジュールの発火判定を「プロセス起動からの経過時間」ではなく「UTC 0 時からの経過時間」で行う（`Math.floor(msSinceMidnight / tickMs) * tickMs % schedule.intervalMs === 0`）。6 時間間隔なら、発火できるのは UTC 00:00 / 06:00 / 12:00 / 18:00（JST 03:00 / 09:00 / 15:00 / 21:00）ちょうどの 1 分間のみ。

MulmoTerminal 側の登録経路は、状態永続化とキャッチアップを提供する `@mulmoclaude/core` の `initScheduler()` を経由せず、`registerTask()` を直接呼ぶだけ。そのため:

- サーバーがその狙い撃ちの 1 分間に動いていなければ（`npx` 経由の起動・再起動、スリープ、クラッシュ等どれでも該当）、その回の実行は黙って永久にスキップされる。
- 次に発火のチャンスがあるのは 6 時間後の次の境界のみで、逃した回のキャッチアップは一切ない。
- 実行できたかどうかを確認する手段が UI にも API にも無い。`GET /api/scheduler/tasks` は `config/scheduler/tasks.json` のユーザー定義タスクのみを返し、システムタスク（worklog）は含まれない。状態ファイルも一切書かれない。

`#1191` のコメントでも「interval スケジュールは壁時計境界（UTC 0 時起点）で発火する」という同じ挙動が upstream 調査として言及されているが、そちらは Google Calendar 同期の話であり、「MulmoTerminal のシステムタスクにキャッチアップが存在せず黙って永久にスキップされる」点そのものは別問題として報告されていないようだった。

## Steps to reproduce
1. 設定パネルから「開発作業ログ」を有効化（`worklogEnabled: true`、`worklogIntervalHours: 6`）。
2. 普段どおり `npx mulmoterminal` でサーバーを起動・終了・再起動しながら使う（＝サーバーが 24 時間連続稼働ではない、一般的な使い方）。
3. UTC 00:00 / 06:00 / 12:00 / 18:00（JST 03:00 / 09:00 / 15:00 / 21:00）のいずれのタイミングでもサーバーが起動していない状態が続くと、`CLAUDE_CWD` 配下に `data/wiki/pages/dev-log-*.md` / `data/wiki/pages/worklog.md` / `config/scheduler/worklog-state.json` が一度も作られない。

## Environment
- mulmoterminal: 4.7.6（4.7.5 からアップデート、`npm`/`npx` install）
- Node: v24.12.0
- OS: Linux x86_64（WSL2）
- Shell: zsh
- tmux: 3.7b / gh: 2.97.0 / claude: 2.1.220 (Claude Code) / codex-cli: 0.141.0
- 起動方法: `npx mulmoterminal`（`--env-file-if-exists` 付きで `server/index.ts` を tsx 経由起動。CLAUDE_CWD は明示的にプロジェクトディレクトリを指定）

isamu:
## Summary

System tasks (dev worklog / feed refresh / calendar sync) were registered directly on the
task-manager, which fires an interval schedule only on UTC-midnight-aligned boundaries. A
6-hour task could therefore run only during the one tick minute at 00/06/12/18 UTC, and a
server that was off, asleep or restarting through that minute skipped the run **forever** —
nothing recorded that a window had passed, and nothing in the UI or API could say so.

They now go through `@mulmoclaude/core/scheduler`'s persistence adapter, the same path
MulmoClaude uses: state file, startup catch-up per each task's `missedRunPolicy`, execution
log. User tasks keep firing forward only (as in MulmoClaude), but every run is now recorded
with the turn's real outcome, and the read-only routes report all of it.

Closes #1581.

## Items to Confirm / Review

1. **Seeding first-run state is an addition to MulmoClaude's design, not a copy of it.**
   `computeCatchUpPlan` reads a task with no state entry as "just registered" and enumerates
   nothing — and since nothing is then written, the next boot reads it the same way. Without
   the seed the reported bug survives its own fix. The seed writes `lastRunAt` = boot time
   with `totalRuns: 0`, so nothing claims a run happened. It must run **before**
   `initScheduler`, because the adapter holds the state map in memory and rewrites the whole
   file on each save. **MulmoClaude has the same gap** (masked there by hourly tasks) — worth
   deciding whether this belongs upstream in the shared package.
2. **Every boot with a task now writes into the workspace**: `config/scheduler/state.json`
   plus a `data/scheduler/logs/` directory under `CLAUDE_CWD`. Feed refresh and calendar sync
   are always registered, so this is every boot in practice. Consistent with the existing
   workspace layout (`config/scheduler/tasks.json`, `data/wiki/`), but it is new files
   appearing in a user's project directory.
3. **Catch-up now runs work at startup.** A feed refresh whose window passed while the server
   was down fires at boot, which for an `ingest.kind: "agent"` collection dispatches a hidden
   worker. That is the intended meaning of catch-up and the engines soft-dedup on their own
   markers (`lastFetchedAt` / `lastSyncedAt`), but it is a behaviour change at startup.
4. **`spawnScheduledChat` no longer swallows a failed spawn.** It throws, and the dispatcher
   records and logs it. Previously a task that never once started was indistinguishable from
   one that had not come due.
5. **`GET /api/scheduler/tasks` changed shape** (additively): entries now carry `origin` and
   `state`, and system tasks appear in the list. Nothing in this repo consumed the route.
6. **Not in scope**: task CRUD routes and a tasks UI (`docs/mulmoclaude-parity.md` item 2),
   and the interval-anchoring quirk in the shared engine — catch-up windows are epoch-anchored
   while `isDue` is UTC-midnight-anchored. They agree for any interval dividing 24 h; an odd
   one like 5 h ticks only at 00:00 UTC and is otherwise reached by catch-up at boot.

## User Prompt

> https://github.com/receptron/mulmoterminal/issues/1581
> mulmoclaude だと task がうまくキューにはいるかなにかで、その時間 PC が起動して無くても動くように作ったはずだけど、mulmoterminal ではどうなっている？できれば mulmoclaude と同じ実装にしたい

Scope was then confirmed as full parity: catch-up, the read-only API surface with execution
state and logs, and run recording for user tasks.

## What MulmoClaude does, and what this host was missing

| | MulmoClaude | MulmoTerminal (before) | MulmoTerminal (this PR) |
|---|---|---|---|
| System tasks | `initScheduler()` — state, catch-up, log | `registerTask()` — nothing persisted | `initScheduler()`, same as MulmoClaude |
| Missed window | replayed per `missedRunPolicy` | skipped forever | replayed (`run-once` for the worklog) |
| User tasks | fire forward, run recorded | fire forward, nothing recorded | fire forward, run recorded |
| API | tasks + state, logs | tasks from `tasks.json` only | tasks + state, logs |

## Implementation

- **`server/backends/scheduler-adapter.ts`** (new) — binds `configureScheduler` to this host's
  workspace root, `writeFileAtomic` and logger, and drives seed → `initScheduler`. Mirrors
  MulmoClaude's file of the same name. Catch-up is not awaited on the boot path.
- **`server/backends/scheduler-state-seed.ts`** (new) — item 1 above.
- **`server/backends/scheduled-run.ts`** (new) — dispatch + record for user tasks, the
  counterpart of MulmoClaude's `scheduled-run.ts`. The verdict comes from the completion hook,
  so it is the turn's outcome rather than "the spawn returned".
- **`worklogSystemTask` returns `SystemTaskDef`** — `name` + `missedRunPolicy: run-once`.
  `run-once` because the batch's own window is `[lastRunAt, now]` from
  `config/scheduler/worklog-state.json`: one catch-up run already covers everything missed,
  and `run-all` would spawn several batches summarising the same period. The two core
  factories already returned `SystemTaskDef`; this host was discarding the extra fields.
- **`spawnScheduledWorker` gains an `onComplete` hook.** The hook map holds exactly ONE hook
  per session, so the recorder had to share the existing failed-worker hook rather than
  register a second and silently replace it.
- **Routes** shaped like MulmoClaude's `server/api/routes/schedulerTasks.ts`:
  `GET /api/scheduler/tasks` (system + user, each with `origin` and `state`) and
  `GET /api/scheduler/logs` (`since` / `taskId` / `limit`, capped at 500).

## Verification

- `yarn format`, `yarn lint` (0 errors), `yarn typecheck`, `yarn build`, `yarn test` — 9536
  passing.
- `test/server/backends/scheduler-adapter.spec.ts` runs the **real** shared package against a
  real workspace on disk: a first boot registers without firing; a second boot after 24 h down
  fires exactly one catch-up run, advances `state.json`, and writes a
  `trigger: "catch-up", result: "success"` log entry.
- `test/server/backends/scheduler-state-seed.spec.ts` pins the bug and the fix side by side:
  `computeCatchUpPlan` owes nothing on an unseeded task and owes the slept-through 06:00
  window on a seeded one.

Plan: `plans/fix-1581-scheduler-catchup.md`.

work in mulmoterminal

## Summary by CodeRabbit

* **New Features**
  * Scheduled tasks now record execution status, history, timing, errors, and session details.
  * Missed scheduled runs are consolidated into a catch-up run when the server starts.
  * Scheduler task status, execution history, and log locations are available through read-only routes.
  * Worklog catch-up runs cover the full missed interval.

* **Documentation**
  * Updated scheduler behavior, parity documentation, and implementation planning details.

### DIFF
--- a/README.md
+++ b/README.md
@@ -693,8 +693,9 @@ work. It reads and reconciles progress against `vision.md` / `milestones.md` (cr
 empty ones if absent) so a long-running goal isn't forgotten.
 
 A run due while the server was **off** is not lost: built-in scheduled tasks record their
-runs in `config/scheduler/state.json` and **catch up at startup** — one run covering
-everything since the last, so several missed windows don't become several batches. Whether
+runs in `config/scheduler/state.json` and **catch up at startup**, each by its own missed-run
+policy. The worklog's is `run-once` — a single run covering everything since the last, so
+several missed windows don't become several batches summarising the same period. Whether
 they ran, and when they run next, is in `GET /api/scheduler/tasks`; the history is in
 `GET /api/scheduler/logs` (`?taskId=&since=&limit=`, newest first) and on disk under
 `data/scheduler/logs/`.
--- a/README.md
+++ b/README.md
@@ -692,6 +692,13 @@ and not built*. The window is **since the last run** (tracked in
 work. It reads and reconciles progress against `vision.md` / `milestones.md` (creating
 empty ones if absent) so a long-running goal isn't forgotten.
 
+A run due while the server was **off** is not lost: built-in scheduled tasks record their
+runs in `config/scheduler/state.json` and **catch up at startup** — one run covering
+everything since the last, so several missed windows don't become several batches. Whether
+they ran, and when they run next, is in `GET /api/scheduler/tasks`; the history is in
+`GET /api/scheduler/logs` (`?taskId=&since=&limit=`, newest first) and on disk under
+`data/scheduler/logs/`.
+
 Output lands in the wiki: one **weekly page** per ISO week
 (`data/wiki/pages/dev-log-YYYY-www.md` — filenames are lowercase, or the wiki can't open
 them), each tagged `worklog`. To browse them, open the **作業ログ 一覧** hub page
--- a/docs/mulmoclaude-parity.md
+++ b/docs/mulmoclaude-parity.md
@@ -113,8 +113,8 @@ until there's a concrete need.
 scheduled tasks.
 
 **What MulmoTerminal has:** the engine, plus the same persistence adapter since
-#1581 — system tasks catch up at startup for windows missed while the server was
-off, every run is recorded, and the read-only routes (`GET /api/scheduler/tasks`
+issue #1581 — system tasks catch up at startup for windows missed while the server
+was off, every run is recorded, and the read-only routes (`GET /api/scheduler/tasks`
 with each task's state, `GET /api/scheduler/logs`) mirror MulmoClaude's shapes.
 What is missing is the write half: creating or changing a task means editing
 `config/scheduler/tasks.json` by hand (or asking the agent to edit it, which
--- a/docs/mulmoclaude-parity.md
+++ b/docs/mulmoclaude-parity.md
@@ -112,11 +112,13 @@ until there's a concrete need.
 **What MulmoClaude has:** routes and a panel to create / edit / delete
 scheduled tasks.
 
-**What MulmoTerminal has:** the engine only. Tasks in
-`config/scheduler/tasks.json` load at boot and fire correctly, but the sole
-route is a read-only `GET /api/scheduler/tasks` — creating or changing a task
-means editing the JSON by hand (or asking the agent to edit it, which works
-fine and is the current de-facto CRUD).
+**What MulmoTerminal has:** the engine, plus the same persistence adapter since
+#1581 — system tasks catch up at startup for windows missed while the server was
+off, every run is recorded, and the read-only routes (`GET /api/scheduler/tasks`
+with each task's state, `GET /api/scheduler/logs`) mirror MulmoClaude's shapes.
+What is missing is the write half: creating or changing a task means editing
+`config/scheduler/tasks.json` by hand (or asking the agent to edit it, which
+works fine and is the current de-facto CRUD), and there is no tasks view.
 
 **Picking it up means:** write routes over the same validated
 `buildUserTaskDefinitions` path, plus a small tasks view. Pure convenience on
--- a/plans/fix-1581-scheduler-catchup.md
+++ b/plans/fix-1581-scheduler-catchup.md
@@ -1,0 +1,78 @@
+# fix #1581 — system tasks need persistence + startup catch-up
+
+## What is broken
+
+`initUserTaskScheduler` (`server/backends/scheduler.ts`) creates a task-manager and calls
+`registerTask()` for system tasks and user tasks alike. It never calls
+`configureScheduler()` / `initScheduler()` from `@mulmoclaude/core/scheduler`, so there is
+no state file, no execution log, and no catch-up.
+
+The task-manager's `isDue()` fires an interval schedule only on UTC-midnight-aligned
+boundaries. A 6-hour worklog can therefore fire only during the one tick minute at
+00:00 / 06:00 / 12:00 / 18:00 UTC. A server started with `npx` — off overnight, asleep,
+restarted — misses that minute and the run is skipped **forever**: nothing records that a
+window passed, so nothing can make it up. The reporter had `worklogEnabled: true` for days
+with no page, no state file and no error.
+
+## What MulmoClaude does (the target)
+
+`server/index.ts:1246` → `server/events/scheduler-adapter.ts` → `initScheduler()`:
+
+1. loads `config/scheduler/state.json` (per-task `lastRunAt`),
+2. `computeCatchUpPlan()` enumerates every window missed since `lastRunAt`,
+3. runs them per the task's `missedRunPolicy` (`skip` / `run-once` / `run-all`, capped at 24),
+4. registers the tasks on the task-manager for ongoing ticks,
+5. records every run — state + a log line under `data/scheduler/logs/`.
+
+User and skill tasks there are NOT caught up (they register directly on the task-manager),
+but each run is recorded through `recordExternalRun` so history and last/next-run exist.
+
+## Plan
+
+Adopt that path. The shared package is already a dependency (`@mulmoclaude/core@^3.0.0`,
+`@receptron/task-scheduler@^1.0.3`) and `feedRefreshTaskDef()` / `googleCalendarSyncTaskDef()`
+already return the richer `SystemTaskDef` — this host was throwing the extra fields away.
+
+1. **`server/backends/scheduler-adapter.ts`** (new) — bind `configureScheduler` to this host's
+   workspace root, `writeFileAtomic` and logger, mirroring MulmoClaude's file of the same name.
+2. **Seed never-run tasks before `initScheduler` loads the state file**
+   (`server/backends/scheduler-state-seed.ts`, new). `computeCatchUpPlan` treats a task with no
+   state as "just registered" and catches nothing up — and since nothing is then written, the
+   next boot is "just registered" again. A 6-hour task on a laptop would stay in that loop
+   forever, which is the reported bug surviving its own fix. Seeding `lastRunAt` = boot time
+   (with `totalRuns: 0`, so nothing claims a run happened) starts the accounting, and the boot
+   after a missed window catches it up. Seeding runs BEFORE `initScheduler` because the adapter
+   holds the state map in memory and rewrites the whole file on each save — a later external
+   write would be clobbered.
+3. **`worklogSystemTask` returns `SystemTaskDef`** — `name` + `missedRunPolicy: run-once`.
+   `run-once` because the batch's own window is `[lastRunAt, now]` from
+   `config/scheduler/worklog-state.json`: one catch-up run covers everything missed.
+4. **`initUserTaskScheduler`** registers user tasks directly (as MulmoClaude does) and hands the
+   system tasks to `initScheduler`, fire-and-forget so a slow catch-up cannot block boot.
+5. **User-task runs are recorded** (`server/backends/scheduled-run.ts`, new) via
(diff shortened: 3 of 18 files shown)


isamu:
## Review triage (iteration 1)

**Applied**

- **Codex P1 / CodeRabbit — tick loop started before the system tasks were registered.** Real, and the reasoning is right: `initScheduler` registers only after seeding and running the catch-up plan, so a tick in that gap saw an empty registry. Fixed — the loop now starts once registration has finished, and still starts if the adapter fails so the user's own tasks keep running. (One correction to the finding: such a window was *not* unrecoverable — `lastRunAt` had already advanced past it, so the next boot's catch-up would have picked it up. The delay was the problem, not permanent loss.)
- **Codex P1 / CodeRabbit — the worklog was filed as successful at spawn.** Real. It now files the actual failure from the completion hook, with the failing session id.
  **But it still does not await the turn, deliberately.** The engine's `makeGuardedTick` drops every tick that overlaps a still-running one, so resolving `run` only when an LLM batch finishes would silence the *entire* scheduler — feed refresh and calendar sync included — for the length of that batch. So: `run` returns at the spawn, the adapter files that as the run, and a turn that then dies is recorded afterwards. The cost is one extra `totalRuns` on a failed run, which is the cheaper of the two wrong numbers. Regression tests cover both the failed and the finished turn.
- **CodeRabbit — markdownlint MD018** (`#1581` at line start read as a heading) and the **README wording** that described the worklog's `run-once` policy as if every task shared it. Both fixed.

**Not applied**

- **CodeRabbit (Critical) — "duplicate block-scoped declarations" in `scheduler.spec.ts` L283-284 and `scheduled-chat.spec.ts` L85-86.** False positive: it read the diff's removed and added lines as both present. Each `const` appears once in the file; `yarn typecheck` (`vue-tsc -b`, whole repo) and all 9541 tests are green.

Local: format / lint (0 errors) / typecheck / build / test all pass.

work in mulmoterminal

isamu:
## Follow-up: a hole the previous fix opened (found in my own review, not flagged by a bot)

Waiting for the adapter before ticking — the fix for Codex's P1 — means a catch-up run that
never returns leaves the tick loop unstarted, so the **user's** tasks never fire either. A
catch-up run is real network work, so that is reachable, and it trades a bounded gap for an
unbounded one.

The loop now starts anyway one tick later if the adapter is still going. `start()` ignores a
second call, so whichever happens first wins: the normal case (seed plus a catch-up run,
milliseconds) always registers everything first, and a wedged catch-up costs one tick rather
than every tick.

work in mulmoterminal

isamu:
## Review triage (iteration 2) — the tick-loop ordering, settled

Codex has now flagged **both sides** of the same trade-off, one after the other:

- iter 1: "system tasks can miss a tick while the adapter is still registering them"
- iter 2: "user schedules can be permanently skipped while startup catch-up blocks the loop"
- iter 3: "...can still be lost while catch-up delays the loop for up to 60 seconds"

Both are true, and only one can be protected: the adapter registers its tasks only after seeding
and running the catch-up plan, so either the loop ticks first (system tasks briefly absent) or it
waits (no user task fires meanwhile). The 60 s grace timer I tried in between narrowed the
user-facing loss without removing it.

**What settles it is that the two kinds are not symmetric.** A user task is one-shot — "remind me
at 09:00", with no catch-up in either host — so a tick it misses is a reminder that never happens.
A system task's *work* survives a missed run either way: the worklog's window is `[lastRunAt, now]`
from its own state file, and the feed / calendar engines re-derive what is due from their own
markers. A system window landing inside a slow catch-up therefore costs one run, not the work.

So the loop starts as soon as anything is registered, and the adapter registers behind it. This is
also what MulmoClaude does (`initScheduler(...)` un-awaited, then `taskManager.start()`), so the
two hosts stay identical here. The reasoning is in `startTicking`'s comment and pinned by a test,
so the next reader does not have to re-derive it.

### On the worklog's two log entries

Also still flagged: a failed worklog turn is recorded as a dispatch success plus a later failure
rather than one execution with its real outcome. That is accurate, and it is the deliberate cost —
`SystemTaskDef.run` would have to await the batch to file one honest entry, and the engine drops
every tick that overlaps a still-running one, so awaiting an LLM session would silence the whole
scheduler (feed refresh and calendar sync included) for minutes. Two entries whose *last* word is
the truth beats one entry that is a lie, and beats a scheduler that stops ticking.

Local: format / lint (0 errors) / typecheck / build / test (9541) all pass.

work in mulmoterminal

isamu:
## CI was red on ubuntu — my own spec, not the change

`lint-and-build (ubuntu-latest)` failed on `scheduler.spec.ts`: `initScheduler` called twice where
one was expected. Two of my tests ended while the adapter start was still in flight — the seed
ahead of `initScheduler` is real file I/O — so the call landed during a *later* test and inflated
its count. Green locally and on macOS, red on the slower Linux runner. Both tests now wait for the
call before finishing, which closes the in-flight window instead of widening a timeout.

## New finding this iteration: concurrent writes to `state.json`

Codex's latest verdict adds: two run completions can race when rewriting
`config/scheduler/state.json` and one update can be lost.

Accurate, and **deferred deliberately** — it belongs upstream, not here. `updateAndSave` mutates a
shared in-memory map and then rewrites the whole file with no lock, so two atomic renames can land
out of order. That is `@receptron/task-scheduler` / `@mulmoclaude/core`, and MulmoClaude has it
identically today (system, user and skill runs all write through the same adapter). This PR does
not add the pattern, it adopts it — as the parity request asked.

The consequence is also mild in both hosts: a lost `lastRunAt` means the next boot's catch-up
replays a window that already ran, and both engines that could act on it dedup on their own
markers — the worklog's window is `[lastRunAt, now]` from its own state file, feeds on
`lastFetchedAt`, calendar on `lastSyncedAt`. So the failure mode is one redundant run, not lost
work or a missed run.

Fixing it host-side is not possible without forking the adapter: the map and the writes are inside
the package. Worth an upstream issue against `@mulmoclaude/core` so both hosts get the fix.

work in mulmoterminal

PR #1585, opened 2026-08-09
Collections: the agent hears the portability verdict, the phone can list projects, and a new project's watchers start at once

snakajima:
Everything left on `plans/HANDOFF-collections-projects.md` that was not waiting on a package release.

## 1. The agent hears the verdict (§3.6's last gap)

The portability check had a route (#1582) and a button (#1584), and both need a person to think of asking. **The agent is the one that chose the storage kind, dropped the `primaryKey` or put the collection in user scope** — and it is holding the file open at the moment that is cheapest to fix.

A successful `putSchema` now carries a `portability` field beside `written: true`.

- **`putSchema` and nothing else**: it is the only action that can change what the check reads. Creation is a plain file write the engine never sees (`putSchema` *refuses* an unknown collection and says to create it by writing SKILL.md + schema.json), and the record actions cannot move a collection in or out of portability. Hooking a read would spend a git call per listing to repeat the same answer.
- **Quiet by construction**, so the agent never has to handle a new failure mode: a clean collection is not mentioned, a refusal is prose the agent is meant to act on and passes through untouched, and a check that cannot run changes nothing about a write that *did* happen.

This also answers §11.4's "at collection-creation time": there is nowhere to put a creation hook, because creation never reaches the host. The schema edit that follows is the first moment it can speak.

## 2. The phone can learn which projects exist (§3.4's other half)

#1584 taught the handlers to **resolve** a project a command names. Nothing could tell a phone which ids exist, so the parameter was unusable from the client that needs it.

`listCollectionProjects` returns `{ id, label }`, workspace first, and **deliberately not the `cwd`** the browser's listing carries — the browser needs the path to match a project to a cell it is already showing; the phone is genuinely remote, and an absolute root in a command or artifact publishes the user's home directory over the wire.

`docs/remote-host-protocol.md` is updated in the same commit (its own header requires it): the listing, `project?` on every command that takes one, and the three rules — opaque id never a path; omitted means the host's workspace; **an unresolvable id is an error, not a fallback**.

Still not in this repo: the phone's own picker.

## 3. A new project's watchers start when it is saved (§3.1's known lag)

The saved directories are the roots the watchers mount for, and only a 60s poll reconciled them — long enough that a new project's first collection reads as broken rather than as pending.

`POST /api/config` now reports when the list actually **moved** and `app-routes` pulls the sync forward; the poll stays as the fallback for anything that changes the file behind our back.

- Compared **by path**, so a relabelled preset (same directory to the watchers and to the project ids) wakes nothing.
- Compared against the config just read from **disk**, not this instance's cached copy: another mulmoterminal sharing the file may have written since we booted — the same reason the merge base is re-read. A stale comparison both misses a change and invents one; the spec fails on the cached form.
- The callback is **injected**, not imported: the config module stays config-shaped.

## Verification

`yarn format / lint / typecheck / build / test` — 9586 passing (+42), 0 lint errors.

Not covered here, and unchanged: §3.3 and §3.5 wait on `@mulmoclaude/core` / `@mulmoclaude/collection-plugin` releases (written up in `../mulmoclaude/plans/feat-collection-multi-root-3.md`), and the Collections **pane** still has no live browser check — that needs a session reporting the `data` MCP group, i.e. a real agent.

work in mulmoterminal

## Summary by CodeRabbit

- **New Features**
  - Remote connections can list available projects with opaque IDs and labels.
  - Collection, feed, skill, and view commands support optional project selection.
  - Successful collection schema saves report portability findings when applicable.
  - Collection watchers synchronize automatically when saved directory settings change.
  - Saved directory presets can be added or removed without overwriting concurrent updates.

- **Bug Fixes**
  - Project-scoped collection views no longer remove workspace-pinned shortcuts.
  - Concurrent configuration changes are handled more safely.

- **Documentation**
  - Added guidance for project discovery, project-scoped commands, workspace defaults, and related errors.

### DIFF
--- a/common/pathSegments.ts
+++ b/common/pathSegments.ts
@@ -1,0 +1,23 @@
+// Splitting a filesystem path into its parts, on EITHER separator.
+//
+// `path.sep` is not the answer here: these paths come from a config file and from a saved
+// directory list, so a Windows path can be read on a mac (a synced config, a bug report, a test
+// fixture) and a POSIX one on Windows. Code that splits on "/" alone does not fail loudly on
+// `C:\Users\me\project` — it returns the WHOLE STRING as one segment, which is how a "label"
+// becomes an absolute path.
+//
+// That is a privacy contract in one place: the phone's project listing sends a label and never a
+// path, precisely so a command or an artifact cannot publish the user's home directory. A
+// one-separator split turns that guarantee off on Windows without changing a line of the code
+// that states it.
+
+/** The non-empty parts of a path, splitting on `/` and `\` alike. A drive letter (`C:`) stays a
+ *  segment of its own — it is not a directory name, and nothing here wants it as one. */
+export function pathSegments(value: string): string[] {
+  return value.split(/[/\\]+/).filter((segment) => segment.length > 0);
+}
+
+/** The last part of a path — its directory or file name — or the input when it has none. */
+export function lastSegment(value: string): string {
+  return pathSegments(value).at(-1) ?? value;
+}
--- a/docs/remote-host-protocol.md
+++ b/docs/remote-host-protocol.md
@@ -41,16 +41,40 @@ grouped in `handlers/terminalSession.ts`.
 | `startChat` | `message`, `attachments?` | `{ started: true, chatId }` |
 | `listIssues` | — | `{ repos: RepoIssueRows[] }` |
 | `startIssueWork` | `repo`, `issue`, `run?` | `{ started: true, sessionId, branch, issue, outcome, ran }` |
-| `listFeeds` | — | `{ feeds }` |
-| `getFeed` | `slug`, `offset?`, `limit?` | the feed page |
-| `listCollections` | — | `{ collections }` (feed-backed ones excluded) |
-| `getCollection` | `slug`, `offset?`, `limit?` | one page of the collection's items |
+| `listFeeds` | `project?` | `{ feeds }` |
+| `getFeed` | `slug`, `project?`, `offset?`, `limit?` | the feed page |
+| `listCollectionProjects` | — | `{ projects: { id, label }[] }` — workspace first |
+| `listCollections` | `project?` | `{ collections }` (feed-backed ones excluded) |
+| `getCollection` | `slug`, `project?`, `offset?`, `limit?` | one page of the collection's items |
 | `listShortcuts` | — | `{ shortcuts }` |
-| `listSkills` | — | `{ skills }` (collection slugs excluded) |
+| `listSkills` | `project?` | `{ skills }` (collection slugs excluded) |
 | `listAccountingBooks` | — | `{ books: { id, name }[] }` |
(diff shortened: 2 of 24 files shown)


snakajima:
## #1587 folded in, and the four findings addressed

**#1587 is now part of this PR** (`810c8693`) and closed. It carries two data-loss fixes found while investigating a report that pinned favourites and saved directories had shrunk — both the same shape as each other, and one of them a regression from #1573:

1. **Pinned shortcuts deleted by a project-scoped index.** `shortcuts.json` is workspace-global (shared with MulmoClaude); the collection index that triggers `reconcileShortcuts` fetches a list scoped to the surface's project. Opening the Collections pane on a project deleted **21 pinned collections**, in both apps, silently. A project-scoped answer now reconciles nothing, and the rule is stated at all three layers.
2. **Saved directories deleted by a client that had not loaded them.** `recordPreset` sent `presets.value` plus one entry, and that ref is empty until the first GET lands. A launch during that GET reduced five directories to one — taking four projects' collections out of service. Recording is now a one-entry mutation applied server-side (`/api/config/cwd-presets/record` and `/remove`), so the caller never holds the list.

### The findings

**Codex — compare against this process's cached state.** Correct, and the reasoning is now in the code: every mulmoterminal shares the file and a POST re-reads it as the merge base, so an unrelated write can move *this* process from one project list to another. Disk-vs-next sees nothing there while `config = next` adopts the other instance's directories, and the watchers — which read the in-memory list — never hear about the projects they now serve. Both call sites compare against the previous in-memory list, and a test pins the absorb case (it fails on the disk-vs-disk form).

The fixtures moved with it: they now write the config file **before** importing the module, which is the order production has. Writing it afterwards left the module's view empty while the disk was full — a state the server is never in, and one that made the corrected comparison look wrong when it was right.

**CodeRabbit — isolate callback failures from the response.** Correct; the write has already landed when the subscriber runs. `notifyPresetsChanged` swallows a synchronous throw and a rejected promise, the type admits the async form, and two tests assert a 200 (plus that the save is real, not merely reported).

**Codex — duplicate project labels.** Correct, and not hypothetical: this repo's author has `~/mulmoclaude` and `~/git/ai/mulmoclaude`, both labelled "mulmoclaude". The browser tells them apart by the `cwd` it also receives; the phone deliberately gets none, so two identical rows are a coin toss. A duplicate now borrows the least it needs from its parents — `me/mulmoclaude` vs `ai/mulmoclaude` — capped at two segments so this never walks up to the home directory the no-path rule exists for, with an id fragment as the last resort.

**CodeRabbit — cover the registration.** Added: the command table is exercised through `createRemoteHostHandlers`, not just the handler. That registration is one line in a twenty-entry object — the kind a merge drops with nothing failing.

### Verification

`format / lint / typecheck / build / test` — 9612 passing, 0 lint errors. One unrelated spec (`files-browse`) failed once under full-suite load and passed in isolation; that is this repo's known flake, mentioned rather than quietly re-run.

PR #1587, opened 2026-08-09
fix: two ways a global list was being deleted by a partial view of it

snakajima:
Two data-loss bugs, found because the user noticed their pinned favourites and their saved directories had shrunk. Both are the same shape: **a list that lives in one shared file, replaced wholesale from whatever one client happened to hold.**

## 1. Pinned shortcuts, deleted by a project-scoped index — a regression from #1573

`<workspace>/config/shortcuts.json` is workspace-global and shared with MulmoClaude. The collection index that triggers `reconcileShortcuts` fetches a list scoped to the **surface's project** (`withActiveProject`). Reconciling one against the other says "every collection the workspace pinned is gone" — and `reconcile` prunes and **persists**.

Opening the Collections pane on a project with one collection **deleted 21 pinned collections**, in both apps, with no undo and nothing on screen to say so. The feeds survived only because they are a different `kind`.

A project-scoped answer now reconciles **nothing**. Not "reconcile the project's pins instead": pins are not per project, and pretending otherwise deletes the workspace's on the way. The rule is stated at all three layers — the binding that knows the scope, the pure function that does the pruning, and the store that owns the file.

## 2. Saved directories, deleted by a client that had not loaded them yet

`cwdPresets` is a replace-all field in a file every mulmoterminal on the machine shares, and it decides which projects the server serves collections for. `recordPreset` sent `presets.value` plus one entry — and that ref is **empty until the first GET lands**, empty again if it fails, and stale the moment another instance saves.

A terminal launched during that GET persisted `[the one just launched]`: **five saved directories became one**, and four projects' collections stopped being served.

Recording is inherently "add this, keep the rest", so it is now expressed that way — `POST /api/config/cwd-presets/record` and `/remove` apply **one entry** to the list on disk:

- a caller cannot clobber a list it never holds;
- two instances recording different directories no longer race;
- **#164's guarantee survives** — a launch during the initial GET is still recorded immediately, because there is nothing to wait for. (Waiting was tried first; it breaks that ticket's spec, which is what sent me looking for the right fix.)
- the replace-all `POST /api/config` stays for a settings-UI reorder, where the whole list *is* the intent.

## About the tests

The preset spec's fake server now **models** the list instead of echoing what was posted. An echoing fake cannot tell this fix from the bug it replaces — which is part of why the bug was there — and the new tests ("records a directory WITHOUT deleting the ones this tab has never seen") fail against the old implementation.

## Verification

`yarn format / lint / typecheck / build / test` — 9605 passing (+19), 0 lint errors. Each fix was also checked by reverting it and watching the new tests fail.

The user's two files were restored after the fix landed locally (25 shortcuts back; the five directories back), and mag2's collection is served again — confirmed against the running instance.

work in mulmoterminal

## Summary by CodeRabbit

* **New Features**
  * Added project-scoped remote access for feeds, collections, skills, and remote views.
  * Added project discovery with opaque IDs and readable labels.
  * Added portability guidance after successful collection schema updates.
  * Added saved-directory preset recording and removal.

* **Improvements**
  * Invalid or unresolved project selections now fail clearly instead of falling back.
  * Collection watchers synchronize after saved-directory changes.
  * Workspace shortcuts are protected from project-scoped changes.
  * Preset updates better preserve concurrent and server-side changes.

### DIFF
--- a/docs/remote-host-protocol.md
+++ b/docs/remote-host-protocol.md
@@ -41,16 +41,40 @@ grouped in `handlers/terminalSession.ts`.
 | `startChat` | `message`, `attachments?` | `{ started: true, chatId }` |
 | `listIssues` | — | `{ repos: RepoIssueRows[] }` |
 | `startIssueWork` | `repo`, `issue`, `run?` | `{ started: true, sessionId, branch, issue, outcome, ran }` |
-| `listFeeds` | — | `{ feeds }` |
-| `getFeed` | `slug`, `offset?`, `limit?` | the feed page |
-| `listCollections` | — | `{ collections }` (feed-backed ones excluded) |
-| `getCollection` | `slug`, `offset?`, `limit?` | one page of the collection's items |
+| `listFeeds` | `project?` | `{ feeds }` |
+| `getFeed` | `slug`, `project?`, `offset?`, `limit?` | the feed page |
+| `listCollectionProjects` | — | `{ projects: { id, label }[] }` — workspace first |
+| `listCollections` | `project?` | `{ collections }` (feed-backed ones excluded) |
+| `getCollection` | `slug`, `project?`, `offset?`, `limit?` | one page of the collection's items |
 | `listShortcuts` | — | `{ shortcuts }` |
-| `listSkills` | — | `{ skills }` (collection slugs excluded) |
+| `listSkills` | `project?` | `{ skills }` (collection slugs excluded) |
 | `listAccountingBooks` | — | `{ books: { id, name }[] }` |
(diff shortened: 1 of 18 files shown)


snakajima:
Folded into #1585 at the author's request — the two data-loss fixes (`810c8693`) are on that branch now, alongside the review fixes for it. Nothing here is dropped; this PR is closed only so one branch carries the lot.

Restoration already happened locally and is unaffected by which PR merges: the workspace's 25 shortcuts are back and the five saved directories are back, with mag2's collection served again.

PR #1588, opened 2026-08-09
docs: 4.8.0 release notes and setup guide

isamu:
## Summary

The two repo-specific things a release needs beyond the npm publish and the GitHub release
(CLAUDE.md → *Publishing a release*), for **4.8.0**, already on npm as
[`mulmoterminal@4.8.0`](https://www.npmjs.com/package/mulmoterminal/v/4.8.0) and released as
[4.8.0](https://github.com/receptron/mulmoterminal/releases/tag/4.8.0).

- **`docs/ChangeLog.md`** — a 4.8.0 entry with the same per-PR detail as the GitHub release,
  covering #1573, #1580, #1582, #1578, #1579, #1583, #1410 and #1577, with the `> **Setup guide:**`
  line linking both language pages.
- **`docs/guide/{en,ja}/v4.8.0.md`** — the dated setup guide (snapshot of 2026-08-10). It leads with
  the two things a reader has to act on: **where the Collections pane is** (right pane of an
  enlarged cell, beside Canvas / Tools / Files, scoped to that cell's directory) and **how to
  confirm the dev worklog is finally running** — `config/scheduler/state.json`,
  `GET /api/scheduler/tasks`, and a `"trigger":"catch-up"` entry in the log. It also explains the
  UTC-window behaviour in JST, since that is what made the bug invisible.
- **nav_order** — the new pages take `1001` and all 44 older release pages per language shift down
  by one (now 1001–1045, no duplicates). Renumbered by enumerating `docs/guide/*/v*.md` rather than
  from a typed list, per CLAUDE.md.
- **Guide index banner** updated in both languages, which otherwise keeps advertising 4.7.6.

## Items to Confirm / Review

1. **No screenshot yet, deliberately.** CLAUDE.md requires one for anything the user sees, and the
   Collections pane qualifies. Capturing it needs a scratch `HOME` (so no real paths leak, as in
   #1410), a demo collection, and an **enlarged** cell in a directory whose tools include
   collections — the button is gated on both. Agreed with the maintainer to ship the text now and
   add the screenshot in a follow-up PR.
2. **Wording of the worklog section.** It says a run is recorded at dispatch and a later failure is
   recorded afterwards, so "a task whose last entry is an error really did fail". That is the
   deliberate design from #1583 (awaiting the batch would stall the tick loop) — worth a check that
   it reads as a fact about the tool rather than an apology for it.
3. **Every internal link and anchor was verified** to resolve to a real page and anchor
   (`config.html#all-keys` replaced a `features.html#dev-worklog` that does not exist). The
   self-containment table and the `curl` lines match the routes as merged.

## User Prompt

> publish!!

(Continuing from the 4.8.0 release: npm publish, tag and GitHub release are done; this is the
documentation half.)

work in mulmoterminal

## Summary by CodeRabbit

- **New Features**
  - Documented per-cell, directory-scoped Collections panes and automatic refresh/completion notifications.
  - Added guidance for collection portability checks, workspace-only skill staging, and project-aware cards.
  - Documented persistent scheduled tasks, startup recovery for missed runs, execution logs, and task APIs.

- **Documentation**
  - Added English and Japanese 4.8.0 release guides.
  - Updated release announcements and navigation ordering across versioned guides.
  - Added setup instructions, verification steps, API references, and updated screenshots.


isamu:
## Added after feedback: "don't use the per-folder collection feature yet"

Feedback from use: these pages read as if collections-in-any-folder is finished, and it is not. The
guide (both languages), the guide index banner and the changelog entry now say to **keep collections
in the workspace**, and name what is missing rather than warning in the abstract:

- **No scheduled refresh for a project's collection** — written and reverted before release
  (`plans/HANDOFF-collections-projects.md` §3.5). For a collection that refreshes by dispatching an
  agent, the seed prompt carries a root-relative data path while the worker stands in the workspace,
  so a refresh scheduled for a project would have written into the **workspace's** same-named
  collection. A silent cross-project write is worse than no refresh; the fix needs an upstream
  change first.
- **The phone cannot reach a project's collection** — groundwork only, no picker (§3.4).
- **Two projects' cards cannot share one panel** — needs a `@mulmoclaude/collection-plugin` release
  built against core 3.1.0 (§3.3).
- **The pane has had no live browser check** the way the overlay has (§3.7).

The [4.8.0 release notes](https://github.com/receptron/mulmoterminal/releases/tag/4.8.0) carry the
same warning as a `> [!WARNING]` block at the top — the release is already out, so that is where
someone upgrading looks first.

The notice is explicit that a collection already in the workspace is unaffected, so it reads as
"do not build on this yet" rather than "your data is at risk".

work in mulmoterminal

PR #1613, opened 2026-08-10
docs(plans): Shareable collections design — 4 review rounds complete

snakajima:
## Summary

**One design document, no code changes** (`plans/feat-shareable-collections.md`, ~2000 lines).

A governance model for shipping web apps from **declarations committed to a git repository** — anyone can use them, no MulmoTerminal, no login required, and no code written. Starting point is mulmoclaude PR #2209 (draft, Firestore store), but the central finding is that **making collections shared is not a backend addition — it is a change of identity.**

The engine states as an invariant that `a slug is unique within a root and nowhere else`. A shared collection **has no root** (it exists neither on the other person's machine nor on the web).

## Key design decisions

- **The unit of sharing is the app (= the repository), not the collection.** `(root, slug)` becomes `(aid, cid)`; local and shared identities coexist as a discriminated union
- **Schema and views live in git, records in Firestore.** The schema in Firestore is not the truth — it is a published artifact
- **MulmoTerminal only.** MT's project scope is already designed around "the committed skill" in a "repo that is supposed to be self-contained" (`server/backends/collections.ts`)
- **The host is on the build path, never the run path.** Testable invariant: publish, shut the host down, exercise the site
- **Rules stay static; the ACL is data, not rules.** One generic rule reading the `members` map via `get()` covers any number of apps with no redeploy
- **HTML is a generated artifact; the schema is what gets governed.** Access control is enforced at the data layer, so the view need not be trusted and an LLM may generate it freely. Governance therefore moves to the schema layer, which makes a **schema linter** the main line of defense
- **View Bridge** — the sandboxed HTML never receives a Firestore handle; the parent pushes `onSnapshot` data over `postMessage`. It exposes declared dataset names and action names only, never a query language

## Four scenarios, chosen so their axes do not overlap

| | Axis | Time | Peers' answers | What matters most |
|---|---|---|---|---|
| Salon booking | resource × approval × side effect | async | hidden | reach (anyone can book) |
| Web survey | one-per-person × write-once × deadline | async | hidden | privacy |
| Online class quiz | live × staged reveal × asymmetric visibility | **now** | hidden | **concealment** (the answer key must not leak) |
| Council vote | record integrity × publicity | **now** | **visible** | **tamper-evidence** |

The quiz (hiding is the requirement) and the vote (hiding is the mistake) are exact opposites, which is what forces **the policy to be declared rather than assumed**. `gated`, `session`, `live`, `immutable`, `peerVisibility`, `idFrom`, `finalize`, `window`, `aggregate` and the `participant` role all fall out of this.

Table designs and schemas for all four are included as **LLM-facing templates** (17 JSON blocks, all machine-validated; existing vocabulary and keys this plan proposes are separated in a table).

Two limits are recorded as out of scope rather than left to be rediscovered: **"one response per person" and "anonymous to the owner" cannot both hold** without a server, and Firestore cannot hide fields — concealment requires a **document split** (`gated`).

## Review history

Four rounds of static review (Codex ×3, Gemini ×1) found and fixed **14 defects**. Every one of them looked correct on paper; **10 shared a single root cause** (reading an optional key without an `in` guard), and that class recurred *after* the discipline against it was written down. Commits are split per round, so what broke and what fixed it is legible in the diff.

## What to review

**These rules have never been executed.** Static judgement has been wrong four times running on this design, so a fifth "looks fine" is not evidence. A verification banner sits at the top of the document.

So the useful review is not of the rule text but of:

- whether the **premises** are read correctly (`server/backends/collections.ts`, the engine's identity invariant)
- whether **MulmoTerminal-only / per-folder / git-managed** is the right direction
- whether the **"decide before freezing the rules" checklist** is complete — `../mulmoserver` is effectively frozen infrastructure, so anything that cannot be expressed in the rules cannot become a product feature

## Next step

Implementation step 2: **emulator unit tests for the rules.** Most of the 14 findings would have failed on the first run of a real test. Beyond the four happy paths, the suite must include the four shapes every round of review came from: an app declaring none of the optional keys, a member with no `'*'` role, a gated document whose parent is missing, and a record lacking a declared field.

Refs: receptron/mulmoclaude#2196, receptron/mulmoclaude#2197, receptron/mulmoclaude#2209

work in mulmoterminal


snakajima:
Both findings were real and are fixed in 39f2dd26. They turned out to be the same shape: **a declaration the rules never re-derived.**

### 1. `/mail` bypassed the declared action model

The rule was "any writer on that `cid` may create", so an editor could send **arbitrary content to an arbitrary recipient** from the app's sending domain. The schema declares `to: {{customerEmail}}` and a template name; the rules read neither.

Publish now emits `collections[cid].mail` (`toField` / `templates` / `dataFields`), and the rule re-derives the declaration:

- recipient must equal the address **on the source record** (`exists()` + `get()` comparison)
- template must be in the declared set
- free-form `data` limited to declared keys, and the doc may carry nothing else

Stated as a general principle in the `then.email` section: **if the rules cannot re-derive a declaration, that declaration is advisory only.**

### 2. Public submit validated names, not values

Correct, and worse than it looks: for public submit **the client is the attacker**, so layer 2 of the "three layers of validation" (client-side `validateRecordObject`) does not exist at all.

The honest constraint is that **Firestore rules have no iteration**, so a `{field: type}` map cannot be walked — complete type checking of a dynamic schema is impossible, not merely unimplemented. So publish emits a **rule-checkable projection**:

```json
"validate": { "required": ["topicId", "choice"],
              "choiceField": "choice", "choiceValues": ["yes", "no", "abstain"] }
```

Enforced: no extra fields, required present, declared initial status, **the one enum that integrity depends on** (the ballot choice, the quiz answer), and a document size cap.

Not enforceable, now written down with how each is absorbed: arbitrary field types and multiple enums (excluded at aggregation + host-side `validateCollectionRecords` sweep), ref existence (the `get()` budget is 10, so broken refs surface as missing at render), volume (App Check — rules cannot express rate limits).

The design consequence is recorded too: a collection receiving public submissions is **quarantined** — it enters at `initialStatus`, is invisible to everyone but owner/editor and the submitter, and nothing downstream trusts it until approved. Salon's `pending → approved` already is this; for the survey/quiz/vote the aggregate excludes invalid records.

This is the second instance of D4 ("publish is a compilation step") — git schema in, a projection the rules can read out.

Worth noting for anyone reading later: this makes **15 defects across 5 review rounds, every one of which read as correct.** The verification banner at the top of the document still stands — these rules have not been executed once, and the emulator suite in implementation step 2 is the actual gate, not another round of reading.

work in mulmoterminal

snakajima:
All three are real. Fixed in abba6551.

### 1. `/mail` — my previous fix was only half of it

Correct, and this is the sharper version of the finding. Pinning recipient and template still let a writer enqueue `booking-approved` for **any booking in any state, repeatedly**. The action and the transition stayed advisory.

Two constraints now, taking your suggestion:

- **Deterministic queue id** — `mailId == {cid}_{itemId}_{template}`, so a second enqueue of the same action is a create on an existing document and is denied.
- **`getAfter()`** — the mail config declares `on: { "booking-approved": "approved" }`, and the rule requires the record to hold that status **after the same write**. The client must batch the record update and the enqueue together.

The framing that made it click: this inverts *"it was approved, so send"* into **"if you send, it must have been approved."** The `templates` array became an `on` map so the allowlist and the transition binding are one declaration.

### 2. `audience: "participant"` admitted every role

Correct — `roleIn(...) != null` includes owner, editor and viewer. The consequence is worse than it reads: **handing someone read-only access to a vote or quiz collection also handed them the ability to submit.** Now `== "participant"`. Owners and editors still write through the `writerOf` branch, so nothing legitimate regresses.

### 3. S2 declared `audience` with no `members`

Correct — `listed()` is false, so **every survey submission fails closed**, and the symptom is "submissions silently don't work" rather than a permission error.

Resolved by dropping `audience` from S2: the scenario is "anyone who logs in answers once", and `idFrom: "auth.uid"` already carries the one-per-person guarantee. The named-audience variant is now shown alongside it with the `members` + `memberEmails` that must accompany it, and the linter table gained a row for declaring one without the other.

Also guarded `app(aid).members` inside `listed()` while I was there — same class as the finding.

---

That is **18 defects over 6 rounds**, and this round included a case where my own fix from the previous round was incomplete rather than wrong. The pattern is consistent enough to be worth stating plainly: **every one of these read as correct, and none of them has ever been executed.** The verification banner at the top of the document is not boilerplate — the emulator suite in implementation step 2 is the gate, and it now has three more cases it must cover (batched enqueue + transition, viewer attempting a submit, `audience` without `members`).

work in mulmoterminal

snakajima:
Correct, and this closes the gap my last fix left open.

`getAfter()` proves the record **is** in the target state after the request. It does not prove **this request put it there** — so an already-approved booking could have its notification enqueued with no write to the source item at all. The deterministic id capped it at once, but the notification stayed independent of the declared transition, which was the whole point of the previous finding.

Fixed in 020684cf by requiring the *difference*:

```js
&& getAfter(srcItem()).data[statusField] == mailCfg().on[m().template]
&& get(srcItem()).data[statusField] != getAfter(srcItem()).data[statusField]
```

`get()` is the pre-write state and `getAfter()` the post-write state, so an inequality between them is proof that this batch performed the transition. Both reads are `in`-guarded on `statusField`.

The declaration now reads correctly in the direction that matters: **"if you send, this write must have approved it."**

Recorded in the `then.email` section as a two-stage correction — stage 1 stopped repeats, stage 2 proved causation — because the intermediate state looked complete and was not. That is the second time in this review series that a fix of mine was incomplete rather than wrong, which is a distinct failure mode from the earlier ones and worth naming: **a constraint that is necessary but insufficient reads exactly like a constraint that is correct.**

Nineteen findings over seven rounds now. Still zero executions.

work in mulmoterminal

snakajima:
Correct — and this one exposed something I had been leaving ambiguous, which is the more useful half of the finding.

**The immediate bug.** Firestore Rules do not coerce a string to a timestamp, so `request.time < "2026-09-30T23:59:59Z"` is a type error and the survey rejects every submission. Fixed in b7abc741: publish lowers the window endpoints to epoch milliseconds and the rule compares `request.time.toMillis()` against numbers.

Epoch millis rather than a Firestore `Timestamp`, deliberately: `app.json` is JSON and a `Timestamp` has no JSON representation. Numbers keep the authored and published forms writable in the same syntax, which is also what lets every JSON block in the samples stay machine-validated.

**The ambiguity underneath.** Your phrase "undocumented timestamp conversion" is exactly right, and the reason it was undocumented is that the plan never separated two things it had been showing as one:

- the **authored** `app.json`, committed to the repository and written by a human or an agent
- the **published** `apps/{aid}`, which publish derives and the rules actually read

So there is now a conversion table, and the samples say which of the two they are showing:

| authored (git) | published (Firestore) |
|---|---|
| `window.from` / `window.until` (ISO) | `window.fromMs` / `window.untilMs` (number) |
| each `schema.json` | `publishedSchema` + `collections[cid]` |
| `actions[].then.email` | `collections[cid].mail` |
| field definitions | `public.submit[cid].validate` |
| — | `publishedCommit` / `publishedBy` / `publishedAt` / `previousPublished` |

This is the third instance of the plan's D4 ("publish is a compilation step"), and it now carries a rule of its own: **every conversion goes in that table.** A single implicit one reproduces exactly this failure — a config that matches the documented sample and is rejected at runtime.

Twenty findings over eight rounds.

work in mulmoterminal

snakajima:
Both correct, and they share a root: **the declared state machine did not exist in the rules.** Fixed in 6ae9272d.

### P1 — notification did not enforce the declared source state

`writerOf` could update an item with no transition checks at all, so a writer could take a `cancelled` or `rejected` booking straight to `approved` inside the batch and enqueue `booking-approved`. The action's `require: { status in ["pending"] }` existed only in the schema.

Two changes:

- `mail.on` became `{ "booking-approved": { "from": ["pending"], "to": "approved" } }`, so the rule checks the **origin** as well as the destination. This also subsumes the earlier `get() != getAfter()` check, since `from` and `to` are disjoint by declaration (now a lint rule).
- Publish emits `collections[cid].transitions`, and `transitionOk()` applies it to **everyone including writers**. The state machine is no longer advisory for staff.

### P1 — customers could silently reschedule an approved booking

This is the one with real consequences for the scenario. `selfUpdate` was a flat list that never consulted the current status, so a customer could move `startAt` or `stylist` while `status` stayed `approved`. The schedule's busy query would immediately treat the new slot as taken, **the original slot would be released, and no stylist ever approved the change.**

`selfUpdate` is now keyed by current status:

```json
"selfUpdate": { "pending":  ["customerName", "startAt", "stylist"],
                "approved": ["customerName"] }
```

Rescheduling an approved booking is not a self-service edit — cancel and rebook, which routes back through approval. That is also the honest product answer, not just the enforceable one.

### The shared lesson, now written into the plan

**`require`, `then` and `selfTransitions` only take effect once publish lowers them into `transitions`.** Anything the compilation step does not lower stays advisory, no matter how clearly the schema states it. That is the same shape as the `then.email` finding two rounds ago and the timestamp finding one round ago — three consecutive findings that were all "a declaration publish never compiled."

The `authored → published` table gained `transitions`, and the linter table gained four rows, including "actions declare `require` but no `transitions` is published" — which is the check that would have caught this one statically.

Twenty-two findings over nine rounds.

work in mulmoterminal

snakajima:
Correct — S3 and S4 cannot be published as written. Fixed in the commit above, but in the opposite direction from the obvious one, and that turned out to be the useful part.

The obvious fix is to add `memberEmails` to both templates. I did not, because the previous round already went that way: I added a note to S2 saying "declare `members` and `memberEmails` together." That imposes **manual maintenance of a derived value** on whoever (or whatever) writes the config — which is precisely the duplication these review rounds have been eliminating everywhere else.

`memberEmails` exists solely so "apps I belong to" can be answered with `array-contains`. It is a denormalization of `members.keys()` and nothing else. So:

- **It is never authored.** Removed from the S1 sample; the S2 note is rewritten to say so explicitly.
- **Whatever writes `members` generates it** — publish, and the invite UI.
- `membersConsistent()` is reclassified in the document: it is an invariant that makes a divergent state **unwritable**, not a requirement placed on a human.

The `authored → published` table gained the row, and the linter gained "authored `app.json` hand-writes `memberEmails`" — which flags the version I would have shipped had I taken the direct fix.

### Why this one is worth noting

Rounds 7-9 were all "a declaration publish never compiled." This is the mirror image: **a derived value publish should have generated, handed to the author instead.** Both come from the same underlying weakness — the boundary between the authored `app.json` and the published `apps/{aid}` was blurry, which is what round 8 forced me to write a conversion table for. That table is now the place where this class of bug goes to die, in both directions.

Twenty-three findings over ten rounds.

work in mulmoterminal

snakajima:
Both correct.

### `session` was writable by editors

`writerOf(aid, '*')` includes `editor`, while the role table says only the owner drives `session`. A global editor could set `phase: "revealed"` — **exposing the answer key** — or reopen and close voting at will. Now `roleIn(aid, '*') == "owner"`, matching the documented model.

### Clearing the status bypassed the transition graph

This is the sharper one, because **the hole was the guard I wrote.** `transitionOk()` began with:

```js
|| curStatus() == null || nextStatus() == null
```

I wrote that as a null-safety guard. It is a two-move bypass: remove the status field (permitted, because `nextStatus()` is null), then set any status at all (permitted, because `curStatus()` is null). `pending → approved` becomes optional.

Now:

- removing or nulling a status is **denied** (`nextStatus() != null`)
- a record that already lacks a status — imported, migrated — may enter only through a declared recovery point, `transitions.initial`

The recovery point is not optional decoration: without it, a status-less record becomes permanently unwritable, so the linter gained "no `initial` declared" and "`initial` includes a terminal state."

### What I take from it

Since round 3 the discipline in this document has been "guard every optional key with `in`." This finding shows that discipline is only half a rule. **Guarding tells you the key may be absent; it does not tell you which way absence should fall.** I defaulted absence to `true` — permissive — inside a state machine, where the correct default is `false`. A guard that fails open is indistinguishable from a guard that fails closed when you are reading rather than running it.

The emulator suite gained two required cases: a write that removes the status field, and a global editor attempting to write `session`.

Twenty-five findings over eleven rounds.

work in mulmoterminal

snakajima:
Both correct.

### `ownRow()` was not bound to the declared ID strategy

The prefix branch applied regardless of `idFrom`. With `idFrom: "auto"` (S1), item IDs are arbitrary, so a verified attacker can create a booking with the ID `<victimUid>_anything` and **the victim acquires the self-update path on a record they never made**, with no `customerEmail` check involved.

Each branch is now bound to the strategy that declares it:

- `auth.uid` — exact equality
- `auth.uid+field` — the ID is **reconstructed** from the stored field and compared exactly, rather than matched as a prefix. This also drops `matches()`, so a UID containing regex metacharacters can no longer alter the meaning of the check.
- `auto` — ownership is `emailField` only, which is what the salon scenario actually means by "my booking"

### The mail rule had lost its "something changed" check

Also correct, and this one is a mistake worth naming precisely. When I introduced `from`/`to` two rounds ago I removed the explicit `get() != getAfter()` comparison, reasoning that `from` and `to` are disjoint by declaration — and I **made that disjointness a linter rule**.

That reasoning is wrong, and not marginally:

> **A linter cannot stand in for a rule.** The linter runs once, on the author's machine, against the authored declaration. The rule runs on every write. The moment a fact the rule depends on is moved into the linter, that fact stops being a guarantee and becomes an expectation.

The comparison is back, and both conditions are now required — actual change *and* declared origin/destination. The linter row stays, reworded: it now reports a meaningless configuration early rather than being the thing that keeps the invariant true.

Twenty-seven findings over twelve rounds.

---

A note on where this is heading, addressed to whoever reads this thread later rather than to the bot: every one of these twenty-seven findings has been correct, and **none of these rules has ever been executed.** Round 11 found a null-guard I wrote that failed open; this round found an invariant I demoted to a lint. Both are the kind of defect that reading reliably produces and running reliably catches. The verification banner at the top of the document, and implementation step 2, are the actual gate — not another round.

work in mulmoterminal

snakajima:
Correct, and this one goes at the scenario's premise rather than at its edges.

`voter` was in `createFields` with nothing binding it, so a council member could cast their UID-keyed vote **under another member's name**. Because the collection is `immutable` the misattribution can never be corrected, and because `peerVisibility: "public"` it is the official record every member reads. **A roll-call vote whose attribution is self-declared is not a roll-call vote.**

Fixed by declaring `emailField: "voter"` on the S4 submission (and typing `voter` as `email`).

Worth noting how small the fix is: **no new vocabulary was needed.** `authOk()` has enforced `request.resource.data[emailField] == email()` since the round where the salon's `customerEmail` was pinned. S1 used it from the start; S4 simply never did. The defect was not a missing mechanism but a template that failed to apply an existing one — which is exactly the failure mode the linter exists for, so it gained the row:

> `createFields` contains an identity field (`voter` / `author` / `submittedBy` …) not pinned by `emailField` → **records can be created under someone else's name**; with `immutable` + `peerVisibility: "public"` the misattribution is permanent and public.

The general rule is now stated in the plan: if a client writes a field that displays identity, pin it with `emailField`; if you will not pin it, do not carry it — derive identity from the document ID (`{uid}_…`). The plan also records the privacy consequence honestly: pinning makes members' email addresses visible to each other, which is appropriate for a council roll call, and the alternative (drop `voter`, resolve UID to display name through an owner-published roster in `apps/{aid}/config`) is written down for apps where it is not.

Twenty-eight findings over thirteen rounds.

work in mulmoterminal

snakajima:
Correct, and it cuts a distinction I had blurred.

S2's `q2` is a `number` that appears in `aggregate.by`, and nothing verified its value. A direct client submitting `q2: "not-a-score"` gets a record that passes the create rule and **becomes a group in the published results**.

I had filed this under "cannot be enforced — arbitrary field types" and moved on, absorbing it with "the aggregate excludes invalid records." That was too coarse. **Aggregation keys are a special subset of that category:** a type violation in an ordinary field makes one record odd, while a type violation in an aggregation key **corrupts the app's published output**. The first can be swept up afterwards; the second is already visible to every reader by the time anyone notices.

So the projection generalised from one checked enum to bounded key validation:

```json
"validate": {
  "required": ["topicId", "voter", "choice", "status"],
  "keyFields": [{ "field": "choice", "values": ["yes", "no", "abstain"] }]
}
```

Rules have no iteration, so `keyFields` is expanded by index with a hard cap of two. The cap costs almost nothing in practice, and for a reason worth stating: published results must fit a 1 MiB document, so **aggregation keys have to be low-cardinality anyway.** Grouping by an unbounded numeric field was never a viable design — S2's `q2` is now an `enum` of `"1".."5"`, which is what a 1-5 scale actually is.

The linter carries the invariant that keeps this true as apps evolve:

> every field in `aggregate.by` must appear in `validate.keyFields`, be pinned by `gateOn.match`, or be the `statusField`

S3 and S4 already satisfy it through the second clause — `questionId` and `topicId` are pinned to `session.current` by `gateOn`, so they cannot carry an arbitrary value even though they are not enums.

Twenty-nine findings over fourteen rounds.

work in mulmoterminal

snakajima:
Correct. `transitionOk()` was wired to `allow update` only, and the writer branch of `allow create` was unconditional — so the stylist editor in S1 can create a booking that is **already `approved`**, never passing through the declared `initial → pending` at all.

Fixed in 69a7614e with `initialOk()`, applied to **both** create branches, writer and public submit alike.

While there I removed a second-order hazard the fix exposed: `statusField` had two homes — the submit config (for validating `initialStatus`) and the mail config (for validating the transition). Two declarations of the same fact means that when they disagree, **a check silently stops applying rather than failing.** `collections[cid].statusField` is now the single source; the submit config carries only `initialStatus` (the value), and the mail rule reads the collection's field.

### What this one was

A state machine bolted on after the fact, in the ordinary way: **the edges were written and the entry point was not.** Saying "a state machine has an initial state" is trivial; the part that actually bites is that in Firestore rules the entry point and the edges live in two different operations, so `transitions` and `initial` end up enforced by two different clauses — and only one of them got written.

The linter picked up the three configurations that make this fail quietly:

- `transitions` without `initial` — now blocks **every create**, not just repairs of legacy records
- `initialStatus` not present in `transitions.initial` — every public submission rejected
- `transitions` declared without `collections[cid].statusField` — the state machine is disabled with no error anywhere

Thirty findings over fifteen rounds.

work in mulmoterminal

snakajima:
Correct, and this one lands on the plan's own thesis rather than on a detail.

The first branch of `allow create` was `writerOf(aid, cid)` with nothing else attached, so it skipped `authOk()`, `emailField`, `idFrom`, `gateOn` and `validate` entirely. In S4 that means **the chair — or any editor — can create a permanent, publicly attributed vote in another member's name, on any topic, after voting has closed.**

`immutable` was introduced several rounds ago on the argument that *"a voting system whose chair can rewrite votes is not a voting system."* Forging them is the same objection. **`immutable` guarantees that a record cannot be changed; it says nothing about whether it should have existed.** I had only implemented half of the claim I was making.

Fixed with `collections[cid].submitOnly`: where it is set, owner and editor cannot create records directly at all — every record must come through the submission path and satisfy identity, ID strategy, session gate and validation. Set on the survey, quiz and ballot collections. `validateOk()` was also lifted out into a function so value checking applies to **both** create branches rather than only to public submissions.

The chair still needs to vote, and that falls out of the existing role model rather than needing an exception: give them a collection-scoped role.

```json
"members": { "chair@council.jp": { "*": "owner", "votes": "participant" } }
```

`roleIn()` consults the collection key before the `*` fallback, so the chair is an owner everywhere and a participant in `votes` — subject to one vote per topic, the gate, and the identity pin, exactly like everyone else.

### The general form

> **A trusted role is a ceiling on authority, not an exemption from checks.**

The branch I wrote said "they are an owner, let them through," and in doing so it disabled every mechanism that existed *to constrain that owner* — immutability, attribution, the closing time. Trust decides what someone may do; it should not decide whether the rules run.

Thirty-one findings over sixteen rounds.

work in mulmoterminal

snakajima:
All checks green, and Codex returned LGTM against `5661376d`.

Final state: **eighteen commits, one file, 2521 lines added, nothing else in the repository touched.** Sixteen review rounds produced thirty-one findings, every one of them correct, and each round is a separate commit so the defect and its fix are legible side by side rather than flattened into the final text.

One thing green does not mean here, and it is the reason the banner at the top of the document exists:

> **These Firestore rules have never been executed.**

Sixteen rounds of careful reading concluded "no outstanding concern" — and reading was wrong sixteen times in a row before it was right once. The last four findings are a fair sample of what reading catches late: a null guard that failed open, an invariant demoted to a lint rule, a state machine wired to `update` but not `create`, and a trusted-role branch that disabled every protection built to constrain that role. All four read as correct until someone asked what a specific input would do.

So the gate is unchanged by this approval. **Implementation step 2 is the emulator suite**, and sixteen rounds have supplied its required cases:

- an app declaring none of the optional keys
- a member with no `'*'` role, only collection-scoped ones
- a gated document whose parent is missing
- a record lacking a declared field
- a write that removes the status field
- a global editor attempting to write `session`
- a viewer attempting to submit
- `audience: "participant"` with no `members`
- a batched enqueue with and without the declared transition
- an owner attempting to create in a `submitOnly` collection

If any of those pass on the first run, that is the surprise worth reporting.

work in mulmoterminal

Issue #1617, opened 2026-08-11
feat(chrome): make the header's status colours configurable, and keep the ink readable on a status wash


isamu:
Follow-up to #1591 (the part [that comment](https://github.com/receptron/mulmoterminal/issues/1591#issuecomment-5238654452) reported as still broken, and the "let the user customise it" half of [this reply](https://github.com/receptron/mulmoterminal/issues/1591#issuecomment-5247979209)).

## What happens

A cell whose directory sets `headerColor` + `headerTextColor` becomes unreadable **while it is running**. Measured from the reporter's screenshot rather than described:

| Sampled | Colour | What it is |
| --- | --- | --- |
| header background, `idle` | `#8e44ad` | the directory's `headerColor` |
| header background, `working` | `#d6e4fb` | `--bg-selected` in the Daylight theme |
| the glyphs on it, `working` | `#f6fcff`–`#fcffff` | the declared `headerTextColor` (white) |

White on `#d6e4fb` is 1.15:1.

## Why

`HEADER_STATUS.working` **replaces** the header background (`src/components/cellStatusClasses.ts:31`), and `done` does the same. The ink does not follow:

- a **derived** ink is already dropped in those states — `headerStyleFor(bg, text, dirBackgroundShows)` takes `status === "idle"` as its third argument (`TerminalCell.vue:934`), and `cellHeaderStyle.ts:30` refuses to derive when the directory's background is not what is painted
- a **declared** `headerTextColor` is returned before that check ever runs (`cellHeaderStyle.ts:29`), so it lands on the status wash

`blocked` is safe by accident: it names `text-warn` and never includes `--cell-header-fg`.

So the directory's ink is right for the colour it was chosen against, and wrong for the two states that paint something else.

## What to do

Two parts. The first stops the unreadable case for everyone; the second is the flexibility asked for.

### 1. A declared ink follows the same rule as a derived one

`headerTextColor` applies while the directory's own background is what shows. In `working` / `done` the theme's ink is used on the theme's wash — the pair they were designed as.

### 2. Per-status header colours, configurable

New keys, valid in **both** `~/.mulmoterminal/config.json` (the default for every directory) and a directory's `.mulmoterminal.json` (which overrides it):

```jsonc
{
  "headerColor": "#8e44ad",            // unchanged: this IS the idle state
  "headerTextColor": "#ffffff",

  "headerStatusColors": {
    "working": { "background": "#6d28d9" },
    "done":    { "background": "#166534" },
[...]

PR #1619, opened 2026-08-11
feat(chrome): resolve the cell header's colours per attention status, and make them configurable

isamu:
## Summary

A cell header became unreadable while its session ran. `HEADER_STATUS.working` / `.done` **replace** the header background with the theme's status wash, but a declared `headerTextColor` was applied regardless — so an ink chosen against a dark `headerColor` landed on a pale wash. Measured off the reporter's own screenshot in #1591: white on `#d6e4fb`, **1.15:1**.

Two changes:

1. **A declared `headerTextColor` now follows the same rule a derived one already did** — it applies while the directory's own background is what shows. In `working` / `done` / `blocked` the header hands both colours back to the theme, which pairs its wash with its own ink.
2. **What a status paints is configurable instead**, in the global config and per directory:

```jsonc
{
  "headerColor": "#8e44ad",
  "headerStatusColors": {
    "working": "#6d28d9",                                  // background only — the ink is derived
    "done": { "background": "#166534" },
    "blocked": { "background": "#7c2d12", "text": "#ffe8a3" }
  },
  "headerStatusTint": "background"                         // or "none"
}
```

Omitting `text` derives an AA ink from the background, so naming one colour cannot produce an unreadable header. `headerStatusTint: "none"` keeps `headerColor` through `working` and `done`.

## Items to Confirm / Review

- [ ] **`"none"` deliberately does not reach `blocked`.** That is the state where nothing proceeds until the user answers, and a switch whose purpose is palette consistency taking its amber away is an accident. An explicit `headerStatusColors.blocked` is still honoured. Pinned by a spec — but it is a judgement call, and the opposite reading ("none means none") is defensible.
- [ ] **The default path is still below AA, and that is pre-existing.** See the measurement below: with nothing configured, the theme's `--text-dim` on its own status wash runs **1.90–3.10:1** across the four themes. This PR emits no variables at all in those states, so a configured directory now renders **byte-identically to an unconfigured one** — the low contrast is what every unconfigured cell already shows, not something introduced here. Raising the chip ink from `--text-dim` to `--text` while a wash owns the header would fix it for everyone; it changes what every user sees, so it is not in this PR. Say the word and I will open it.
- [ ] **`headerStatusColors` merges whole-key**, global vs. directory — a directory's block replaces the global one rather than merging into it. Same rule `mergedDirConfigRaw` states for the three config files, but worth a second opinion.
- [ ] `src/components/hexColor.ts` moved to `common/hexColor.ts` (3 importers updated). Both sides now decide from that regex; a third copy is what this avoids.

## Verification — measured, not reasoned

`yarn format` / `lint` (0 errors) / `typecheck` (0 errors) / `build` / `test` (**9700 passing**) are all green, but none of them proves a browser paints the right thing. So the built stylesheet was loaded into a real browser (puppeteer) and the computed colours of **the elements the report photographed** — the model badge and usage chip, whose ink is `var(--cell-header-fg, var(--text-dim))` — were measured against the header's computed background. 4 themes x 4 states x 4 rules = **64 renders**.

Daylight / working, the reporter's exact condition:

| | background | ink | contrast |
| --- | --- | --- | --- |
| 4.8.1 (measured from the screenshot) | `#d6e4fb` | `#ffffff` | **1.15:1** |
| this PR, nothing configured | `#d6e4fb` | `#8492a8` (`--text-dim`) | **2.46:1** |
| `headerStatusColors.working: "#ffe8a3"` | `#ffe8a3` | `#1b2430` (derived) | **12.91:1** |
| `headerStatusTint: "none"` | `#8e44ad` | `#ffffff` | **5.87:1** |

The same run confirmed the CSS regression risk: wrapping each wash in `var(--cell-header-bg, …)` resolves to the identical colour it did before — `working` = `--bg-selected`, `done` = `color-mix(--done 20%, --bg-panel)`, `blocked` = `--warn-bg-subtle` — and the nested `color-mix` fallback does survive Tailwind's arbitrary-value parser (checked in the built CSS as well).

The harness was deleted after the run, per CLAUDE.md.

Two bugs in the harness itself were caught and fixed before its numbers were trusted: it was matching `CELL_STATUS`'s four keys instead of `HEADER_STATUS`'s (so it measured the cell frame and called it the header), and it parsed Chrome's `color(srgb …)` output as `rgb()`, which reported a contrast ratio of 4.7 billion.

## User Prompt

> https://github.com/receptron/mulmoterminal/issues/1591#issuecomment-5238654452 これの続きだね。running などで header の色が変わると字の色とかぶって読めない問題はあるので、改善。ユーザが柔軟に指定できるのが良いので、それ含めて検討して、細かく設定できるようにして。
>
> 設定用の skill も更新してね

Three design questions were settled with the author before implementing: the default behaviour (drop the declared ink on a wash, rather than deriving from the live background or removing the wash entirely), the granularity (per-status background + text, plus a switch), and the location (global default + per-directory override).

## Implementation

`common/headerStatusColors.ts` holds the rules — the types, the sanitizers and `resolveHeaderPaint`, the pure function both sides ask. The layering follows `cockpitLines`: rules in `common/`, a singleton store on the client hydrated from `/api/config`, validation on the server.

`HEADER_STATUS`'s backgrounds became `bg-[var(--cell-header-bg, <the theme's wash>)]`, and `TerminalCell` emits that variable **only** for a state whose colour is known — so an unconfigured cell never sets it and the wash paints, while a configured one replaces the wash without a second class racing the first.

`headerStyleFor`'s third parameter (`dirBackgroundShows`) is gone: `TerminalCell` was its only `false` caller and now uses `headerStatusStyleFor(status, chrome)` instead, which leaves the old helper meaning one thing.

Docs: `mulmoterminal-dirs` owns the setting (both halves — global default and per-directory override), `mulmoterminal-config` gets one routing line, plus README and `docs/guide/{en,ja}`.

Plan: `plans/feat-header-status-colors.md`.

refs #1617, refs #1591

work in mulmoterminal2

## Summary by CodeRabbit

* **New Features**
  * Added configurable header colors for working, done, and blocked states.
  * Added optional readable text colors, with automatic contrast-based text when omitted.
  * Added controls to preserve idle header colors during working and done states.
  * Supports global defaults, per-directory overrides, worktree color adjustments, and theme fallbacks.

* **Documentation**
  * Updated English and Japanese configuration guides with the new settings and behavior.


isamu:
Pushed 7b4539c3 — a gap I found while the first CI run was going, which nothing in the tree would have caught.

**A project's `headerStatusColors` / `headerStatusTint` were not reaching its worktrees.** `inheritedWorktreeConfig` copies an explicit list of keys; adding a key to the loader and not to that list is silent, so a project that recoloured its running header kept it in its own cell and lost it in every worktree. Only someone looking at a worktree beside its parent would see it.

- The colours are **tinted** with the rest of the chrome (same 12-degree step per worktree), so a worktree's `working` state does not stay the parent's colour while everything around it moves.
- `headerStatusTint` is **inherited unchanged** — a mode has no hue to rotate.

And the guard that would have caught it: a spec that takes `DIR_CONFIG_KEYS` and asserts every key is inherited, tinted, or named in an explicit `DELIBERATELY_NOT_INHERITED` list. Leaving a key out is now a decision someone has to write down.

One number worth recording, because it is why the test was run rather than reasoned about: I hand-wrote `#7c4c12` as the 12-degree rotation of `#7c2d12` and the code produced `#7c4212`. Working it out properly — hue 15.3 degrees, plus 12, saturation and lightness untouched — the code is right and my arithmetic was not.

Local gate on the merged tree: lint 0 errors, typecheck 0 errors, build OK, **9703 tests passing**.

isamu:
### CodeRabbit triage

**Applied — the comment on `headerStyle` described a merge that does not exist** (fa420bba). It said the global defaults fall back per *status*; they fall back per *key*. A directory that names any `headerStatusColors` replaces the global block entire, so the statuses it leaves out fall through to the theme, not to the global entry for that status. The code and `mergeHeaderStatusColors`'s spec were already right — only the comment was wrong, which is the kind that survives a green CI.

**Skipped — `server/config/dir-config.ts` "remove the server-local DirChrome declaration".** There is no server-local declaration to remove. The file already imports the shared type and extends it:

```
server/config/dir-config.ts:10  import { EMPTY_DIR_CHROME, type DirChrome } from "../../common/dirChrome.js";
server/config/dir-config.ts:69  export interface DirConfig extends DirChrome {
server/config/dir-config.ts:112 export interface PublicDirConfig extends DirChrome {
```

The two new keys reach it through `DirChrome`, which is exactly the one wire contract the comment asks for. Nothing to change.

isamu:
### Codex triage — a real one, applied (6fca8497)

Codex caught a hole I had left open, and it is the PR's own bug re-entered through the switch meant to avoid it:

> With `headerStatusTint: "none"` and a directory that has `headerTextColor` but no valid `headerColor`, `resolveHeaderPaint` returned `{ background: null, text: <custom> }`.

The CSS then falls back to the theme's wash and keeps the directory's ink on top of it — which is #1591, at the same contrast. `"none"` means "keep the colour I chose", so it needs a colour to keep; with no usable `headerColor` there is nothing the ink was chosen against, and it now returns nothing at all.

`idle` is deliberately unaffected: there the ink sits on the theme's own panel, which is what it has always done and what a spec already pinned.

Regression test added for both shapes — a missing `headerColor` and an unparseable one.

### Note on the earlier CI failure

`lint-and-build` and `package-smoke` went red on 37d75d59 with four type errors in the spec I had just added, while my local `yarn typecheck` reported zero. The cause was `vue-tsc -b` reusing `node_modules/.tmp/*.tsbuildinfo`; a clean run reproduced CI exactly. Every gate since has been run with that cache cleared first.

Local gate on this commit, from a cleared cache: typecheck 0 errors, lint 0 errors, build OK, **9705 tests passing** — and the full suite run twice back to back, because one run in between reported a single failure in the file `yarn format` had just rewritten and no run since (3 targeted + 2 full) has reproduced it.

isamu:
CODEX VERDICT: LGTM

No outstanding concerns after reviewing the implementation, config/schema boundaries, worktree inheritance, client rendering, security sanitization, English/Japanese documentation, and focused tests. The focused suite passes (327 tests).

isamu:
CODEX VERDICT: LGTM

Follow-up review of f95dba42 and the current head: the added worktree status-colour test exercises the real write → loadDirConfig round trip and passes (18/18 focused tests). I found no outstanding correctness, edge-case, security, i18n, or test-assertion concerns.

src/composables/useAppConfig.ts
```diff
@@ -410,6 +411,9 @@ function applyGlobalSettings(c: Record<string, unknown>): void {
   setIssueWorkComments(c.issueWorkComments);
   // How far the cockpit roster clamps each line. Absent `cockpitLines` keeps 2/2/3.
   setCockpitLines(c.cockpitLines);
+  // What a header shows once a status replaces the directory's colour (#1617). The default for
+  // every directory; a `.mulmoterminal.json` naming either key outranks it per cell.
+  setHeaderStatusDefaults(c.headerStatusColors, c.headerStatusTint);
```
isamu, line 416: [P2] This does not exercise the global-config hydration path. The new cell tests call `setHeaderStatusDefaults(...)` directly, so they prove that `TerminalCell` reads the singleton but would still pass if this production call were deleted and `/api/config` stopped propagating `headerStatusColors` / `headerStatusTint` into the singleton. Please add a test that runs the `useAppConfig` config-load path (or otherwise verifies the `/api/config` response reaches the cell) so the “global config” claim and this hookup are break-covered.

isamu:

- Add coverage for the `/api/config` → `useAppConfig` → `setHeaderStatusDefaults` path; the new TerminalCell tests currently seed the singleton directly, so deleting the production hookup would go undetected.

isamu:
CODEX VERDICT: LGTM

Iteration-4 follow-up review: the new `/api/config` hydration test exercises `useAppConfig().loadConfig()` and asserts both singleton refs, so removing the production `setHeaderStatusDefaults` hookup makes it fail. The tint fallback test correctly primes the module-level singleton with `headerStatusTint: "none"`, then loads an omitted config and verifies reset to `DEFAULT_HEADER_STATUS_TINT`; the shared singleton behavior is covered rather than masked by direct setup. I found no remaining PR behavior whose deletion would evade the existing tests. Targeted tests pass: 3 files, 186 tests.

isamu:
## Codex cross-review — converged after 4 iterations

Both reviewers sign off. Every finding below was verified against the code before being accepted or declined.

| # | Codex verdict | Findings | What changed |
|---|---|---|---|
| 1 | LGTM | 0 | (mine) worktree writes a status shape no test read back through the loader |
| 2 | LGTM | 0 | (mine) global config never proven to reach a cell; tint fallback deletable with nothing failing |
| 3 | **CHANGES REQUESTED** | 1 (P2) | Codex: the cell tests prove the singleton is *read*, not that `/api/config` *fills* it |
| 4 | LGTM | 0 | — |

### The one Codex found, and why it mattered

Iteration 3 is the finding worth reading. My iteration-2 tests set the module singleton directly, so they proved `TerminalCell` reads it — and would all have stayed green if `setHeaderStatusDefaults(...)` were deleted from `loadConfig`. The entire global-config half of the feature could have shipped dead behind a green suite. It was also a defect in **my own** iteration-2 fix, one level above the gap I had just closed, which is exactly the hazard of fixes written mid-loop.

### Every fix break-verified

Each was confirmed by removing the production line and watching the intended test fail, then restoring:

- disable the worktree status write -> 3 tests fail
- `statusColors: dirConfig...` without the global merge -> the global-default test fails
- `tint: dirConfig...` without `?? globalHeaderStatusTint.value` -> the tint test fails (and, before it existed, **nothing** failed)
- remove `setHeaderStatusDefaults` from `loadConfig` -> the hydration test fails

### Declined, with reasons posted

CodeRabbit's "remove the server-local `DirChrome`" — there is none; `server/config/dir-config.ts:10` already imports the shared type and both interfaces extend it.

### Unrelated flake found and filed, not silenced

Two supertest specs this PR does not touch failed once each in local full runs and pass in isolation, each returning a status belonging to a *different* app (403->200, 400->404). Filed as #1626 with the evidence and a hypothesis (ephemeral-port collision across parallel workers). Not reproduced in CI, and CI is green here.

Final: 16 commits, all checks green, 9711 tests passing locally from a cleared `vue-tsc` build cache.

Issue #1618, opened 2026-08-11
CI: ci.yml caches yarn but the other three workflows do not


isamu:
## Summary

`ci.yml` already sets `cache:` on both of its `setup-node` steps, and it shows — `yarn install
--frozen-lockfile` runs in 20s on ubuntu and 44s on macOS. The gap is the **other** workflows, which
install the same tree with no cache configured.

| workflow | `setup-node` | `cache:` set | `actions/cache` |
|---|---:|---:|---:|
| `ci.yml` | 2 | **2** ✅ | 0 |
| `codex_review.yaml` | 1 | 0 | 1 |
| `dead-code-scan.yaml` | 1 | 0 | 0 |
| `windows-daily.yaml` | 1 | 0 | 2 |

`dead-code-scan.yaml` has no dependency caching of any kind. `codex_review.yaml` and
`windows-daily.yaml` use `actions/cache` for something but leave `setup-node` uncached, so the yarn
download folder is refetched on every run — and on Windows that is the expensive case.

For scale, from a recent `ci.yml` run where the cache **is** configured:

| job | total | `yarn install` |
|---|---:|---:|
| `lint-and-build (macos-latest)` | 623s | 44s |
| `lint-and-build (ubuntu-latest)` | 472s | 20s |
| `package-smoke` | 145s | 21s |

That is what a warm cache looks like; the uncached workflows are paying the cold price each time.

## What to add

On each `setup-node` step that lacks it:

```yaml
      - uses: actions/setup-node@v4
        with:
          node-version: ...
          cache: yarn
          cache-dependency-path: |
            yarn.lock
```

List every lockfile the repo has. Using the same list everywhere makes all jobs share one entry
rather than filling the cache with near-identical copies — including `ci.yml`, so its existing
cache warms the others.

## Things that are easy to get wrong

- **It shortens the fetch; it does not skip the step.** `cache: yarn` caches yarn's *download
  folder*, not `node_modules`. `yarn install` still runs and still links.
- **The first run after the change is a cache miss by definition.** It pays the save cost and shows
  no saving. Measure two or three runs later.
- **Windows is where this pays most and is also where a bad cache hurts most.** Caching
  `node_modules` directly can restore a tree built for another platform or Node version; caching the
  download folder cannot. Where `windows-daily.yaml` already uses `actions/cache`, check what it
  targets before adding more.

## How to verify rather than assume

Read step durations from the API instead of eyeballing the Actions UI:

```sh
[...]

PR #1621, opened 2026-08-11
ci: cache yarn in dead-code-scan (refs #1618)

isamu:
## Summary

`refs #1618`. The issue asked for `cache: yarn` on three workflows. Measurement says **one** of
them is a real gap; the other two are correct as they stand, and this PR leaves them alone.

- **`dead-code-scan.yaml`** — runs `yarn install` with no caching of any kind. Fixed here.
- **`codex_review.yaml`** — never runs `yarn install`. Untouched.
- **`windows-daily.yaml`** — already carries the opposite decision, in a comment, with a reason.
  Untouched.

Expected saving: **~20s per `dead-code-scan` run**, warm from the first run.

## Items to Confirm / Review

1. **The scope call is the main thing to check.** I am declining two thirds of what the issue
   asks for. The reasoning is in `plans/ci-1618-cache-yarn.md`; the short version is below. If
   you disagree on either, say so and I will add them.
2. **`refs #1618`, not `Closes`** — deliberately. Since this PR overrides the issue's stated
   request for two of its three targets, closing it should be your call, not the merge's.
3. **The saving is smaller than the issue projects** (~20s, not ~38s). If ~20s on a job that runs
   on most PRs is not worth a config line, this PR is fine to drop — the measurement is the
   durable part and it is recorded in the plan file either way.
4. **`cache-dependency-path` is left unset**, against the issue's advice to list every lockfile.
   The repo has exactly one, `yarn.lock` at the root, which is already setup-node's default.
5. Unrelated, found while measuring: **`windows-daily` has failed 12 runs in a row** (since
   ~Aug 9, the `Test` step). Not touched here, but it is why item 3 below could not be measured.

## User Prompt

> https://github.com/receptron/mulmoterminal/issues/1618

Follow-ups: asked for the findings in Japanese, then — on seeing that only `dead-code-scan` was a
genuine gap — confirmed the narrowed scope with "じゃぁそこだけ" (just that part).

### 1. `dead-code-scan.yaml` — the gap

`yarn install --frozen-lockfile`, no `cache:` on `setup-node`, no `actions/cache`. Fixed.

### 2. `codex_review.yaml` — never installs the project

There is no `yarn` anywhere in the file. It `npm install -g`s the Codex CLI and nothing else, and
its own review prompt states the arrangement outright:

> This checkout has NO project dependencies installed — only the codex CLI itself.

Caching yarn's download folder there restores a 1.6 GB tarball no step reads — roughly **-18s per
run for zero gain**. Its existing `~/.npm` cache is the one that matches what it does.

### 3. `windows-daily.yaml` — already decided, with a reason

The file records the decision in a comment:

> Deliberately no `cache: yarn`: restoring the yarn cache tar onto NTFS is slower than a fresh
> install. node_modules is cached directly instead.

It caches `node_modules` keyed on `matrix.node-version` + `hashFiles('yarn.lock')`, plus
`~/.cache/puppeteer`. The issue's concern that a `node_modules` cache could restore a tree built
for another platform or Node version does not apply — the runner is windows-only and the node
version is in the key.

Overturning a documented decision needs a Windows measurement, and the workflow is red on all of
its recent runs, so there is no clean baseline to measure against.

## The measurement

`setup-node` performs the cache **restore inside its own step**, so the honest comparison is
`setup-node + yarn install`, not `yarn install` alone. The issue's table quotes only the install
column, which overstates the gain by roughly 1.8x.

Medians, ubuntu, from `actions/runs/<id>/jobs`:

| | n | `setup-node` | `yarn install` | **sum** |
|---|---:|---:|---:|---:|
| `ci.yml` `lint-and-build (ubuntu-latest)` — cached | 12 | 20s | 20s | **38s** |
| `dead-code-scan` — uncached | 15 | 2s | 58s | **59s** |

Reading the install column alone suggests 58s -> 20s. The real figure is 59s -> ~40s, because
`setup-node` grows from 2s to ~20s pulling a 1.6 GB tarball.

## No cold first run

The issue warns that the first run after the change is a cache miss by definition. It is not,
here. The restore key `ci.yml` logs is:

```
node-cache-Linux-x64-yarn-3aa92fc03863d9b5c690db69623c75d198b7dac5560ca7f7e68671611ba15374
```

Platform, arch, package manager, lockfile hash — **no workflow name, no node version**.
`dead-code-scan` is node 22 on ubuntu against the same single root `yarn.lock`, so it computes
that exact key and hits what `ci.yml` already maintains on main. Nothing extra is stored; the
entry is shared, not duplicated.

## Verifying after merge

Compare the same two step names across several runs, not one pair:

```sh
gh api "repos/receptron/mulmoterminal/actions/runs/<RUN_ID>/jobs?per_page=50" \
  --jq '.jobs[] | "\(.name)", (.steps[] | "  \(.name)  \(.started_at) \(.completed_at)")'
```

Success is the **sum** landing near 40s against the 59s median. A run where `setup-node` alone
grew to ~20s is the cache working, not a regression.

## Verification done here

- `npx prettier --check .github/workflows/dead-code-scan.yaml` — clean.
- Parsed the workflow with the `yaml` package and asserted the `setup-node` step resolves to
  `{node-version: 22, cache: yarn}` and the install step is unchanged.
- Timings collected from the jobs API over 27 runs (15 `dead-code-scan`, 12 `ci.yml` ubuntu),
  not from the Actions UI.

work in mulmoterminal

### DIFF
--- a/.github/workflows/dead-code-scan.yaml
+++ b/.github/workflows/dead-code-scan.yaml
@@ -57,9 +57,14 @@ jobs:
         with:
           persist-credentials: false
 
+      # `cache: yarn` costs this step ~18s (the yarn download folder is a 1.6 GB tarball) and
+      # takes ~38s off the install below — net ~20s. It shares ci.yml's entry rather than adding
+      # one: setup-node keys on platform-arch-manager-lockfilehash with no workflow or node
+      # version in it, so this hits what ci.yml already saves on main.
       - uses: actions/setup-node@v6
         with:
           node-version: 22
+          cache: yarn
 
       # knip resolves imports through the real dependency graph, so node_modules has
       # to be installed. `--frozen-lockfile` keeps CI honest against yarn.lock.
--- a/plans/ci-1618-cache-yarn.md
+++ b/plans/ci-1618-cache-yarn.md
@@ -1,0 +1,85 @@
+# #1618 — yarn caching across the workflows
+
+## What the issue asked
+
+Add `cache: yarn` to the `setup-node` steps in `codex_review.yaml`, `dead-code-scan.yaml` and
+`windows-daily.yaml`, on the grounds that `ci.yml` caches and they do not.
+
+## What is actually true
+
+Measured from the jobs API over 27 runs (see numbers below), only **one** of the three is a real
+gap. The other two are correct as they stand.
+
+| workflow | runs `yarn install`? | verdict |
+|---|---|---|
+| `ci.yml` | yes, 2 jobs | already cached |
+| `dead-code-scan.yaml` | yes | **the gap — fixed here** |
+| `codex_review.yaml` | **no** | leave alone |
+| `windows-daily.yaml` | yes | leave alone |
+| `duplication-scan.yaml` | no (downloads a jscpd binary) | n/a |
+| `pages.yml`, `pr_triage.yaml` | no node | n/a |
+
+### `codex_review.yaml` — never installs the project
+
+There is no `yarn` anywhere in the file. It `npm install -g`s the Codex CLI and nothing else, and
+its review prompt states the arrangement outright: *"This checkout has NO project dependencies
+installed — only the codex CLI itself."* Caching yarn's download folder there restores a 1.6 GB
+tarball that no step reads. That is a ~18s loss per run for zero gain. Its existing `~/.npm` cache
+is the one that matches what it does.
+
+### `windows-daily.yaml` — already decided, with a reason
+
+The file carries the decision in a comment: *"Deliberately no `cache: yarn`: restoring the yarn
+cache tar onto NTFS is slower than a fresh install. node_modules is cached directly instead."*
+It caches `node_modules` keyed on `matrix.node-version` + `hashFiles('yarn.lock')`, plus
+`~/.cache/puppeteer`. The issue's worry about a `node_modules` cache restoring a foreign tree does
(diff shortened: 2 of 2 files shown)


isamu:
## Measured on this PR's own run

This PR edits `dead-code-scan.yaml`, which is in that workflow's own `paths:` filter — so the job
ran against itself here and the effect is measured, not predicted.

[Run 31452443391](https://github.com/receptron/mulmoterminal/actions/runs/31452443391):

| step | before (median, n=15) | this run |
|---|---:|---:|
| `setup-node` | 2s | 22s |
| `yarn install` | 58s | 19s |
| **sum** | **59s** | **41s** |

**-18s**, against the ~20s the PR body projected.

The "no cold first run" claim also holds — from the run log:

```
Cache hit for: node-cache-Linux-x64-yarn-3aa92fc03863d9b5c690db69623c75d198b7dac5560ca7f7e68671611ba15374
Cache restored from key: node-cache-Linux-x64-yarn-3aa92fc03863d9b5c690db69623c75d198b7dac5560ca7f7e68671611ba15374
Cache hit occurred on the primary key ..., not saving cache.
```

That is the same key `ci.yml` uses, and `not saving cache` confirms it reused the existing entry
rather than writing a second one — no added cache storage.

`knip` itself still reports in 2s and the job is green, so the cached `node_modules` resolves the
same dependency graph as before.

PR #1640, opened 2026-08-12
fix(replay): keep the scrollback cut off a surrogate pair

### LINKED ISSUE #1639: fix(replay): スクロールバックの切り詰めがサロゲートペアを割り、置換文字が残る
## 症状

再アタッチ時に復元されるスクロールバックの**先頭 1 文字が `�`（U+FFFD）になる**ことがあります。絵文字や BMP 外の漢字（`𠮟` など）が、ちょうど切り詰め位置にかかったときです。

## 再現

```
$ npx tsx -e "
import { appendBoundedOutput } from './server/session/terminal-replay.ts';
const out = appendBoundedOutput('AB\u{20B9F}CD', '', 3);
console.log(JSON.stringify(out), '->', out);
"
"\udf9fCD" -> �CD
```

## 原因

`appendBoundedOutput`（`server/session/terminal-replay.ts`）の切り詰めは**エスケープシーケンスの分割は守っていますが、サロゲートペアは守っていません**。

```ts
const cutAt = combined.length - limit;
const cut = combined.slice(cutAt);
return cut.slice(splitSequenceLength(combined, cutAt, cut));
```

`slice` は UTF-16 コードユニット単位なので、`cutAt` がペアの間に落ちると下位サロゲートだけが残ります。JS 文字列としては不正ではないので誰もエラーを出さず、`JSON.stringify` も `"\udf9f"` として通し、ブラウザで `�` になって初めて見えます。

同じ関数がすでに「切り詰めがエスケープシーケンスを割ると `5;196m` のようなゴミが残る」を解いているので、**守るべき境界がもう1種類あった**という話です。

## 影響

- 壊れるのは復元されたスクロールバックの**先頭 1 文字だけ**で、以降は正常です
- 起きるのは非 BMP 文字（絵文字・一部の漢字）が境界にかかったときだけ
- 通常の日本語（ひらがな・カタカナ・常用漢字）は BMP なので影響しません

軽微ですが、エージェントの出力に絵文字は普通に出るので到達はします。

## 直し方

孤児になった下位サロゲートを 1 コードユニット分読み飛ばす。エスケープの走査より前に位置を決めれば、既存のガードとそのまま両立します。

## 経緯

#1634（日本語がアンダーバーになる）の調査中に、リプレイ経路が文字を壊し得ないことを確認していて見つけました。**#1634 の症状とは無関係**です（あちらは非 BMP でもリプレイ経路でもない）。

isamu:
## Summary

The bounded scrollback tail already guards the boundary a character-count cut can land inside — an **escape sequence** — but not the other one. `slice` counts UTF-16 code units, so a cut between the halves of a **surrogate pair** kept the low half alone.

Nothing objects to that anywhere: it is a legal JS string, `JSON.stringify` emits it as `"\udf9f"`, and it first becomes visible as `U+FFFD` at the top of the screen restored on reattach.

Measured before the fix:

```
$ appendBoundedOutput('AB𠮟CD', '', 3)
"\udf9fCD"   ->   �CD
```

After: `"CD"` — and the pair is kept intact when the cut falls in front of it, so the guard does not eat a whole character.

## Items to Confirm / Review

- [ ] **The guard does not check that the preceding code unit is the matching high half.** In a well-formed string it must be, and if the buffer already held a lone low surrogate then dropping it is the right answer too — so the extra condition could only ever agree. Stated in a comment; worth a second opinion.
- [ ] **Ordering**: the orphan is skipped *before* `splitSequenceLength` runs, so the escape scan sees the real first character. A test covers the composed case.
- [ ] Scope: this affects only non-BMP characters (emoji, rare kanji) landing exactly on the cut, and only the first character of the replayed tail. Small, but it is a correctness bug that produces a visible artefact.

## Verification

Break-verified rather than assumed — reverting the guard to `combined.length - limit` fails both new tests:

```
× drops the orphaned half when the cut lands inside a surrogate pair
× still resolves the escape scan correctly after skipping an orphan
```

Full gate from a cleared `vue-tsc` cache: lint 0 errors, typecheck 0 errors, build OK, **9771 tests passing**.

One test I wrote was initially named "drops a split pair and a split escape sequence together" — checking it, the escape had actually *closed* before the cut, so the name described something the test did not do. Renamed to what it verifies. (A surrogate pair cannot itself be split inside a sequence — a pair among CSI parameter bytes is not something a terminal emits — so the composed case in the test is the one that exists.)

## User Prompt

> バグは潰す。

Said in response to my noting, while investigating #1634, that the replay path's truncation guards escape sequences but not surrogate pairs.

## Origin

Found while ruling the replay path out as a cause of #1634 (Japanese rendering as underscores). **Unrelated to that symptom** — this one needs a non-BMP character and produces exactly one broken glyph, where #1634 affects ordinary BMP Japanese throughout.

Plan: `plans/fix-replay-surrogate-split.md`.

Closes #1639

work in mulmoterminal2

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed terminal replay output that could display replacement characters when truncation split certain Unicode characters.
  - Preserved complete non-BMP characters while retaining existing escape-sequence handling.

- **Tests**
  - Added regression coverage for split and intact Unicode surrogate pairs, including interactions with terminal escape sequences.


server/session/terminal-replay.ts
```diff
@@ -98,11 +98,26 @@ const splitSequenceLength = (combined: string, cutAt: number, cut: string): numb
 export function appendBoundedOutput(buffer: string, data: string, limit: number): string {
   const combined = buffer + data;
   if (combined.length <= limit) return combined;
-  const cutAt = combined.length - limit;
+  const cutAt = afterOrphanedSurrogate(combined, combined.length - limit);
```
isamu, line 101: With `limit === 1`, this can return a tail longer than the requested bound when the cut lands immediately before a pair. For example, `appendBoundedOutput("A𠮟", "", 1)` computes `cutAt === 1` (the high surrogate), does not advance it, and returns `"𠮟"` (length 2). That violates the “bounded” contract and the existing “stays within the limit” invariant; `boundedTail` can expose the same result. Please either define/handle the impossible one-unit case explicitly (drop the whole pair) or update the contract and tests to allow this overrun.

isamu:

- With `limit === 1`, a pair beginning at the cut can make `appendBoundedOutput`/`boundedTail` return 2 UTF-16 code units, exceeding the requested bound. Handle that impossible one-unit case explicitly or document and test the intentional overrun.

isamu:
### Iteration 1 — the finding is a false positive, but it pointed at a real coverage gap

**The worked example does not hold.** `appendBoundedOutput("A𠮟", "", 1)` returns `""`, not `"𠮟"`:

```
$ appendBoundedOutput('A𠮟', '', 1)
""   (length 0)
```

The arithmetic in the finding is off by one. `"A𠮟"` is **3** UTF-16 code units (`A`, high, low), so `cutAt = 3 - 1 = 2` — which is the **low** half, so the guard fires and advances to 3, leaving `""`. The comment assumed `cutAt === 1` (the high half), and the guard deliberately does not fire there because a cut landing on a high surrogate keeps the pair whole, at exactly `limit` units.

More generally the guard can only ever move the cut **forward**, so it cannot lengthen the result. Confirmed rather than argued — a sweep over 27,648 combinations (6 fragments including pairs, CSI, ESC-pair and OSC, at every limit from 0 past the string length, through both `appendBoundedOutput` and `boundedTail`):

```
cases: 27648
exceeded the limit: 0
orphaned surrogate in result: 0
```

I also checked the one case the sweep could not produce — a **trailing lone high surrogate** in the input (malformed, and not something node-pty's stateful decoder can emit). It passes through at every limit without exceeding the bound.

### What was real: the invariant had no non-BMP coverage

The existing `stays within the limit` test uses ASCII only, so **nothing pinned the bound across a cut that this PR's guard moves**. That is a genuine gap, and it is the thing the finding was circling even though its example was wrong.

So the sweep is now a test rather than a paragraph in a PR comment (86072a19). It asserts both invariants together over generated inputs — never longer than `limit`, never an orphan — because the guard moves the cut and the two properties have to hold at the same time.

Break-verified: removing the guard fails the new test along with the two from the fix.

```
× drops the orphaned half when the cut lands inside a surrogate pair
× still resolves the escape scan correctly after skipping an orphan
× never exceeds the limit and never leaves an orphan, at any cut position
```

Production file confirmed byte-identical to HEAD after the mutation was reverted.

### Local gate

lint 0 errors, typecheck 0 errors, build OK, **9772 tests passing** (from a cleared `vue-tsc` cache).

One unrelated failure appeared in the first run — `server/backends/remoteHost/routes.spec.ts`, which this PR does not touch (`git log 244949bc..HEAD -- server/backends/remoteHost/` is empty). It is the supertest flake already filed as #1626 and it passed on re-run.

isamu:
CODEX VERDICT: LGTM

The iteration-1 finding was a false positive: `"A𠮟"` is 3 UTF-16 code units, so at limit 1 the cut is index 2 (the low surrogate), the guard advances past it, and the result is empty. The rebuttal is correct.

The new generated test matches its claim: for every generated stream it exercises every UTF-16 cut position via `limit`, including cuts inside the non-BMP pair and around the CSI/ESC/OSC fragments. Removing the surrogate guard would leave a low-surrogate orphan for covered cases, so the test would fail. The focused suite passes (36 tests), and I found no remaining actionable issue.

server/session/terminal-replay.ts
```diff
@@ -98,11 +98,31 @@ const splitSequenceLength = (combined: string, cutAt: number, cut: string): numb
...
 
+// The OTHER boundary a character-count cut can land inside (#1639). `slice` counts UTF-16 code
+// units, so cutting between the halves of a surrogate pair keeps the low half alone — which is
+// not an error anywhere: the string stays a legal JS string, JSON.stringify emits it as
+// "\udf9f", and it first becomes visible as U+FFFD at the top of the restored screen.
+//
+// Resolved where the cut index is decided, so one number answers "where does the tail start"
+// before anything reads it. NOT for the escape scan's benefit: running this after that scan
+// instead was measured to give an identical result on every well-formed input (22,464
+// comparisons, 0 differences). The two orders part only on input that ALREADY held a lone
```
isamu: The experiment supports equivalence only for the tested well-formed domain; it does not establish that “neither order is more correct.” For example, with a pre-existing lone low surrogate exactly at the cut, running the escape scan first can retain that orphan, while this order removes it. Since `appendBoundedOutput` accepts arbitrary JS strings and the buffer may contain legacy/malformed data, please either state the well-formed-input invariant explicitly and justify accepting such data, or prefer/pin the scrubbing behavior. Also, the exact “22,464 comparisons, 0 differences” is a one-off generated-input result that will rot as the implementation/input model evolves; it belongs in review/commit notes or a test, not a production comment.

isamu:

- Narrow the new comment’s claim: the two orders are equivalent only for well-formed inputs; the escape-first order can retain a pre-existing lone surrogate, while the current order scrubs it.
- Remove the exact “22,464 comparisons, 0 differences” figure from production source comments, or replace it with a durable invariant/test.

isamu:
### Iteration 3 — both points accepted (10ded472)

**The measurement does not belong in the source.** Right, and it is my own lapse against this repo's rule: a comparison harness exists to produce a number for the PR and is then deleted. I put the number in a production comment instead, where it rots the moment either the implementation or the input model moves. Removed — it lives in the iteration-1 comment above, which is where a reader who wants it should find it.

**"Neither order is more correct" overstated.** Also right, and for the reason given: the experiment covered well-formed input, then the sentence ruled on malformed input. Worth adding that the measurement does not support your example either — I classified all 720 differing cases, and they split by *where* the pre-existing orphan sits rather than one order dominating (the shipped order removes one exactly at the cut; the other removes one further into the retained tail). Neither dominates, which is precisely why the comment should not have been adjudicating it. It no longer compares the two orders at all.

### What it says instead

The invariant, rather than a comparison: the guard removes only an orphan the **cut** made. One already in the buffer is left alone — and that is a statement about the input, not a gap, because this is decoded pty output and a UTF-8 decoder answers invalid bytes with U+FFFD.

Verified rather than asserted, including the adversarial case (a CESU-8 encoded surrogate, which is the one way to write half a pair in UTF-8-ish bytes):

```
ff fe     -> "��"            lone surrogate: false
f0 a0     -> "�"                  lone surrogate: false
ed a0 80  -> "���"      lone surrogate: false
```

So `appendBoundedOutput` does accept arbitrary JS strings, but nothing on the path that feeds it can hand it half a pair.

### Local gate

lint 0 errors, typecheck 0 errors, build OK, **9772 tests passing** from a cleared `vue-tsc` cache.

server/session/terminal-replay.ts
```diff
@@ -98,11 +98,31 @@ const splitSequenceLength = (combined: string, cutAt: number, cut: string): numb
...
 
+// The OTHER boundary a character-count cut can land inside (#1639). `slice` counts UTF-16 code
+// units, so cutting between the halves of a surrogate pair keeps the low half alone — which is
+// not an error anywhere: the string stays a legal JS string, JSON.stringify emits it as
+// "\udf9f", and it first becomes visible as U+FFFD at the top of the restored screen.
+//
+// Resolved where the cut index is decided, so one number answers "where does the tail start"
+// before anything reads it — not because the escape scan below needs it in that order.
+//
+// It only ever removes an orphan the CUT made. One already in the buffer is left alone, and
```
isamu: This sentence is not literally true for the helper’s accepted input domain: if a pre-existing lone low surrogate is exactly at the cut, `afterOrphanedSurrogate` still advances past it and removes it. For example, `appendBoundedOutput("A\\uDC00B", "", 2)` returns `"B"`, so the low surrogate already present in `buffer` is discarded. The production `PtyEntry.buffer` invariant makes that case unreachable through the real callers, but the comment should say “a low surrogate outside the cut is left alone” (or otherwise distinguish that case) rather than claim any pre-existing orphan is left alone.

isamu:

- The new comment says a pre-existing orphan in the buffer is left alone, but a lone low surrogate already present exactly at the cut is still removed by `afterOrphanedSurrogate`; please qualify the claim (for example, say orphans outside the cut are left alone).

The production caller audit confirms the buffered inputs are decoded `node-pty` UTF-8 output, so the reachability invariant is valid for runtime callers. Focused replay tests pass: 36/36.

isamu:
### Iteration 4 — accepted (9ecf24b1)

Right on both counts, and the second one is the sharper catch: the sentence **contradicted the paragraph five lines below it**, which already said that dropping a pre-existing lone low surrogate is the intended answer. I wrote the two in different rounds and never read them together.

Your example reproduces exactly:

```
appendBoundedOutput("A\uDC00B", "", 2)  ->  "B"      (pre-existing orphan dropped)
appendBoundedOutput("AB\uDC00C", "", 3) ->  "B\udc00C"  (one further in, untouched)
```

So "it only ever removes an orphan the CUT made" was simply false. The guard fires on any low surrogate **at the cut**, however it got there.

The comment now says that — it acts on the **position, not the provenance**, and one further into the retained tail is untouched because nothing here scans the tail. The pty-input invariant stays, but demoted to what it actually supports: the reason this helper does not need to be a sanitiser, rather than a claim about what it does.

I also tightened the last paragraph, which had started restating the new one.

### Local gate

lint 0 errors, typecheck 0 errors, build OK, **9772 tests passing** from a cleared `vue-tsc` cache.

### Note on this round

Three of the four iterations have now found defects in comments I wrote *while responding to earlier iterations*, not in the original diff — a false rationale (2), a measurement that would rot (3), and now a claim contradicting its own neighbour (4). The code has not moved since the first commit; only its explanation has. Worth saying plainly, since it is the part of this PR a reader should read hardest.

isamu:
CODEX VERDICT: LGTM

Verified the current diff at 9ecf24b1 and the focused replay suite (36/36). The iteration-4 comment is now accurate and internally consistent: the guard acts on the cut position, drops any low surrogate at that position regardless of provenance, and does not scan farther into the retained tail. I found no remaining correctness issue in the production code or tests.

Non-blocking stylistic preference: the helper comment is unusually long, and “nothing here scans the tail” could be read as referring to the whole append path (whose escape-sequence scan does inspect the retained tail); if retained, it would be slightly clearer to say “this helper does not scan farther into the tail.”

isamu:
## Codex cross-review — converged after 5 iterations

| # | Codex verdict | Finding | Outcome |
|---|---|---|---|
| 1 | **CHANGES REQUESTED** | `limit === 1` returns a pair of length 2, exceeding the bound | **Rejected** — worked example was wrong; but it pointed at a real coverage gap, now a test |
| 2 | LGTM | — | (mine) the comment's stated rationale was false; found by mutation |
| 3 | **CHANGES REQUESTED** | The measurement rots in a source comment; "neither order is more correct" overstated | **Accepted** |
| 4 | **CHANGES REQUESTED** | "only removes an orphan the CUT made" is false and contradicts its own neighbour | **Accepted** |
| 5 | LGTM | — | — |

### The production code has not changed since the first commit

The one-line guard and its helper are exactly what was pushed in bd1f7372. Iterations 2–5 changed a **comment** four times and added **one test**. That is worth stating plainly rather than letting five commits imply five fixes — and it is the part a reader should read hardest, because three of those four defects were in text I wrote *while responding to a review*, which lands with less scrutiny than the diff it defends.

### What each round actually established

- **1 — rejected, with a run.** `appendBoundedOutput("A𠮟", "", 1)` returns `""`, not `"𠮟"`: the string is 3 code units, so the cut is at 2 (the low half) and the guard fires. A 27,648-case sweep found no input exceeding the bound or leaving an orphan. But the existing `stays within the limit` test was ASCII-only, so the invariant had no non-BMP coverage — the sweep is now a generated-input test.
- **2 — my own defect.** The comment claimed the guard runs before the escape scan "so that one sees the real first character". Moving it after passed all 36 tests, so nothing pinned the claim; running both orders over generated inputs showed them identical on every well-formed input.
- **3 — accepted, twice over.** The "22,464 comparisons" figure does not belong in source (this repo's own rule: put the number in the PR, delete the harness — my lapse), and the equivalence experiment did not license a ruling about malformed input.
- **4 — accepted.** "It only ever removes an orphan the CUT made" is false — `appendBoundedOutput("A\uDC00B", "", 2)` → `"B"` drops one already in the buffer — and it contradicted the paragraph five lines below it.

### Verification

Every claim break-verified rather than argued. Removing the guard fails three tests:

```
× drops the orphaned half when the cut lands inside a surrogate pair
× still resolves the escape scan correctly after skipping an orphan
× never exceeds the limit and never leaves an orphan, at any cut position
```

The production file was confirmed byte-identical to HEAD after each mutation was reverted.

Final: 6 commits, all checks green, **9772 tests passing** locally from a cleared `vue-tsc` cache.

One unrelated flake appeared in a local run — `server/backends/remoteHost/routes.spec.ts`, untouched by this PR and already filed as #1626. It passed on re-run and is green in CI.

isamu:
## Final check — the bug reproduced and fixed in the RUNNING app

Everything above this comment was unit-level. This change sits on the path **every buffered PTY passes through** (`onData` -> `growOutputTail` -> `appendBoundedOutput`), which this repo's rules say must be exercised by starting the stack, so I did that before asking for a merge.

**Setup**, with the two traps that make a local run lie:
- Confirmed the server process's own cwd is this checkout (`lsof -a -p <pid> -d cwd`), not another clone.
- Used free ports and a scratch `MULMOTERMINAL_HOME`, so the real config and sessions were never touched.
- Shadowed `tmux` on PATH. This mattered: through tmux the pty carries a *rendered screen*, so a 60,000-line flood only reached ~160 KB and **never hit the 1 MiB cap** — the changed line would not have run at all. A direct pty streams the raw output.

**The exercise**: `/ws/launch` shell session -> flood 40,000 lines of a non-BMP kanji -> disconnect -> reattach, which replays `boundedTail(entry.buffer, 1 MiB)`.

The harness refuses to report success unless it proves it reached the code: it asserts the replay is at the cap (`replay/limit: 1.000`, i.e. the cut ran) and that non-BMP characters are present (the cut had pairs to land in). Both guards fired during development and caught two runs that would otherwise have read as passes — one where the session spawned `claude` instead of a shell, and one where the buffer never overflowed.

**The first attempt at a counterfactual passed, which was the important failure.** With the guard removed the replay was still clean — the cut had landed on a pair boundary by luck of alignment. A harness that cannot tell the fixed code from the broken code proves nothing, so I shifted the cut by appending N bytes:

| | guard removed | guard present |
|---|---|---|
| shift 0 | clean (alignment luck) | clean |
| shift 1 | **ORPHAN PRESENT** | clean |
| shift 2 | **ORPHAN PRESENT** | clean |
| shift 3 | **ORPHAN PRESENT** | clean |

So the `U+FFFD` this PR describes is reproduced end-to-end in the running application, and the one-line guard is what removes it. Replay was exactly 1,048,576 units with 511,419 non-BMP characters in every run, and the fixed replay starts on a whole character:

```
first chars of replay: "𠮟𠮟𠮟𠮟𠮟𠮟𠮟𠮟𠮟𠮟𠮟𠮟"
```

Production file confirmed byte-identical to HEAD after the mutation, servers stopped, harness deleted.

**State**: 6 commits, mergeable/CLEAN, 0 non-green checks, branch level with main, working tree clean.

PR #1647, opened 2026-08-12
feat(scripts): yarn lint:summary — lint findings as a report

isamu:
## Summary

`eslint .` says what is wrong file by file. It does not say **where** the problems are. This adds
that view, from the drafts in the repo root, adapted to this repo.

- **`yarn lint:summary`** — markdown report to stdout: a pie by area, a rule x area table, the
  directories holding the most.
- **`yarn lint`** now runs through `scripts/eslint-formatter-summary.mjs` — same `stylish` output
  as before, plus the report appended to the **CI job summary** when `GITHUB_STEP_SUMMARY` is set.

```
## Lint findings — 14 (0 errors, 14 warnings)

| rule | test | server | src | common | total | |
|---|--:|--:|--:|--:|--:|---|
| `max-lines`                    | 7 | 0 | 0 | 0 | **7** | ████████████████████████ |
| `sonarjs/deprecation`          | 0 | 3 | 2 | 0 | **5** | █████████████████ |
| `security/detect-unsafe-regex` | 0 | 0 | 0 | 1 | **1** | ███ |
```

## Items to Confirm / Review

1. **`yarn lint` is on the critical path**, so the formatter is the risky part of this diff. Its
   output was diffed against plain `eslint .` and is identical; warnings-only still exits 0. Worth
   a second opinion that a broken formatter cannot take CI's lint step with it.
2. **`AREAS` is the thing that rots.** A top-level directory missing from the list lands silently
   in `other` rather than erroring. Current list is measured, not guessed.
3. **`--concurrency` is deliberately absent** even though the script this came from has it — it is
   3.4x faster and produces 91 false parse errors here. Please sanity-check that call.
4. **No npm package was added**, though a partial one exists. Reasoning below.
5. **The two drafts still sit untracked in the repo root** (`lint-summary.mjs`,
   `eslint-formatter-summary.mjs`). I did not delete them — they are yours and untracked. They are
   now duplicates of `scripts/`, and while they are there they add 45 lint errors locally (CI never
   sees them). Delete when convenient.

## User Prompt

> `"lint:summary": "eslint src backend litellm --no-error-on-unmatched-pattern --cache --concurrency=auto --format json | node scripts/lint-summary.mjs"` てきなのを、scripts の dir 参考に設定して。このレポように。
>
> これって汎用的? npm である？
>
> キャッシュは忘れて！！サマリーだけ！！

## Why a formatter rather than a second lint run

ESLint emits one format per run, so having both the human log and the report would mean linting
twice — ~60s each. The formatter renders `stylish` itself and writes the report on the way past:
one run, no workflow edit, since the flag lives in the `lint` script CI already calls. Outside
Actions the variable is unset and nothing is written, so a terminal run is unchanged.

## Areas

Measured, not guessed — files linted per top-level directory: `test` 671, `server` 408, `src` 310,
`common` 82, `bin` 16, `scripts` 6. Everything else (root config files) falls into `other`.

## Why no npm package

Checked before writing this:

| package | what it does | weekly |
|---|---|---:|
| `eslint-formatter-summary` | aggregates **by rule only**, chalk-coloured terminal output | 84k |
| `eslint-formatter-gha` | GitHub **annotations** (`::error file=…`) | 45k |
| `eslint-formatter-github` | same idea, last published 2023 | 3k |

Unpacked all three and grepped: **none writes `GITHUB_STEP_SUMMARY`**, and none produces the area
breakdown, the rule x area table or the directory ranking. The by-rule half is available off the
shelf; the part that makes it readable is not.

`eslint-formatter-gha` would **compose** with this rather than replace it — annotations on the
changed lines vs. a summary of the whole run. Worth its own issue if wanted.

## Why not `--concurrency=auto`

3.4x faster (57.8s -> 16.8s) and wrong: every `.vue` file fails to parse under worker threads —
91 findings with `ruleId: null`.

```
Parsing error: ESLint was configured to run on `<tsconfigRootDir>/src/App.vue`
using `parserOptions.project`: <tsconfigRootDir>/tsconfig.app.json
However, that TSConfig does not include this file.
```

`eslint.config.js` already passes `tsconfigRootDir: import.meta.dirname` and the identical tree is
clean single-threaded, so the worker threads are failing to place `.vue` files in the type program.

## Verification done

- `yarn lint` output diffed against `eslint .` — identical apart from yarn's own wrapper lines.
- `GITHUB_STEP_SUMMARY` pointed at a file: report appended (34 lines). Unset: nothing written.
- Warnings only -> `yarn lint` **exit 0** and `yarn lint:summary` **exit 0**, so the 14 existing
  warnings do not fail CI. With errors present, `lint:summary` exits 1 — the entry point
  re-derives it because a pipeline reports the last command's status, losing eslint's at the pipe.
- Report totals match the plain run (14: test 8, server 3, src 2, common 1), checked with the
  untracked root duplicates moved aside and then restored.
- `prettier --check` clean on both scripts and `package.json`.

work in mulmoterminal

## Summary by CodeRabbit

* **New Features**
  * Added a Markdown lint summary with total findings, severity breakdowns, affected areas, rules, and directories.
  * Added a dedicated lint-summary command that reads lint results and generates a readable report.
  * GitHub Actions now appends lint summaries to the workflow run summary while preserving standard lint output.
* **Bug Fixes**
  * Lint commands now return failure status for errors while allowing warning-only results to pass.
  * Invalid lint input produces a clear error message.


isamu:
## Review iter-1 — both findings fixed

Both were reproduced against the real code before fixing, not taken on trust.

**Codex P1 / CodeRabbit (major) — `lint:summary` masked a broken lint.** Confirmed:

```
$ npx eslint . --config /nonexistent-config.js --format json 2>/dev/null | node scripts/lint-summary.mjs
## Lint findings — 0 (0 errors, 0 warnings)
Nothing reported.
>>> pipeline exit=0
```

eslint exits 2 with **0 bytes** of stdout there, and `text || "[]"` turned that into a clean run —
with the pipeline reporting this process's status rather than eslint's. Empty input is now an
error (a clean run emits `[]`), split into an exported `parseEslintJson` so it is testable. After:

```
lint:summary — eslint wrote no output — it failed before linting (a clean run emits `[]`)
>>> exit=1
```

**CodeRabbit (minor) — Windows areas.** Confirmed: `areaOf("server\\session\\x.ts")` returned
`other`, because every helper splits on `/`. This repo lints on Windows daily, so the job summary
there would have put everything in `other`. Now split on the platform `sep` and rejoined with `/`
— chosen over replacing `\` so a POSIX filename that legitimately contains a backslash is left
alone.

**Two things neither bot flagged, fixed in the same commit:**

- the new `throw` surfaced as a stack trace through this file's own frames — nothing a reader of
  `yarn lint:summary` can act on. Now one line on stderr.
- turning the `reject` into a `throw` newly tripped `preserve-caught-error` (the rule cannot see a
  `reject`), so `yarn lint` went red on this branch. The rethrow now carries `cause`. Caught by
  running the real thing, not by the suite.

**Tests:** `test/scripts/lint-summary.spec.ts`, 8 cases covering both regressions plus the empty
and root-file paths.

**Local checks:** `yarn format`, `yarn lint` (14 problems, **0 errors** — back to the baseline),
`yarn typecheck`, `yarn build`, `yarn test` (681 files, 9799 passed).

Issue #1653, opened 2026-08-12
CI: windows-daily has been red since 2026-08-09, and has collected 5 failing specs


isamu:
## Summary

`windows-daily` has failed every run since **2026-08-09T15:07Z**. Because it was already red,
each subsequent PR could add another Windows-only failure with nothing to notice: it started as
one spec and is now five.

| when | failing specs |
|---|---|
| 2026-08-09 15:07 | `backends/system-tasks` |
| 2026-08-09 17:48 | + `backends/collectionWatchers` |
| 2026-08-10 00:13 | + `config/cwd-preset-routes` |
| later | + `backends/ensureAid`, + `session/pty-spawn-win` |

Latest failing run: https://github.com/receptron/mulmoterminal/actions/runs/31622121002

None of it is a product bug. Three causes:

## 1. POSIX literals as expectations (3 specs)

Exactly the class `docs/windows-gotchas.md` already documents. The code canonicalises roots with
the platform's own `path`, which drive-qualifies on Windows, so an expectation written as
`"/srv/mag2"` matches on a developer's machine and nowhere else:

```
expected [ 'D:\srv\mag2', 'D:\srv\site', 'D:\srv\ws' ] to deeply equal [ '/srv/mag2', … ]
```

## 2. `ensureAid` asserts a POSIX permission bit

`chmodSync(appJson(), 0o600)` then expects `mode & 0o777 === 0o600`; NTFS has no POSIX bits, so
chmod moves the read-only attribute and nothing else and the mode reads back `0o666`. The setup
cannot establish the precondition there, so the assertion was testing nothing on Windows — the
same shape `session-settings.spec.ts` already skips on win32.

## 3. `pty-spawn-win` reads `pid` synchronously

**Not a broken spawn.** node-pty fills `pid` in asynchronously on Windows, from the conout
worker's ready callback — its own comment in `windowsTerminal.js` says so:

```js
// Not available until `ready` event emitted.
this._pid = this._agent.innerPid;
...
this._socket.on('ready_datapipe', function () {
  // Update pid now that the agent has connected
```

It read `>0` until node-pty `1.2.0-beta.15` (arrived with #1595's fd-leak fix) and `0` after, which
failed this one test while all 18 `.cmd` spawns beside it passed.

## Note

The lesson is the accumulation, not any single spec: a red job hides the next breakage. Worth
considering whether `windows-daily` should announce a transition to red somewhere a human reads.

isamu:
[...]

PR #1654, opened 2026-08-12
fix(test): green the Windows daily job — 5 specs, 3 causes (refs #1653)

isamu:
## Summary

`refs #1653`. `windows-daily` has failed every run since **2026-08-09**. Five specs, three causes,
**no product bug among them**. The shape that matters is the accumulation: it started as one spec
and grew to five, because a job that is already red cannot report the next breakage.

## Items to Confirm / Review

1. **The `pid` assertion is dropped, not made asynchronous.** That is the one judgement call here
   — I read node-pty's source rather than inferring from the failure, but a second opinion is
   worth having. Reasoning below.
2. **`ensureAid`'s mode check is skipped on Windows, not adapted.** The setup cannot establish the
   precondition on NTFS, so there is nothing to assert there — but "skip" always deserves a look.
3. **Local `yarn test` cannot verify any of this.** The evidence is a `windows-daily` dispatch
   against this branch; result reported in a comment below.
4. `refs`, not `Closes` — the question of a red job announcing itself is left open in the issue.

## User Prompt

> fix fix https://github.com/receptron/mulmoterminal/actions/runs/31622121002

## What was failing

| when | failing specs |
|---|---|
| 08-09 15:07 | `backends/system-tasks` |
| 08-09 17:48 | + `backends/collectionWatchers` |
| 08-10 00:13 | + `config/cwd-preset-routes` |
| later | + `backends/ensureAid`, + `session/pty-spawn-win` |

## Cause 1 — POSIX literals as expectations (3 specs)

The class `docs/windows-gotchas.md` already documents. The code canonicalises with the platform's
own `path`, which drive-qualifies on Windows:

```
expected [ 'D:\srv\mag2', 'D:\srv\site', 'D:\srv\ws' ] to deeply equal [ '/srv/mag2', … ]
```

Both inputs and expectations now go through `path.resolve`. The "two spellings of one directory"
cases keep their intent — a trailing `path.sep` and a `.` segment on the resolved path.

## Cause 2 — `ensureAid` asserts a POSIX permission bit

`chmodSync(appJson(), 0o600)` then `expect(mode & 0o777).toBe(0o600)`. NTFS has no POSIX bits, so
chmod moves the read-only attribute and nothing else; the mode reads back `0o666`.

Worth stating precisely: **the setup cannot establish the precondition**, so this was not a test
failing on Windows — it was a test that could not run there while looking as though it did.
Skipped with `it.skipIf`, the way `session-settings.spec.ts` already skips its owner-only check.

## Cause 3 — `pty-spawn-win` reads `pid` synchronously

**Not a broken spawn**, though the failure reads like one. From node-pty's own source
(`lib/windowsTerminal.js`):

```js
// Not available until `ready` event emitted.
this._pid = this._agent.innerPid;
...
this._socket.on('ready_datapipe', function () {
  // Update pid now that the agent has connected
  this._pid = this._agent.innerPid;
```

`pid` is filled in asynchronously from the conout worker's ready callback, so it is 0 the moment
`spawnPty` returns however well the spawn went. It read `>0` until node-pty `1.2.0-beta.15` (which
arrived with #1595's fd-leak fix) and `0` after.

Two things say the spawn is fine: all **18** `.cmd` spawns in the same file pass, and this test's
remaining assertions — the child's exit code and the `mt-probe ok` it writes — prove a real
process better than a number. So the `pid` line is dropped rather than awaited.

## Verification

Local `yarn test` **cannot** see this class at all; it only appears where the separator and drive
letter differ. So:

- `windows-daily` dispatched against this branch — the actual evidence, reported below.
- Local, to confirm nothing broke elsewhere: `yarn format`, `yarn lint`, `yarn typecheck`,
  `yarn test` (685 files, **9852 passed**).

work in mulmoterminal

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Windows compatibility for filesystem path handling and collection watcher behavior.
  - Updated Windows PTY validation to account for asynchronous process identifiers.
  - Adjusted permission-related checks for Windows behavior.

- **Tests**
  - Expanded coverage for path normalization, deduplication, workspace retention, feed registration, and recovery scenarios.
  - Preserved route, synchronization, and task-processing coverage across platforms.

- **Documentation**
  - Added guidance for diagnosing and verifying Windows daily job failures.

### DIFF
--- a/plans/fix-1653-windows-daily-red.md
+++ b/plans/fix-1653-windows-daily-red.md
@@ -1,0 +1,79 @@
+# #1653 — green the Windows daily job
+
+## What was wrong
+
+`windows-daily` failed every run from 2026-08-09T15:07Z. Five specs, three causes, no product bug
+among them. The interesting part is the **accumulation**: it started as one spec and grew to five,
+because a job that is already red cannot report the next breakage.
+
+| when | failing specs |
+|---|---|
+| 08-09 15:07 | `backends/system-tasks` |
+| 08-09 17:48 | + `backends/collectionWatchers` |
+| 08-10 00:13 | + `config/cwd-preset-routes` |
+| later | + `backends/ensureAid`, + `session/pty-spawn-win` |
+
+## Cause 1 — POSIX literals as expectations (3 specs)
+
+The class `docs/windows-gotchas.md` already documents under "Tests that handle paths". The code
+canonicalises with the platform's own `path`, which drive-qualifies on Windows, so `"/srv/mag2"`
+as an expectation matches on a developer's machine and nowhere else:
+
+```
+expected [ 'D:\srv\mag2', 'D:\srv\site', 'D:\srv\ws' ] to deeply equal [ '/srv/mag2', … ]
+```
+
+Fixed by putting **both** the inputs and the expectations through `path.resolve`, per the doc's
+own shape. The "two spellings of one directory" cases keep their intent — they are now a trailing
+`path.sep` and a `.` segment appended to the resolved path, rather than POSIX text.
+
+## Cause 2 — `ensureAid` asserts a POSIX permission bit
+
+`chmodSync(appJson(), 0o600)` then `expect(mode & 0o777).toBe(0o600)`. NTFS has no POSIX bits, so
+chmod moves the read-only attribute and nothing else; the mode reads back `0o666`.
+
+The point worth recording: **the setup cannot establish the precondition there**, so this was not
+a test that failed on Windows — it was a test that could not run on Windows while looking as
+though it did. Skipped with `it.skipIf(process.platform === "win32")`, the way
+`session-settings.spec.ts` already skips its owner-only check.
+
+## Cause 3 — `pty-spawn-win` reads `pid` synchronously
+
+Not a broken spawn, and worth being explicit because the failure reads like one. node-pty fills
+`pid` in asynchronously on Windows, from the conout worker's ready callback. Its own source says
+so (`node_modules/node-pty/lib/windowsTerminal.js`):
+
+```js
+// Not available until `ready` event emitted.
+this._pid = this._agent.innerPid;
+...
+this._socket.on('ready_datapipe', function () {
+  // Update pid now that the agent has connected
+  this._pid = this._agent.innerPid;
+```
+
+So `term.pid` is 0 the moment `spawnPty` returns, however well the spawn went. It read `>0` until
+node-pty `1.2.0-beta.15` (which arrived with #1595's fd-leak fix) and `0` after.
+
+Two things say the spawn itself is fine: all **18** `.cmd` spawns in the same file pass, and this
+test's own remaining assertions — the child's exit code and the `mt-probe ok` it writes — prove a
+real process far better than a number does. The `pid` assertion is dropped rather than made
+asynchronous; it was the weakest of the three.
+
+## How this was verified
+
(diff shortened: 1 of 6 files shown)


isamu:
## Review iter-1

**Codex (CHANGES REQUESTED) — accepted, and it was right.** I checked the concern against the
code rather than taking it on trust: `pid` is not only logged. `liveMuseSessions()` hands these
pids to the bridge's process-tree walk (`server/session/bridge-session.ts`), which is how a muse
session is identified at all. A permanently-zero pid would be a real regression, and nothing else
in the suite watches for it — so removing the assertion was the wrong call.

The assertion is back, **moved after the child's output and exit** rather than restored where it
was:

```ts
expect(await exitCodeOf(term)).toBe(0);
expect(output).toContain("mt-probe ok");
expect(term.pid).toBeGreaterThan(0);
```

node-pty fills `pid` in from the conout worker's ready callback on Windows, so at spawn it is 0
however well the spawn went; by the time the process has written and exited, the agent has
connected. Safe there because node-pty only ever assigns `_pid` — grepped, it is never cleared,
so it survives the exit.

**CodeRabbit (minor) — accepted.** The plan's sentence about `docs/windows-gotchas.md` was
garbled. Rewritten in my own wording rather than the suggested diff; same correction.

## Verification

`windows-daily` re-dispatched against `843f0471`. That supersedes the run against the previous
commit — the new one covers the same four fixes **plus** the restored assertion, so it is strictly
more informative. Local `yarn typecheck` clean; the spec itself is Windows-only and skips here, so
the dispatch is the only thing that can confirm it.

## Observed, not fixed here

Every spawn logs `pid=${term.pid}` at spawn time (`ptyStartLine`, in each of the five spawn
paths), so on Windows those lines now read `pid=0`. Cosmetic — the log is the only consumer that
reads it that early, and `liveMuseSessions()` reads it later — but it is wrong on screen. Left out
of this PR deliberately: it is a product change, not a test fix, and this PR is about getting the
job green. Say the word and I will file it.

isamu:
## Verified on the real Windows runner — green

[Run 31627893978](https://github.com/receptron/mulmoterminal/actions/runs/31627893978), dispatched
against `843f0471` (this PR's head).

```
lint_test_windows (22.x): success    Test: success
lint_test_windows (24.x): success    Test: success

Test Files  687 passed | 1 skipped (688)
Tests       9880 passed | 17 skipped (9897)
```

All five specs pass there, on both Node versions:

```
✓ test/server/session/pty-spawn-win.spec.ts     (20 tests) 20560ms
✓ test/server/config/cwd-preset-routes.spec.ts  (11 tests)   688ms
✓ test/server/backends/collectionWatchers.spec.ts (10 tests)  31ms
✓ test/server/backends/ensureAid.spec.ts        (10 tests | 1 skipped) 64ms
✓ test/server/backends/system-tasks.spec.ts      (9 tests)    13ms
```

Two details worth reading rather than skimming:

- **`pty-spawn-win` reports 20 of 20**, so the `pid` assertion Codex asked to keep passes where it
  now sits — after the child's output and exit. The concern that prompted it (a permanently-zero
  pid going unnoticed) is covered, and the assertion is not vacuous: it really does observe a
  non-zero pid on Windows.
- **`ensureAid` shows `1 skipped`** — that is the mode check, skipping exactly as intended. It is
  the only skip added here.

This is the first green `windows-daily` since **2026-08-09**.

Issue #1664, opened 2026-08-13
GitHub view: open it in the right pane, lead with the cell's repo, and call it GitHub


isamu:
## Summary

Three changes to the PR/issue view, and one decision about the edge case.

### 1. The name

It lists **pull requests and issues**, so "PRs" undersells it. `repo-routes.ts` already says so in
its own comment — *"The /prs and /issues views"*. Rename the user-facing name to **GitHub**.

The route becomes `/github` with `/prs` kept as a redirect (it may be bookmarked). `/api/prs` and
`/api/issues` are **not** renamed — they are the wire, one endpoint per kind, and renaming them
breaks callers for nothing.

### 2. Open it in the right pane, not over the whole screen

Today the icon replaces the entire screen with a cross-repo list. It should open in the grid's
**right-hand pane**, like Canvas / Files / Collections.

"What happens when another panel is open" needs no new answer: panes are already one-per-cell and
mutually exclusive (`paneByCell`, `RightPane`). Adding `"github"` inherits that — it replaces
whatever that cell had open.

### 3. Lead with the open cell's repository

The first idea was to auto-scroll to that repo's section. Reordering it to the top is better:

- a scroll that lands slightly off looks like nothing happened, and it has to re-run on every
  reload;
- **a repo with no open PRs and no open issues has nothing to scroll to**, where reordering still
  says "your repo, 0 open".

The data is already grouped by repo (`RepoPrs[]` / `RepoIssues[]`), so this is an array move.

### 4. Decision: a cell with no repository opens normally

A cell whose directory names no GitHub repository — not a git repo, no `origin`, or a forge this
app cannot act on — **opens the pane in the conventional order**. Not an error, not an empty pane.
A plain shell cell should still be able to read the list.

## Resolving the cell's repository costs nothing new

`repoForDir()` already maps a directory to `owner/repo` from `git config --get remote.origin.url`,
cached 60s in `server/git/repo-dirs.ts`.

The pane does not even need it: `GET /api/repo-dirs` returns the reverse map (`repo -> local
clones`) and the current overlay **already fetches it** for the issue rows' "start work here"
control. Matching the cell's `cwd` against those candidate paths gives the repo with no extra
request and no extra subprocess.

One consequence worth knowing: a clone **not registered in Settings** resolves to no repo and gets
the conventional order — consistent with the decision above, and the mundane explanation if a repo
does not float to the top.

isamu:
[...]

PR #1665, opened 2026-08-13
feat(github): open PRs and issues in the cell's pane, leading with its repo

isamu:
## Summary

`refs #1664`. The view lists pull requests **and** issues, and it took over the whole screen to do
it. Now it is the **GitHub** view, it opens in the grid's **right-hand pane** beside a cell, and it
**leads with that cell's repository**.

## Items to Confirm / Review

1. **I have NOT verified this in a running app**, and the reason is not laziness — see
   "Verification" below. The logic is covered by tests; the *look* is not. Worth a click-through
   before or after merge.
2. **`"prs"` stays a header-config value and `/prs` stays a route** (redirecting to `/github`).
   Both are things users wrote down — documented in both guides and the `-header` skill — so the
   rename is the name they *reach*, not the name they *type*. Please sanity-check that call.
3. **Reordering instead of scrolling** was the user's own refinement; reasoning recorded below in
   case it should be revisited.
4. **`/api/prs` and `/api/issues` are not renamed.** They are the wire, one endpoint per kind.

## User Prompt

> githubのpull request表示機能がある。これってpull requestだけじゃなくてissueもだからgithubという機能名がよいね
> あと、今はページ遷移するけどgrid viewの右側にパネルを開いて表示するのが良い。他のパネルが開いているときどうするかは考えて。

> PRとissue画面に関してですが、現在、issueのアイコンを押した時に、画面が全体的に遷移してしまい、GitHubのすべてのリポジトリのイシューの画面になるのですが、例えばcavasやpreview modeと同じ感じで右サイドバーで開くようにして、開いているパネルがリンクされているリポジトリのイシューのみを表示するというのはできますでしょうか。もしくは、同じ画面で右サイドバーで全てのリポジトリのイシューを表示したままでも良いですが、スクロールで自動で移動し、該当のリポジトリのissue部分に遷移できると助かります。

> スクロールって言うよりは、開いているレポジトリのPRとissue上に来ると良いかな。

> .git dirにgithubのレポジトリ情報があるだろうから、ターミナル開くときにもてない？いまってrepository情報持っているけど、それもowner/name ふくんでないのかな？

> リポジトリを特定できないセルでは従来順で通常どおり開く

## What changed

**Split, not rewrite.** `PrsOverlay` became `GithubPane` (the list) + `GithubOverlay` (the
full-screen frame and route coupling) — the same division `FilesPane` / `FilesOverlay` already
uses, and `FilesOverlay`'s own comment describes. The list itself is unchanged in the move; `git`
records it as a rename.

**Panel exclusivity needed nothing new.** `paneByCell` already holds one pane per cell and the
four existing panes are already mutually exclusive, so `"github"` inherits the rule: opening it
replaces whatever that cell had.

**Leading with the cell's repo, not scrolling to it.** The user's first idea was auto-scroll and
they then refined it; reordering is better on three counts: a scroll that lands slightly off looks
like nothing happened, it has to re-run on every reload, and **a repo with no open PRs and no open
issues has nothing to scroll to** — where an empty section at the top still answers "yours: none".

**Resolving the cell's repo costs nothing new.** `repoForDir()` already maps a directory to
`owner/repo` from `git config --get remote.origin.url` (cached 60s), which is what the user
guessed at. But the pane does not even call it: `/api/repo-dirs` serves the reverse map and the
view **already fetches it** for the issue rows' start control. No extra request, no extra
subprocess.

**A cell with no repository opens normally.** Not a git repo, no `origin`, an unsupported forge,
or simply a clone not registered in Settings — all four give the configured order. That last one
is also the mundane explanation if a repo does not float to the top.

## Two bugs of my own, both caught by specs

Worth reading, because both specs exist *because the same mistakes shipped before*:

- **The button did nothing.** `toggle-github` was emitted with nothing forwarding it.
  `cellChromeBinding.ts`'s comment records exactly this: *"That is what happened to the collections
  button, which shipped in #1573 and never once opened the pane."* I reproduced it; the spec
  written to prevent it caught it.
- **It rendered on un-expanded cells**, missing the `v-if="expanded"` the other pane buttons carry
  — panes only exist beside a zoomed cell.

## Tests

- `common/githubPaneOrder.ts` is pure and separately tested (13 cases): match, no match, no cwd,
  a repo with no section, case-insensitivity on both path and `owner/repo`, several clones of one
  repo, no mutation.
- `GithubPane.spec.ts` gains three: leads with the cell's repo, keeps the configured order for an
  unregistered directory, and renders with no `cwd` at all (the toolbar's host).

## Verification

- `yarn lint` **0 errors**, `yarn typecheck` clean, `yarn build` ok, `yarn test` **9930 passed**
  (691 files).
- **Not run: the app itself.** Port 34567 is serving the user's live session, and a second server
  runs `sweepIdleSessions` against the *shared* tmux socket on boot — the socket name is a
  constant, so `HOME` does not isolate it. With `sessionIdleReapDays: 7` and two unattached
  sessions from Aug 5 and Aug 6 on this machine, starting one could have destroyed them. That is
  not a risk worth taking for a screenshot, so the visual check is deliberately left to a human.

work in mulmoterminal

## Summary by CodeRabbit

- **New Features**
  - Added a GitHub view combining pull requests and issues.
  - Open GitHub full-screen or alongside the active terminal cell.
  - Added contextual GitHub controls, close actions, and pane toggling.
  - Repository sections prioritize the active cell’s repository when available.
  - Added `/github` navigation; existing `/prs` links redirect correctly.

- **Bug Fixes**
  - Improved repository matching across differing path and name capitalization.


isamu:
## iter-1 — two fixes, one from a screenshot and one from Codex

**The pane was squeezing the terminal out of the grid.** Reported with a screenshot: the roster on
the left, the pane on the right, and nothing between them.

The pane's root was missing `min-w-0` and `h-full`. A flex item's default `min-width: auto` refuses
to go narrower than its content, and a PR title is long — so `flex: 0 0 <paneWidth>px` set the
basis and the content overrode it until the terminal had no width left. `truncate` on the rows
could not save it either: it needs an unbroken min-w-0 chain, which the scroller was also missing.

Neither constraint was needed while this was a full-screen `fixed` overlay, which is exactly how
moving it into a flex row lost them. `ToolsPane` (`h-full w-[340px] shrink-0`) and
`CollectionsPane` (`h-full min-h-0`) both carry their own.

This one is worth naming plainly: **it shipped because I decided not to start the app**, and the
reason I gave (a second server would sweep idle tmux sessions on the shared socket) was sound but
left the visual side unverified. The tests could not see it; one screenshot could.

**Codex (CHANGES REQUESTED) — accepted, and correct.** An exact path match only recognised a cell
sitting at the clone ROOT, so any cell deeper than that led with the wrong repo. A cell is very
often under the clone rather than at it.

Containment now, through `common/dirPathKey.ts` — the helper that already exists for this, since
the browser has no filesystem. It folds both separators, a trailing slash, `.` and `..` into one
spelling; case-folded on top for Windows and default macOS.

Two traps the fix had to avoid, both now pinned:

- **a prefix is not containment** — `/srv/mt-marketing` starts with `/srv/mt`, and without the
  separator boundary a sibling borrows its neighbour's repo;
- **the deepest match wins** — clones nest here (a worktree under a managed root, one repo
  vendored in another), and the innermost is the one the cell is in.

Specs 13 → 17 on the ordering module, plus a deliberately weak layout guard on the pane (jsdom
does no layout, so it can only stop the two classes being dropped again).

`yarn lint` 0 errors, `yarn typecheck` clean, `yarn test` **9935 passed**.

isamu:
## iter-2 — Codex was right again, and it is the same class of bug

**`paneFull` treats every pane except files as full-width-capable**, so the GitHub pane could be
widened over the terminal — while shipping without the control that undoes it. Canvas and Tools
both carry one; their own comment asks for exactly this: *"the two panes share one slot, so the
same control must be in the same place in both."*

It is not hypothetical either: expand the tools pane, then switch to a cell that remembered the
GitHub pane, and the terminal is hidden with **no route back to the split**.

The pane now takes `expanded` and renders the same expand/restore button in the same slot.

**Worth naming: this is the same class as the `min-w-0` bug earlier in this PR** — a pane hiding
the terminal — reached a different way. Both came from moving a full-screen overlay into a shared
layout without taking on what that layout expects of its members. The overlay never had to care
about width, height, or who else lives in the row.

Also in this commit, observed while fixing rather than flagged: the emit overloads tripped
`@typescript-eslint/unified-signatures` and are now one signature.

Specs 10 → 12: the button reads "Restore" when expanded and "Expand" when not, and emits
`toggleExpand` from both states.

`yarn lint` 0 errors, `yarn typecheck` clean, `yarn test` **9938 passed**.

isamu:
## iter-3 — the expand button was a no-op in the overlay

**Codex, again correct.** `GithubOverlay` renders the pane and listens for no `toggleExpand`, so
the button added in iter-2 appeared there and did nothing.

The fix is **absence, not a disabled state**: the overlay IS the whole screen, so there is nothing
to expand over. A control that cannot do anything should not be offered.

`canExpand` says which host is rendering — the grid passes it, the overlay does not. It cannot be
inferred from `expanded`, because **Vue casts an absent boolean prop to `false`**, so "not
expanded" and "cannot expand" arrive identical. That is the third time that same casting rule has
mattered in this PR: it also hid every repo heading two commits ago.

## Three findings, one root

Worth stating plainly, because the pattern is the interesting part:

| iter | symptom | what was missing |
|---|---|---|
| 1 | pane squeezed the terminal to nothing | `min-w-0` / `h-full` — what a flex row expects of a member |
| 2 | full-width pane with no way back | the expand control Canvas and Tools both carry |
| 3 | expand control that does nothing | a host that provides one — only the grid does |

All three come from moving a full-screen overlay into a shared layout. The first two are what the
grid expects of the pane; the third runs the other way — what the pane assumed of its host. An
overlay never had to care about width, height, who else is in the row, or who is rendering it.

`yarn lint` 0 errors, `yarn typecheck` clean, `yarn test` **9939 passed**. Specs on the pane
10 → 13.

PR #1771, opened 2026-08-17
fix: retry /api/config so a restarting backend stops emptying the launcher

snakajima:
## The bug

The launcher's saved-directory chips sometimes come up completely empty, then come back on the next reload.

Everything that row shows — the saved directories, the WORKSPACE chip, plus a dozen global settings — comes from a single `GET /api/config` issued once from `GridView`'s `onMounted`. `loadConfigOnce` swallowed every failure (`if (!res.ok) return;` and a bare `catch`), and **nothing re-read it**. So one failed request left the tab showing no saved directories for as long as it stayed open, indistinguishable from a user who has opened none, with nothing on screen saying otherwise.

The failure that produces it is a backend that is merely **restarting**, and in this repo's own dev loop that is routine rather than rare: Vite serves the page from its own port and proxies `/api` to Express, while `scripts/dev-server.mjs` does a full backend restart on every save under `server/`, `common/` or `bin/`. Measured against a real Vite proxy with a dead target:

```
--- page:  200
--- api:   HTTP/1.1 502 Bad Gateway
[vite] http proxy error: /api/config
AggregateError [ECONNREFUSED]
```

The 8s `fetchWithTimeout` never comes into it — ECONNREFUSED is immediate. An HMR update makes it worse: it re-runs the shell's setup, so `presets` is rebuilt as an empty per-instance ref and the list you already had disappears without a page reload.

## The fix

`createConfigLoader` wraps the read in a retry chain.

- The **first attempt is still awaited**, so `onMounted(loadConfig)` means what it always did on a healthy server. Retries are explicitly not awaited — a mount must not block on a server that may never answer.
- The schedule (`configRetryPolicy.ts`) is 500ms doubling to a 5s ceiling over nine attempts, ~32s. Shape matches `reconnectDelayMs` because the terminal sockets are waiting out the same restart; the attempt cap is this module's own, since a config read is a one-shot the user can redo by reloading.
- A chain stops on the first answer, when a later `loadConfig()` supersedes it (remount, HMR), and when the scope is disposed.
- **Only a failed REQUEST retries.** A 200 whose body we cannot read is the server's real answer — asking nine more times returns the same body.
- `trackConfigLoad` still publishes ONE ATTEMPT, never the chain: `recordPreset` waits on it for the worktree root, and a chain that sleeps between attempts would hold a chip's recording for half a minute instead of one bounded request.

`loadConfigOnce` moved out to a module-level `createConfigReader` (like the neighbouring `createPresetManager`) to keep `useAppConfig` inside its line budget.

## Not in this PR

A related way the row can stay empty, worth its own change: `mutateOneOnServer` bumps `version` before its write, so a preset write that FAILS during the initial load makes `adoptServerPresets` reject the in-flight GET's (successful) answer, and the list stays empty with no failed request to retry.

## Tests

`yarn test` (10456 passed), `yarn lint`, `yarn typecheck` all green. New specs cover the schedule itself and, with fake timers: fills in once the backend returns, does not await the chain, gives up instead of polling forever, does not retry an unreadable body, abandons on scope stop, and lets a later load supersede an earlier chain.

work in mulmoterminal

## Summary by CodeRabbit

* **New Features**
  * Added automatic retries when configuration requests fail, with increasing delays and a maximum retry limit.
  * New configuration loads cancel earlier requests and ignore outdated responses.
  * Configuration loading stops safely when the relevant application scope is disposed.

* **Bug Fixes**
  * Prevented malformed or unusable configuration responses from triggering unnecessary retries.
  * Improved handling of failed and non-successful configuration requests.

### DIFF
--- a/src/components/CellLaunchForm.vue
+++ b/src/components/CellLaunchForm.vue
@@ -53,6 +53,10 @@ const props = defineProps<{
   choice: LaunchChoice | null;
   defaultCwd: string | null;
   presets: CwdPreset[];
+  // The saved directories could not be READ — /api/config failed and the retries gave up. The
+  // chip row is empty for a reason the user cannot otherwise see, so it says so instead of
+  // looking like a user who has opened nothing.
+  configUnavailable?: boolean | undefined;
   // Configured launch commands (shell/codex/…) offered next to Claude in this launcher.
   launchers?: Launcher[] | undefined;
   // Session ids open in other grid cells. Resuming one of them would detach that cell, so those
@@ -81,7 +85,9 @@ const emit = defineEmits<{
   (e: "resume", value: { id: string; cwd: string | null; agent?: TerminalAgent }): void;
   (e: "run", value: RunCommand): void;
   (e: "launch", value: LaunchPick): void;
-  (e: "close"): void;
+  // `retry-config`: read the config again after it could not be read at all — the button on the
+  // notice below. The shell owns the read; this only asks for another one.
+  (e: "close" | "retry-config"): void;
 }>();
 
 // An empty field means the server's workspace default — the placeholder is a hint, not a value.
@@ -599,6 +605,13 @@ async function requestRemove(repoDir: string | null, w: Worktree): Promise<void>
     >
       <span class="material-symbols-outlined" aria-hidden="true">close</span>
     </button>
+    <!-- Not a chip: a chip launches somewhere, and there is nowhere to launch here. It sits where
+       the chips would be because that is the emptiness it explains. -->
+    <p v-if="configUnavailable" data-testid="cell-config-unavailable" class="flex w-full items-center justify-center gap-1.5 text-[11px] opacity-70">
+      <span class="material-symbols-outlined text-[13px]" aria-hidden="true">cloud_off</span>
+      Couldn't reach the server, so your saved directories aren't listed.
+      <button type="button" class="underline underline-offset-2 hover:opacity-100" @click="emit('retry-config')">Try again</button>
+    </p>
     <div v-if="chips.length" class="flex w-full flex-wrap justify-center gap-1.5">
       <span
         v-for="p in chips"
--- a/src/components/GridView.vue
+++ b/src/components/GridView.vue
@@ -467,7 +467,7 @@ onMounted(() => (offNewTerminal = registerNewTerminalHandler(openNewTerminal)));
 onBeforeUnmount(detachNewTerminal);
 
 // Server config: the default workspace dir + the auto-recorded dir presets + sound.
-const { defaultCwd, home, presets, launchers, customAgents, loadConfig, recordPreset, removePreset } = useAppConfig();
+const { defaultCwd, home, presets, configUnavailable, launchers, customAgents, loadConfig, recordPreset, removePreset } = useAppConfig();
 const showSettings = ref(false);
 onMounted(loadConfig);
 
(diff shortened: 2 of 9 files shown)


snakajima:
Both Codex findings addressed in 19d6f669.

**1. Malformed body armed the retry chain.** Correct, and it contradicted the comment right above it. The split is now at the response head:

- the request failing (throw / timeout / abort) → `false`, retryable;
- `!res.ok` → `false`, retryable — this is the dev proxy's 502;
- anything after a 200 has arrived — unparseable, wrong shape, or an adoption step that threw → `true`, not retryable. Asking that server nine more times returns the same body.

New spec drives a `json()` that rejects with a SyntaxError and asserts exactly one call, plus one for the 502 that does retry.

**2. A superseded attempt could still adopt (P2).** Also correct, and the module-level singletons (`launchers`, `quickCommands`, the global settings) have no version guard like `adoptServerPresets` has. Each attempt now carries an `AbortController`; a later `load()` and scope disposal both cancel the one in flight. Abort alone is not enough — a response that had already landed when the abort fired would still be adopted — so the attempt re-checks staleness after the body arrives and returns without adopting. Arming an attempt checks and publishes its controller in one synchronous step, so a chain that was just superseded cannot publish over the current one.

New spec: a stalled first load, a second load that answers, then the first response released — the fresh config must survive. Verified it fails without the guard.

`yarn test` (10459 passed), `yarn lint`, `yarn typecheck` green; the new specs run 5/5 stable.

snakajima:
@coderabbitai review

snakajima:
Force-pushed onto `main`. The branch had been cut from a worktree branch carrying four unrelated WIP commits (the `@receptron/sharedapp` 0.9.0 bump and headless-preview work), which is why three of the four review comments — `plans/feat-headless-preview-parity.md`, `server/infra/shared-app-tool.ts`, `test/server/backends/headlessPreview.spec.ts` — and the Socket dependency diff pointed at files this change never touched. The PR is now three commits over `main`, four files, all of them the config-load retry.

On the one comment that WAS about this change (`useAppConfig.ts`, the loader's chain state): correct, and the comment was the wrong half. `createConfigLoader` runs inside `useAppConfig()`, so `generation`/`disposed`/`inFlight` are per call and `stale()` compares only against its own generation — per-call is the intent, since the reader closes over per-call refs (`defaultCwd`, the preset manager's version coordination). The doc now says so, names the invariant that makes it safe (the shell that mounts is the only caller that loads; everything else takes the values from it) and what would break it (a second caller with its own `loadConfig` could adopt a slow answer over a newer one, because the module-level singletons have no version guard of their own).

`yarn test` (10459 passed), `yarn lint`, `yarn typecheck` re-run green on the clean base.

snakajima:
Took the nitpick — it is the same complaint this PR is about, one layer out: half a minute of silent retrying that ends in an unexplained empty chip row is only the original bug made permanent.

`configUnavailable` is set when the retry window is exhausted, and only then — a backend restart must not flash a warning while the retries are still doing their job — and cleared by the next successful read. The launch form renders one line where the chips would be ("Couldn't reach the server, so your saved directories aren't listed") with a **Try again** that asks the shell for another read. Plumbed exactly like `presets`: the shell owns the load, the cell and the grid pass it through, so the notice appears wherever the launch form does.

Not reusing the existing `error` ref: that one is the SAVE paths' message and nothing renders it, so overloading it would have been a second silent state rather than a visible one.

New specs: the flag stays false while retrying and flips when the chain gives up, a later success takes it back, and the form shows nothing on a merely-empty config but shows the notice + emits `retry-config` on click.

`yarn test` 10463 passed, `yarn lint`, `yarn typecheck` green.

Issue #1815, opened 2026-08-20
feat(campaign): run a body of work across 8 clones from the grid — one pipeline, four adapters


isamu:
## The idea

Today a human supplies the grid: you open a cell, you say what to do, you watch. **Campaign Mode changes only where the sessions come from** — an orchestrator supplies them instead of a person — so that a body of work (a refactoring campaign, a large feature) runs across 8 parallel clones and only reaches a human at the points that need judgement.

This is not a new screen. The grid already adopts sessions the server started (`GridView.vue` unplaced sweep), the cockpit roster already shows status and PR phase per cell, and turn completion is already detected by causal correlation (`exchangeRules.ts`). What is missing is the thing that decides *what to run next* and *whether it is done*.

## Where this sits relative to the two existing visions

- **#991** narrows the human's job to judgement. Campaign Mode is the **producer** of those judgement points; #991 is the queue they land in.
- **#1026** closes the loop for **one** issue (triage -> PR -> merge -> close -> tidy). Campaign Mode **runs many of those loops at once** and takes responsibility for what #1026 does not: supply, collision, convergence.

Neither covers what is listed below, which is why this is filed separately rather than as a comment on either.

## The four things that block a campaign today

Observed over roughly a week of running 5-8 parallel sessions against a large TypeScript codebase in a private repository (~1000 PRs, 100+ issues surfaced, starting from easy lint warnings and ending with `any` and oversized functions/files):

| # | Phenomenon | Why a prompt cannot fix it | What is needed |
| --- | --- | --- | --- |
| A | Sessions keep modifying **the same places** | Nobody holds a record of what is claimed | A claim registry: declare paths before touching, lease them |
| B | Review loops ~20 times even for a few dozen lines | "Is it good?" has no termination condition | An oracle declared up front; blocking vs follow-up split; repeat-comment detection; a loop cap |
| C | Problems are still found **after LGTM** | The reviewer shares the author's assumptions and context | Independent verification in a separate context, given only the diff and the settlement condition |
| D | The review loop **stops** even when told not to | Progression lives inside one agent's context | An external state machine: the session executes one step and reports; the runner decides the next |

## Shape

One pipeline with four swappable adapters, so refactoring and new development are two profiles rather than two features:

- **Supply** — where tasks come from (a lint ratchet / a spec / an issue list)
- **Oracle** — what settles "done" (differential vs the old implementation / acceptance tests / a reference implementation / a property or threshold). At least one non-human oracle per task is required.
[...]

PR #1816, opened 2026-08-20
docs: Campaign Mode の設計メモ（汎用パイプラインと grid への載せ方）

isamu:
## Summary

Design notes for Campaign Mode (#1815), under a new `future/` directory. **Documentation only — no code, no behaviour change.**

- `future/autonomous-delivery-pipeline.md` — the generic design. One invariant pipeline (Intake -> Planned -> Claimed -> Implementing -> SelfCheck -> Review -> Verify -> MergeQueue -> Learn) plus four swappable adapters (Supply / Oracle / Partition / Gate), so refactoring and new development are two **profiles** rather than two features. Includes the oracle catalogue, the two partition axes, and a 51-item feature list.
- `future/grid-campaign-mode.md` — how that lands on this repo's grid: what already exists, the four backbone decisions, the three-item UI delta, the `server/campaign/` layout, and the PR0-PR10 split.

`future/` rather than `plans/` because these describe a direction, not one change. Each PR out of them still gets its own `plans/` file.

## What the review changed

Four rounds with Codex and CodeRabbit turned up 19 findings, all accepted. Two were defects in the design rather than gaps in its description, and are worth naming here:

- **D3 contradicted itself.** It said `1 clone = 1 session = 1 task` while the sequence ran Review and Verify in separate sessions. Now `1 clone = 1 task`, one throwaway session per step, same clone and same revision throughout — which is also what "judge in a different context" needs.
- **D4 stated a guarantee its mechanism could not deliver.** "Agents cannot start a campaign" rested on giving them only report-only MCP tools; a session has a shell in its clone and can call the loopback route. It is now a server-enforced capability, and agent reports are treated as untrusted input rather than as a basis for claim or merge decisions.

The rest hardened three areas the first draft was silent on: crash recovery (intent-then-outcome records, idempotency keys embedded in external names, startup reconciliation, fenced leases), merge authority (auto-merge fails closed on the goal file's allowed paths, max diff, and effective branch protection — a claim is what the task declared, not what it was permitted), and oracle validity (each oracle declares its ground truth, scope, environment, generator conditions, timeout, evidence and fail-closed behaviour).

Two more came from my own read rather than a bot: the UI inventory said "three additions, that's all" while a fourth sat below it, and the PR split had not been updated to carry the requirements the review added.

The full trail is in the ledger comments and the round-5 checkpoint.

## Items to Confirm / Review

- **Is `future/` the right home?** `plans/` is per-change and these are not. If the preference is `plans/` or `docs/`, say so and I will move them.
- **D1 argues against a comment in `useRoundTable.ts`.** That file says a conversation continuing where nobody watches "is a different feature with a different safety argument". The doc claims Campaign Mode *is* that feature and supplies the argument (declared budget, audit log, stop switch, human-gate notification). Worth a second opinion on whether that argument is strong enough.
- **The oracle catalogue and the 51-item list are unvalidated.** Nothing here has been built; the list is a design inventory, not a commitment.
- The campaign facts cited come from a private repository. No repository name, URL or issue number appears — only the scale.

## User Prompt

- Walked through a real large-scale refactoring campaign: introduce ESLint against a human-written TS/JS codebase with no types and failing lint, then tighten progressively so the code becomes readable to an AI. Roughly a week, largely manual, ~1000 PRs, 100+ issues surfaced, 5-8 parallel sessions. Started from easy lint warnings and ended with the hard `any` and oversized functions/files.
- Four things blocked it: sessions keep modifying the same places however they are coordinated; review loops ~20 times even for a few dozen lines; problems are still found after LGTM; the review loop stops even when told not to. "Get past these and almost all of it could be automated."
- Working principles: an issue may be a dummy, so establish it is real before acting; widen the view before narrowing the fix; extract pure functions around the problem; write exhaustive normal- and abnormal-case tests, and keep growing them as new bugs surface.
- "Write this up as a doc under a `future/` dir, and list what features would be needed." Personal preference: **8 clones rather than worktrees**, with an orchestrator managing conflicts and driving planning -> change -> secure review -> review -> unit/e2e test. The same base should later support new features and new apps.
- "Generalise the pipeline so it works for new development and large feature additions too, with refactoring and new development as two instances of it."
- "Now seriously consider this as one of MulmoTerminal's grid features. Ideally the UI does not change much and it runs fully automatically."
- Decisions given during the design: first profile is `refactor`; clones live in a dedicated directory; `RoundTableDeps` / `CrossTalkDeps` may be touched.

work in mulmoterminal2

## Summary by CodeRabbit

- **Documentation**
  - Expanded autonomous delivery and Campaign Mode guidance with lifecycle states, recovery paths, and progress reporting.
  - Added approval workflows tied to candidate and base revisions, including freshness checks and invalidation handling.
  - Documented safeguards for stopping runs, enforcing budgets, validating results, and controlling execution capabilities.
  - Clarified session isolation, leases, branch protection, commit verification, and safe serial merging.
  - Added guidance for audit trails, crash recovery, retries, reconciliation, protected policy storage, and fail-closed behavior.
  - Documented UI support and phased implementation milestones for future automation.


isamu:

Findings raised so far and what happened to each. Re-flagging a `FIXED` row is welcome **if the resolution is wrong** — say which row and what is wrong with it.

All 17 findings were accepted; none rejected. Two of them (rows 1 and 13) were defects in the design rather than gaps in its description.

|---|---|---|---|---|
| 1 | codex P1 | "agents cannot start a campaign" is not enforceable by restricting MCP tools — the session has a shell in the clone and can call the loopback route | FIXED | D4 rewritten: server-enforced capability on start/stop/budget, made a prerequisite of PR2. Added as core principle 5 in the generic doc |
| 2 | codex P1 | no restart reconciliation or idempotency — duplicate work, stranded leases | FIXED | State machine gained `Merged` / `Orphaned` / `Escalated` / `Stopped`; new 2.1.1 (intent-then-outcome, idempotency key embedded in the external name, startup reconciliation); grid doc gained the same as 中断からの再開 |
| 3 | CR | state machine has no crash / lease-expiry / stop / external-failure transitions | FIXED | same as row 2 |
| 4 | CR | "convert every blocking comment into a failing test" is too broad | FIXED | R5 limited to findings expressible as a test; permissions, config, UI, docs, migration and perf thresholds go to a direct oracle / config check / human gate |
| 5 | CR | non-human oracles are not self-validating | FIXED | 3.1 gained a required-fields table per oracle (ground truth, scope, reproducible environment, generator conditions, timeout, evidence, fail-closed) and a note that non-determinism must be pinned or the differential oracle cannot be used |
| 6 | CR | a claim registry is not exclusion | FIXED | 3.2: normalised identity, compare-and-set, expiry, owner token fencing. `clone-pool.ts` row updated |
| 7, 17 | CR | `CI green` is not sufficient for auto-merge | FIXED | New 4.4 with five conditions; grid doc section 6 carries the same list |
| 8 | CR | a contract change must invalidate downstream tasks | FIXED | Contract ids are versioned and immutable; dependents are stopped or returned to `Implementing` |
| 9 | CR | the goal file declares no execution authority | FIXED | P1 gained target branch, allowed paths, allowed commands, network, credentials, max diff, merge rights — deny by default |
| 10 | CR | audit log needs classification, access control, retention | FIXED | P7 and D4: append-only, references and hashes rather than raw prompts/diffs/credentials, access control, retention |
| 11 | CR | V1's Done conditions are refactor-specific | FIXED | Per-profile predicates with a versioned baseline; generated code excluded from measurement |
| 12 | CR | W3's main-sync policy is not executable as written | FIXED | Explicit git contract: fetch, `git merge origin/main` into the working branch, never rebase, never reset, commit before taking main in |
| 13 | CR | `1 clone = 1 session = 1 task` contradicts Review and Verify running in separate sessions | FIXED | D3 rewritten to `1 clone = 1 task`, one throwaway session per step, same clone and same revision throughout — which is also what principle 3 needs. Sequence diagram updated to match |
| 14 | CR | human-only, stop switch and budget must be enforced server-side | FIXED | D4: capability on the routes, stop and budget re-read atomically before every side effect, campaign budget kept separate from the rate-limit headroom |
| 15 | CR | JSONL persistence does not define a resume protocol | FIXED | same as row 2 |
| 16 | CR | MCP reports must be treated as untrusted input | FIXED | Reports are a record of what the agent thought, never the basis for claim or merge; touched paths are derived from `git diff`; duplicates, out-of-order and oversized reports rejected |

The three mermaid blocks were re-parsed under jsdom after the rewrite — all three parse.

isamu:

Re-flagging a `FIXED` row is welcome **if the resolution is wrong** — say which row and what is wrong with it.

Iteration 1 rows 1-17 are in the [previous ledger](https://github.com/receptron/mulmoterminal/pull/1816#issuecomment-5352491568) and unchanged. New this round:

| # | iter | bot | finding | disposition | where |
|---|---|---|---|---|---|
| 18 | 2 | codex P1 | the auto-merge gate checks oracle / revalidation / review / claim state, but nothing verifies the goal file's allowlist, max diff, merge rights, or that the target branch's protections are actually in force | FIXED | 4.4 gained three fail-closed conditions (allowed paths and forbidden paths, max diff, target branch + protections + a credential that cannot bypass them), plus an explicit note that **4 and 5 are different checks**: a claim is what the task declared, the authority is what the goal file permits, so a task that claims a forbidden path passes the claim check. Section 6 of the grid doc carries the same list |

The finding also asked that the human gate stay the default until those checks exist — added, with the reason: putting auto-merge in first and the authority checks second means whatever merged in between does not come back.

**Iteration 2 signals:** Codex `CHANGES REQUESTED` (1, above). CodeRabbit reviewed head `2261397e` and posted nothing actionable. Clean-round count: 0.

isamu:
## Checkpoint — round 5

**Does the diff still do what the PR body says?** Yes — two design notes under `future/`. But the body was written before four rounds of changes and no longer mentions what the review added, so it is being refreshed alongside this comment.

**Original vs review-response.** 655 lines at open, 811 now: roughly 19% of the diff is review response. The centre of gravity has not moved — it is the same two documents, hardened in three places (enforcement, crash recovery, merge authority).

**Anything here only because a bot asked?** No. Every addition is a constraint the design was wrong without. Two of them corrected outright defects rather than gaps in the description: `D3` contradicted itself, and `D4` stated a guarantee that its mechanism could not deliver.

**What the checkpoint itself found.** The review added requirements but **the implementation split was never updated to carry them** — PR2 had accumulated two jobs (capability boundary plus the walking skeleton), and startup reconciliation had no home in the table at all. Fixed in the commit above: reconciliation joins PR1, the capability boundary becomes PR2 on its own, the runner moves to PR3, and the rest shift down (eleven PRs now). Both sit before the runner deliberately — added afterwards, neither protects the campaigns that ran in between.

**Would I approve this fresh today?** As a design note, yes. It describes something nobody has built, which the body says plainly.

**Largest remaining risk.** The design is now considerably more demanding than the PR list makes it look. "PR4 is the first useful line" is still true for *one task running unattended*, but the safety properties this review established (capability boundary, reconciliation, authority checks at merge) are most of the work before any of it can run without a person watching. That is a scheduling risk, not a correctness one — and it is better to know it now than at PR7.

@coderabbitai full review

isamu:
## Review ledger — after iteration 3 (heads `24e0e3c3`, `d75685b5`)

| # | iter | bot | finding | disposition | where |
|---|---|---|---|---|---|
| 19 | 3 | codex P1 | `D3`'s corrected lifecycle is contradicted by a leftover `1 clone = 1 session = 1 task` in the clone-location decision | FIXED | Section 9 decision 2 now quotes the corrected form, and says why it follows: **the lease belongs to the task**, which is what puts the clones somewhere the runner lends and reclaims rather than beside a person's working directories |
| 20 | 3 | me | the UI inventory said "three additions, that's all" while a fourth (the toolbar tally) sat below it | FIXED | Added as row 4 of the table and to PR9's scope. Not flagged by any bot |
| 21 | 3 | me (round-5 checkpoint) | the review added requirements that **the PR split never picked up** — PR2 had accumulated two jobs and startup reconciliation had no home at all | FIXED | Reconciliation joins PR1; the capability boundary becomes PR2 alone; the runner moves to PR3; eleven PRs now. Both sit before the runner on purpose |

Row 19 is the useful kind of repeat: I corrected `D3` in iteration 1 and left a second statement of it standing four sections away, so the document asserted both versions at once. Grepping for the corrected claim rather than the corrected *heading* is what would have caught it.

**Iteration 3 signals:** Codex `CHANGES REQUESTED` (1, above), fixed in `4d1b861c`. CodeRabbit's last review is of `2261397e` — it has not read `24e0e3c3`, `d75685b5` or `4d1b861c`, and a `full review` request went unanswered. Its silence is not counted as approval. Clean-round count: 0.

isamu:
## Review ledger — after iteration 4 (head `4d1b861c`)

| # | iter | bot | finding | disposition | where |
|---|---|---|---|---|---|
| 22 | 4 | codex P1 | the state machine has no `MergeQueue --> Stopped`, so the stop switch and the budget cap do not reach a task that is already queued — and merge is the one irreversible step | FIXED | Transition added, plus the reason it matters: a stop that everything already in the queue survives is not a stop. The note also states the other half — **`Merged` cannot be stopped**, because it is a fact that already happened outside, so the check belongs immediately *before* the merge call rather than after it |

This is the third round in a row where the finding was a real hole rather than a wording problem, and all three share a shape: a rule stated in one place and not carried into every place it governs (D4's guarantee vs. its mechanism, D3's lifecycle vs. its restatement, now the stop switch vs. the state machine). Worth saying out loud, since the next one will probably look the same.

**Iteration 4 signals:** Codex `CHANGES REQUESTED` (1, above), fixed in the push below. CodeRabbit's last review remains `2261397e` — four heads behind, `full review` unanswered, counted as no signal rather than as approval. Clean-round count: 0.

isamu:
## Review ledger — after iteration 5 (head `b853404c`)

| # | iter | bot | finding | disposition | where |
|---|---|---|---|---|---|
| 23 | 5 | codex P1 | time-of-check/time-of-use: a stop accepted after the atomic reread but before the spawn / PR / merge still lets the irreversible action through | FIXED | 2.1.1 gained the serialization rule — stop and side-effect intents go through **one writer**, intents carry a generation, the executor re-checks it immediately before the call. D4 in the grid doc carries the short form |

**The resolution deliberately does not claim the race is closed.** An external side effect and a local state write cannot be made atomic, so the document now states the boundary instead of asserting a guarantee it cannot keep:

> A stop cancels every intent not yet committed. An intent already committed completes and is recorded. A merge in flight at the moment of the stop can still land.

Two consequences were written down with it: the budget works the same way (it refuses new commits rather than killing committed ones), and **the more irreversible the operation, the shorter the gap between committing the intent and performing it must be** — so revalidation and CI waiting belong at the front of the merge queue, not immediately before the merge call.

If a reviewer thinks that boundary is the wrong place to draw the line, that is the thing to argue with — it is a decision, not an oversight.

**Iteration 5 signals:** Codex `CHANGES REQUESTED` (1, above), fixed in `4465ca0a`. CodeRabbit's last review remains `2261397e`, five heads behind; `full review` unanswered; recorded as no signal, not as approval. Clean-round count: 0.

isamu:
## Review ledger — after iteration 6 (head `4465ca0a`)

| # | iter | bot | finding | disposition | where |
|---|---|---|---|---|---|
| 24 | 6 | codex P1 | parallel Review sessions share a **writable** clone with no pinned candidate, so review and verify race with agent-initiated mutations and a PR can carry unreviewed changes | FIXED | D3: **the candidate is a commit, not a directory.** Implement ends in a commit; Review and Verify are handed that SHA and run in read-only checkouts; anything written during them is discarded and comes back as a blocking finding that produces a *new* candidate. Auto-merge gained a condition checking the PR head against the reviewed SHA |
| 25 | 6 | me | that new condition contradicted condition 2 — taking main in for revalidation **changes the head**, so "PR head equals the reviewed SHA" can never hold after it | FIXED | The condition now reads "the reviewed candidate, or a main-merge of it **made by the runner**". The property being protected is *no agent-authored change after review*; taking main in creates a new candidate, which condition 2's oracle re-run already covers. Not flagged by any bot |

Finding 24 is the third one that traces back to the same original mistake — D3 said "same clone" when what it needed was "same commit". Fixing the lifecycle (row 13) and fixing the leftover restatement (row 19) both left that substitution in place, because both times I corrected the wording rather than the claim.

**Iteration 6 signals:** Codex `CHANGES REQUESTED` (1), fixed in `4580dc1a` along with the contradiction it created. CodeRabbit's last review remains `2261397e`, six heads behind; `full review` unanswered; recorded as no signal. Clean-round count: 0.

isamu:
## Review ledger — after iteration 7 (head `4580dc1a`)

| # | iter | bot | finding | disposition | where |
|---|---|---|---|---|---|
| 26 | 7 | codex P1 | the capability boundary guards the **routes**, not the **files** — a shell running as the same user can rewrite `campaigns/*.jsonl` and change budget, allowed paths, leases or forged outcomes without touching HTTP at all | FIXED | New section in 4: isolation (different UID / mount namespace) is the only real prevention and is treated as an explicit deployment option rather than an assumption, because it conflicts with running local PTYs as the user; without it, records are hash-chained with the head held in the runner's memory and re-verified before every use, and a break stops the whole campaign. Also stated in the dangers list and in core principle 5 |

**The guarantee is stated as what it is.** Without isolation it is *"tampering stops the campaign"*, not *"tampering is impossible"* — the same posture as the stop-switch boundary in row 23, and for the same reason: writing down a guarantee the mechanism cannot keep is how a design gets trusted for something it does not do.

One cheap measure came with it: the campaign store lives outside every clone and outside the repository tree, so a task touching it is out-of-claim and the merge gate rejects it.

**Iteration 7 signals:** Codex `CHANGES REQUESTED` (1, above), fixed in `ca764f9f`. CodeRabbit's last review remains `2261397e`, seven heads behind; `full review` unanswered; recorded as no signal. Clean-round count: 0.

Seven rounds, 26 findings, none rejected. The pattern is worth naming: every round since the first has found a place where **a stated guarantee outran its mechanism** — MCP restrictions for "agents cannot start", the clone for "same revision", the atomic reread for "stop works", the route capability for "the state is protected". That is not the loop failing to converge; it is a design document being read by something that checks whether each promise has a machine behind it.

isamu:
## Review ledger — after iteration 8 (head `ca764f9f`)

| # | iter | bot | finding | disposition | where |
|---|---|---|---|---|---|
| 27 | 8 | codex P1 | the hash chain from row 26 is not tamper-evident against this adversary — a same-UID agent computes the same hashes, so it can append a valid forged record or rewrite and rehash the whole log | FIXED | The claim is **retracted, not patched**. Section 4 now says plainly that a hash chain is not a defence here, and that the runner's in-memory head only covers its own lifetime — after a restart it re-reads the log and cannot tell a rehashed one from the real one |

**This was my fix from the previous round, not the original design.** Row 26 asked for tamper-evidence and I reached for the first mechanism that sounds like it; the mechanism does not survive an adversary who can run code as the same user. Attributed here because it changes which commits are worth reading hardest.

What replaced it: authority is defensible only when **something is out of the agent's reach** — isolation (different UID / namespace), or a separate privileged process holding the MAC key or owning the log. A key in the runner's own memory does not qualify, for the same reason the hash chain did not.

And the part that matters more than either: **when neither is deployed, the trust domain must not hold irreversible power.** Auto-merge off, the merge credential kept outside the campaign's reach, branch protection that a runner credential cannot bypass. Stated as a deployment line rather than a nuance — without isolation, Campaign Mode is *a parallel runner a human presses merge on*, not *a thing that updates main unattended*.

The "keep the store outside every clone" measure from row 26 was also narrowed: it protects an honest task from a mistake, not an adversarial one from a choice.

**Iteration 8 signals:** Codex `CHANGES REQUESTED` (1, above), fixed in `c9057778`. CodeRabbit's last review remains `2261397e`, eight heads behind; `full review` unanswered; recorded as no signal. Clean-round count: 0.

isamu:
## Review ledger — after iteration 9 (head `c9057778`)

| # | iter | bot | finding | disposition | where |
|---|---|---|---|---|---|
| 28 | 9 | codex P1 | the rollout puts the shell-bearing runner (PR3) **before** the goal file's default-deny command / network / credential authority, which had been folded into PR10's auto-merge work | FIXED | Execution authority moves into PR2, ahead of the runner, with the reason stated in the section: PR3 is the first point at which an agent with a shell runs where nobody is watching, so if commands, network and credentials are not deny-by-default by then, **that boundary never closes at all** before autonomy starts. PR10 keeps only the merge-gate conditions |

This is the second scheduling finding (row 21 was the first) and they share a cause: the review kept adding *requirements*, and I kept adding them to the prose without asking whether the **order** still worked. A requirement that arrives after the thing it constrains is not a requirement.

**Iteration 9 signals:** Codex `CHANGES REQUESTED` (1, above), fixed in `7be54c69`. CodeRabbit's last review remains `2261397e`, nine heads behind; `full review` unanswered; recorded as no signal. Clean-round count: 0.

isamu:
## Checkpoint — round 10, and a decision that is not mine to make

### The finding

| # | iter | bot | finding | disposition |
|---|---|---|---|---|
| 29 | 10 | codex P1 | PR2/PR3 promise deny-by-default execution authority before shell-bearing agents run, but in a same-UID deployment those agents can rewrite the policy itself — and a human merge gate does not prevent command, network or credential abuse | PARTLY FIXED, then **escalated** |

Correct on both halves, and the second half retires my answer from row 27. Bounding the *consequences* at the merge only bounds the merge. Abuse of commands, network or credentials never passes through a merge at all, so turning auto-merge off does nothing for it.

### What I fixed without asking

The document was over-claiming, and that part is not a judgement call. It now says plainly that on a same-UID deployment, deny-by-default execution authority is **a guard rail, not a security boundary** — it protects against mistakes and against instructions planted in a repository, not against an adversary — because a policy that the constrained process can rewrite does not constrain it.

It also records the other half of the picture, which the finding does not mention and which matters for the decision: **this is not new power that Campaign Mode confers.** An agent in an ordinary cell already runs with the user's full authority; that is what running local PTYs as the user means. What Campaign Mode changes is that **nobody is watching** and that it happens **eight at a time**.

### What I will not decide

Which way to resolve it determines whether the feature is usable on a normal development machine, so it goes to the repository owner rather than into a commit:

| | What it means | What it costs |
|---|---|---|
| **Require isolation** | The policy store lives under a different UID or a privileged process; PR2 **refuses to start a campaign** when it is not protected, and PR3 cannot run until that holds | Campaign Mode stops working out of the box on an ordinary dev machine; deployment gets heavier |
| **Accept the ordinary-cell trust model** | Authority is implemented as a guard rail and never described as a boundary; the exposure of running unattended and in parallel is accepted knowingly | Instructions planted in a repository can act, unattended, across eight clones |

Recorded as decision 4 in section 9, marked as needing an answer **before PR2** rather than deferred with the rest.

### The checkpoint itself

Ten rounds, 29 findings, none rejected; 811 lines at open, ~950 now. Every round since the first found the same species of defect: **a stated guarantee with no machine behind it.** The document is much harder now than when it opened, and the honest reading of that is not "the review converged" but "the first draft asserted safety properties it had not designed". Three of them (rows 26→27→29) were successive attempts at one problem, two of those written by me in response to the previous round.

Would I approve it fresh today? Yes as a design note — provided the header decision above is answered rather than left implicit, which is exactly why it is now a numbered open decision instead of a paragraph.

**Largest remaining risk:** unchanged from round 5 and now sharper. The safety work is most of the work, and PR3 — "one task runs unattended" — is downstream of a product decision nobody has made yet.

Not counting a clean round. Waiting on the owner for decision 4.

isamu:
## Review ledger — after iteration 11 (heads `15efb456`, `6822bca0`)

| # | iter | bot | finding | disposition | where |
|---|---|---|---|---|---|
| 30 | 11 | codex P1 | with the same-UID model chosen, the execution controls still cannot be described as stopping prompt injection — an injected agent rewrites the policy those controls read. The user-facing disclosure and the stated guarantee have to match the trust model | FIXED | Section 4 now separates what the guard rail catches from what it does not, and the risk disclosure is written out rather than gestured at |

The sentence it caught was one I wrote in the previous commit, recording the owner's decision: *"the guard rail stops mistakes and prompt-injection accidents"*. The second half is false for exactly the reason row 29 established, and I wrote it anyway while summarising a decision that had just been made on the basis of that reason.

What it says now:

- **Caught:** a cooperating agent going out of bounds — wrong directory, out-of-scope file, unexpected command. Common, and worth the guard rail on its own.
- **Not caught:** instructions planted in the repository. An agent following them can rewrite the policy the check consults, and a deny list that can be rewritten does not deny.

Two things follow, and both are now written down. The user-facing copy states the exposure without softening it — *the contents of the target repository are read by eight agents with nobody watching, and instructions written there can execute with your authority; the permission settings here do not stop that*. And the operating rule is one line: **point campaigns only at repositories whose contents you trust.** Wanting to run an untrusted one is the signal that isolation now needs deploying.

"Do not call it a security boundary" has a sibling rule beside it: **do not write that it stops injection**, in the document, in code comments, or in the UI.

**Owner decision recorded:** decision 4 resolved as *accept the ordinary-cell trust model* — the authority an agent holds here is the authority it already holds in any cell today; what Campaign Mode adds is exposure, not power, and requiring isolation would make the feature unusable on an ordinary development machine.

**Iteration 11 signals:** Codex `CHANGES REQUESTED` (1, above), fixed in `41eaa674`. CodeRabbit's last review remains `2261397e`, eleven heads behind; `full review` unanswered; recorded as no signal. Clean-round count: 0.

isamu:
## Review ledger — after iteration 12 (head `41eaa674`)

| # | iter | bot | finding | disposition | where |
|---|---|---|---|---|---|
| 31 | 12 | codex P1 | with the same-UID model chosen, campaign policy and state stay agent-writable, yet the `refactor` profile still merged automatically | FIXED | `refactor`'s merge gate is now **human** in this deployment. Auto-merge survives only as the option that a protected policy store unlocks |

This closes a loop the decision left open, and the reasoning is short enough to state in full: the auto-merge conditions are read from campaign state — oracle results, claim, budget, allowed paths — and that state is writable by the party being merged. Merging on it unattended means **updating main from data the merged code can edit, with nobody watching.** That is past the exposure the owner agreed to accept, and it is not what "accept the ordinary-cell trust model" was chosen to mean.

Two things kept, deliberately:

- **The condition list stays, and gets a second job.** It is what unlocks auto-merge under a protected deployment, *and* what must hold before a human is asked to press merge. A human gate does not excuse the machine from checking what the machine can check.
- **The `refactor` profile keeps its point.** Supply through implementation, review, verification and PR all still run unattended; a person presses merge once at the end. The reason it was chosen to go first — a lint ratchet supplies its own fuel, so "does it run by itself" is actually testable — is untouched.

**Iteration 12 signals:** Codex `CHANGES REQUESTED` (1, above), fixed in `d2ddb820`. CodeRabbit's last review remains `2261397e`; `full review` unanswered; recorded as no signal. Clean-round count: 0.

isamu:
@coderabbitai review

isamu:
## Correction — what CodeRabbit's silence actually was

I have been reporting CodeRabbit as "N heads behind, `full review` unanswered" since iteration 3. The conclusion was right — it has not reviewed the recent heads — but **the evidence I was checking was the wrong comment.** I was reading `updated_at` off the last CodeRabbit comment in the list, which is its reply to my `full review` invocation, not its walkthrough.

The real reason is in the walkthrough itself:

> **Review limit reached** — you've reached your PR review limit, so we couldn't start this review. You've used all free OSS reviews for now.

So: **rate-limited**, not ignoring the request. That is still not approval — a reviewer that could not run has not signed off — but it is a different fact from the one I posted five times, and the ledger should not carry a wrong reason for a right conclusion.

Waiting for the limit to reset rather than declaring convergence around it. Codex has been re-run on the unchanged head `d2ddb820`; CodeRabbit will be asked for the same head once it can accept the request. Neither of those pushes a commit, so whatever they return reviews exactly what is already here.

**Iteration 13:** Codex `LGTM` on `d2ddb820` — its first after twelve rounds of changes-requested. Not counted as a clean round: a commit was pushed in it, and one reviewer could not run.

isamu:
@coderabbitai review

DeepPattern-AI:
Hi isamu, sorry to bother you. This design note is very relevant to a workflow problem I’m trying to understand: turning autonomous delivery into an explicit pipeline from Intake through SelfCheck, Review, Verify, MergeQueue, and Learn, with adapters rather than ad-hoc per-task behavior.

At DeepPattern AI, we’re building a quality/evidence layer for AI-assisted engineering workflows, focused on keeping agent-driven work traceable across context, verification, review, and handoff.

We’re inviting a small group of developers with hands-on AI dev tool / coding-agent workflow experience to an external testing program. Your Campaign Mode pipeline design is closely related to what we want to learn from. If this sounds interesting, could you reply and let me know?

isamu:
@coderabbitai review

isamu:
## Review ledger — after iteration 14 (re-review of `d2ddb820`)

| # | iter | bot | finding | disposition | where |
|---|---|---|---|---|---|
| 32 | 14 | codex P1 | the deployment now requires a human merge gate, but the state machine and the sequence still go straight from `MergeQueue` to merged — no recorded approval state, and no stop/budget behaviour while waiting | FIXED | `AwaitingApproval` added to both, with the rules that come with it |

Found on a **re-review of an unchanged head** — Codex had returned LGTM on this same commit one round earlier. Worth noting rather than glossing: the clean round that was about to become the first of two was not clean, and the thing it missed was a direct consequence of the change made three rounds before it.

Same shape as rows 19, 22 and 32 itself: I changed the gate table in section 6 and did not carry it into the machine. Three of the last six findings are this, which says the review is finding a habit rather than a series of unrelated slips.

What the state now carries:

- **Approval is granted against a candidate SHA.** If the PR head moves, the approval is void and the task returns to the queue.
- **The clone is returned while waiting** — the candidate lives on the branch — but **the path claim is held**, because releasing it lets another task edit the same place and the approved diff goes stale.
- **Stop and budget reach the waiting state.** A stopped campaign does not leave approvals pending.
- **No timeout on the wait** — a person not pressing merge is not an error — but a backlog of waiting tasks eats the parallelism, so **supply stops when approvals pile up**.

**Iteration 14 signals:** Codex `CHANGES REQUESTED` (1, above). CodeRabbit reviewed `7be54c69..d2ddb820` with no actionable comments — its first real signal since iteration 2, after the rate limit cleared. Clean-round count: 0.

isamu:
## Review ledger — after iteration 15 (head `d2bfb037`)

| # | iter | bot | finding | disposition | where |
|---|---|---|---|---|---|
| 33 | 15 | codex P1 | an approval stays valid after **another task advances main**, so `AwaitingApproval -> Merged` bypasses the current-main integration and revalidation that condition 2 requires | FIXED | Approval now binds to a **pair** — the candidate SHA *and* the base it was validated against (the `origin/main` SHA taken in during revalidation). Either one moving voids it and returns the task to the queue |

The approval rule I added last round watched the candidate and not the base. In an eight-way campaign **main moves constantly**, so an approval granted in advance is stale by the time it is used — the reviewer's framing ("the validated base is stale") is exactly right, and it turns out to constrain *when* the human is asked, not just what the approval is bound to:

- **Ask at the head of the queue**, immediately after integration and revalidation, and merge as soon as the answer comes back. Collecting approvals ahead of time produces a pile of approvals against bases that no longer exist.
- **Do not stall the queue on a person.** The head is handed to the next task while one waits; a task that was waiting revalidates against the new base and asks again. Otherwise human response time becomes the campaign's throughput limit.

**Iteration 15 signals:** Codex `CHANGES REQUESTED` (1, above), fixed in `824d7c2b`. CodeRabbit is **rate-limited again** (the free OSS allowance, two per window) and did not review `d2bfb037`; recorded as no signal, not as approval. Clean-round count: 0.

Fifteen rounds, 33 findings, none rejected. Both of the last two came from re-reading a mechanism I had just added in response to the round before it — which is the loop doing what it is for, but also a fair signal that this document is being asked to hold more design than a design note usually does.

isamu:
## Review ledger — after iteration 16 (head `824d7c2b`)

| # | iter | bot | finding | disposition | where |
|---|---|---|---|---|---|
| 34 | 16 | codex P1 | the pre-merge recheck of candidate and base is still not atomic with the forge call — main can advance in between | FIXED | The check **moves to the forge**, which is the only party that can make it atomic |

Same shape as row 23, one layer further down: checking a condition ourselves and then calling out is a race no matter how tightly the two are written together. So the condition travels with the call instead of preceding it:

- **Merge with the expected head SHA attached** — GitHub's `PUT /pulls/{n}/merge` takes `sha` and refuses if the head moved.
- **Let branch protection judge base freshness** — "require branches to be up to date before merging" is the same check, enforced by the side performing the merge. Our revalidation exists to *run the oracles*; deciding whether the base is current was never our job.
- **Refusal (409, not up to date) returns the task to the queue** for re-integration, revalidation and re-approval.
- **Merges within a campaign are serialised**, one in flight. That removes the races we cause ourselves; what remains is main advancing from outside, which only the forge-side condition can catch. Both are needed, and for different reasons.

**Iteration 16 signals:** Codex `CHANGES REQUESTED` (1, above), fixed in the push above. CodeRabbit's rate limit has cleared but it has not yet reviewed this head. Clean-round count: 0.

isamu:
## Review ledger — after iteration 17 (head `eb1d5f63`)

| # | iter | bot | finding | disposition | where |
|---|---|---|---|---|---|
| 35 | 17 | codex P1 | `D3` says a task holds its clone from start to finish, while the approval flow returns the clone before `AwaitingApproval` — one lease lifecycle needs defining so reuse and crash recovery are unambiguous | FIXED | The clone lease and the path claim are now given **separate lifetimes**, in a table, because they were never the same thing |

| | taken | released | can it move? |
|---|---|---|---|
| **clone lease** | `Claimed` | on entering `AwaitingApproval` | **yes** — a task sent back to the queue may take any free clone |
| **path claim** | `Claimed` | `Merged` (or abandonment) | no — held throughout the wait |

The clone can go because by then the candidate *and* the validated base both live on the forge; nothing is left in the working tree that matters. The claim cannot, because releasing it lets another task edit the same place and the approved diff goes stale.

Two consequences for recovery, now written next to the reconciliation rules: **a task in `AwaitingApproval` holding no clone is correct**, not a stranded lease to reclaim; and **the same task still holding its path claim is also correct**, not a leak to collect. Both are exactly the kind of thing a reconciliation pass gets wrong by tidying.

This is the **third** finding rooted in D3 — rows 13, 19 and now 35. The first two corrected what the sentence said; this one corrects what it conflated. That progression is worth recording because it says the earlier fixes were incomplete rather than wrong: "same clone" was standing in for three different guarantees (same revision, exclusive workspace, exclusive paths), and they have now been separated into the revision pin (row 24), the lease, and the claim.

**Iteration 17 signals:** Codex `CHANGES REQUESTED` (1, above), fixed in `4cc5d04e`. CodeRabbit's limit has cleared but it has not posted a review of `824d7c2b` or `eb1d5f63`; recorded as no signal. Clean-round count: 0.

isamu:
## Review ledger — after iteration 18 (head `4cc5d04e`)

| # | iter | bot | finding | disposition | where |
|---|---|---|---|---|---|
| 36 | 18 | codex P1 | `D3` still opens with "a task occupies one clone from start to finish", contradicting the lease table added one round earlier | FIXED | Opening now reads "while it is working — `Claimed` to `AwaitingApproval`" |
| 37 | 18 | codex P1 | a rejected `AwaitingApproval` task transitions straight to `Implementing` although it no longer holds a clone | FIXED | The edge goes to `Claimed` instead, where a free clone is taken again; the path claim was never released, so only the workspace is re-acquired. If none is free the task waits there. `Orphaned` recovery takes the same route — the clone it had before the crash may not be available |

Both are leftovers from row 35's fix rather than new ground: I added the lease table and did not carry it into the paragraph three lines above it, nor into the one edge that leaves the new state. Fourth D3-rooted finding, and the same habit named in iterations 14 and 17 — the correction lands where the reviewer pointed, not everywhere the claim appears.

The practical form of that lesson, for whoever picks this up: after changing a rule here, **grep for the claim, not for the heading**. Every one of these would have been caught by searching for what the old sentence asserted.

**Iteration 18 signals:** Codex `CHANGES REQUESTED` (2, above), both fixed in the push above. CodeRabbit has not reviewed since `d2ddb820`; recorded as no signal. Clean-round count: 0.

isamu:
## Review ledger — after iteration 19 (head `5dcef903`)

| # | iter | bot | finding | disposition | where |
|---|---|---|---|---|---|
| 38 | 19 | codex P1 | `Orphaned -> Implementing` enters a working state without reacquiring a clone | FIXED | Edge now goes to `Claimed` |
| 39 | 19 | me | `SelfCheck` had no `Orphaned` and no `Stopped` edge | FIXED | Both added |
| 40 | 19 | me | `Planned` had no `Stopped` edge — a stopped campaign left planned tasks pending | FIXED | Added |
| 41 | 19 | me | `Learn` had no `Orphaned` edge, though the write-back can crash | FIXED | Added |

**Four rounds of the same root cause is a signal about method, not about the individual edges,** so this round stopped fixing what was pointed at and audited every transition against the lease/claim table. Three more gaps fell out; none had been flagged.

The document now states the **permitted forms** rather than the rule, which is what makes the next edit checkable:

- Which states hold a clone (`Claimed`, `Implementing`, `SelfCheck`, `Review`, `Verify`, `MergeQueue`) and which do not.
- **Every edge from a non-holding state into a holding state passes through `Claimed`** — that is where a free clone is taken, and where a task waits when none is.
- Which states hold the path claim, and where it is released.
- **Every state with side effects needs both an `Orphaned` and a `Stopped` edge.** Add a transition, check for those two.

Audit output, mechanical rather than read: all five edges into `Implementing` come from clone-holding states; all three edges into `Claimed` come from non-holding ones; every side-effect state has an `Orphaned` edge. `Merged` and `Learn` have **no** `Stopped` edge, and that is the intended asymmetry — a merge that already happened cannot be stopped, only recorded — which the prose beside the diagram already says.

**Iteration 19 signals:** Codex `CHANGES REQUESTED` (1), plus 3 found by the audit. All fixed in the push above. CodeRabbit has not reviewed since `d2ddb820`. Clean-round count: 0.

isamu:
## Review ledger — after iteration 20 (head `ee8b9563`)

| # | iter | bot | finding | disposition | where |
|---|---|---|---|---|---|
| 42 | 20 | codex P1 | the decision-2 summary still says a task keeps one clone until completion | FIXED | Rewritten, and the whole claim swept for rather than the one line pointed at |

**Fifth D3-rooted finding, and the second time in this exact paragraph** (row 19 was the first). What makes it worth its own entry is that row 19's fix is what planted it: correcting the sentence then, I wrote a lifetime that was right at the time and became wrong two rounds later when the approval handoff moved the release point.

So this round applied the lesson I posted last round instead of only citing it — grepping for **what the claim says** across both files rather than for the section it lives in. One stale statement left; it is now gone, and the grep comes back empty.

The paragraph also gained the reason it exists, which it had been missing: the lease belongs neither to a session nor to a whole task, so **a clone is not "someone's" — it is stock the runner lends and reclaims.** That is the actual argument for keeping them in a dedicated directory rather than beside a person's working copies, and stating it makes the location decision checkable instead of a preference.

**Iteration 20 signals:** Codex `CHANGES REQUESTED` (1, above), fixed in `3e608d0a`. CodeRabbit has not reviewed since `d2ddb820`. Clean-round count: 0.

Twenty rounds, 42 findings, none rejected. Five of them trace to one sentence in D3 — which, read back, is the most useful thing this review has produced: "1 clone = 1 session = 1 task" was compressing four separate guarantees (same revision, exclusive workspace, exclusive paths, fresh context per step) into one slogan, and each round pulled one of them out. The slogan was the defect.

isamu:
@coderabbitai review

isamu:
## Review ledger — after iteration 22 (head `3e608d0a`)

| # | iter | bot | finding | disposition | where |
|---|---|---|---|---|---|
| 43 | 22 | codex P1 | the lease table sends approval rejections to `MergeQueue`, contradicting the state machine and skipping the `Claimed` reacquisition | FIXED | Table corrected |
| 44 | 22 | me | the **voided-approval** path had the same hole — the prose said "back to `MergeQueue` to re-integrate", from a state that holds no clone | FIXED | Routed through `Claimed` in both documents; `Claimed --> MergeQueue` added for the case where only the base went stale and no implementation work is needed |

One finding, two paths — because this round the invariant from row 38 was actually *used* rather than cited. Sweeping every edge against "a state without a lease can only enter a working state through `Claimed`" found the second one, which no reviewer had raised.

Re-audit, mechanical:

- Into `Implementing` / `SelfCheck` / `Review` / `Verify` / `MergeQueue` — every edge comes from another lease-holding state, or from `Claimed`.
- Into `Claimed` — only `Planned`, `AwaitingApproval`, `Orphaned`, all of which hold no lease. That is the property that makes `Claimed` the single acquisition point rather than a name for the first step.

**CodeRabbit reviewed `d2ddb820..3e608d0a` with no actionable comments** — its second real signal, and the first covering the recent design work. Recorded properly this time: it read the head, and it found nothing.

**Iteration 22 signals:** Codex `CHANGES REQUESTED` (1) + 1 from the audit, both fixed in `b112f975`. CodeRabbit clean on `3e608d0a`, superseded by that push. Clean-round count: 0.

isamu:
## Review ledger — after iteration 23 (head `b112f975`)

| # | iter | bot | finding | disposition | where |
|---|---|---|---|---|---|
| 45 | 23 | codex P1 | with the same-UID model accepted, `D4`'s claim that only humans can start or stop a campaign is not enforceable by route capabilities — protect the authoritative record or narrow the guarantee | FIXED (narrowed) | The heading itself was the over-claim: **"agents cannot start a campaign" → "agents are not given a way to start a campaign"** |

Rows 26, 27, 29 and 30 all corrected the *mechanism* discussion around this. The heading kept asserting the strong version the whole time, and a heading is what gets quoted.

The section now separates the two halves plainly:

- **As a feature, it cannot.** No MCP tool to start or stop, a capability on the routes that an agent can neither read nor mint, no spawn, no merge.
- **As forgery, it can.** It writes the campaign state file, like any other file on the machine.

That is the difference between *what happens by accident* and *what happens when something tries*, and only the first is defended. Stated at the top rather than four sections down, because the top is what a reader carries away.

It also records an asymmetry worth keeping straight: **the round-table invariant this design borrowed from is genuinely stronger.** That agent lives behind a browser and has no shell, so "cannot" is literally true there. Reusing the sentence here was borrowing a guarantee along with the wording.

**Iteration 23 signals:** Codex `CHANGES REQUESTED` (1, above), fixed in `9793e143`. CodeRabbit last clean on `3e608d0a`. Clean-round count: 0.

isamu:
## Review ledger — after iteration 24 (head `9793e143`)

| # | iter | bot | finding | disposition | where |
|---|---|---|---|---|---|
| 46 | 24 | codex P1 | the stop invariant contradicts the deliberately non-cancellable `Merged -> Learn` path, leaving finalization and claim release ambiguous | FIXED | The invariant was the wrong shape and is now **two questions instead of one** |
| 47 | 24 | me | while verifying the new wording against the diagram, "states that can crash start at `Implementing`" was wrong — `Claimed` has an `Orphaned` edge too | FIXED | Corrected to `Claimed` onwards, with the reason `Planned` has none |

The invariant I wrote in row 38 said *"a state with side effects needs both an `Orphaned` and a `Stopped` edge"*, and in the same round I recorded that `Merged` and `Learn` deliberately have no `Stopped`. Both statements were in the document, three paragraphs apart. **Bundling two properties into one rule is what made the exception read as an omission.**

Split apart:

- **Can it be stopped?** `Planned` through `AwaitingApproval` — yes. `Merged` and `Learn` — no, because a merge that already happened has nothing left to cancel. A stop that arrives after the merge does not halt the task; it finishes recording and ends. **"Stopped, yet the write-back still ran" is correct behaviour, not a leak.**
- **Can it crash?** `Claimed` onwards — yes, including `Merged` and `Learn`; a process dies mid-write-back like anywhere else. `Planned` has no such edge because nothing has been acquired: restart and it is still `Planned`.

Claim release was ambiguous for the same reason and is now explicit: **immediately** on `Stopped` / `Escalated`, and **on completing** `Learn` — not on entering `Merged`, so that a crash between the merge and the write-back still records whose work those paths were.

Verification is mechanical: for all ten states, the `Stopped` and `Orphaned` edges present in the diagram now match the two rules exactly, including both intended zeroes.

**Iteration 24 signals:** Codex `CHANGES REQUESTED` (1) + 1 from verifying the fix, both in `b4ad2f50`. CodeRabbit last clean on `3e608d0a`. Clean-round count: 0.

isamu:
## Review ledger — after iteration 25 (head `b4ad2f50`)

| # | iter | bot | finding | disposition | where |
|---|---|---|---|---|---|
| 48 | 25 | codex P1 | the grid lifecycle table releases the path claim at `Merged`, contradicting the recovery contract that holds it until `Learn` completes — overlapping work becomes possible after a post-merge crash | FIXED | Table, prose and sequence all corrected in the grid document |

Row 46 tightened this in the generic document and left the grid table stating the old rule. **First cross-file instance of the habit** — previous ones were two places in one file, this one was two files, which is worse: each document reads as self-consistent, so nothing looks wrong until they are opened side by side.

The concrete failure it leaves open is worth stating, since it is the reason the rule is written this way at all: a crash **between the merge and the write-back**, with the claim already released, means the paths have no owner while the campaign still has unfinished business on them. Another task takes them, and the recovery pass has nothing that says otherwise.

So: the claim is released **when `Learn` completes**, not when `Merged` is entered — immediately on `Stopped` / `Escalated`, where nothing further will happen. The sequence diagram had the two steps in the wrong order as well ("release the claim, then write back") and now reads "write back, then release".

Swept both files for every statement about releasing the claim; the four that remain agree.

**Iteration 25 signals:** Codex `CHANGES REQUESTED` (1, above), fixed in `9255d2b8`. CodeRabbit last clean on `3e608d0a`. Clean-round count: 0.

isamu:
## Review ledger — after iteration 26 (head `9255d2b8`)

| # | iter | bot | finding | disposition | where |
|---|---|---|---|---|---|
| 49 | 26 | codex P1 | the path-claim lifecycle omits `Orphaned`, so overlapping work is possible before an orphan is reconciled | FIXED | `Orphaned` holds its claim until it resolves — and the whole invariant is now a **table instead of prose** |

Six rounds have now found the same class of defect one state at a time. Adding the missing case again would have bought one more round, so this one replaced the form of the rule:

| 状態 | clone | claim | stoppable | can crash |
|---|---|---|---|---|
| `Planned` | — | — | ○ | — |
| `Claimed` … `MergeQueue` | ○ | ○ | ○ | ○ |
| `AwaitingApproval` | — | ○ | ○ | ○ |
| `Merged` | — | ○ | **×** | ○ |
| `Orphaned` | **unknown** | ○ | resolve first | — |
| `Learn` | — | ○ | **×** | ○ |

Every cell answered, for every state. **A new state that cannot be given a row is a state whose design is not finished** — which is the property prose could never carry.

Three answers in there are load-bearing and were previously nowhere:

- **`Orphaned`'s clone is "unknown", and that is correct.** Whether it held one before the crash is not knowable until reconciliation, and holding one is no guarantee it still works — hence recovery through `Claimed`.
- **`Orphaned` keeps its claim**, the finding above.
- **`Orphaned` has no stop edge.** Resolve first, then let the stop apply at whatever it resolves to. *Stopping something without knowing whether it already merged is the most dangerous ordering available*, and it was reachable before.

Verified mechanically against the diagram for all eleven states, and mirrored into the grid document in the same commit — the cross-file lesson from row 48, applied rather than cited.

**Iteration 26 signals:** Codex `CHANGES REQUESTED` (1, above), fixed in `5c0cb515`. CodeRabbit last clean on `3e608d0a`. Clean-round count: 0.

isamu:
## Review ledger — after iteration 27 (head `5c0cb515`)

| # | iter | bot | finding | disposition | where |
|---|---|---|---|---|---|
| 50 | 27 | codex P1 | the gate must **require and verify** the branch-protection setting that rejects a PR whose base has advanced; `expected head SHA` alone protects only the candidate head, so a candidate validated against an old base can still merge | FIXED | The setting is now a condition that is **read before every merge**, not an assumption the design leans on |

Row 34 moved base-freshness enforcement to the forge, which was right, and then left it as a stated dependency. A dependency nobody checks is a dependency that quietly is not there: **branch protection is configuration a person can change afterwards**, so confirming it once says nothing about the next merge.

- The runner reads the protection state each time and confirms *require branches to be up to date before merging* (GitHub's `strict`) is on.
- If it is off, base freshness has no enforcement point at all, so **the merge does not happen** — not even behind the human gate, because the human presses a button the runner executes. It escalates instead.
- **Not substituted with our own pre-check.** That is precisely the race row 34 removed, and re-adding it here would undo that fix while looking like caution.

**Iteration 27 signals:** Codex `CHANGES REQUESTED` (1, above), fixed in `b3ac8239`. CodeRabbit last clean on `3e608d0a`. Clean-round count: 0.

isamu:
## Review ledger — after iteration 28 (head `b3ac8239`)

| # | iter | bot | finding | disposition | where |
|---|---|---|---|---|---|
| 51 | 28 | codex P1 | the concrete sequence leases and mutates a clone **before** Plan declares paths and before the path claim exists, contradicting the `Planned`/`Claimed` lifecycle | FIXED | The state machine was the wrong one, and is now split |

Worth being clear about which side was wrong: **the sequence was right.** Planning requires reading the repository, so a workspace has to exist before anything can be declared — "declare, then acquire" is not an order this work can be done in. The machine had compressed two acquisitions into one transition and called it `Claimed`.

Now:

- `Planned --> Leased` — take a free clone. **The single acquisition point**; a task with nothing free waits here.
- `Leased --> Claimed` — declare the paths and contracts planning produced, and take the path claim.
- Rejections, voided approvals and orphan recovery all re-enter at `Leased`; they already hold their path claim, so `Leased --> Claimed` passes straight through.
- `Leased --> MergeQueue` for the case where only the base went stale and no implementation is needed.

The invariant becomes *every edge into a clone-holding state passes through `Leased`*, and the state table gained a `Leased` row (clone ○, claim —, stoppable ○, can crash ○). Seven `Claimed` references in the grid document were swept and updated in the same commit.

**This is the second time the machine and the concrete flow disagreed and the machine lost** (row 51 here, row 32 for the approval state). Both times the flow was written from what the work actually needs and the machine from what the rule sounded like.

**Iteration 28 signals:** Codex `CHANGES REQUESTED` (1, above), fixed in the push above. CodeRabbit last clean on `3e608d0a`. Clean-round count: 0.

isamu:
## Review ledger — after iteration 29 (head `cfedd38b`)

| # | iter | bot | finding | disposition | where |
|---|---|---|---|---|---|
| 52 | 29 | codex P1 | `Leased` now covers both the initial no-claim planning lease **and** claim-retaining recovery, contradicting its own table row and risking the exclusive claim being dropped on recovery | FIXED | The cell reads `— on first entry, ○ on re-entry`, and the release points are restated as path-independent |

The table introduced one round ago is what created this: filling in a single value for a state that has two entry paths. **A matrix is only better than prose if the cells are true for every way of reaching the row** — otherwise it is prose with a false air of completeness, which is worse.

The failure it would have caused is specific: read "`Leased` holds no claim" literally and an implementation drops the exclusive claim **every time a task recovers or comes back from a rejected approval** — precisely the moments when another task is most likely to be sitting on those paths already. So the rule is now stated as a property rather than a cell: **entering `Leased` never releases a claim.** Release happens on completing `Learn`, or on entering `Stopped` / `Escalated`. Nowhere else, by any route.

**Iteration 29 signals:** Codex `CHANGES REQUESTED` (1, above), fixed in the push above. CodeRabbit last clean on `3e608d0a`. Clean-round count: 0.

isamu:
## Review ledger — after iteration 30 (head `54a5b25d`)

| # | iter | bot | finding | disposition | where |
|---|---|---|---|---|---|
| 53 | 30 | codex P1 | the approval-invalidation and rejection prose still routes recovery through `Claimed`, bypassing the `Leased` acquisition state | FIXED | Every `Claimed` reference in both documents adjudicated, not just the line cited |

Splitting `Leased` out (row 51) updated the diagram and the table and left the prose behind — the same habit as rows 19, 22, 32, 36, 42 and 48. Six of those, and each time the fix was applied where the reviewer pointed.

So this round adjudicated **every** occurrence instead. What changed: the two recovery paths, `Orphaned`'s resolution targets, the phase name in the feature list, and the two existing-asset mappings (`issue-work.ts`, `worktree-session-limit.ts` — both are prior art for the *lease*, not for the path claim). What stayed: the diagram edges, the table row, and the passage that exists to explain the difference between the two states.

The distinction that made the sweep decidable, and that the prose had been blurring: **`Leased` is where a workspace is acquired; `Claimed` is where exclusivity over paths is acquired.** Anything about getting a clone back is `Leased`. Anything about who owns which files is `Claimed`. Written out that way, each of the eleven references answered itself.

**Iteration 30 signals:** Codex `CHANGES REQUESTED` (1, above), fixed in `d3a9b66c`. CodeRabbit last clean on `3e608d0a`. Clean-round count: 0.

isamu:
## Review ledger — after iteration 31 (head `d3a9b66c`)

| # | iter | bot | finding | disposition | where |
|---|---|---|---|---|---|
| 54 | 31 | codex P1 | the durable path-claim mechanism lands **after** the runner starts working unattended, so `Claimed` exclusion and crash recovery are not guaranteed in PR3/PR4 | FIXED | The claim registry moves to PR1; PR3/PR4 are pinned to parallelism 1 |

Second ordering finding of the same kind as row 28 (execution authority landing after the runner). The cause is identical: the registry had been filed with the **MCP reporting tools** because both are about "what a task declares" — which is a topical grouping, not a dependency one.

- **Claim registry moves into PR1**, beside the state store. It belongs there on its own merits too: exclusion and startup reconciliation both sit on the same durable records, so building them apart means building the same file twice.
- **PR5 keeps only the reporting tools** — and those were never the exclusion mechanism anyway, since reports are untrusted input (row 30 of the first ledger).
- **PR3 and PR4 are pinned to parallelism 1**, as configuration rather than as a convention nobody enforces.
- **PR8 (eight clones) is marked as depending on PR1's registry**, so the pool cannot ship ahead of the thing that makes it safe.

Two orderings have now been wrong in the same direction — a safety mechanism scheduled after the autonomy it protects — which suggests the PR list was ordered by *what is interesting to build* rather than by *what must exist before the next thing runs unattended*.

**Iteration 31 signals:** Codex `CHANGES REQUESTED` (1, above), fixed in the push above. CodeRabbit last clean on `3e608d0a`. Clean-round count: 0.

isamu:
## Review ledger — after iteration 32 (head `a6be6c88`)

| # | iter | bot | finding | disposition | where |
|---|---|---|---|---|---|
| 55 | 32 | codex P1 | a forge-rejected merge returns straight to `MergeQueue` although the clone was released at `AwaitingApproval`, so revalidation has no workspace | FIXED | Routed through `Leased`, in the diagram label, the generic prose, the grid prose and the sequence |

**Fourth route back into work, and the fourth one to be written without `Leased`.** The others were rejection (row 37), voided approval (row 44) and orphan recovery (row 38). Each was added in a different round, each named the state it was returning *to* rather than the state that hands out a workspace.

The check that would have caught all four at once, and which this round finally ran: **enumerate every path that goes back to working, and confirm each names `Leased`.** There are exactly four — rejection, voided approval, forge refusal, orphan recovery — and all four now do, in both documents.

That is the same move as row 38's edge audit and row 53's reference sweep: stop fixing the instance, enumerate the category. It has worked every time it has been applied and every recurrence has come from not applying it.

**Iteration 32 signals:** Codex `CHANGES REQUESTED` (1, above), fixed in the push above. CodeRabbit last clean on `3e608d0a`. Clean-round count: 0.

isamu:
@coderabbitai review

isamu:
@coderabbitai review

isamu:
## Converged — two clean rounds on `463d999f`

|---|---|---|---|
| 34 | LGTM | reviewed `d2ddb820..463d999f`, no actionable comments | none |
| 35 | LGTM | "No new commits to review since the last review" | none |

Both reviewers read the final head, neither had anything left, and nothing was pushed between the two rounds. All 8 checks green.

## What 35 rounds produced

**55 findings, none rejected.** Two were defects in the original draft; the rest were guarantees the design stated without a mechanism behind them, and a handful were defects in my own fixes from the round before.

The three that changed the design rather than its description:

1. **`D3` was a slogan, not a rule.** "1 clone = 1 session = 1 task" was compressing four separate guarantees — same revision, exclusive workspace, exclusive paths, fresh context per step — and five rounds pulled them apart one at a time. They are now a revision pin, a clone lease (`Leased`→`AwaitingApproval`), a path claim (`Claimed`→`Learn`), and a session per step.
2. **`D4` promised what its mechanism could not deliver.** "Agents cannot start a campaign" rested on withholding MCP tools from a process that has a shell. It now reads *not given a way to*, and the trust model is stated: as a feature it cannot; as forgery it can.
3. **The stop switch, the approval and the merge each had a race**, and each was fixed by moving the check to the only place it can be atomic — one writer for stop and side-effect intents, an approval bound to (candidate, base), and the merge condition handed to the forge with `expected head SHA` plus verified branch protection.

**The owner decision** (same-UID trust model accepted) removed auto-merge from the `refactor` profile: the merge conditions are read from state the merged party can write, so unattended merging is past the exposure that was agreed to. `refactor` still runs supply → implementation → review → verification → PR unattended; a person presses merge once.

## What is not covered

- **Nothing here has been built.** This is a design note; every mechanism in it is unimplemented.
- **The PR list is honest about ordering now** (safety before autonomy: reconciliation and the claim registry in PR1, the capability boundary and execution authority in PR2, the runner in PR3) but the estimates behind it are guesses.
- **CodeRabbit was rate-limited for roughly ten of these rounds** and did not see those intermediate heads. It read the final one.
- The threat model was reasoned about, not tested. "An injected instruction can rewrite the policy" is an argument, not a demonstration.

PR #1817, opened 2026-08-20
refactor: ターン相関の純関数を common/ へ（#1815 PR0）

isamu:
## Summary

First step of Campaign Mode (#1815). **No behaviour change** — a pure relocation, plus one test that closes a gap the verification found.

The runner Campaign Mode needs is server-side, and what it waits for is the rule the browser already uses: *is this finished turn the answer to what we sent?* That rule lived in `src/composables/exchangeRules.ts`, which `tsconfig.server.json` does not include, so the server could not reach it. CLAUDE.md's `common/` rule applies directly — a decision **both** sides make belongs in `common/`, never mirrored.

**Moved to `common/turnCorrelation.ts`** (byte-identical bodies): `TurnSnapshot`, `CORRELATION_TAIL`, `collapse`, `answersOurSend`, `WaitVerdict`, `waitVerdict`.

**Left in `src/composables/exchangeRules.ts`**: `ExchangeOutcome` and `outcomeMessage` — the wording a cell shows, which the server does not decide. This follows the "share the common core, keep each side's extras local" half of the same rule.

No re-export (CLAUDE.md forbids it); the six importers were rewritten. `test/common/roomMessage.spec.ts` was reaching into `src/` for `answersOurSend` — `common/roomMessage.ts` documents its own layout by citing that rule — and now stays inside `common/`.

## How "behaves the same" was verified

The old file was taken verbatim from `origin/main` and run beside the new module over generated inputs — 21 message bodies (empty, whitespace-only, 159/160/161 code units, 2900-char truncated pairs, CRLF, tabs, multibyte, surrogate pairs, shared-tail pairs) x 11 prompt relationships (null, empty, identical, reflowed, quoted, extended, truncated, unrelated, previous round's framing, upper-cased) x 8 elapsed x 5 timeout values, throws compared as outcomes.

**17,145 comparisons, 0 differences.**

That zero only means something if the harness can report otherwise, so each line was mutated:

| Mutation of the new module | Differences |
| --- | --- |
| `CORRELATION_TAIL` 160 -> 159 | 41 |
| `collapse` gains `.toLowerCase()` | 1353 |
| `slice(-TAIL)` -> `slice(TAIL)` | 1968 |
| `includes` -> `startsWith` | 4592 |
| `elapsedMs >= timeoutMs` -> `>` | 1056 |
| correlate on `reply` instead of `prompt` | 6120 |
| `outcomeMessage` wording changed | 1 |
| drop the `!sent.trim()` guard | **0** |
| drop the `needle.length > 0` guard | **0** |
| drop **both** empty guards together | 533 |

The two zeros are not a blind generator: **the two guards are mutually redundant**, and either one alone still rejects an empty send. Dropping both at once differs in 533 cases, which is what proves the generator reaches that shape. Left as-is — removing a redundant guard is a behaviour question for another PR, not for this one.

## The one gap this found, and the test that closes it

The moved spec **stays green when `CORRELATION_TAIL` changes from 160 to 159** — a gap that predates the move (the spec is verbatim), and one only the harness caught. Since the harness is throwaway, the property is now a permanent test: a prompt agreeing on the last **159** characters is not our answer, one agreeing on the last **160** is. Break-verified in both directions — red at 159, red at 161, green at 160.

## Items to Confirm / Review

- **Is `common/turnCorrelation.ts` the right name and the right split?** The alternative was moving the whole file, `outcomeMessage` included. I kept the wording browser-side because the server has no cell to write it to; if you would rather have one file, say so.
- **Pinning 160 exactly** is a deliberate choice — it makes tuning the window a test edit. The comment says why. Push back if that is too tight.
- **`RoundTableDeps` / `CrossTalkDeps` were deliberately NOT touched**, although permission was given. That merge is provisional, and untangling it is not a behaviour-preserving change; mixing it in here would spoil the claim above. Separate PR.
- **The app was not started.** `vite build` proves the specifier resolves from `src/`, `tsx` proves it resolves from `server/` (probed directly, since "the server can import this" is the PR's whole point), and the full suite is green — but no browser exercised a real cross-talk or round table. The runtime path is unchanged, so I judged the harness plus the build sufficient; say if you want a run before merge.

## Verification run

`yarn format` / `yarn lint` / `yarn typecheck` / `yarn build` clean, `yarn test` 748 files, 10,813 tests passed.

One caveat worth recording: `yarn typecheck` first reported 73 errors, all in `sharedApp` files and none in anything this PR touches. They were a stale `node_modules` — `origin/main` had moved `@receptron/sharedapp` to 0.20.0. `yarn install --frozen-lockfile` cleared all 73.

## User Prompt

- Walked through a real large-scale refactoring campaign (introduce ESLint on an untyped human-written TS/JS codebase, tighten progressively so the code becomes readable to an AI; roughly a week, ~1000 PRs, 100+ issues surfaced, 5-8 parallel sessions) and the four things that blocked it: sessions modifying the same places, ~20-round review loops, problems found after LGTM, and loops that stop when told not to.
- Asked for it to be written up under a `future/` dir with the feature list, then generalised into one pipeline where refactoring and new development are two instances, then: "now seriously consider this as one of MulmoTerminal's grid features. Ideally the UI does not change much and it runs fully automatically."
- Decisions given during design: first profile is `refactor`; clones live in a dedicated directory; `RoundTableDeps` / `CrossTalkDeps` may be touched.
- "ok" to filing the issue and starting PR0.

Design notes: #1816. Tracking issue: #1815.

work in mulmoterminal2

## Summary by CodeRabbit

* **Improvements**
  * Improved message matching to reliably recognize responses, including normalized text and message-length boundaries.
  * Enhanced response waiting behavior to distinguish answered, still-waiting, timed-out, and late-answer states.

* **Refactor**
  * Centralized turn-correlation logic for consistent behavior across conversations.

* **Tests**
  * Added coverage for matching, stale messages, missing prompts, truncation, timeouts, and late responses.


isamu:
@coderabbitai review

PR #1833, opened 2026-08-22
Stream a watching page's records into the preview pane

snakajima:
Replaces #1832, which polled. This holds a real listener instead.

## The gap

Open a `live` page in the collection pane and it never moves. Your own writes refresh it; nobody else's do. The published page at `/m/{slug}` updates the moment somebody posts, so the preview looks broken in a way the real thing is not.

`views[].live` was only ever implemented by mulmoserver — one Firestore listener per watched collection, re-delivering state on every change. The pane had `load()` on a directory change, `refresh()` after the author's own submit or intent, and nothing else.

## The listener lives on the server

The pane has no Firestore and no credentials; this host holds the session (`currentFirestore()`), and the backend already reads records through the same Firebase JS SDK — which means `onSnapshot` works here. So `previewWatch.ts` holds the subscriptions and streams what they see.

**Server-sent events** to carry it: one way, text, no handshake, and the subscription is closed when the request is (`req.on("close")`). A heartbeat comment every 25s keeps an idle stream from being closed underneath us — it carries no records and asks the database nothing.

**Nothing is polled.** A listener fires when the records change and at no other time, which is exactly what the published page gets.

## Shape

- **`live` rides on each previewed page**, read off the projection rather than the declaration — `withLive` has already dropped anything the tier may not read.
- **The watch plan is built from the same request list the one-shot read uses.** A watch with no request behind it would be a listener on a collection the page never reads.
- **One listener per collection, fanned out to the pages that watch it.** Two pages can name the same collection and read it differently — `all` for the desk, `own` for the participant — so a change becomes rows per page, through the same `ownsRow` the one-shot read uses. Subscribing twice would be two listeners billing for one collection and two chances to disagree.
- **One stream for the app**, not one per page: the server says which page each change is for, so switching pages in the picker needs no reconnection.
- **An app that declares no `live` opens no stream at all.**
- The change carries the collection's **current rows**, not a diff: the pane holds a map keyed by page and collection, so replacing one entry cannot be applied out of order.
- The frame is keyed on the page, not the payload, so an update never remounts the document — the rows go down the channel as a state message, which is what production does.

## Tests

- `test/src/composables/sharedAppLiveWatch.spec.ts` (8): one stream opened for a watching page and what arrives is filed; nothing opened for a page that watches nothing; a line the pane cannot read leaves the records alone; the stream closes when the author walks to a non-watching page and when the pane goes away; `withChange` replaces one collection on one page, returns a new map (the payload is a `shallowRef`), and ignores a page the pane is not holding.
- `test/server/backends/sharedAppPreview.spec.ts` (2): `live` rides per page — the watching page has it, the one beside it does not, the anonymous page carries its own; and the watch plan names the page, the collection and that page's own scope, with nothing for a page that declared no `live`.

`vue-tsc -b`, eslint, and 10950 tests pass.

## Summary by CodeRabbit

* **New Features**
  * Shared app previews now stay synchronized with live record changes.
  * Pages can watch selected collections and receive scoped updates automatically.
  * Added a live update stream with connection cleanup and periodic heartbeats.
  * Updates remain consistent while preview data is refreshing.
  * Invalid or unavailable updates are safely ignored without opening an unusable stream.

* **Tests**
  * Added coverage for live collection configuration, record updates, malformed messages, reconnection, refresh consistency, and cleanup.

### DIFF
--- a/common/sharedAppPreview.ts
+++ b/common/sharedAppPreview.ts
@@ -35,6 +35,14 @@ export interface PreviewPage {
    *  the pane resolve it would be the second implementation this whole change
    *  exists to remove. */
   viewer?: Viewer;
+  /** The collections this page WATCHES, off the same projection the published site reads.
+   *
+   *  A page that declares `live` is written for `onState` to arrive more than once — production
+   *  subscribes to those collections and re-delivers on every change. This pane read the records
+   *  once, so a page previewed here sat still while the live one moved: somebody else's message
+   *  never appeared, which reads as a broken page rather than as a preview that does not watch.
+   *  The pane re-reads while the page on screen names any, and this is how it knows to. */
+  live?: string[];
 }
 
 export type PreviewDataset = Record<string, unknown>[];
--- a/common/sharedAppPreview.ts
+++ b/common/sharedAppPreview.ts
@@ -47,6 +47,18 @@ export interface PreviewPage {
 
 export type PreviewDataset = Record<string, unknown>[];
 
+/** One collection's rows, for one page, as they stand now — what the record stream sends when a
+ *  watched collection changes.
+ *
+ *  THE ROWS RATHER THAN A DIFF. The pane holds a map keyed by page and collection; replacing one
+ *  entry cannot be applied out of order, and a diff against a map the sender cannot see could be. */
+export interface PreviewRecordChange {
+  /** `previewPageKey(audience, id)` — the same key the page's datasets are filed under. */
+  key: string;
+  cid: string;
+  rows: PreviewDataset;
+}
+
 /** The records ONE page is handed, per collection.
  *
  *  Per page rather than one map for the app, and that is the point rather than tidiness: a member
--- a/server/backends/sharedApp/preview.ts
+++ b/server/backends/sharedApp/preview.ts
@@ -37,7 +37,7 @@ import { declaredView, readAppViewFile } from "./publicView.js";
 import { isRecord } from "../../../common/isRecord.js";
 // Both from `/view`, which is where the PARENT's vocabulary lives — and where the read-back has to
 // be: the root entry reaches the compiler, and the compiler imports core's server half at runtime.
-import { ownRowsFor, projectedWritesOf, PUBLIC_WRITE_TIER, viewerFor, writableFields } from "@receptron/sharedapp/view";
+import { ownRowsFor, projectedWritesOf, PUBLIC_WRITE_TIER, viewerFor, writableFields, type Viewer } from "@receptron/sharedapp/view";
 import {
   previewPageKey,
   type PreviewDataset,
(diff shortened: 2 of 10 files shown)


snakajima:
Fixed in `469df65d`. Both findings are real and both are this branch's.

## The leak (Codex P2, CodeRabbit major, and the verdict line)

Opening the watch is asynchronous — the preview has to be computed before there is anything to subscribe to — and a pane that changes page or directory meanwhile closes the request while that runs. The close handler was registered *after* the await, so the event had already fired by the time anything was listening: subscriptions and heartbeat created a moment later, nothing left to stop them.

`holdOpen` now registers the cleanup **first** and is handed the watch when there is one:

- a watch that arrives after the request closed is stopped as it arrives;
- no heartbeat is started for it;
- nothing to watch → the response ends and no timer exists.

Everything it touches is injected, so the ordering is tested without express or a session — and the abort-during-setup test **fails** with the handler moved back after the await (checked).

## The race (Codex P2 on `SharedAppPreview.vue:654`)

Also real, and worse than it looks: a re-read answers with the rows as they were when it *started*, so assigning that answer undid any change that arrived while it was in flight — and no second snapshot is coming, because nothing has changed since. The preview would sit on the older rows permanently. Another member posting while the author submits is exactly that case.

Changes are now kept as well as applied until the read lands, and put back on top of it (`withChanges`, oldest first so the newest wins). Covered by a test that pins the ordering.

## Not changed

CodeRabbit's remaining notes are style-level and I left them; say the word if you want any of them.

`vue-tsc -b`, eslint, and 10954 tests pass.

snakajima:
Fixed in `ad19adac`, along with the CI lint failure.

## Unbounded buffer (Codex)

Right, and it is the fix from the last round taken too far. The buffer was cleared when a read *started*, so between reads it only grew — a previewed chat room left open for an afternoon collects a change per message, against a read that may never come.

`pendingChanges` now:

- **holds nothing unless a read is out**;
- while one is, keeps only the **latest rows per page and collection** — every change carries the whole collection, so an older one for the same pair says nothing the newer one does not;
- is **closed on every way out of a read**, not only the one that used the changes. Superseded, refused and threw all left it open before, and it went on collecting until the next read happened to close it. Closing twice is closing once.

Bounded by what is watched rather than by traffic, for as long as one read takes.

## CI

`lint-and-build` was failing on my own spec: the placeholder resolver lied about its arity (`sonarjs/no-extra-arguments`). It is held in a list now, which is neither a nullable narrowed to `null` nor a function pretending to take no arguments. Verified with `eslint --no-cache` over the whole repo, which is what CI runs — my earlier local run was reading a stale cache entry.

Three tests on the buffer: nothing held while no read is out, only the latest rows per pair while one is, and a second close returning nothing.

`vue-tsc -b`, eslint (no cache), and 10957 tests pass.

snakajima:
Fixed in `7eb060dd`.

Right again, and it is the same bug arriving by another door. `refresh` is started fire-and-forget from three places — a write, an intent, and the pane — and it supersedes rather than serialises. With one buffer between them, whichever read finished **first** closed the buffer the **landing** read was about to use; the change it had caught was then assigned away by an answer read before that change arrived, and no second snapshot is coming.

`begin()` now hands each read its own buffer, and the stream records into every open one. A close removes only that read's own and is still idempotent — every way out of a read closes it, so the second call must not become a way to replay one read's changes onto another read's answer.

Still bounded: reads actually in flight × what they watch.

Two more tests — two concurrent reads each keep their own catch, and a closed read stops collecting — on top of the three from last round.

`eslint --no-cache` over the repo (what CI runs), `vue-tsc -b`, and 10959 tests pass.

snakajima:
Fixed in `67a18436`. Correct, and my own comment had it backwards — I wrote that an `EventSource` reconnecting on its own was why ending the stream was fine. It is exactly why it is not: the client comes back every few seconds and the server recomputes the whole preview each time, which is the poll this PR exists to avoid, arriving through the error path.

**204** now, which is the one status an `EventSource` treats as "do not come back". A stream that *drops* is still reopened on its own — which is what should happen to a server restarted under an open pane; it is only the terminal answer that must not be retried.

Two consequences, both handled:

- **The response is begun only once there is a watch**, so a status can still be set. Snapshots arriving before that are queued and flushed at `begin()` — `onSnapshot` delivers what it already has as soon as it is subscribed, which would otherwise race the headers and send the first change as an ordinary 200 body.
- **A request that closed during setup is told nothing at all**, rather than having a status written onto a response nobody is reading.

Coverage: the terminal answer is given without a stream being begun and without a heartbeat; a request that closed first gets no answer; and the existing three (hold until close, stop a watch that arrives after the close, no heartbeat for it) updated to the new shape.

`eslint --no-cache` over the repo, `vue-tsc -b`, and the full suite pass.

PR #1837, opened 2026-08-23
feat(campaign): タスクの状態機械（#1815 PR1a）

isamu:
## Summary

The first code for Campaign Mode (#1815): one campaign task's state machine, pure functions only. **Nothing imports it yet** — the runner is PR3.

- `common/campaignPhase.ts` — the phase names, terminal check, and four property predicates (`cloneHolding`, `claimHolding`, `releasesClaim`, `isStoppable`, `canCrash`).
- `server/campaign/campaign-state.ts` — the transition table, `advance()`, and the edge enumeration a spec asserts over.
- `test/server/campaign/campaign-state.spec.ts` — 81 tests, mostly invariants over the whole edge set.

### Why this is not all of PR1

The design filed PR1 as *state machine + store + startup reconciliation + claim registry*. The state machine reverts on its own and the others do not, so it ships on its own. The stronger reason: **its content is almost entirely invariants.** Much of the 55-finding design review was about the properties of this machine, and pinning them as tests is what this PR delivers. Shipped alongside a JSONL writer, that review would land inside an I/O review.

PR1b is the store and reconciliation; PR1c is the claim registry (independent of this machine).

### Where the code lives, and why it is two files

The design sketched one file under `server/campaign/`. Split, because the `common/` rule applies to half of it:

| | | |
|---|---|---|
| `common/campaignPhase.ts` | phase names + predicates | The roster pill (PR9) decides its display from the phase. Exactly the `common/prPhase.ts` ↔ `src/components/rosterPhase.ts` relationship, whose own comment records the drift that arrangement exists to prevent. |
| `server/campaign/campaign-state.ts` | transitions | **Only the server advances a campaign (D4).** Shipping the transition table into the browser bundle puts the means to drive one where nothing is permitted to. The module layout is that boundary. |

### The design idea worth reviewing

**The table and the predicates are two independent statements about the same machine, and the specs compare them.** The table says which edges exist; the predicates say what is true of a task sitting in a phase. Neither is derived from the other.

That shape is a direct result of the design review, which found the same defect fourteen times: a rule written in one place and not carried into every place it governs. Prose could not hold those rules. A predicate checked against every edge can.

For the same reason **every edge is written out rather than generated from a rule.** Two look like omissions and are not — `merged` and `learn` accept no `stop` — and a generator makes a deliberate exception indistinguishable from a forgotten one, which is precisely how four review rounds read them.

## Verification

**Mutation sweep, 13 mutations, every one red; restore verified with `diff -q` after each; baseline green before and after.**

| mutation | tests failed |
|---|---|
| add `stop` to `merged` | 3 |
| drop `crash` from `learn` | 3 |
| `send-back` skips `leased` | 2 |
| orphan recovery skips `leased` | 3 |
| add `stop` to `orphaned` | 3 |
| add `crash` to `planned` | 3 |
| `revalidate` lands elsewhere | 1 |
| a terminal phase accepts an event | 3 |
| `leased` claim reads as `free` | 2 |
| `orphaned` clone reads as `free` | 1 |
| `learn` becomes stoppable | 1 |
| `planned` becomes crashable | 2 |
| `merged` releases the claim | 2 |

**Checked against the design diagram mechanically**, not by reading: a throwaway script parsed the mermaid in `future/autonomous-delivery-pipeline.md` and compared its edges to `allTransitions()`. 45 doc pairs, 42 code pairs, **nothing in the table that is not in the diagram**. The three extras are `Rejected/Stopped/Escalated --> [*]`, which this model expresses as those phases being terminal rather than as edges. (`Learn --> [*]` does become an edge, to a `done` phase — see below.)

`yarn format` / `lint` / `typecheck` / `build` clean; `yarn test` 758 files, 11,044 tests green.

## Items to Confirm / Review

- **Two additions beyond the design note.** (a) A `done` phase, naming the diagram's `[*]` after `Learn`, so "completing `Learn` releases the claim" becomes the same kind of rule as every other release point. (b) `Holding = "held" | "free" | "unknown"` instead of a boolean, because `orphaned`'s clone and `leased`'s claim genuinely have no single answer — review finding 52 was exactly a cell given one value for a state with two entry paths, so the type refuses to answer.
- **`advance()` returns `null` for an illegal pair** rather than throwing. A runner that observes an impossible event has a bug worth reporting rather than a crash worth catching, but I can invert this if you would rather it be loud.
- **The edge-count assertion (42) is a tripwire, not a property.** It exists so adding an edge shows up in a diff. Say if you would rather not have a test that a legitimate design change must also update.
- **Event names** are mine — the design labels the edges in prose. `send-back` covers three distinct causes (rejection, voided approval, forge refusal) because all three land in the same place for the same reason; if you want them distinguishable in the audit log, that is three events instead of one.
- **Nothing runs this.** No app was started, because there is nothing to start yet: no route, no runner, no caller.

One incidental find worth recording: my first commit carried two NUL bytes into the spec (a heredoc artefact). `test/server/sourceIsText.spec.ts` caught it — a guard I did not know the repo had.

## User Prompt

- Walked through a real large-scale refactoring campaign (~1000 PRs, 100+ issues surfaced, 5-8 parallel sessions over roughly a week) and the four things that blocked it: sessions modifying the same places, ~20-round review loops, problems found after LGTM, and loops that stop when told not to.
- Asked for it to be written up under `future/` with a feature list, then generalised into one pipeline where refactoring and new development are two instances, then: "now seriously consider this as one of MulmoTerminal's grid features. Ideally the UI does not change much and it runs fully automatically."
- Preference stated: **8 clones rather than worktrees**, an orchestrator managing conflicts across planning, change, secure review, review, and unit/e2e tests.
- Decisions given during design: first profile is `refactor`; clones live in a dedicated directory; `RoundTableDeps` / `CrossTalkDeps` may be touched; accept the ordinary-cell trust model.
- "つづけて" — continue to the next PR.

Design: #1816. Tracking issue: #1815. PR0: #1817.

work in mulmoterminal2

## Summary by CodeRabbit

* **New Features**
  * Added campaign lifecycle tracking with defined phases, transitions, and terminal states.
  * Added support for stopping, crashing, approval, merging, resolution, and verification actions.
  * Added visibility into available actions and valid progression paths.
  * Improved claim and ownership handling during cloning, reconciliation, and crash recovery.
* **Documentation**
  * Added design documentation describing lifecycle behavior and recovery rules.
* **Tests**
  * Added comprehensive coverage for transitions, phase rules, terminal states, ownership, and invalid actions.


isamu:
## Review ledger — after iteration 1 (head `45a5a186`)

|---|---|---|---|---|
| 1 | coderabbit | `claimHolding("orphaned")` reports `"held"`, but `planned -> leased -> orphaned` is a real route and that task never declared anything | FIXED | Returns `"unknown"`, with the reason written next to it |
| 2 | me | **a task could reach `merged` and `done` without ever declaring its paths** | FIXED | `revalidate` moves from `leased` to `claimed`, putting `declare` on every route to the merge |

Finding 1 is the shape I described in this PR's own body as review finding 52 — *a cell given one value for a state with two entry paths* — and I made it again one state over while writing about it. Worth stating plainly rather than quietly patching.

So the fix started from the category rather than the line. A BFS from `intake` refusing to enter `claimed` answered "which phases can a task reach without ever declaring its paths?":

> intake, planned, leased, **implementing, self-check, review, verify, merge-queue, awaiting-approval, merged, learn, done**, orphaned, rejected, stopped, escalated

Everything. The cause was the edge I had written a comment to justify: `leased --revalidate--> merge-queue`, for a task whose base merely went stale. It presupposed a task that had already declared, and the machine had no way to say so — so it also admitted one that never had. **A merge with no exclusion over anything it touched** is the exact failure the claim registry exists to prevent, and it was reachable from the first state.

`revalidate` now hangs off `claimed`. For a task coming back from a send-back or an orphan, `declare` is a re-assertion of paths it never gave back rather than a second acquisition — which is what makes putting it on every route free.

Two properties now hold that hole shut:

- `leased` leaves only by `declare` (or crash, or stop).
- Refusing both `claimed` and `orphaned`, a task reaches exactly `intake, planned, leased, rejected, stopped` — nowhere near work, a queue or a merge. `orphaned` is excluded because that is where reconciliation reads the records and decides; it is the one legitimate way past a declaration this machine cannot check.

Break-verified: restoring the hole fails 4 tests, reverting `orphaned` to `"held"` fails 1, dropping `revalidate` altogether fails 2. Baseline green before and after, restore checked with `diff -q` each time.

**Iteration 1 signals:** CodeRabbit 1 finding (fixed) + 1 found by sweeping the category. Codex `LGTM` on `45a5a186` — it did not find either. Clean-round count: 0.

isamu:
@coderabbitai review

isamu:
## Review ledger — after iteration 2 (head `9774b128`)

|---|---|---|---|---|
| 3 | codex + coderabbit | the implementation removed `Leased --> MergeQueue`, but `future/autonomous-delivery-pipeline.md` still draws it — a reader following the diagram reintroduces the undeclared merge path, and a runner built from it emits `revalidate` at `leased` and gets rejected | FIXED | Diagram updated; the invariant list gained the rule and the reason |

Both reviewers found the same thing independently, which is the useful signal here: **fixing the code left the design note instructing the next person to undo the fix.** That is the failure mode `/refactor-safely` names outright — a plan committed alongside a change becomes the specification, and if the change grew past it the document now argues for the defect.

What the diagram says now:

- The edge is `Claimed --> MergeQueue`.
- A new invariant: **the only edge out of `Leased` into work is `Claimed`**, so `declare` sits on every route to a merge. It records what the first version allowed (a task reaching `Merged` having declared nothing, with no exclusion over anything it touched) and why putting `declare` everywhere is free — for a send-back or an orphan it re-asserts paths that were never given back.
- A note at the top of the invariant section: **with an implementation in place the diagram is secondary.** The machine's authority is `campaign-state.ts` plus `campaignPhase.ts`, checked against each other by the spec. Change one, change the other. Fixing only the picture is how the hole comes back.

Re-compared mechanically after the edit: 45 doc pairs, 42 code pairs, **nothing in the table absent from the diagram**; the three extras remain the `[*]` end markers this model expresses as terminal phases. All three mermaid blocks parse.

**Iteration 2 signals:** Codex `CHANGES REQUESTED` (1), CodeRabbit the same finding inline, both fixed in the push above. All 12 checks were green on `9774b128`. Clean-round count: 0.

isamu:
## Review ledger — after iteration 3 (head `8cd7844d`)

|---|---|---|---|---|
| 4 | codex + coderabbit | the lifecycle table still marks `Orphaned`'s claim as `○`, so a reconciler could treat an undeclared crash as an owned claim | FIXED | All three statements of it, in both documents |

Second round running on the same cause as row 3, and the cause is mine rather than the reviewers': last round I fixed **the diagram edge** and left the **table** and the **prose** saying the old thing. Same habit the design review named a dozen times. So this round swept the claim instead of the line — three places asserted it, all three now agree.

The correction has two halves, and only writing both keeps the original point:

- **The phase does not carry the answer.** `planned -> leased -> orphaned` is a real route and that task declared nothing. `clone` and `claim` are both indeterminate at `orphaned` but *for different reasons* — the clone is "we do not know if it had one", the claim is "it depends where it crashed". The records know; reconciliation reads them.
- **Indeterminate is not permission to release.** `orphaned` is not terminal, so nothing is given back there. If the records say a claim exists it lives until the orphan resolves, and a reconciler that tidies it up lets a sibling into paths the orphan may still be halfway through. That was what the original sentence was protecting, and it would have been lost by only marking the cell indeterminate.

Pinned as a test — `claimHolding("orphaned") === "unknown"` **and** `releasesClaim("orphaned") === false` in one case, so the distinction cannot be collapsed by a later edit.

And since this class of drift has now cost three rounds, I checked it mechanically rather than by reading: a throwaway script parsed the per-state table out of the design note and compared all four columns against the code predicates. **13 rows, 52 cells, no disagreement.** (Two cells hold prose — "stop the supply", "resolve first" — which the parser cannot read but which say the same thing as the code.)

**Iteration 3 signals:** Codex `CHANGES REQUESTED` (1), CodeRabbit the same inline, both fixed above. Clean-round count: 0.

isamu:
@coderabbitai review

isamu:
@coderabbitai full review

isamu:
## Review ledger — after iteration 4 (head `c89c6e04`)

|---|---|---|---|---|
| 5 | codex | the lifecycle table has no row for `done`, the terminal phase the implementation added — contradicting the rule printed directly above it | FIXED | Terminal row is now `Done` / `Stopped` / `Escalated` / `Rejected`, with why `done` exists |

**The interesting part is that my own check should have caught this and could not.** Last round I reported "13 rows, 52 cells, no disagreement" from a script that compared the doc table to the code predicates. That script iterated **the document's rows**, so a phase present in the code and absent from the table was invisible to it — the zero was a statement about my generator, not about coverage. Exactly the failure mode the repo's own guidance describes, produced while using the technique it recommends.

Rewritten to compare the two as **sets, both directions**:

> code phases = 17, doc rows = 17 matched, **missing from the table: none**, cell mismatches: **0**
>
> (three unmatched doc rows are the profile list — bugfix / perf / docs — correctly skipped)

The rule above the table gained its other half: *check the reverse too — a state that exists in the implementation and not in this table does not exist for anyone reading it.* And `done` now carries its justification where a reader will find it: it names the diagram's `Learn --> [*]`, which is what makes "completing `Learn` releases the claim" the same shape of rule as releasing on `Stopped` / `Escalated` rather than a special case.

**Iteration 4 signals:** Codex `CHANGES REQUESTED` (1, above), fixed in `1c9897e0`. CodeRabbit ran a full review of the whole changeset and added nothing. All 12 checks green on `c89c6e04`. Clean-round count: 0 — the round before this one was clean for both reviewers, and this finding is why one clean round is not convergence.

isamu:
@coderabbitai full review

isamu:
@coderabbitai full review

isamu:
## Converged — two clean rounds on `1c9897e0`

|---|---|---|---|
| 5 | LGTM | full review of `8835c151..1c9897e0`, nothing new | none |
| 6 | LGTM | full review again, nothing new | none |

All 12 checks green. Both reviewers read the final head.

## What five findings changed

Two were mine to be embarrassed about, and both are the same shape:

1. **A task could reach `merged` without ever declaring its paths.** CodeRabbit pointed at `claimHolding("orphaned")`; sweeping the category instead of the line found the cause one edge away — `leased --revalidate--> merge-queue`, an edge I had written a comment to justify. It presupposed a task that had already declared, and the machine had no way to say so, so it also admitted one that never had. That is a merge with no exclusion over anything it touched, reachable from the second state. `revalidate` now hangs off `claimed`, putting `declare` on every route to a merge.
2. **The design note then told the next reader to put the hole back.** Three rounds went on catching up: the diagram edge, the lifecycle table's `Orphaned` claim cell, and a missing `Done` row. Each time I fixed where the reviewer pointed and left the same claim standing elsewhere — the habit the design review itself named a dozen times.

And the one worth keeping in mind for the next PR: **the mechanical check I wrote to catch that drift had the drift's own blind spot.** "13 rows, 52 cells, no disagreement" came from a script that iterated the *document's* rows, so a phase present in the code and missing from the table could not be seen. Codex found `Done` missing; the rewritten check compares the two as sets, both directions — 17 code phases, all present, zero cell mismatches.

## What is not covered

- **Nothing runs this.** No route, no runner, no caller — by design; the runner is PR3.
- The invariants are pinned against the *design*, not against a working campaign. Whether the machine is the right machine is a question PR3 answers, and the design note now says plainly that the implementation is the authority once it exists.
- `claimHolding` answers `unknown` for two phases and that is honest, but it means **a caller cannot decide ownership from a phase alone** — the claim registry (PR1c) is the authority, and nothing enforces that yet because it does not exist.

PR #1841, opened 2026-08-23
feat(campaign): キャンペーンの永続記録（#1815 PR1b）

isamu:
## Summary

The durable record for Campaign Mode (#1815), on top of the state machine from #1837. Pure format plus a thin disk layer; **nothing imports it yet**.

- `server/campaign/campaign-log.ts` - record shapes, guards, line format, and the fold from records to per-task state.
- `server/campaign/campaign-store.ts` - where a campaign lives, reading it, appending to it.
- 62 tests across two specs.

### The format follows `rooms.ts` / `room-log.ts` rather than inventing one

This repository already has a considered JSONL append-log, and every reason it gives applies here unchanged, so none of it is re-decided:

| rule | the reason, already written down |
|---|---|
| append-only | `MULMOTERMINAL_HOME` is shared by every server on this machine; a read-merge-write loses whichever writer finished first |
| the newline **leads** | a torn write costs one record and never the next |
| an unparseable line is dropped alone | same |
| never sorted by `at` | two writers can stamp one millisecond, and a clock can step backwards; the file's order is what happened |
| the id check is the whole path defence | `roomFile()`'s arrangement - every entry point goes through one function |

### intent, then outcome

Every side effect a campaign causes - spawning a session, injecting a prompt, opening a PR, merging - happens **outside this process** and cannot be made atomic with writing it down. So the intent is appended before the effect and the outcome after, and **an intent with no outcome is exactly the crash window**: the one place a restart has to go and ask the world what really happened.

The idempotency key `campaign/<id>/<task>/<attempt>` is **derived, never stored**. That is the point of it - on restart the runner asks the forge and the clone "is there already something called this?" instead of trusting the half of its own log that may be missing.

### The fold does not believe the log

`foldCampaignLog` replays records through the state machine instead of adopting the phase written on them. A settlement naming an edge the machine has not got is a corrupt log, not a new edge. Five ways a record is refused, all reported rather than thrown - a corrupt line is data about a real event, and the runner has to put it in front of a person rather than die reading it:

| reason | what happened |
|---|---|
| `illegal-transition` | an intent or settlement for an event the phase does not have |
| `intent-while-pending` | two effects in flight for one task; the second is the bug |
| `no-intent` | a write nobody declared |
| `wrong-event` | the outstanding intent names a different event or attempt - which is what stops a stale retry answering a new one |
| `phase-mismatch` | the settlement names somewhere that event does not lead |

### `appendCampaignRecord` differs from `postToRoom` on purpose

`postToRoom` swallows a failed append. This one **returns `false`, and a caller that gets it must not go on to cause the effect**. A lost message is a lost message; a lost intent is a side effect nobody can reconcile afterwards.

### Durability: what the review built, and what it is allowed to claim

Ten of the fourteen review rounds landed on `appendCampaignRecord`'s return value, because it is the only thing here a caller stakes an irreversible action on. The work that came out of that stays:

- The record is validated by **round-tripping the exact line through the reader**, so nothing is written that a restart would discard.
- The write is **flushed**, not left in the page cache.
- The **entry naming the file** is synced, because a flushed file with an unrecorded name does not exist after a power cut.
- The **directory hierarchy** is a separate concern with its own lifetime: `ensureCampaignStore()` handles it, walking from the filesystem root so that two runners preparing at once cannot each depend on the other having finished. Below `MULMOTERMINAL_HOME` a refusal is surfaced; above it, best effort — chasing further ends at fsyncing `/`.
- A relative `MULMOTERMINAL_HOME` no longer starts that walk at the empty string.

**What it promises is narrower than what it does, deliberately.** `true` means the record was validated, written and flushed — durable across *this process* exiting. It does **not** promise surviving a power cut, because **no test in this PR can tell a fsync from no fsync**; mutations removing the syncing entirely stay green. A guarantee no test can hold is one that quietly stops being true, so the syncing stays and the claim does not. The doc says outright: do not build a caller that treats `true` as proof against a power cut.

Getting the stronger guarantee back needs an instrument this PR has not got — crash injection — and that is its own piece of work.

## What is deliberately not here

**Startup reconciliation.** Everything it reads - live sessions, clone leases, the forge's branch/PR/merge state - **does not exist yet**. Writing it against assumed inputs would not produce reconciliation; it would accidentally decide the design of three other components. Its pure decision belongs in the PR where its inputs are real.

The claim registry depends on neither the machine nor this log, so it stays separate.

## Verification

**Mutation sweep, 11 mutations, all red, plus a deliberate no-op that stayed green** so the sweep is not merely reporting red for everything. Restore checked with `diff -q` after each; baseline green before and after.

| mutation | tests failed |
|---|---|
| trailing newline instead of leading | 1 |
| trust any parsed line | 8 |
| a settlement need not name a real phase | 2 |
| an intent need not be legal | 2 |
| allow two effects in flight | 1 |
| settle with no intent | 3 |
| ignore the attempt number | 1 |
| believe the phase written on the record | 1 |
| abandon with no intent | 1 |
| loosen the campaign-id check | 7 |
| swallow an unreadable campaign | 1 |
| **no-op control** | **0 (green, as it must be)** |

`yarn format` / `lint` / `typecheck` / `build` clean; `yarn test` 763 files, 11,109 tests green.

## Items to Confirm / Review

- **`abandoned` is a record kind I added**, not something the design names. It is how "the effect did not happen" gets written down without moving the task - otherwise the only way to clear a pending intent is to succeed, and a failed spawn would look like a crash forever.
- **The fold reports corruption rather than throwing.** That is a judgement: fail-closed says stop, and reporting lets the runner stop *and* show a person what it found. If you would rather it throw, the change is one line, but the caller then cannot say which record was bad.
- **The permission tests skip on Windows and as root**, matching `rooms.spec.ts` and #1484 - `chmod` on Windows moves only the read-only attribute, and root ignores the bit, so the write that must fail simply succeeds. Windows therefore never exercises "a failed append returns false"; said here because silence would read as covered.
- **Nothing runs this.** No route, no runner, no caller - as with #1837, by design.

One repeat worth recording: my first commit again carried a literal NUL byte into a spec - this time a deliberate test case for a null byte in a campaign id, written as a raw character rather than as an escape. `test/server/sourceIsText.spec.ts` caught it again, and it is now written with a unicode escape. While there I checked whether the id regex has JavaScript's trailing-newline hole, by running it rather than reasoning: it does not, because `$` in JS matches only at the very end of the string (unlike Python's).

## User Prompt

Consolidated from the conversation: a large-scale parallel refactoring campaign written up under `future/`, generalised into one pipeline where refactoring and new development are two instances, then designed as a MulmoTerminal grid feature ("UI does not change much, ideally fully automatic"), with 8 clones rather than worktrees and an orchestrator across planning, change, secure review, review and tests. Decisions given during design: first profile `refactor`; clones in a dedicated directory; the ordinary-cell trust model accepted. Then "つづけて" - continue to the next PR.

Design: #1816. Tracking issue: #1815. PR0: #1817. State machine: #1837.

work in mulmoterminal2

## Summary by CodeRabbit

* **New Features**
  * Added append-only campaign state storage that preserves progress across restarts.
  * Campaign actions can be recorded as completed or abandoned.
  * Added campaign validation and safe listing of stored campaigns.
  * Added detection and reporting of invalid transitions, duplicate attempts, mismatched records, and corrupted log entries.
  * Preserved valid campaign history when individual log entries are unreadable or incomplete.
* **Tests**
  * Added comprehensive coverage for persistence, validation, ordering, malformed records, and error handling.


isamu:
## Review ledger — after iteration 1 (head `08e079b4`)

|---|---|---|---|---|
| 1 | codex | `listCampaigns()` treats storage it cannot read as empty, so a restart skips reconciliation for campaigns it could not inspect | FIXED | `existsSync` removed from the file entirely; only `ENOENT` is empty |
| 2 | codex | `listCampaigns()` accepts any name ending `.jsonl`, including a **directory** — reported as a campaign, then `EISDIR` when read | FIXED | `withFileTypes` + `isFile()`, matching `listRooms()` |
| 3 | me | `readCampaign()` had the same fail-open as (1) | FIXED | Same treatment, in the same commit |

Both of the reviewer's findings are the same mistake from opposite ends. **(1) is a fix I had already made in `readCampaign` and did not carry into the function beside it** — the habit this project keeps catching. **(2) is a divergence from `listRooms()` in a PR whose stated argument is that it follows `listRooms()`**; I copied the reasons and not the code.

So the fix swept the category rather than the two lines, and that turned up (3): `readCampaign` guarded with `existsSync` too, which answers `false` for a path that does not exist **and** for one it cannot traverse. Same fail-open, one function over. `existsSync` is now gone from the module and both functions answer through the same predicate:

> only the filesystem saying *there is nothing here* is empty; *I could not find out* reaches the caller.

The divergence from `listRooms()` that remains is deliberate and now says so in a comment: that one swallows every failure, which is right for a conversation listing nobody depends on, and wrong here — an empty list reads as "no campaigns" and a restart believes it.

**Break-verified:** listing directories as campaigns fails 1 test, `readCampaign` swallowing every error fails 2, `isMissing` always true fails 3.

One note on the method, since it bit me in the same round: **the first sweep reported one mutation as green when it had not applied at all** — the patch text no longer matched after formatting, and my harness printed a result anyway. A mutation that does not apply is indistinguishable from a coverage gap, which is exactly the trap the repo's guidance names. The harness now refuses to print a result it did not earn, and the re-run gave the 2 failures above.

**Iteration 1 signals:** Codex `CHANGES REQUESTED` (2) + 1 from the sweep, all fixed in `3e4d740b`. CodeRabbit posted no inline findings. Clean-round count: 0.

isamu:
## Review ledger — after iteration 2 (head `3e4d740b`)

|---|---|---|---|---|
| 4 | codex | after an abandonment, an attempt number could be **reused**, so a retry gets the same derived idempotency key as the try that failed | FIXED | Attempts strictly increase per task; the parse layer bounds them |

This one is a hole in the mechanism the PR exists to build, so it is worth stating in full rather than as a table row.

The key is `campaign/<id>/<task>/<attempt>`, and its whole job is to let a restart ask the forge and the clone *"is there already something called this?"* instead of trusting a log that may be missing its second half. **An abandoned attempt may well have left a branch or a PR behind under that name** — abandonment says the effect did not complete, not that it left nothing. Reusing the number aims the retry at exactly those leftovers, and the restart reads them as proof the retry already happened.

So attempts now increase strictly, which is what makes a key single-use. Two consequences came with it:

- **The parse layer bounds an attempt to a positive safe integer.** It arrives off a file, so it is unbounded until something bounds it, and `1e20` would render into a key and compare in ways no reader expects — the reasoning `isIssueNumber` in `common/prPhase.ts` is already written from.
- **The watermark is now `record.attempt`, not `Math.max(...)`.** With increase guaranteed, a max can only ever mean "quietly tolerate going backwards", which is the thing being prevented.

Tests cover it as a property as well as by case: **no task ever uses one idempotency key twice**, across a log that tries to.

**Break-verified:** removing the check fails 3 tests, loosening `<=` to `<` fails 2, accepting any number as an attempt fails 4, not advancing the watermark fails 6.

**Iteration 2 signals:** Codex `CHANGES REQUESTED` (1, above), fixed in `84998a7a`. CodeRabbit added no inline findings. Clean-round count: 0.

isamu:
## Review ledger — after iteration 3 (head `84998a7a`)

|---|---|---|---|---|
| 5 | codex | `CAMPAIGN_ID` admits both `Lint` and `lint`, which are **one file** on macOS and Windows — two campaigns append into one log, and reconciliation derives keys under an id the writer never used | FIXED | Lowercase only |

**`ROOM_ID_RE` is lowercase for exactly this reason and says so in its comment.** I copied its shape, widened it to `A-Za-z`, and reintroduced the defect the comment exists to warn about.

That is the third divergence in this PR from the file it claims to follow — `listRooms`' `isFile()` filter (finding 2), `listRooms`' catch (finding 1, where diverging was right but only by accident of my writing it differently), and now the id shape. The pattern is worth naming since it will apply to the next module too: **I read the reference for its reasoning and then wrote the code from memory.** Reading the reasoning is what stops a wrong decision; reading the code is what stops a wrong detail, and this PR needed both.

Why rejecting outright rather than normalising: the two spellings are only *sometimes* the same file. A rule that folds case would be right on macOS and Windows and pointless on Linux, and one that detects the filesystem makes the id namespace depend on where the server happens to run. Refusing the variant is the only answer that does not vary by platform.

**Break-verified:** re-admitting upper case fails 5 tests; adding the `i` flag fails the same 5.

**Iteration 3 signals:** Codex `CHANGES REQUESTED` (1, above), fixed in the push above. CodeRabbit added no inline findings of its own this round. Clean-round count: 0.

isamu:
@coderabbitai full review

isamu:
## Review ledger — after iteration 4 (head `606860cd`)

|---|---|---|---|---|
| 6 | codex | `appendCampaignRecord` returns `true` for records the reader later discards — `attempt: 0` type-checks as a `number`, `at: Infinity` becomes `null` in JSON | FIXED | The line is put through the reader before it is written |
| 7 | me | the first version of that fix contained a comparison **no input could fail** | FIXED | Removed |

Finding 6 breaks the one promise this return value exists to make. `CampaignRecord` is **looser than the file format** — the type says , the format says positive safe integer — so a caller can hand over something that type-checks, lands on disk, and is invisible after a restart. It gets `true`, causes the side effect, and the intent that was supposed to make it reconcilable is not there.

The fix asks the reader directly: **parse the exact line about to be written, and refuse unless it comes back as one record.** A round trip rather than a second predicate, because a second predicate can drift from what `readCampaign` really accepts and a round trip cannot.

Finding 7 is the more interesting half. I first wrote that check as a round trip **plus** a `JSON.stringify` comparison against the original record — and the mutation sweep reported *"do not compare what came back"* as **green**. Not a gap in the tests: once the line parses back as one record,  **is** the JSON projection of , and so is `JSON.stringify(record)`. The comparison could not fail for any input. It was dead code wearing a safety check's clothes, which this repo rates as worse than no check at all — and the sweep is exactly what turns "looks careful" into "provably does something".

**Break-verified after the trim:** removing the check fails 4 tests; loosening `!== 1` to `> 1` fails 4.

**Iteration 4 signals:** Codex `CHANGES REQUESTED` (1) + 1 from the sweep, both fixed in `a165dde0`. CodeRabbit full-reviewed `90f2bf1f..606860cd` and added nothing of its own. Clean-round count: 0.

isamu:
## Correction to the previous ledger, and iteration 5

**First, a correction.** My iteration-4 ledger was posted through a shell that ate three backticked words, so it reads with gaps — "the type says , the format says positive safe integer". The missing words were `number`, `readBack` and `record`. The substance stands; the comment is just missing them. (Cause: `gh pr comment --body "…"` in double quotes lets the shell run backticked text. Heredoc from here on, as this one is.)

## Iteration 5 (head `a165dde0`)

|---|---|---|---|---|
| 8 | coderabbit | `appendFileSync` does not flush, so a host failure after `true` loses the intent | FIXED | `{ encoding: "utf8", flush: true }` |

The most on-point finding of this PR, because it attacks the promise rather than the code: **the whole reason this function returns a boolean is that `true` licenses a caller to go and cause an irreversible side effect.** A write sitting in the page cache makes that licence a guess about the next few seconds of the machine's life. Node has supported `flush` since 21.x and this package requires `>=22.9`.

`rooms.ts` does not flush, and is right not to — the second deliberate divergence from it in this PR, both recorded in comments where the code differs:

| | `rooms.ts` | here | why |
|---|---|---|---|
| a failed append | swallowed | returns `false` | a lost message is a lost message; a lost intent is a side effect nobody can reconcile |
| durability | none | `flush: true` | same reason |

That is now three findings in this PR from the same place — following `rooms.ts` where it is right and *not* following it where this module's guarantee is stronger. Findings 2 and 5 were the first kind (I diverged where I should have copied); this one is the second (I copied where I should have diverged).

Codex returned `LGTM` on `a165dde0` this round and did not raise it.

isamu:
## Review ledger — after iteration 6 (head `dd17fb2d`)

|---|---|---|---|---|
| 9 | codex | flushing the file does not make its **name** durable — a power cut can leave a fsynced file that no directory entry points at, so a restart sees no intent while the side effect happened | FIXED | The containing directory and its parent are fsynced after the append |

The natural next layer under finding 8, and the same question one level down: *what exactly does `true` promise?* A flushed file with an unrecorded name is, after a power cut, a file that does not exist.

- Both directories, not one: fsyncing `campaigns/` does not make `campaigns/` itself durable inside **its** parent, and the first append creates both.
- **Windows is skipped**, because a directory cannot be opened for this there and nothing substitutes. So the doc comment now states the weaker guarantee outright — on Windows `true` means *written and flushed*, and a file created in that same moment can still be lost — rather than leaving a reader to assume the strong one.
- Everywhere else a failed directory sync **throws**. A name that may not survive is precisely what a caller must not be told `true` about.

**On verification, honestly:** durability is not unit-testable — it needs a real power cut — so what is pinned is that the path providing it is reached and that its failure propagates. Mutating the code to fsync a directory that does not exist fails 3 tests. The mutation that removes the Windows guard stays **green on macOS**, and that is not coverage: only the `test_windows` job exercises it.

Two of the last three findings have been about this one boolean, which is the right place for them — it is the only thing in the PR a caller stakes an irreversible action on.

Codex `CHANGES REQUESTED` (1, above), fixed in the push above. Clean-round count: 0.

isamu:
## Review ledger — after iteration 7 (head `02a107d5`)

|---|---|---|---|---|
| 10 | codex | the sync covered `campaigns/` and its parent, but `MULMOTERMINAL_HOME` itself can be created by the same call — and the entry naming **it** lives one level higher again | FIXED | The walk is bounded by what was actually created |

Third finding in a row on the same boolean, each one level further down, and this one is where the regress had to be stopped deliberately rather than by adding another level: **in principle it runs to the root.**

What bounds it is information Node already returns. `mkdirSync(…, { recursive: true })` hands back **the topmost directory it created**; everything below that is new and everything above already existed, so every entry above is already durable. The walk therefore runs from *that directory's parent* down to the file's own directory, and no further.

Printed for a three-level home, to check the shape rather than argue it:

```
/tmp/x                          <- the first parent that already existed
/tmp/x/deep                     <- topmost created
/tmp/x/deep/er
/tmp/x/deep/er/still            <- MULMOTERMINAL_HOME
/tmp/x/deep/er/still/campaigns  <- holds the file's own entry
```

The documented guarantee is now the one the code keeps: the data is flushed, the entry naming it is synced, **and so is every directory this call had to create**. Windows still lacks the second half and still says so.

**Break-verified:** making the walk touch a directory that is not there fails 4 tests; breaking `chainDown`'s intermediate paths fails 1. A new case drives it — `MULMOTERMINAL_HOME` three levels of nothing deep, append, read back, list.

Codex `CHANGES REQUESTED` (1, above), fixed in the push above. Clean-round count: 0.

isamu:
## Review ledger — after iteration 8 (head `1463e83d`), and a pattern worth naming

|---|---|---|---|---|
| 11 | codex | a concurrent writer whose `mkdirSync` reports creating nothing syncs only its own directory, so it can return `true` while the hierarchy another writer just created is not yet durable | FIXED — **by stating a rule rather than adding a level** | Hierarchy durability moved out of the append |

**Four consecutive rounds on this one boolean, each a level deeper than the last.** That is the loop telling me the fixes were enumerating bad cases instead of stating the rule, so this round stopped adding levels.

The rule: **a directory hierarchy and a record have different lifetimes, so they get different calls.** Durability of a directory is a property of its whole ancestor chain *and of every other process writing there* — chasing it inside `appendCampaignRecord` leaves only two endings, syncing to the filesystem root or racing, and both are answers to the wrong question.

- `ensureCampaignStore()` creates the hierarchy and makes the names it needed durable. Idempotent, and it **throws** rather than reporting a hierarchy that may not survive. The runner calls it once before a campaign runs.
- `appendCampaignRecord` syncs its own file's entry, and its promise is now stated with its precondition: durable *given a durable campaigns directory*.
- `mkdirSync` stays in the append so a first write works without setup, but the comment says plainly that the strong promise is not its to make.

### What this PR does not measure

Worth stating because the sweep would otherwise read as complete. **Two mutations that remove directory syncing entirely stay green**, and no test can do better: telling a fsync from no fsync requires a power cut. What goes red is a mutation that makes the sync *fail loudly* — syncing a path that is not there fails 3 tests — and that proves **the code path is reached**, not that durability is achieved.

So the fsync work in this PR is **argued, not demonstrated**. The tests cover reachability and error propagation. If that is not good enough for a guarantee this load-bearing, the honest alternative is to weaken the documented promise rather than to add tests that cannot discriminate.

Codex `CHANGES REQUESTED` (1, above), fixed in the push above. Clean-round count: 0.

isamu:
## Review ledger — after iteration 9 (head `854485f2`)

|---|---|---|---|---|
| 12 | codex | the same race survives inside `ensureCampaignStore`: a second caller whose `mkdirSync` reports creating nothing syncs less and can return while the first is still working | FIXED | The chain now starts from a fixed point |

**Fifth round on durability, and this one has a general answer rather than another level.** While the walk starts from *what this call created*, two callers necessarily sync different ranges, so a race is structural — no amount of extending it helps. Starting from a **fixed point** removes it outright: every caller syncs the same chain itself and needs to know nothing about what the others did.

### The line, stated because it is a choice

The chain starts at `MULMOTERMINAL_HOME`'s parent. That is a decision, not a derivation:

> This app makes durable the tree it owns — `MULMOTERMINAL_HOME` and below. Above that is the machine's own filesystem, which it neither created nor manages.

Following the regress past that line ends at fsyncing `/`, which is where it stops being engineering. The residual is real and is written into the code rather than left implicit: **if `MULMOTERMINAL_HOME` itself points somewhere that had to be created, the entries above HOME are outside this promise.**

For a home three levels deep the chain is exactly:

```
/tmp/x/deep/er                  <- HOME's parent, the fixed start
/tmp/x/deep/er/still            <- MULMOTERMINAL_HOME
/tmp/x/deep/er/still/campaigns
```

### Verification, and what it cannot reach

Making the chain touch a directory that is not there fails 4 tests — the path is reached and failures propagate. **Reducing the chain to the leaf, which is exactly the racy shape, stays green.** As with the previous round: no test distinguishes a fsync from no fsync, and none can detect the absence of a race either. The argument rests on the *shape* — each caller establishing the property for itself — not on a measurement, and saying so is more useful than a test that cannot discriminate.

If a reviewer thinks the line is drawn in the wrong place, that is the thing to argue with; it is a policy decision about how much of the filesystem this app is responsible for, and I have made it explicit rather than converged on it silently.

Codex `CHANGES REQUESTED` (1, above), fixed in the push above. Clean-round count: 0.

isamu:
## Checkpoint — round 10

**Does the diff still do what the PR body says?** Yes: a record format, a disk layer, and their tests. But the body predates `ensureCampaignStore`, which round 9 added — refreshed alongside this.

**Original vs review-response.** The two modules opened at ~110 lines of code and are ~120 now; almost none of the growth is behaviour. What grew is *durability*, which arrived entirely through review: nine of the twelve findings landed on it.

**Anything here only because a bot asked?** No, but three comments were only there because a bot *had* asked, which is different and worse. The checkpoint found:

- The file header still said "the two things anybody does with them" — there are four exports.
- `syncDirectory`'s comment predates `ensureCampaignStore` and pointed at "the guarantee below", which by now belongs to a different function. Its claim that directory syncing "only matters for the append that creates the file" also stopped being true once setup existed.
- **Two comments were narrating the review** — *"written that way at first; the mutation sweep found…"*, *"the residual is stated rather than hidden"*. `CLAUDE.md` forbids that explicitly: a comment must not reference the current task or fix, because that belongs in the PR description and rots. Trimmed to the conclusions; the history is here in the thread, which is where it belongs.

**Would I approve it fresh today?** Yes for what it does. The comment-to-code ratio is high for this repo's taste, and I would defend most of it — each remaining comment records a decision that cost a round — but it is the fair thing to flag rather than let a reviewer discover.

**Largest remaining risk, unchanged and now sharper:** the durability guarantees are **argued, not demonstrated**. No test in this PR can tell a fsync from no fsync, and two mutations proving that stay green. What is tested is that the paths are reached and that failures propagate. A reviewer who wants that guarantee actually verified needs a different instrument than a unit test, and it is better to say so than to let 174 green tests imply otherwise.

Not counting a clean round: this pushed a commit.

isamu:
@coderabbitai full review

isamu:
@coderabbitai full review

isamu:
## Review ledger — after iteration 12 (head `041b21ef`)

|---|---|---|---|---|
| 13 | codex | `ensureCampaignStore()` can create an ancestor above `dirname(MULMOTERMINAL_HOME)` and never sync it | FIXED | The chain starts at the filesystem root, with two different failure rules |

This is the residual I documented last round as out of scope, re-raised. Before arguing the line I checked whether it could be closed cheaply — and it could, which makes the argument unnecessary.

**Why it needed a new shape rather than one more level.** Removing the race needs a starting point that does not depend on who created what, and `MULMOTERMINAL_HOME` can point anywhere, so the only point covering every directory this call might create is the **root**. But syncing to the root *fatally* turns "the OS will not let us fsync `/`" into a campaign that cannot start — a worse failure than the one being fixed.

So the chain runs from the root, and the two halves fail differently because they are not equally this app's business:

| | | on failure |
|---|---|---|
| **HOME and below** | the tree this app owns | **throws** — a hierarchy that may not survive is exactly what a caller must not be told is ready |
| **above HOME** | the machine's own filesystem | **best effort** — attempted, because HOME itself may have just been created, but never fatal |

For a home three levels deep:

```
best   /            best   /tmp        best   /tmp/x
best   /tmp/x/deep  best   /tmp/x/deep/er
fatal  /tmp/x/deep/er/still            <- MULMOTERMINAL_HOME
fatal  /tmp/x/deep/er/still/campaigns
```

### The test I wrote for this was worthless, and the sweep said so

I added a case for the best-effort half using `chmod 0o500`, and the mutation *"make every ancestor fatal"* stayed **green**. The reason is the permission bit: `0o500` still lets a directory be opened for reading, which is exactly what a directory fsync does, so the sync it was supposed to defeat simply succeeded. A break-check that cannot break says nothing about the code and looks identical to coverage.

`0o100` — execute-only — still allows traversing *into* the directory, which the rest of the call needs, while refusing to open it. With that, **both directions discriminate**: making every ancestor fatal fails 1 test, making every ancestor best-effort fails 1 test.

The third mutation, *"start below the root again"*, stays green — the durability of ancestors above HOME is still not observable by any test, as with every other fsync claim in this PR.

Codex `CHANGES REQUESTED` (1, above), fixed in the push above. Clean-round count: 0 — and the round before this one was clean for both reviewers, which is the second time in this PR that a clean round was not convergence.

isamu:
@coderabbitai full review

isamu:
@coderabbitai full review

isamu:
## Review ledger — after iteration 14 (head `46e64a36`), and a question for the reviewers

|---|---|---|---|---|
| 14 | codex | with a **relative** `MULMOTERMINAL_HOME`, `path.parse("state/campaigns").root` is `""`, so the chain starts at the empty string, fails there quietly, and the directory actually holding the new home — the working directory — is never synced | FIXED | `path.resolve` on both before walking |

A real bug and a different one from the preceding rounds, so it is fixed rather than argued. `mulmoterminal-home.ts` does permit a relative home, and the mechanism is exact:

```
root of "state/campaigns"          ""
root of the resolved path          "/"
```

**Neither mutation discriminates it**, and that is worth stating rather than glossing: undoing the resolve, and resolving only one of the two, both stay green. A relative chain — `["", "state", …]` — resolves against the working directory, so the *observable* behaviour is identical; the only difference is whether the working directory itself is made durable, which no test in this PR can see. The relative-home case was added anyway, because it covers something genuine (a relative home works end to end), but it does not cover this fix.

---

## The standing question

**Ten of the fourteen rounds on this PR have been about power-failure durability, and no test in the PR can verify any of it.** Every finding has been correct and every fix has been an improvement, so this is not a complaint — but the pattern says the *promise* is what keeps outrunning the code, and a promise no test can hold is the kind that quietly stops being true.

Two ways to end it, and this is a policy call rather than a defect, so I am putting both up rather than picking one silently:

**(a) Keep the guarantee.** `true` means the record survives a host or power failure, with the caveats already documented (Windows, and above `MULMOTERMINAL_HOME`). Verified by argument and by mechanism, never by test. Future rounds keep finding levels because the surface is large.

**(b) Narrow it to what is demonstrable.** `true` means the record is written, validated against the reader, and flushed — durable across *this process* exiting. Say plainly that surviving a power cut also requires the directory entries, which `ensureCampaignStore` makes a best effort at, and stop promising more than can be shown. Codex itself offered this in an earlier round.

(b) is smaller and more honest for something that has never run; (a) is what a runner acting on `true` would want if it can be trusted. The work already done is not wasted under either — under (b) it stays, described accurately.

I have kept (a) for now, since that is what the code does today. Say which you want.

Codex `CHANGES REQUESTED` (1, above), fixed in the push above. Clean-round count: 0 — this follows a round that was clean for both reviewers, the third such reset in this PR.

isamu:
## Decision: the guarantee is narrowed to what can be shown

The repository owner picked **(b)**. The code is unchanged; the claims it makes about itself are not.

### What `appendCampaignRecord` returning `true` means now

> The record was validated against the reader, written, and flushed — so it survives **this process exiting**, and the next process can read it.

And explicitly not:

> Surviving a **power cut**. That needs the entry naming the file, the entry naming its directory, and so on up. This call syncs the first; `ensureCampaignStore` makes a best effort at the rest. None of it is promised, because none of it can be demonstrated — telling a fsync from no fsync needs a real power failure, not a test.

The doc comment now ends with the instruction that follows from it: **do not build a caller that treats `true` as proof against a power cut.**

`ensureCampaignStore` is narrowed the same way — it promises the directory exists and throws if it cannot be created; the durability of the names leading to it is best effort. One consequence worth noting: syncing below `MULMOTERMINAL_HOME` still throws, but the *reason* changed. It was "do not report a hierarchy that may not survive as ready"; that is no longer a promise, so it is now "a refusal there usually means something else is wrong with the tree this app owns", which is still worth stopping for.

### Why this rather than another round

Ten of fourteen rounds went to this guarantee. Every finding was correct and every fix improved the code — the flush, the entry sync, the fixed-point chain, the ownership line, the relative-home bug are all real work and all stay. What did not hold up was the **promise**, which kept outrunning what the code could be shown to do, and a promise no test can hold is one that quietly stops being true.

So the fsync work remains and is described accurately, which is the outcome (b) was chosen for. If a future runner needs the stronger guarantee, it needs an instrument this PR does not have — a crash-injection harness — and that is a different piece of work with its own PR.

Clean-round count: 0 (this pushed a commit).

isamu:
## Review ledger — after iteration 15 (head `6f43d6c8`)

|---|---|---|---|---|
| 15 | codex | `appendCampaignRecord` returns `false` **after** the record is on disk when the directory sync throws, leaving a pending intent for an effect the caller then declines to cause | FIXED | Nothing after the write may report failure |

A real correctness bug, and one **narrowing the promise is what made visible**. While `true` claimed power-failure durability, failing on an unsyncable directory looked principled. With the promise stated as *validated, written, flushed*, the directory sync is plainly outside it — and a step outside the promise must not be able to fail the operation.

The damage the old behaviour did is worth spelling out, because it is not "a false negative":

- the caller sees `false` and correctly does **not** cause the effect;
- the log nonetheless holds an intent for it;
- the fold rejects the retry as `intent-while-pending`, so the task cannot simply try again;
- and the next restart reads that intent as a crash window and goes off reconciling an effect that never happened.

So: **once the record is on disk, nothing afterwards reports failure.** A refusal to sync becomes a warning naming the campaign, and the call returns `true`.

**Driven, not asserted:** `chmod 0o300` on the campaigns directory — write and execute let the append land, the missing read bit makes the directory fsync fail. That is the narrow gap where the two outcomes differ, and the test sits in it. Break-verified: letting a sync refusal fail the append fails 1 test.

This is the first finding in several rounds that is not about durability at all, which is what choosing (b) was for — the promise stopped moving, so the review could look at what the code actually does.

Codex `CHANGES REQUESTED` (1, above), fixed in the push above. Clean-round count: 0.

isamu:
## Review ledger — after iteration 16 (head `81853c0e`)

|---|---|---|---|---|
| 16 | coderabbit | as root the guarded `expect` is skipped but the assertion **after** the guard still runs, so the test fails in root-based CI containers | FIXED | The guard shape is gone from the file, not just from that test |

Correct, and a defect in a test I added last round. The inner guard — `if (process.getuid?.() !== 0) expect(…)` — is a shape that *invites* this: it silently permits an assertion to sit outside it, depending on work that never happened.

So rather than patch the one test, all **eight** permission tests moved to `it.skipIf`, with the condition named once:

```ts
const CANNOT_REFUSE_ACCESS = process.platform === "win32" || process.getuid?.() === 0;
```

Naming it states the reason in one place — *a test that needs an access to be refused cannot be written where nothing refuses* — and removes the trap class rather than one instance of it. Worth noting the inner-guard shape was copied from `rooms.spec.ts`, where it is fine because nothing there asserts outside the guard; that is how a pattern gets carried past the conditions that made it safe.

**Verified in both directions by actually running them**, since a skip is otherwise indistinguishable from a pass:

- `CANNOT_REFUSE_ACCESS` pinned to `true` (a root container, simulated): **48 passed, 8 skipped** — nothing outside a skip depends on refused access.
- The old shape restored under the same simulation: the test fails with `AssertionError: expected [] to deeply equal [ { kind: 'intent', … } ]`, exactly as reported.

Codex returned `LGTM` on `81853c0e` and did not raise it — this one was CodeRabbit's, and it is the sort only a reviewer thinking about *where CI runs* would catch. Clean-round count: 0.

isamu:
## Synced with main, and re-verified on the merged result

`origin/main` had moved 11 commits ahead. Merged in (never rebased) and the gates re-run on the combination CI will actually build, rather than on the branch alone:

- `yarn install --frozen-lockfile`, then `lint` / `typecheck` / `build` clean
- `yarn test` — **764 files, 11,225 tests green**

Nothing in this PR needed changing for it; the merge is the point, not a fix.

Also correcting the record: a confirmation round I said was running never actually started — the session it was launched from ended first, so neither the CodeRabbit request nor the Codex re-run happened. The clean-round count therefore stands at **0**, not 1. Restarting it properly below.

isamu:
@coderabbitai full review

isamu:
## Review ledger — after iteration 17 (head `068e1d72`)

|---|---|---|---|---|
| 17 | coderabbit | the plan says "five rejection reasons"; the table beneath it lists six | FIXED | Count corrected, and all three statements of it checked against each other |

Small, and the same habit as most of this PR's self-inflicted findings: `attempt-reused` was added to the table two rounds ago and the count two lines above it was left alone.

Rather than edit the number and move on, the three places that state it are now checked as **sets** instead of by eye:

```
prose says    : 6
table rows    : 6   illegal-transition, intent-while-pending, no-intent, wrong-event, phase-mismatch, attempt-reused
type declares : 6   no-intent, wrong-event, illegal-transition, phase-mismatch, intent-while-pending, attempt-reused
prose == table: true
table == type : true
```

Counting by eye is what produced the error; comparing the sets is what says the fix is complete rather than merely applied.

**Iteration 17 signals:** Codex `LGTM` on the merged head. CodeRabbit 1 finding, fixed above. Clean-round count: 0.

isamu:
@coderabbitai full review

isamu:
## Review ledger — after iteration 18 (head `701c7467`)

|---|---|---|---|---|
| 18 | coderabbit | `isCommon` accepts a non-finite `at`, because `JSON.parse` turns `1e309` into `Infinity` and `typeof Infinity === "number"` — but `JSON.stringify` writes it back as `null`, so such a record does not survive re-serialisation | FIXED | `Number.isFinite` |

Verified before fixing rather than taken on trust:

```
parsed.at            : Infinity  | typeof: number
passes typeof number : true
passes isFinite      : false
re-serialised        : {"kind":"intent","at":null,…}
reader accepts it    : true
```

**Why it matters more than a stray value.** `appendCampaignRecord` validates by round-tripping the line *through the reader* — the point being that the check cannot drift from what `readCampaign` accepts. That argument only holds while the reader is not **laxer** than the writer, and here it was: the writer refuses `at: Infinity` (a test covers it), the reader took it. So the round-trip was measuring against a reader more permissive than the one that runs after a restart.

The fix is one predicate, but the property is what was missing, so it is now pinned as one: **the reader refuses exactly what the writer refuses** — driven over the non-finite values plus a zero and a fractional attempt, each fed back through `recordLine` and expected to read as nothing.

**Break-verified:** restoring `typeof raw.at === "number"` fails 2 tests (4 across the earlier per-line cases).

**Iteration 18 signals:** Codex `LGTM` on the unchanged head — its second in a row, but not a clean round, because CodeRabbit found this in the same round. Clean-round count: 0.

isamu:
@coderabbitai full review

isamu:
@coderabbitai full review

isamu:
## Converged — two clean rounds on `8e100ec6`

|---|---|---|---|
| 19 | LGTM | full review of `c19d0876..8e100ec6`, nothing new | none |
| 20 | LGTM | full review again, nothing new | none |

All checks green. Both reviewers read the final head, which includes the merge of the 11 commits `main` gained mid-review.

## What eighteen findings changed

**Nine of the eighteen landed on one boolean** — `appendCampaignRecord`'s return value — because it is the only thing here a caller stakes an irreversible action on. That is also where the most useful outcome came from, and it was not a fix:

> **The promise was narrowed to what can be demonstrated.** `true` means validated, written, flushed — durable across *this process* exiting. Surviving a power cut is attempted and not promised, because no test in this PR can tell a fsync from no fsync, and mutations removing the syncing entirely stay green. The fsync work all stays; only the claim about it changed.

That decision immediately paid: with the promise no longer moving, the very next round found a **real** bug it had been hiding — `false` was being returned after the record was already on disk, which left an intent nobody would cause, blocked the retry as `intent-while-pending`, and sent the next restart to reconcile a crash window that never happened.

The other findings worth naming:

- **A task could reach `merged` having declared no paths at all** — a merge with no exclusion over anything it touched, reachable from the second state. `revalidate` moved off `leased` so `declare` sits on every route.
- **An attempt number could be reused after an abandonment**, aiming a retry at whatever the failed try left behind under that same idempotency key, which a restart then reads as "already done".
- **`Lint` and `lint` were one file** on macOS and Windows, so two campaigns could share one log.
- **The reader was laxer than the writer** on non-finite timestamps, which quietly undermined the round-trip check's whole argument.

## Habits this PR exposed, since they will recur

- **Three findings came from following `rooms.ts` in spirit and not in code.** I read the reference for its *reasoning* and then wrote from memory — the reasoning stops a wrong decision, the code stops a wrong detail, and this needed both.
- **Two mutation sweeps reported green for mutations that had not applied**, until the harness was made to refuse a result it had not earned.
- **One test that could not fail**: `chmod 0o500` still permits the directory read a fsync performs, so the case meant to defeat it simply succeeded.
- **A `JSON.stringify` comparison that no input could fail** shipped in a fix and was caught by the sweep rather than by reading.

## What is not covered

- **Nothing runs this.** No route, no runner, no caller — as with #1837, by design.
- **Startup reconciliation is deliberately absent.** Everything it reads — live sessions, clone leases, forge state — does not exist yet, and writing it against assumed inputs would decide those designs by accident rather than reconcile anything.
- **Durability is argued, not demonstrated**, and the code now says so where a reader will find it.

PR #1845, opened 2026-08-23
feat(campaign): パスの排他が決めること（#1815 PR1c）

isamu:
## Summary

What a claim registry **decides**, with no disk under it — the answer to phenomenon A, "sessions keep modifying the same places". Pure functions only; **nothing imports it yet**.

The design lists four things acquisition must be: a normalised identifier, a compare-and-set, an expiring term, and an owner token that fences a stale holder out. This PR is the deciding half of that; **the filesystem CAS is the next PR**.

### Why split there

PR #1841 showed that review of the I/O drowns out review of the decisions — nine of its eighteen findings landed on one boolean about durability. And the decisions are where a mistake is *invisible*: a prefix test that ignores segment boundaries makes `src/foo` and `src/foobar` one exclusion, and passes every test anyone would think to write.

### Normalisation is not reinvented

`server/infra/canonical-path.ts` already does it: realpath through symlinks, and on a case-insensitive filesystem the spelling the disk actually has — which is *more* correct than folding case by hand, since on a case-sensitive filesystem those are genuinely different paths. This module takes paths that are already canonical and says so.

**A second normaliser would be a second answer**, and two answers to "is this the same path" is precisely what exclusion cannot survive.

## The decisions, and why each is shaped that way

| | |
|---|---|
| `covers` | does one path contain the other — **at a segment boundary** |
| `pathsConflict` | either direction counts |
| `conflictingPaths` | which of the wanted paths a holding collides with |
| `isLive` | the term is **exclusive**: at exactly `expiresAt` it is over |
| `blockingClaims` | only live overlaps block; an expired claim is not in the way, and not gone either |
| `acquisitionOrder` | the order paths must be taken in |
| `fencingVerdict` | may this holder still act |
| `nextToken` | what a takeover writes |
| `renewed` | what a renewal changes |

Three of those carry an argument rather than a preference:

- **`acquisitionOrder` is a deadlock argument.** Two tasks wanting overlapping sets always meet on the same path first, so one loses there instead of each holding half of what the other needs. **By code unit**, because the argument only holds if the order is the same in *every* process — `localeCompare` makes it depend on who is running the app, which is exactly what `common/byCodeUnit.ts` exists to avoid.
- **`fencingVerdict` keeps three refusals apart** because a runner does different things with them: `not-the-owner` is a bug in the caller, `superseded` means somebody took the claim over and this task must stop, `expired` means nobody holds it and the task must reacquire before touching anything.
- **The generation rises even when the same task takes its own claim back.** The previous holder may still be running, and the number is the only thing that will tell its next write from the new one's.

## Verification

**Mutation sweep, 8 mutations, every one red.** Restore checked after each; baseline green before and after.

| mutation | tests failed |
|---|---|
| sort by locale instead | 1 |
| raw prefix rather than segment-aware | 1 |
| expiry inclusive | 2 |
| ignore the generation | 2 |
| ignore the owner | 1 |
| same owner keeps its generation | 1 |
| renew from the old deadline | 2 |
| expired claims still block | 1 |

`yarn format` / `lint` / `typecheck` clean; 30 tests.

**One test was worthless before it was fixed.** The ordering case first used `ab` / `a-b`, which sort the same way under both comparators on this machine, so it demonstrated nothing. Replaced with a case pair after checking empirically which pairs actually differ. It asserts **only our own order** — pinning what `localeCompare` returns would make the test's answer depend on where it runs, which is the property being ruled out.

## Items to Confirm / Review

- **`covers` uses `path.sep`**, so a claim recorded on one platform and read on another would compare wrongly. That cannot happen today (claims are per-machine, and paths come from `canonicalPath` on the same host), but it is an assumption worth naming rather than discovering later.
- **Expiry is exclusive.** At exactly `expiresAt` the claim is over. An inclusive reading would give two holders for one instant; if you would rather the boundary went the other way, it is one character and two tests.
- **The registry is not here**, so nothing enforces any of this yet. That is the split, not an omission.
- **Nothing runs it.** No route, no runner, no caller — as with #1837 and #1841.

## User Prompt

Consolidated from the conversation: a large-scale parallel refactoring campaign written up under `future/`, generalised into one pipeline where refactoring and new development are two instances, then designed as a MulmoTerminal grid feature ("UI does not change much, ideally fully automatic"), with 8 clones rather than worktrees and an orchestrator across planning, change, secure review, review and tests. Decisions given during design and implementation: first profile `refactor`; clones in a dedicated directory; the ordinary-cell trust model accepted; the durability promise narrowed to what can be demonstrated. Then "つづけて" — continue.

Design: #1816. Tracking issue: #1815. PR0: #1817. State machine: #1837. Durable record: #1841.

https://claude.ai/code/session_01RhVjP9cFKVS18kowoTyLgz

work in mulmoterminal2

## Summary by CodeRabbit

- **New Features**
  - Added rules for detecting overlapping campaign paths, including platform-specific path behavior.
  - Added claim expiration, blocking, deterministic acquisition ordering, renewal, and takeover handling.
  - Added ownership verification to identify valid, expired, superseded, or unauthorized claims.
  - Added generation-based claim tokens for safer ownership transitions.

- **Documentation**
  - Added a Japanese design document describing claim behavior and planned component boundaries.

- **Tests**
  - Added comprehensive coverage for path conflicts, expiration, ordering, ownership, takeovers, and renewal edge cases.


isamu:
## Review ledger — after iteration 1 (head `52a4a9d9`)

| 1 | 1 | codex | `fencingVerdict` answers `not-the-owner` for a stale holder when the claim was taken over by **another** task, and `superseded` when the same task took it back | FIXED | The generation is asked first — commit on this branch, tests `claim-rules.spec.ts` |

Correct, and **the function's own doc comment says so**: `not-the-owner` is documented as *a bug in the caller*, `superseded` as *somebody took this over, stop*. A stale holder is not a caller bug, so the code was not doing what the comment beside it promised.

Confirmed by running it rather than by reading:

```
cross-owner takeover, old holder asks : not-the-owner   <- wrong
same-owner takeover,  old holder asks : superseded      <- right
a stranger at the current generation  : not-the-owner   <- right
```

One situation, two answers, decided by who happened to take the claim over — which reads to a runner as a bug in its own code instead of "somebody has this now".

**The fix is the ordering, and the ordering is the distinction.** Ask the generation first: anything below the current one is superseded whoever holds it now. A generation *above* the current one was never issued by any registry, so it stays `not-the-owner` — an invented token, not a late holder.

**Break-verified:** restoring the owner-first order fails 2 tests; treating a future generation as superseded fails 1; dropping the generation from the owner line fails 1.

Also synced `origin/main` (3 commits) and re-ran the gates on the merged result: lint / typecheck / build clean, 215 campaign tests green.

**Iteration 1 signals:** Codex `CHANGES REQUESTED` (1, fixed). CodeRabbit reviewed `87c0de34..52a4a9d9` and posted no findings of its own. Clean-round count: 0.

isamu:
## Review ledger — after iteration 2 (head `aca2646d`)

| 1 | 1 | codex | `fencingVerdict` answered `not-the-owner` for a stale holder after a cross-owner takeover | FIXED | generation checked first — `claim-rules.ts` |
| 2 | 2 | **me** | `covers` gave the two spellings of one directory **different answers** depending on direction | FIXED | both sides trimmed of trailing separators |
| 3 | 2 | **me** | the trim's root floor was a branch **no input could reach**, and the test I wrote for it passed either way | FIXED | both removed |

Both reviewers were clean this round — Codex `LGTM` on `aca2646d`, CodeRabbit reviewed `7140c87d..aca2646d` and posted nothing. Rows 2 and 3 came from reading the module again rather than from the thread.

**Row 2.** `covers("/a/", "/a")` was `false` while `covers("/a", "/a/")` was `true`. One directory, two spellings, answered by which way round it was asked. `pathsConflict` hid it by asking both ways, so exclusion itself was never at risk — but `covers` is exported, and a caller reaching for it directly would have been told the wrong thing. Both sides are now trimmed first.

The trim is deliberately **not a regex**: `[/\\]+$` is the anchored-quantifier shape that backtracks super-linearly, and a path is a value from outside. (`sonarjs/super-linear-regex` flagged the first attempt, which is the rule earning its place.)

**Row 3 is the more interesting one, and it is a mistake inside the fix for row 2.** I gave the trim a floor at the filesystem root — `/` stays `/` — and the mutation sweep reported *"trim past the root"* as **green**. Not a coverage gap: the caller appends a separator before comparing, so `/` and `""` answer identically for every absolute path. **No input distinguishes the branch.**

So the guard went, and so did the test I had written for it, which passed under both implementations and therefore said nothing while reading as coverage. Verified by printing the behaviour rather than arguing it:

```
covers("/","/")=true   covers("/","/x")=true    covers("/a/","/a")=true
covers("/a","/a/")=true covers("/a","/ab")=false covers("/a//","/a/b")=true
```

**Iteration 2 signals:** Codex `LGTM`, CodeRabbit clean, 2 findings from my own read. Not a clean round — commits pushed, and the findings were real. Clean-round count: 0.

isamu:
## Review ledger — after iteration 3 (head `c86a8fe9`)

| 1 | 1 | codex | `fencingVerdict` answered `not-the-owner` for a stale holder after a cross-owner takeover | FIXED | generation checked first |
| 2 | 2 | me | `covers` answered differently for the two spellings of one directory | FIXED | both sides trimmed |
| 3 | 2 | me | that fix's root floor was a branch no input could reach, and its test passed either way | FIXED | both removed |
| 4 | 3 | codex | **`renewed` could revive an expired claim in place**, leaving the generation where it was | FIXED | `null` once the term has run out |

Row 4 is the most serious finding on this PR. Confirmed by running it rather than reading:

```
at 1500, before renewal : expired
renewed while expired   : {"token":{"owner":"t1","generation":3},"expiresAt":2000}
at 1500, after renewal  : ok
```

A holder that fencing had already turned away comes back as `ok` **with the same generation** — while another task may hold those paths, distinguishable only by a higher generation that never moved. That does not merely weaken expiry; it removes the point of having a term at all.

`renewed` now extends only a live claim and returns `null` otherwise. **There is no renewing your way out of a lapse** — a failed claim is reacquired through `nextToken`, which is what moves the generation and fences the old holder out for good.

### On the sweep, because it misreported twice

- The first attempt's two mutations **did not apply** — formatting had changed the text — and the harness printed *"MUTATION DID NOT APPLY (not a result)"* rather than a verdict it had not earned.
- The rewritten first mutation then returned `exit=1` **from a syntax error**, not a test failure: dropping the ternary left the arrow body as a brace block. An exit code from a compile failure is not evidence that the tests discriminate, so it was rewritten with parentheses and re-run.

With both applying and compiling: reviving a lapsed claim in place fails 2 tests; checking liveness against the *new* deadline instead of now fails 2.

`lint` / `typecheck` / `build` clean, 218 campaign tests green, and `origin/main` is already in.

**Iteration 3 signals:** Codex `CHANGES REQUESTED` (1, fixed). CodeRabbit's walkthrough still shows `7140c87d..aca2646d` — it has not read `c86a8fe9`, so its silence this round is not a signal. Clean-round count: 0.

isamu:
@coderabbitai full review

isamu:
## Checkpoint — round 5

**Does the diff still do what the PR body says?** Yes — the deciding half of a claim registry, no storage. Nothing has drifted in scope.

**Original vs review-response.** ~35 lines of code, four findings against it, and none of them added a function. All four changed *what an existing rule answers*, which is what a rules-only PR should expect.

**Anything here only because a bot asked?** No — but three comments were stale, and the checkpoint is what found them:

- `ClaimToken.generation` still explained itself as the thing that distinguishes holders **"when both name the same owner"**. That was true before iteration 1; the whole point of that fix was that the generation now decides *regardless* of owner. The doc was describing the behaviour the reviewer had asked me to remove.
- `withoutTrailingSeparator` and `covers` carried **review narration** — *"a floor at the root was written first and removed"*, *"untrimmed, `/a/` did not cover `/a`"*. `CLAUDE.md` forbids a comment that references the current task or fix: it belongs in the PR and rots in the source. Trimmed to the rules and their reasons.

Worth naming: **I flagged this same habit in my own checkpoint on #1841 and then did it again here.** The thread is where a fix's history belongs; the source gets the conclusion.

**Would I approve this fresh today?** Yes for what it does. The comment-to-code ratio is high for this repo's taste — ~150 lines around ~35 — and I would defend most of it, since each remaining comment carries an argument (why segment-aware, why code-unit order, why the generation is asked first) rather than restating the code. But it is the fair thing to flag rather than let a reviewer discover.

**Largest remaining risk:** nothing enforces any of this. These are decisions a registry will consult, and whether they are the *right* decisions is answered by the filesystem layer that has not been written. In particular `acquisitionOrder`'s deadlock argument is only sound if the storage layer actually acquires in that order and releases everything on failure — a property this PR states and cannot test.

**CodeRabbit is rate-limited** ("all 2 included reviews currently available" used) and has not read past `aca2646d`. Its silence on the last two heads is not a signal, and the loop is not counting it as one.

Checkpoints do not count toward the two clean rounds, but this one pushed a commit, so the count stays at 0 regardless.

isamu:
@coderabbitai full review

isamu:
@coderabbitai full review

isamu:
## Review ledger — after iteration 4 (head `b6a3e2ea`)

| 1 | 1 | codex | `fencingVerdict` answered `not-the-owner` for a stale holder after a cross-owner takeover | FIXED | generation checked first |
| 2 | 2 | me | `covers` answered differently for the two spellings of one directory | FIXED | superseded by row 6 |
| 3 | 2 | me | that fix's root floor was a branch no input could reach | FIXED | superseded by row 6 |
| 4 | 3 | codex | `renewed` could revive an expired claim in place | FIXED | `null` once the term has run out |
| 5 | 5 | me (checkpoint) | three comments described behaviour the fixes had removed | FIXED | trimmed to the rules |
| 6 | 4 | coderabbit → me | **`covers` was a ninth hand-rolled copy of `isWithin`** | FIXED | delegated; the hand-rolled trim deleted |

### Row 6, which started as a case question and ended somewhere else

CodeRabbit asked about `canonicalPath` keeping the case of components that **do not exist yet**. Verified, and it is real — a claim is declared *before* the work, so a path to a file about to be created is the ordinary case:

```
A: /tmp/…/NewThing.ts
B: /tmp/…/newthing.ts     (one file on this volume)
pathsConflict(A, B) = false
```

Two tasks can both claim the same new module — exclusion failing exactly where a campaign is most likely to collide.

Looking into how to fold case properly turned up the larger thing. **`server/infra/path-within.ts` already implements this comparison**, and its header opens with:

> It was hand-rolled in eight places … the two where the sides come from DIFFERENT sources were both wrong on Windows (#802).

I had written the ninth. `isWithin` carries what mine did not: the platform's own `path` implementation, case folded on Windows, trailing separators and roots handled by `resolve`. My hand-rolled trim had meanwhile failed twice on outside input — a regex that backtracked super-linearly, then recursion that threw `RangeError` at 20,000 separators (rows 2 and 3's territory). `path.resolve` has neither problem.

**All 37 existing tests passed against the shared helper unchanged**, including the 50,000-separator regression — which is the useful signal that this is a substitution rather than a rewrite.

The residual is written down rather than left to be found: **on a case-insensitive macOS volume the two spellings of a not-yet-existing file are still two keys.** `path-within.ts` deliberately folds only on Windows (a case-sensitive APFS volume is supported, and folding on a guess widens a containment guard), so keying missing components against the volume's actual behaviour is the registry's job — this layer cannot decide it. Both platform arms are now pinned by tests.

**Break-verified:** a hand-rolled prefix fails 4 tests; swapped arguments fail 6; ignoring the platform fails 1.

**Iteration 4 signals:** Codex `LGTM` on `b6a3e2ea`. CodeRabbit was rate-limited for two rounds and its full review has now landed — 2 findings, both above. `lint` / `typecheck` / `build` clean, 221 campaign tests green. Clean-round count: 0.

isamu:
@coderabbitai full review

isamu:
@coderabbitai full review

isamu:
## Converged — two clean rounds on `c151b309`

|---|---|---|---|
| 5 | LGTM | reviewed `b6a3e2ea..c151b309`, nothing new | none |
| 6 | LGTM | **full review** of `7140c87d..c151b309` (merge base to head), nothing new | none |

All checks green. Both reviewers read the final head, and round 6's was a full re-review of the whole diff rather than an incremental one.

## Six findings, and what they were really about

Two came from Codex, two from my own reading, one from a checkpoint, one from CodeRabbit — but four of the six trace to the same root: **I wrote comparisons and orderings by hand that this repository had already solved.**

- **`covers` was a ninth hand-rolled copy of `isWithin`** (row 6), in a module whose own header records that the previous eight included two that were wrong on Windows. Delegating removed my hand-rolled trim, which had already failed twice on outside input — a regex that backtracked super-linearly, then recursion that threw `RangeError` at 20,000 separators. The 37 existing tests passed against the shared helper **unchanged**, which is what makes it a substitution rather than a rewrite.
- **The sort needed `byCodeUnit`**, which exists for precisely this reason — a locale-dependent order would let two runners disagree, which is the deadlock the ordering prevents.

The two genuine logic defects were both about **the fencing token doing its job**:

- `fencingVerdict` answered `not-the-owner` for a legitimate stale holder when the claim had changed hands to another task, and `superseded` when the same task took it back — one situation, two answers, and the function's own doc said which was right. Asking the generation first is the fix and the distinction.
- **`renewed` could revive an expired claim in place**, leaving the generation where it was, so a holder that fencing had already turned away came back as `ok` while somebody else might hold the paths. That removes the point of having a term at all.

## What is not covered

- **Nothing runs this.** No route, no runner, no caller — the storage layer is the next PR.
- **`acquisitionOrder`'s deadlock argument is stated, not tested.** It is only sound if the registry actually acquires in that order and releases everything on a partial failure — a property this layer cannot check.
- **The case gap is a known limitation with a named owner.** `canonicalPath` keeps the case of components that do not exist yet, and a claim is declared before the work — so on a case-insensitive macOS volume, `NewThing.ts` and `newthing.ts` are two keys for one file. `isWithin` folds on Windows and deliberately not on macOS; keying missing components against the volume's real behaviour belongs to the registry.

PR #1863, opened 2026-08-25
feat(shared-app): an eighth template, for a discussion AI agents hold

snakajima:
## What

Adds `templates/ai-council.md` — the shape where the participants are **AI agents rather than people**: a human host posts a question, several agents argue it in public under personas they chose, and the host ends the thread when they have heard enough.

The existing seven templates are all "people fill in a form"; the participants are human and go unsaid. This is the first one where they are not, and it is the only worked example of two keys the skill documents in prose and demonstrates nowhere.

## Why it earns a slot

Each template in this set is built around one declaration key it teaches (`assignee`, `stampField`, `idFrom: "field"`, `views[].live`, `writerDelete`, `views[].limit`). By that standard this one carries two nobody shows:

- **`agents[]`** — had a ten-line JSON snippet at `SKILL.md:307` and no template using it. It is what makes an app self-describing: an agent at another MulmoTerminal reads the brief with `useSharedApp describe` and takes a seat **from the slug alone**, with no copy of the repository.
- **the close triad** — `refIn` + a `transitions` map with no exit from `closed` + `sealed`. Explained at `SKILL.md:970` and declared by no sample. It is the highest-trap-density thing in the skill: **each key alone is theatre**, walked around in two writes, and the failure is silent because an app carrying only `refIn` looks closed.

Note `agents[]` already appears in `salon.md` and `append-feed.md` as prose, but there the agent *operates* the app for the owner (a desk agent approving bookings). Here the agents' output **is the content**, and only that case needs the triad.

## The part worth reviewing carefully

This is the only shape that has to bind **its own owner**, because every agent signs in as the owner. The template states what the fence does *not* reach, and I would rather that stayed in than be trimmed:

- `publish` rewrites all three declarations — the gate holds through the app's DATA and nothing else
- identity is a claimed record, not a credential; nothing stops one agent registering five speakers
- there is no floor control; two agents can answer the same message blind to each other

Left out, the template teaches a false sense of safety.

## Provenance

Drawn from the live app at `~/git/ai/apps/roles`, published as `ai-roundtable`, **whose rules are deployed and whose close has been verified in production**. The template renames collections for clarity (`roles` -> `speakers`, which collided with the members-role vocabulary) and swaps the avatar emoji for a monogram, per this repo's no-emoji rule.

## Verification

Beyond the declaration gate the other seven get:

- **`skillTemplates.spec.ts`** — the declaration through the real `check`, plus the sandbox / canvas / protocol / unique-hue checks. `--hue: 140`, chosen as the widest gap in the set (taken: 25, 65, 95, 175, 230, 265, 295, 330).
- **`skillTemplateAiCouncil.spec.ts`** (new, 17 cases) — **both pages RUN in jsdom**, over the claims a declaration cannot carry: the thread is in the server's stamp order whatever order rows arrive in, an unstamped row sorts last, the close is armed before it fires, controls are drawn from `viewer.can` rather than the role, `createdAt` is never sent, and the composer does not eat a question typed while the last was in flight.
- **Mutation-checked**, because a green test on first run is worth distrusting: reverting the sort sentinel and removing the close arming each fail exactly the case that names them.
- `yarn format` / `lint` / `typecheck` clean; full suite 11,359 passed.

## Two stale counts I did NOT touch

Both predate this change and are adjacent enough to mention rather than fold in:

- `live-poll.md:10` says "the other four templates" (there were already seven).
- `design.md` measures contrast over "the six template hues" and lists six, omitting `append-feed`'s 175 — and now 140. The claim it supports is that the palette holds at *any* hue, and 140 is mid-gamut green, well clear of the cyan region the file names as the tight one; but the table is a measurement and re-measuring it is a separate change.

Say the word and I will fix both here.

work in mulmoterminal

## Summary by CodeRabbit

* **New Features**
  * Added an AI council shared-app template for public discussions between AI agents hosted by a human.
  * Includes topic management, agent personas, public viewing, live host controls, and permanent discussion closure.
  * Added setup guidance and deployment-ready data configuration.

* **Tests**
  * Added coverage for template availability, deployment, schemas, discussion behavior, permissions, controls, and validation.

* **Documentation**
  * Updated the template catalog to reflect eight available templates.

### DIFF
--- a/server/skills/mulmoterminal-shared-app/SKILL.md
+++ b/server/skills/mulmoterminal-shared-app/SKILL.md
@@ -37,7 +37,7 @@ Eight shapes are written out in full — declaration, schemas, and the reasoning
   rules.
 - **[templates/survey.md](./templates/survey.md)** — **collecting answers**, with nothing to run out
   of (a survey, a quiz, an application form, a sign-up with no cap). The shortest declaration of the
-  seven, and the shape most often written with a public page and nothing else — so this one is built
+  eight, and the shape most often written with a public page and nothing else — so this one is built
   around its `member` page, which is where the answers are read. It also spells out the three-way
   trade above, and what a tally may and may not claim about values a respondent typed.
 - **[templates/meeting-room.md](./templates/meeting-room.md)** — a bookable unit you can LIST IN
--- a/server/skills/mulmoterminal-shared-app/SKILL.md
+++ b/server/skills/mulmoterminal-shared-app/SKILL.md
@@ -26,7 +26,7 @@ the user turns this down.
 
 ## Start from a template when one fits
 
-Seven shapes are written out in full — declaration, schemas, and the reasoning behind each key:
+Eight shapes are written out in full — declaration, schemas, and the reasoning behind each key:
 
 - **[templates/salon.md](./templates/salon.md)** — a request that a NAMED PERSON approves, and only
   their own (a salon's bookings, interviews, repairs, review assignments). This is what `assignee`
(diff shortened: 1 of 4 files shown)


snakajima:
Both findings were right; fixed in 8d4b510.

**codex, `ai-council.md:884` (P2) — real defect, mine.** The composer guarded *both* fields on the title alone, so a host who left the title and rewrote only the question lost the question to the clear — precisely what the comment above it claims to prevent. Each field is now judged separately, and there is a new case that tells the two apart: it passes under the old code with a title-only edit and fails with a question-only one. Mutation-checked, so it is not a green that cannot go red.

Worth noting for anyone reading the template: the wrong version *looks* equivalent, which is why it also sits in the app this template was drawn from. The comment now names the near-miss rather than just the rule.

**CodeRabbit, `SKILL.md:29` — right, and the one I missed.** Line 40 read "The shortest declaration of the seven". It wraps as `of the` / `seven`, so the line-based grep that caught the other two counts walked straight past it. Now "the eight".

Re-checked: `format` / `lint` / `typecheck` clean, template specs 33 passed (18 page cases, up from 17).

The two stale counts named in the PR description (`live-poll.md`'s "other four templates", `design.md`'s six measured hues) are still untouched — both predate this change. Happy to fold them in if wanted.

PR #1908, opened 2026-08-28
fix(ui): CopyCodeBlock が渡された class をボタンに届ける (#1895)

### LINKED ISSUE #1895: bug: CopyCodeBlock drops the class it is given (fragment root), so the copy button is unstyled
## What happened

`TerminalCell` passes the shared cell-button styling to `CopyCodeBlock`:

```vue
<!-- src/components/TerminalCell.vue:1609 -->
<CopyCodeBlock v-if="sessionId" :class="CELL_BTN" :session-id="sessionId" :cwd="cwd" :agent="agent" />
```

but `CopyCodeBlock`'s template has **two root nodes** — the `<span class="relative inline-flex">` and the `<Teleport to="body">` added for the manual-copy modal (#968). Vue cannot auto-inherit an attribute onto a fragment root, so the class is dropped and the browser console says so on every agent cell that mounts:

```
[Vue warn]: Extraneous non-props attributes (class) were passed to component but could not be
automatically inherited because component renders fragment or text or teleport root nodes.
  at <CopyCodeBlock class="inline-flex items-center justify-center rounded-md border-none leading-none
      bg-transparent h-[26px] w-7 text-[16px] cursor-pointer
      text-[var(--cell-btn,var(--text-secondary))] hover:bg-hover hover:text-fg" ... >
  at <Terminal ... > at <TerminalCell ... > at <TerminalGrid ... > at <GridView> at <App>
```

**It is not only a warning — the button is visibly wrong.** The inner `<button>` carries `class="cell-btn"`, and `.cell-btn` has no CSS rule anywhere in the repo; it is a marker/hook only. So the copy button loses `bg-transparent`, `border-none`, its `h-[26px] w-7 text-[16px]` box, the `--cell-btn` colour and the hover treatment, and renders with the browser's default button chrome — a bordered, filled box sitting among flat icon buttons in the same toolbar row.

This is the failure mode `CLAUDE.md` already documents under Styling, from #787: *"shared CSS silently stops applying when a component's template has a fragment root — Vue gives the parent's scope id to a single root element only."* Same cause, different symptom: here it is an inherited `class` rather than a scope id.

## What I expected

The copy button to look like every other button in that row, and no Vue warning on mount.

## Steps to reproduce

1. Start any **agent** cell (claude / codex / a custom agent — a shell cell uses `LauncherCell` and does not render this button).
2. Look at the cell's second toolbar row, to the right: the copy-code-block button has a visible border and background while its neighbours are flat.
3. Open the browser console: the `[Vue warn]` above is logged once per cell that mounts.

## Suggested fix

Either give the component a single root — move the `<Teleport>` inside the wrapping `<span>`, which is legal and changes nothing about where it renders — or set `inheritAttrs: false` and bind `v-bind="$attrs"` onto the inner `<button>` explicitly. The first is smaller; the second makes the intent explicit if a second root is ever wanted again.

Worth a spec either way: nothing currently asserts that the button receives the class it is given, which is why this survived.

## Environment

- mulmoterminal 4.11.0, run from a local dev server (`yarn dev`)
- Node v24.19.0, macOS arm64, Chromium (Playwright)
- vue ^3.5.42
- `src/components/CopyCodeBlock.vue` last changed in #1300 (an unrelated `void`-prefix sweep), so this has been present since the Teleport was added

## Notes

Found while verifying an unrelated change (#1890). Not caused by it — that PR does not touch either file.

isamu:
## Summary

`TerminalCell` styles this button by passing `CELL_BTN` — the same class every other button in that toolbar row carries — and the class was reaching **nothing**. The template has two roots (the `<span>` and the `<Teleport>`), so Vue could not auto-inherit it.

`.cell-btn` is a marker with no CSS rule of its own, so the button fell back to the browser's default chrome. Measured in the real app against its neighbour:

| | class | border | background | size |
|---|---|---|---|---|
| copy button, **before** | `cell-btn` only | `2px outset` | `rgb(239, 239, 239)` | 19.2 × 17.3px |
| its neighbour (Move terminal left) | full `CELL_BTN` | `0px none` | transparent | 26 × 28px |
| copy button, **after** | full `CELL_BTN` + `cell-btn` | `0px none` | transparent | 26 × 28px |

## The repair the issue suggested first would have looked fixed and not been

The issue offered two options. Option 1 was "move the `<Teleport>` inside the wrapping `<span>` for a single root". I applied it and measured:

```text
ROOT html: <span class="relative inline-flex MARKER_CLASS"><button class="cell-btn" ...
button has MARKER_CLASS: false
warnings: 0
```

**The warning goes away and the button stays unstyled** — the class lands on the wrapper. Slightly worse, in fact, since `h-[26px] w-7` then sizes the wrapper.

That follows from what `CELL_BTN` contains: `bg-transparent border-none h-[26px] w-7 cursor-pointer hover:bg-hover` is button decoration. Every other use in the repo puts it on the `<button>` itself (`TerminalCell.vue:1434,1437`, `LauncherCell.vue:77`, …); `CopyCodeBlock`'s button was the only one carrying just the marker.

So this takes **option 2**: `defineOptions({ inheritAttrs: false })` and `v-bind="$attrs"` on the inner `<button>` — the destination is named rather than left to inheritance, and a second root can be added back later without breaking it again. The wrapping `<span class="relative inline-flex">` stays, because the transient note is positioned against it.

## Items to Confirm / Review

1. **The specs assert on the BUTTON, never on the wrapper, and that is the point rather than a detail.** A spec written against `wrapper.classes()` passes option 1's repair. The mutation table below includes that repair for exactly this reason.

2. **My first draft of two of those specs was vacuous.** They used `wrapper.element`, which for a multi-root component is test-utils' own `<div data-v-app>` container — an element the class can never land on, so they would have passed whatever the component did. They now go through a `rootOf()` helper that takes `firstElementChild`, with the reason recorded next to it.

3. **In a production build the warning does not appear at all** — measured 0 before the fix, because dev-only warnings are stripped. So for anyone running a release, the *only* symptom was the visual one, and the console evidence in the issue came from `yarn dev`. The warning is pinned at the unit level, where it does fire.

4. `defineOptions` is new to this repo's components (`inheritAttrs: false` exists only in the plugin runtime today). Say if you would rather it were spelled with a separate `<script>` block.

## User Prompt

- 「ではつぎはこれ https://github.com/receptron/mulmoterminal/issues/1895」

## 実装のアプローチ

The issue is precise and its diagnosis is right; only its first suggested remedy is not, and that was settled by applying it and looking rather than by reasoning about it. Both wrong shapes are kept as mutations so the next person does not have to redo that.

Design notes and the full record: [`plans/fix-1895-copy-button-attrs.md`](plans/fix-1895-copy-button-attrs.md).

## 検証

Tree: `5682f673` (origin/main) + this change.

**Gates** — `format` / `lint` / `typecheck` / `build` / `test` all exit **0**. `yarn test` **11659 passed / 50 skipped**.

**break-verify** (restores checked byte-identical)

| mutation | result |
|---|---|
| revert to the shipped state (no `inheritAttrs`, no `$attrs`) | 2 red |
| **the issue's option 1** — single root, class inherited onto the span | 2 red |

**Real app** — a claude cell in Chrome, `getComputedStyle` on the copy button and on a sibling `button.cell-btn`:

| | border | background | height | width | font | cursor |
|---|---|---|---|---|---|---|
| copy button | `0px none` | transparent | 26px | 28px | 16px | pointer |
| Move terminal left | `0px none` | transparent | 26px | 28px | 16px | pointer |

Identical, and zero `Extraneous non-props attributes` warnings in the console. The before-state in the same setup is the first table in this description.

Closes #1895

https://claude.ai/code/session_01Ry9BpQtMSkZ9D64fv3cztP

work in mulmoterminal4

### DIFF
--- a/plans/fix-1895-copy-button-attrs.md
+++ b/plans/fix-1895-copy-button-attrs.md
@@ -1,0 +1,87 @@
+# fix(ui): CopyCodeBlock が渡された class を落とす (#1895)
+
+> 時系列のログであって現状の仕様書ではない。
+
+## 症状（実測で再現）
+
+`TerminalCell` は他のセルボタンと同じ `CELL_BTN` を渡している:
+
+```vue
+<CopyCodeBlock v-if="sessionId" :class="CELL_BTN" :session-id="sessionId" ... />
+```
+
+`CopyCodeBlock` のテンプレートはルートが 2 つ（`<span>` と `<Teleport>`）なので Vue は
+attribute を自動継承できず、**class はどこにも載らずに消える**。unit で観測:
+
+```text
+ROOT html: <span class="relative inline-flex"><button class="cell-btn" ...
+button has MARKER_CLASS: false
+warnings: 1   ← [Vue warn] Extraneous non-props attributes (class)
+```
+
+実アプリ（claude セル、Chrome）でも確認。`.cell-btn` は CSS 規則を持たないマーカーなので、
+ボタンはブラウザ既定の chrome で描かれる:
+
+| | class | border | background | サイズ |
+|---|---|---|---|---|
+| 修正前の copy ボタン | `cell-btn` のみ | `2px outset` | `rgb(239,239,239)` | 19.2 × 17.3px |
+| 隣のセルボタン | CELL_BTN 一式 | `0px none` | transparent | 26 × 28px |
+
+## issue が提案していた案1 は「直ったように見えて直らない」
+
+issue は 2 案を挙げていた。案1 は「`Teleport` を `<span>` の中へ入れて単一ルートにする」。
+**実際に当てて測った:**
+
+```text
+ROOT html: <span class="relative inline-flex MARKER_CLASS"><button class="cell-btn" ...
+button has MARKER_CLASS: false
+warnings: 0
+```
+
+**警告は消えるが、class は外側の `<span>` に載り、ボタンは unstyled のまま。** しかも
+`h-[26px] w-7` がラッパーに付くので、状況としては少し悪くなる。
+
+理由は `CELL_BTN` の中身がボタンの装飾だから —— `bg-transparent border-none h-[26px] w-7
+cursor-pointer hover:bg-hover`。repo の他の使い方も全て `<button class="cell-btn"
+:class="CELL_BTN">` と**ボタン自身**に載せている（`TerminalCell.vue:1434,1437`,
+`LauncherCell.vue:77` 等）。`CopyCodeBlock` の button だけがマーカーしか持っていなかった。
+
+## 採った直し方 —— 案2
+
+`defineOptions({ inheritAttrs: false })` + 内側の `<button>` に `v-bind="$attrs"`。
+**行き先を推測（継承）に任せず名指しする。** ラッパーの `<span class="relative inline-flex">`
+はそのまま残す —— 一時表示される note がそこに絶対配置されるため。
+
+副次的な利点として、将来 2 つ目のルートを足しても壊れない。
+
+## 検証
+
+**unit（`test/src/components/codeBlockCopy.spec.ts` に 5 件追加）**
+
+すべて **button** に対して assert する。これは細部ではなく要点で、`wrapper.classes()` に
+書いたテストは案1 の「直ったように見える」修正を通してしまう。
+
+> 最初 `wrapper.element` に対して書いたが、multi-root コンポーネントではそれが test-utils の
+> コンテナ `<div data-v-app>` で、**class が絶対に載らない要素を見る空振りのテスト**だった。
+> 実際のルート要素（`firstElementChild`）を見るヘルパに直し、その理由をコメントに残した。
+
+| ミューテーション | 結果 |
+|---|---|
+| 出荷時の状態に戻す（`inheritAttrs` も `$attrs` も無し） | 2 red |
+| **issue の案1**（単一ルート化、class は span へ） | 2 red |
+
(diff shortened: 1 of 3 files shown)


isamu:

| — | 1 | codex | none — `CODEX VERDICT: LGTM` on `000a9235` | — | "The explicit attribute forwarding correctly places `CELL_BTN` on the interactive button while retaining the wrapper needed for the transient note. The added tests cover the prior fragment-root regression and the incorrect single-root alternative." |
| — | 1 | coderabbit | **no review — rate limited** | NOT A SIGN-OFF | "Review limit reached… next included review available in 14 minutes." It has not seen this head. Recorded rather than counted; will re-trigger once the window clears. |

Nothing raised, and nothing I found re-reading the diff myself. 12 checks green.

One clean round is not convergence, and half of this one is a bot that never looked — so this is not ready on the strength of it.

### Two things in this change that are worth a reviewer's eye

**The specs assert on the `<button>`, never on the wrapper.** That is load-bearing: the repair the issue suggested first (single root) silences the Vue warning and puts the class on the `<span>`, leaving the button as unstyled as before. A spec written against `wrapper.classes()` passes that repair, so the mutation table includes it deliberately.

**Two of those specs were vacuous when first written.** They used `wrapper.element`, which for a multi-root component is test-utils' `<div data-v-app>` container — an element the class can never land on, so they would have passed whatever the component did. They go through a `rootOf()` helper now, with the reason next to it.

PR #2000, opened 2026-09-07
feat(shapescript): import shapescript-plugin 1.1.0 and serve its file dispatch

snakajima:
## What

`@mulmoclaude/shapescript-plugin` 1.1.0 turns a ShapeScript model into a **file**: `presentShapeScript` saves a new `script` under `artifacts/shapes/<slug>-<epoch-ms>-<token>.shape` and names it as `data.filePath`, accepts `path` instead of `script` to present a `.shape` that already exists, and the View's source editor loads and saves through a new `loadShape` / `saveShape` dispatch.

None of that follows from bumping the range. The host has to hand the package its file capabilities and serve the dispatch, so this PR is the port.

- `@mulmoclaude/core` `^4.7.0` → `^4.8.0` (the plugin peers on it), `@mulmoclaude/shapescript-plugin` `^1.0.0` → `^1.1.0`.
- **`FILES_CONTEXT` is now per-tool.** Its own comment called this shot: *"The moment a second plugin wants `byPath`, FILES_CONTEXT has to become per-tool."* shapescript-plugin 1.1.0 is that second plugin. `byPath` is extension-scoped, so one shared object can only ever carry one plugin's version — handing `presentShapeScript` the html-scoped instance would refuse every `.shape` it was pointed at, and widening one object to all extensions would let each tool open the others' files. A tool with no entry gets `artifacts` alone.
- `shapeScriptByPath` in `backends/openPath.ts` — a `.shape`-scoped byPath, beside the markdown / html / mulmoscript ones.
- **`backends/shapescript.ts`** (new) — intercepts `loadShape` / `saveShape` on `/api/plugin/presentShapeScript` before the generic `/api/plugin/:toolName` catch-all, and publishes a file-change after a save so an open View live-refreshes. A request with no `kind` falls through to the package's tool-call execute. It mirrors `backends/html.ts` deliberately: the two plugins share this contract, and the reasoning behind each step is documented there.

## What is NOT here

`renderShapeScript` — the new MCP tool that rasterises a model to a PNG so the agent can look at it — is a MulmoClaude **host** tool, not part of the package. MulmoTerminal gets the persistence half only. Porting the renderer would mean bringing over the Puppeteer driver and the render page; worth doing separately if it turns out to be useful here.

## Test change

`test/scripts/mulmoclaudePeerRanges.spec.ts` loses its `@mulmoclaude/shapescript-plugin` entry from `NO_CORE`. That is the spec working as designed rather than a change to accommodate: the plugin previously declared no `@mulmoclaude/core` peer, 1.1.0 does, and the spec requires a plugin that starts declaring one to lose its recorded entry. It failed until the entry was removed.

## Verification

`yarn typecheck`, `yarn lint`, `yarn build` clean. `yarn test`: **12,188 passed**, 3 files skipped, 0 failed.

Not exercised end-to-end in the running app — the dispatch route is a straight port of the html one and is covered by the same shape of wiring, but nobody has yet opened a ShapeScript view here and pressed Apply.

https://claude.ai/code/session_01RvsMjzTPP41UUAKS1R3Saa

## Summary by CodeRabbit

* **New Features**
  * ShapeScript models are now saved as editable `.shape` files and can be reloaded in place.
  * Existing ShapeScript files can be opened and updated directly.
  * Larger ShapeScript conversions are supported.
  * ShapeScript load and save actions now update the app’s file state automatically.

* **Documentation**
  * Added unreleased changelog notes describing the new ShapeScript file workflow.


snakajima:
Pushed `b42ace46` with the regression coverage codex asked for: `backends/shapescript.spec.ts` (load/save for an artifact and for a `.shape` outside `artifacts/shapes`, missing path, refusal to create, a save with no `script`, traversal, and fallthrough for a `kind`-less tool call), plus the `.shape` scope in `fileChange.spec.ts` and session-relative rooting in `presentPathRoot.spec.ts`. 12,199 tests pass.

**On the absolute-path finding (`presentPathRoot.ts:59`)** — not changing, and worth being explicit about why.

`absolutizePresentPath` leaves an absolute path alone deliberately; the existing spec asserts exactly that (`"leaves an absolute path alone"`), and it is the documented contract of the `path` argument these three tools share: *"a workspace-relative path or, where the host permits it, an absolute one."* `presentDocument` opens and overwrites any `.md` on disk today and `presentHtml` any `.html` — that is what makes "open my repo's README and edit it" work at all. Constraining `presentShapeScript` alone would make it the odd one out among the tools it was ported from, and the user would find that `.shape` files are the one kind they cannot point the app at.

Constraining all three is a coherent position, but it is a product decision about what the `path` capability means — it would change three tools across two hosts, and belongs in its own PR with that framing rather than being introduced as a side effect of a dependency import.

What this PR does tighten in the same area: `saveShape` is overwrite-only (the package refuses to CREATE), so the dispatch channel can edit a `.shape` the user already had open but cannot mint one at a path of the caller's choosing — asserted by the new spec.

PR #2010, opened 2026-09-08
feat(shapescript): serve renderShapeScript, so an agent can look at its model

snakajima:
## Why

MulmoTerminal could *show* a 3D model but not **check** one. `renderShapeScript` was a MulmoClaude host tool, and a terminal opened in the workspace reaches MulmoTerminal's own MCP surface — which never had it. This closes that.

## Shape of it

A **host tool**, like `manageCollection`, because its execute needs server internals a plugin is not handed (the workspace artifacts root). Everything a *model* sees comes from `@mulmoclaude/shapescript-plugin/render` (1.3.0) — schema, defaults, the four-view sheet, the sentence naming the file — so the two hosts cannot describe the same tool differently. This repository supplies only the two things that are genuinely its own:

- **Reading** follows `presentShapeScript` exactly: the artifacts root for a model this app saved, `shapeScriptByPath` for any other `.shape` on disk. The two tools accept the same paths.
- **Saving** answers with an **absolute** path — and this is precisely why the package leaves saving to the host. MulmoTerminal's sessions run in per-project directories, so a workspace-relative path resolves to nothing from the cwd the agent is actually in, or worse to a different file of the same name. Renders go to `artifacts/renders/` rather than beside the models: a `.shape` is a source someone edits, a PNG is a by-product regenerated on demand.

In the **`render` group beside `presentShapeScript`**, per the owner's decision — a cell that can show a model should be able to check one, and splitting the pair would leave an agent able to present a model it cannot look at first.

It is **not** in `AUTO_ALLOWED_TOOLS`. It is the one render tool that starts an external *process* — a headless browser — which is the far side of the bar that list draws ("saves an artifact and draws it, and calls nothing external"). The prompt is answered once per project.

## Two stale comments fixed

Both were found while wiring this up, and both had been quietly false since shapescript-plugin 1.1.0:

- The `render` group was documented as *"draws into the Canvas panel and stops there. No side effect outside it."* `presentShapeScript` has saved its `.shape` since 1.1.0, and `renderShapeScript` saves a PNG. The line now says what is actually true: a render tool touches only its own output, never the workspace's data or anything off this machine.
- `AUTO_ALLOWED_TOOLS` justified listing `presentShapeScript` on the grounds that *"it does not even save one — the ShapeScript source travels in the result and is rendered client-side"*. That was the reasoning behind a tool running **unattended**, and it stopped being true at 1.1.0. It stays on the list, but now for the same reason as the other three rather than a stronger one that no longer holds.

## Verification

`yarn format` / `lint` / `typecheck` / `build` clean. **12,211 tests pass** (820 files, 3 skipped).

Driven end to end against a real temp workspace — an inline `script` renders, a saved `artifacts/shapes/*.shape` renders by path, a `.shape` outside the artifacts root renders through `byPath`, the answer is absolute, the PNG exists on disk with content, and a non-`.shape` path is refused with a message the agent can act on.

New `test/server/infra/shapescriptRenderTool.spec.ts` (8 tests) covers all of that plus the traversal refusal and the `script` XOR `path` contract; `toolGroups.spec.ts` gains the group membership and the deliberate auto-allow exclusion.

Rasterisation quality itself is not asserted — it needs a real browser and the suite must stay runnable without one.

https://claude.ai/code/session_01RvsMjzTPP41UUAKS1R3Saa

## Summary by CodeRabbit

* **New Features**
  * Added `renderShapeScript` to render ShapeScript models into labeled PNGs from four camera angles.
  * Rendered images are returned for inspection and saved to workspace artifacts.
  * Supports inline scripts and saved ShapeScript files, including relative paths.
  * Added first-use permission handling because rendering launches a headless browser.

* **Documentation**
  * Added an Unreleased changelog entry and updated guidance on rendering behavior.

### DIFF
--- a/common/toolGroups.ts
+++ b/common/toolGroups.ts
@@ -194,12 +194,25 @@ export const LEGACY_GUI_SERVER_IDS: readonly string[] = ["mulmoterminal-gui"];
 // but now for the same reason as the other three rather than a stronger one that no
 // longer holds.
 //
-// `renderShapeScript` is deliberately ABSENT. It is the one tool here that starts an
-// external PROCESS — a headless browser — and can occupy a core for the length of its
-// render budget. The bar this list draws is "saves an artifact and draws it, and calls
-// nothing external"; spawning Chromium is the far side of it, and the prompt is
-// answered once per project.
-export const AUTO_ALLOWED_TOOLS: readonly string[] = ["presentForm", "presentChart", "presentHtml", "presentShapeScript"];
+// `renderShapeScript` is here too, and the reasoning is worth stating because the first
+// attempt got it wrong. It starts a PROCESS — a headless browser — which reads like the
+// far side of "calls nothing external", so it was left off. But this bar is about REACH
+// and COST, not about process creation: the tools kept off it spend money
+// (presentDocument's image fill), publish to the internet, or act in someone else's
+// account. A local Chromium rendering a page WE construct, with every request it makes
+// aborted unless it is one of our two assets, does none of that. It writes a PNG under
+// the workspace artifacts and burns CPU for at most the render budget.
+//
+// Leaving it off was also not the half-measure it looked like. This list governs GRID
+// cells; the workspace passes `allowedToolNames()`, which auto-approves everything not
+// in NEVER_AUTO_APPROVED_TOOLS — so the tool prompted in a cell and ran unattended in
+// the workspace, which is the worst of both and matches no stated policy (codex on
+// #2010). The two paths now agree.
+//
+// And the friction had no payoff: the tool exists so an agent can CHECK a model before
+// showing it, which is a render-look-fix loop. A prompt in the middle of that costs the
+// user attention to approve the agent looking at its own work.
+export const AUTO_ALLOWED_TOOLS: readonly string[] = ["presentForm", "presentChart", "presentHtml", "presentShapeScript", "renderShapeScript"];
 
 /** Tools that must keep the agent's permission prompt on EVERY claude session, including the
  *  workspace.
--- a/common/toolGroups.ts
+++ b/common/toolGroups.ts
@@ -8,7 +8,12 @@
 // config already written against it.
 
 // The groups, ordered by how much damage a call can do.
-//   render   — draws into the Canvas panel and stops there. No side effect outside it.
+//   render   — draws a model or document for the user, and writes nothing but the
+//              artifact it just drew. It USED to mean "no side effect outside the
+//              Canvas panel"; presentShapeScript saves its `.shape` and
+//              renderShapeScript saves a PNG, so the honest line is that a render
+//              tool touches only its own output, never the workspace's data or
+//              anything off this machine.
 //   data     — reads/writes the workspace's structured data (collections, accounting).
 //   media    — generation that is slow, costly, and lands files on disk.
 //   external — reaches a third-party account or API.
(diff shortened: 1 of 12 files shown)


snakajima:
Pushed `907e61c0`. Two things, and the first **corrects the PR description above** — please read this rather than that paragraph.

**Permissions (codex, `toolGroups.ts:202`).** I claimed the tool "asks permission the first time". That was wrong, and codex showed why: `AUTO_ALLOWED_TOOLS` governs GRID cells only, while the workspace passes `allowedToolNames()`, which approves everything not in `NEVER_AUTO_APPROVED_TOOLS`. Leaving the tool off the first list without adding it to the second meant it prompted in a cell and ran unattended in the workspace — no stated policy, and the opposite of what I wrote.

Of the two coherent resolutions I've taken **auto-allow**, which also reverses my earlier reasoning. The bar this list draws is about *reach and cost*, not process creation: the tools kept off it spend money, publish to the internet, or act in someone else's account. A local Chromium rendering a page we construct — with every request aborted unless it is one of our own two assets — does none of those. It writes a PNG under the workspace artifacts and burns CPU for at most the render budget. And the friction had no payoff: the tool exists so an agent can *check* a model before showing it, so a prompt in the middle of that loop costs you attention to approve the agent looking at its own work. Say the word if you'd rather it always prompt; that is `NEVER_AUTO_APPROVED_TOOLS` and a one-line change.

The invariant is now asserted, because nothing did.

**A pre-existing gap that check uncovered — worth your decision, not mine.** `presentDocument` is deliberately kept off `AUTO_ALLOWED_TOOLS` because its execute runs `fillImages`, a **paid** generation call, and the comment says the prompt is wanted in front of it. But it is not in `NEVER_AUTO_APPROVED_TOOLS` either — so in the workspace session it is auto-approved and spends that money without asking. Same shape as the bug codex just found in my code, except this one predates this PR. I have **not** changed it: that is a policy call about money, not a detail of a ShapeScript port. It is pinned in the spec as the current state, so whenever it is decided the test fails and says so.

**My own spec broke CI**, for exactly the reason this PR's description gives: CI has no Chromium, so `rendered` is false there and four assertions failed. The suite now probes once and splits — argument contract, path routing and refusals always run; the pixel cases run only where a browser exists. The probe sits at MODULE scope because `it.runIf` is evaluated at collection time; in a `beforeAll` it left the flag false everywhere and skipped silently, which is worse than the failure it was replacing. Verified both ways: 8 pass here, 4 pass and 4 skip under a deliberately broken `PUPPETEER_EXECUTABLE_PATH`.

snakajima:
`3005c2fa` — Windows CI found a case my probe did not cover, and it is worth recording because it says something about the renderer, not just the test.

A **missing** browser answers `rendered: false` with install guidance, which is the tidy path I had designed for. But on Windows CI the browser **launches and then cannot navigate** — `Navigation timeout of 30000 ms exceeded` — which throws instead. My probe sat at module scope without a guard, so that took down the entire spec file, including the cases that need no browser at all. It now treats any failure as "cannot rasterise here", which is the right reading either way: the flag means "can I assert on pixels", not "why not".

Two things I did **not** change, deliberately:

- **The renderer maps that condition to a hard error, not to `RenderUnavailableError`.** A browser that starts but cannot navigate is an environment failure, exactly like a missing one, and arguably belongs on the graceful path so the model gets "carry on without me" instead of a stack trace. That is a change in `@mulmoclaude/shapescript-plugin/render` and another publish cycle, so I have not slipped it in here.
- **`--no-sandbox` is not in the launch args.** It is the usual reason Chromium misbehaves on CI, and it may well be what Windows is hitting. Same story: a package change, and one that loosens a browser sandbox, which deserves its own decision rather than riding along with a port.

Also, for the record on the local run: `headlessPreview.spec.ts` failed once alongside this and passed both in isolation and on a re-run of the full suite. That is browser contention between two specs that each drive Chromium, not a regression — I checked the baseline without my change passes too.

PR #2067, opened 2026-09-14
feat(shapescript): port publishShapeScript's update-by-id to the host writer

snakajima:
## What

Takes `@mulmoclaude/shapescript-plugin@4.0.0` ([receptron/mulmoclaude#3158](https://github.com/receptron/mulmoclaude/pull/3158)), where `publishShapeScript` accepts an optional `id` and rewrites the user's own gallery post in place under the same URL. Only the account that published it can update it; with `id` every other argument is optional, a field given replacing the post's and one omitted keeping it.

The plugin's `ShapeGalleryWriter` now requires two more members, supplied here exactly as in MulmoClaude's host adapter:

- `readPost(id)` — a `getDoc` of `shapes/{id}`, `null` when absent.
- `updatePost(id, patch, expect)` — a `runTransaction` that re-reads the post, refuses it with `POST_CHANGED_MESSAGE` unless it still carries the owner and object ids the plugin read, and applies a field-level patch with a server `updatedAt` and no `createdAt` (the rules freeze it). Two edits racing on one post cannot orphan each other's objects.

Dependency bumped via `yarn add`; changelog entry under Unreleased.

## Tests

- The tool's schema now leads with `id`.
- `postUpdateOf` sends only the fields given plus a server `updatedAt`, never `createdAt`.
- `postStillMatches` accepts the read state and rejects a changed script, thumbnail, owner, or a missing document.
- The writer exposes `readPost` / `updatePost`.
- `yarn format`, `yarn lint`, `yarn typecheck` pass; the full Vitest suite passes (860 files, 12687 tests).

https://claude.ai/code/session_0172wthUbXnVQZJY7wymr2dk

## Summary by CodeRabbit

- **New Features**
  - Posts can now be updated in place using their existing gallery URL instead of creating duplicate copies.
  - Updates support partial changes, allowing individual fields to be replaced while leaving omitted fields unchanged.
  - Only the original publisher can update a post.

- **Bug Fixes**
  - Added protection against conflicting simultaneous edits, preventing one update from overwriting another or leaving related content out of sync.

- **Documentation**
  - Added unreleased documentation for the updated publishing behavior.


snakajima:
Re the package.json nitpick: the range was changed with `yarn add @mulmoclaude/shapescript-plugin@^4.0.0`, not by hand — the generated `yarn.lock` change is in the same commit (ba73fc12, 5 files).

PR #2083, opened 2026-09-15
feat(transcript): the phone reads a copilot conversation

isamu:
## Summary

The FOURTH per-agent implementation of the phone's conversation view (#1822). claude and codex landed in #2078 with the `not-supported` status, cursor in #2081; copilot is the last of the measurable four. A copilot cell's conversation now reaches the phone instead of answering "not supported".

**Copilot is the first source that is not a file**, and that is the actual cost of this PR. Every source so far is located from the session key and read as a byte window off a file's tail. Copilot has no file — it has `turns` in `$COPILOT_HOME/session-store.db`, whose natural window is `ORDER BY turn_index DESC LIMIT n`. So `TranscriptSource` becomes a two-member union discriminated by `kind`, and the two kinds meet at the **scan**: both hand the same `TranscriptScan` to the same `transcriptViewOf`, so the line budget, the byte cap and the eviction rule stay in one place rather than being reimplemented per agent.

The fold itself is the cheapest of the four, because copilot's shape is **declared in SQL** rather than hidden in an untyped blob — `turns(session_id, turn_index, user_message, assistant_response, timestamp)`, one row per turn with the prompt and the reply already separated. There is no part type to recognise, no wrapper to strip, no double-write to guard, and **no `unknown` row**: cursor's is load-bearing because a new part type would otherwise vanish, and copilot has no part axis for one to catch.

**The limitation, stated rather than papered over: a copilot turn shows no tools at all.** `forge_trajectory_events` — the table whose columns are literally `tool_call_id`, `command`, `output`, `exit_code` — is EMPTY, including for the two sessions on this machine that demonstrably ran tools. So a copilot turn carries what was asked and what was said, and nothing about what ran: thinner than cursor (which shows WHICH tools ran but never what they answered). `session_files` is not the missing half and is deliberately not pressed into service — `UNIQUE(session_id, file_path)` makes it a deduped per-session index, so rendering it per turn would under-report and mis-attribute.

## Items to Confirm / Review

1. **The union is a change to the shared read path every agent goes through.** "The three file sources behave the same" is a claim, so it was RUN, not argued: both trees answered for **7,083 (cwd, id) pairs** — every claude transcript of this project, every cursor chat, every codex rollout on this machine — and whole `TranscriptView` JSON was compared, not a summary. **One id differed, and it is this session's own live transcript growing between the two runs**; re-running both back to back against a file whose size did not move produced identical output. Please sanity-check that reasoning rather than the number.
2. **The cwd is in the SQL, and it has to be.** Copilot keeps ONE store for the machine, so `WHERE session_id = ?` alone would let a hand-edited `?session=` put another project's conversation in this cell — the hole `copilotSessionExistsForCwd` closed on the resume path (Codex review on #2063). The query JOINs `sessions` on `cwd`. Verified against the real store: a real id asked from a different directory answers `none`, not the conversation.
3. **Is `COPILOT_TURNS_READ_LIMIT = 256` the right bound?** Sized so the QUERY never decides what the phone sees — the 250-line budget does. Two things about it changed in the CodeRabbit round and are worth a look: the read asks for **one more row than it keeps**, because "as many rows as the limit" is not "a turn is missing" (a session of exactly 256 turns is complete, and marking it `truncated` tells the phone there is more before this when there is not); and both TEXT columns are read through `substr(col, 1, TRANSCRIPT_MAX_BYTES)`, because `LIMIT n` bounds ROWS and `all()` was materialising 256 unbounded values into the JS heap every poll. Nothing renderable is lost — the whole view is capped below that.
4. **Ordering copilot last.** Measured at 200 calls each under load ~30: the copilot source costs **0.96 ms** whether it hits or misses (the sqlite open dominates), against **8.72 ms** for the whole four-source miss path. It is last so that `node:sqlite`'s lazy import stays off the path the overwhelming majority of cells take.
5. **The cwd string form.** The reader is handed `ptys.get(id)?.cwd ?? sessionCwd(id)`; the shipped resume path compares the same `sessionCwd(...)` value against the same column (`ws-routes.ts:774-778`), so the forms are already known to match in practice. Flagging it because a mismatch would fail *silently* — an empty view, not an error.
6. **A small store, and why it is enough here when cursor's 35 barely were.** 8 turns across 6 sessions, six of them this project's own probes. The risk a small store carries elsewhere is "a shape we have not seen"; here the shape is a schema, and a change to it is a migration that fails loudly.

## Measurements

Copilot CLI **1.0.83**, `~/.copilot/session-store.db`, read through the WAL **on a copy** so an uncheckpointed page could not hide rows:

| table | rows |
|---|---|
| `sessions` | 8 |
| `turns` | 8 |
| `session_files` | 1 |
| `checkpoints` / `session_refs` / `forge_trajectory_events` | **0** |
| `assistant_usage_events` | 11 |

Break-verification, each mutation applied to a file checked against a pristine copy first and restored and re-checked after (all three matched at the end):

| mutation | went red |
|---|---|
| the prompt is left to the fold (the trap #2081 paid for) | 6 |
| a prompt-less row folds into the previous turn | 1 |
| an empty row still opens a turn | 2 |
| `String()` instead of `readString` on the columns | 1 |
| the cwd drops out of the query | 1 |
| the oldest turns are kept instead of the newest | 1 |
| copilot never joins the source list | 1 |

The denominator is honest: this reader is new, so every test above is new too. The number says these decisions are catchable, not that a pre-existing suite catches them.

Read through the REAL reader against the REAL store: the two-turn Japanese conversation in `~/mulmoclaude` and a two-turn probe session both fold correctly, prompt and reply in order with their timestamps.

## Approach and key decisions

- **`TranscriptSource` is a union, not a widened single shape.** A file source keeps `locate` + `createFold` byte-for-byte; a query source answers a `TranscriptScan` or null. `viewFromSource` branches once. Widening one interface to cover both would have put `locate: null` on copilot and taught every reader to check.
- **The source emits BOTH rows.** `foldTurnRecord` supplies the prompt row only when the boundary record rendered nothing, so handing it a copilot row's reply alone would print the answer and drop the question. That is exactly the defect #2081 found in the cursor fold; copilot's one-row-per-turn shape walks straight into it.
- **`prompt` is a string and never null**, which is what says "every row opens a turn": copilot keys rows by `turn_index`, so a reply is never a continuation of the row before it. A row with an empty `user_message` still gets its own turn rather than being folded into someone else's question.
- **Truncation is marked by the source, not the fold.** The fold's `truncated` means "the budget evicted something"; a filled row limit is the different statement that the READ stopped early. `transcriptViewOf` ORs the two so the phone sees one answer.
- **`hasReader` needed no change** — it is derived from `TRANSCRIPT_SOURCES`, so wiring copilot removes it from `not-supported` by construction. The spec line moving from the not-supported test to the wired-agent test IS that assertion.
- **No `unknown` row for copilot**, deliberately. Documented in the file with the reason, because its absence looks like an omission next to cursor.

## Not in this PR

grok, muse and antigravity, unchanged from #2081: no session of any of them exists on this machine to count shapes from, and the #1822 lesson is that a fold written without counting the store is how codex's `custom_tool_call` (52% of rollouts) was missed. They keep answering `not-supported`, which is the honest answer and already ships.

Copilot's `turns` table would also serve row 12 of the capability matrix (last turn → header prompt, handoff, round table, prompts pane). Not wired here — that is a different consumer and a separate change.

## User Prompt

- 「https://github.com/receptron/mulmoterminal/issues/1822 対応できるものは全て対応。対応できないものは not supported をだす？」 — handle everything that can be handled for #1822, and emit "not supported" for the rest.
- 「はい、すすめて」「すすめる」 — proceed with the implementations, one agent at a time.
- 「merge」 — merge #2081 (cursor), which this PR builds on.

Part of #1822. Not `Closes` — grok, muse and antigravity are still unimplemented.

## Summary by CodeRabbit

- **New Features**
  - Added Copilot conversation support to the phone’s transcript view.
  - Conversations are scoped to the working directory and display prompts and replies chronologically.
  - Added limits to keep transcript loading efficient; older content may be truncated.
  - Copilot sessions without readable conversation data show no transcript.
  - Copilot transcripts do not include tool-call records.

- **Bug Fixes**
  - Corrected truncation reporting so conversations at the exact limit are not marked as truncated.

- **Documentation**
  - Updated capability and remote-host documentation to describe Copilot transcript support and its lack of tool-call records.


isamu:
## CodeRabbit のレビュー対応

**2 件とも当たりで、両方直しました。** 1 件目のほうが示唆的です。

### 1. `truncated` が「欠けていないセッション」にも立っていた (`transcript-view-read.ts`)

最初の版は「読んだ行数が limit と同数なら truncated」でした。しかし **「limit と同じ行数」は「ターンが欠けている」ではありません** — ちょうど 256 ターンのセッションは完全なのに、電話に *there is more before this* と表示されてしまう。

読み取りは **保持する数より 1 行多く** 要求するようになり、`more` は「その余分な 1 行が実在したか」になりました。

この区別こそが「truncated を fold ではなく source が立てる」理由そのもの（fold の `truncated` は「予算が evict した」、こちらは「読み取りが早く止まった」という別の文）なので、そこを間違えていたのは論点を自分で崩していました。

### 2. `LIMIT n` は行数の上限であってテキスト量ではない (`copilot-sessions.ts`)

`queryReadOnlySqlite` は `all()` で終わるので、**5 秒ごとのポーリングで、上限のない `assistant_response` を 256 行ぶん JS ヒープに実体化**していました。

両カラムを `substr(col, 1, TRANSCRIPT_MAX_BYTES)` 経由で読むようにしました。**表示され得るものは何も失われません** — ビュー全体がこれより小さい上限で切られるので、これを超えた値はターンがどう並んでも表示されることがない。`substr` は文字数、cap はバイト数なので UTF-8 では常に緩い側に外れます（cap より厳しくなることがない）。

### 破壊検証（どちらも pristine コピーとの一致を前後で確認）

| 元に戻す変異 | 赤くなったテスト |
|---|---|
| 古い `rows.length >= limit` の truncation 判定 | 1 |
| 両カラムから `substr` の上限を外す | 1 |

新規テスト 2 本（「ちょうど read limit ちょうどのセッションを欠損扱いしない」「ビューが表示し得ない大きさの値を実体化しない」）が、それぞれの検出器です。

### `test_windows` の失敗について

**この変更のものではありません。** 落ちているのは `test/server/infra/shapescriptRenderTool.spec.ts` の 2 本で、いずれも ~31 秒のタイムアウト。同じファイル・同じタイムアウトが **main の `75867e56` でも落ちています**（`cc520dfa` と `e8c19dd9` は緑）。この PR は transcript の読み取りと copilot の sqlite しか触っておらず、shapescript のレンダリングとは接点がありません。ジョブは再実行中です。

PR #2089, opened 2026-09-15
update

isamu:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved communication between the shared preview and its host page.
  * Preview messages are now accepted only from the currently displayed preview frame, preventing unrelated messages from being processed.
  * Replaced preview frames are handled correctly, and message listening stops when the preview is unloaded.

* **Chores**
  * Updated editor, plugin, linting, and testing dependencies to newer patch versions.


isamu:
## CI fixes on this branch

Two failures, neither of them a defect in the dependency bumps themselves.

### 1. `lint-and-build` — `sonarjs/post-message` on `SharedAppPreview.vue`

The `eslint-plugin-sonarjs` bump widened the rule: it previously called `resolveFunction` on the listener argument directly, so a listener passed as an **identifier** was never inspected. It now resolves the identifier to its function, so the preview pane's `message` handler is examined for the first time.

Comparing `event.origin` there verifies nobody. The frame is `sandbox="allow-scripts"` with no `allow-same-origin`, so its document has an opaque origin: every message it sends arrives with `origin === "null"`, as does one from any other opaque-origin document. `event.source` is the sending window itself, which only that frame can be, and the rule recognises no such check.

Rather than scope the exception to a 900-line component, the listener moved to `src/utils/sharedAppPreviewChannel.ts` — the shape `eslint.config.js` already prescribes for a per-file exception, and the same reason `writeTerminalSelection` was lifted out of `useTerminalConnections.ts`. The guard logic is unchanged. The new spec covers the case an origin check cannot: a **second** opaque-origin iframe carrying the same `origin === "null"` is rejected.

### 2. `test_windows` — `copilot-turns.spec.ts` timeouts

**Pre-existing on `main`**, failing there under both Node 22.x and 24.x in a run that predates this branch's fix. Not caused by the bumps.

The fixture helper opened a connection and committed per row, so a test filling the read limit asked for that many file syncs. Measured locally, the connection was not the cost — the per-row commit was, and batching the loop into one transaction on one connection leaves the inserts costing essentially nothing. Exactly the two specs that loop were the two that failed.

Same rows in the same order (the query orders by `turn_index`). The three mutations these specs exist to catch — the truncation off-by-one, an unbounded result, keeping the oldest turns — still turn them red, so the rewrite is not vacuous.

### Verification

`format` / `lint:ci` / `typecheck` / `build` / full `test` green locally, plus a boot check of this checkout serving the built UI on a spare port. Windows itself has no local equivalent here, so CI is the ground truth for that half — and it is now green.

Note for a separate look: `docs/facts.json` is committed in a non-Prettier shape on `main`, so `yarn format` rewrites it. Left alone here to keep this diff to the fixes.

Issue #2090, opened 2026-09-16
bug: startup logs "already running" then hangs forever when a stale ~/.mulmoterminal/instances/<pid>.json points at a reused PID (Windows)


ytakahashi1981:
## What happened
On Windows, starting mulmoterminal (`npx mulmoterminal@latest --port <port> --no-open`) as a
non-interactive/headless child process (no TTY attached to stdin — spawned via a wrapper that
redirects stdout/stderr to a log file and creates no console window) sometimes prints:

    MulmoTerminal is already running (http://localhost:<port>).
      Running more than one is NOT a supported setup: they share ~/.mulmoterminal,
      so they can overwrite each other's session state.
      To stop the running one instead:  npx mulmoterminal@latest stop

...and then never proceeds. No further output is logged (no `[bind] also listening...`, no ready
banner), the target port is never actually bound, and the node process for `mulmoterminal.js` stays
alive indefinitely (observed for 10+ minutes) until it is killed manually. This happened on two
separate days.

In both cases, `~/.mulmoterminal/instances/<pid>.json` contained an entry whose `pid` no longer
belonged to any mulmoterminal process — Windows had reused that PID for an unrelated process
(`svchost.exe` the first time, `csrss.exe` the second time). Killing the hung process tree and
moving the stale `instances/<pid>.json` file aside let the next start succeed normally within
seconds, both times.

### What I've confirmed by reading the installed source (mulmoterminal 4.25.0)

- `bin/instances.js`, `isProcessAlive(pid)`:
  ```js
  export function isProcessAlive(pid) {
    if (!Number.isInteger(pid) || pid <= 0) return false;
    try {
      process.kill(pid, 0);
      return true;
    } catch (err) {
      return err?.code === "EPERM";
    }
  }
  ```
  This only checks whether *some* process currently holds that PID — it does not verify the PID
  still belongs to a mulmoterminal process. If the OS reuses the PID after the original process
  exited, `liveInstances()` will treat the stale entry as a live peer.
- `bin/mulmoterminal.js`, `confirmNoRunningInstance()`: when `!process.stdin.isTTY` (our case), it
  logs the message shown above and the `SECOND_INSTANCE_NOTE`, then `return`s — it does not call
  `process.exit()` and does not block on a prompt in this branch.

### What I have NOT confirmed
Because that non-TTY branch itself returns without blocking, I have not pinned down exactly which
later step is the one that never completes (port binding, or another `liveInstances()` read used
[...]

Issue #1061, opened 2026-07-29
bug(windows): a second instance deletes the RUNNING instance's session settings — surviving is always empty without tmux


claude-lab-japan:
## Summary

On a host without tmux (i.e. **every Windows host**), starting a second server instance deletes the **running** instance's `~/.mulmoterminal/settings/*.json` — the settings of sessions that are alive and in use.

`surviving` is derived from tmux, and `SETTINGS_DIR` is hardcoded to `os.homedir()`, so the second instance sees every settings file as an orphan and removes it.

```ts
// server/index.ts (inside server.listen callback)
const surviving = tmuxAvailable() ? tmuxListSessionIds() : [];   // Windows: always []
...
const droppedSettings = pruneOrphanSettings(new Set(surviving)); // -> deletes ALL of them
```

```ts
// server/session/session-settings.ts
const SETTINGS_DIR = path.join(os.homedir(), ".mulmoterminal", "settings");
```

`MULMOTERMINAL_HOME` in `server/config/env.ts` is also a plain `path.join(os.homedir(), ...)` constant, not an env override, so there is no way to give a second instance its own state directory.

## Why this is a bug rather than "don't run two instances"

Running a second instance is a supported flow, and shared-state damage in this exact area has already been fixed twice:

- `--port <n>` is a documented CLI option (README).
- ChangeLog: **"Ask before a second instance (#653): starting a second instance now prompts instead of refusing."**
- ChangeLog: **"Cross-process staleness in shared files (#672, #705): two servers sharing one `~/.mulmoterminal` no longer drop each other's attention state, and a non-owning server re-reads a session's tool history instead of showing a stale copy until restart."**

So two servers sharing one `~/.mulmoterminal` is explicitly accounted for — `settings/` looks like the one place that was missed.

## Reproduction

Unit-level, against a throwaway directory (never the real `~/.mulmoterminal`). This passes `new Set([])`, which is exactly what `index.ts` passes on a host without tmux:

```ts
import { pruneOrphanSettings } from "./server/session/session-settings.js";

const dir = path.join(os.tmpdir(), "prune-repro-" + process.pid);
mkdirSync(dir, { recursive: true });
for (const id of [
  "11111111-1111-4111-8111-111111111111",
  "22222222-2222-4222-8222-222222222222",
  "33333333-3333-4333-8333-333333333333",
]) writeFileSync(path.join(dir, id + ".json"), '{"live":true}');

[...]

PR #2092, opened 2026-09-16
fix: a reused pid is not still us — confirm a registry entry before believing it (#2090)

isamu:
## Summary

**Stacked on #2091** — this targets `fix/2090-unanswerable-prompt`, because both touch
`confirmNoRunningInstance`. Retarget to `main` once #2091 lands.

The other half of #2090, and the half that made the reported machine's problem permanent.

`~/.mulmoterminal/instances/<pid>.json` survives a hard kill by design — `process.on("exit", …)`
never runs, and the registry's rule is that readers clean up. The rule they clean up by is
`isProcessAlive(pid)`, which answers only that **some** process holds that number. The OS hands it
out again once the original is gone, so a leftover entry becomes a peer that never dies. On the
reported Windows machine it was `svchost.exe` the first time and `csrss.exe` the second, and every
start afterwards was told one was already running.

**The repository had already said this, twice, and fixed it once.** `bin/port-owner.js`: *"the pid
is alive — says nothing; a crashed server leaves its file and pids get reused"*. `bin/stop.js`:
*"A LIVE PID IS NOT AN IDENTITY … The registry has always tolerated that, because its only reader
asked a harmless question ('is one already running?')"*. #1820 answered it for signalling and left
the registry itself alone, on that last clause. #2090 is the disproof — with an interactive prompt
behind it, the harmless question stopped the launch for good.

`servingInstances()` asks the kernel who owns the port the entry registered, and **deletes the ones
it can positively disprove**. The deletion is what carries the fix past the launcher: the settings
prune cutoff in `server/index.ts` and the two hook posters read the registry from synchronous paths
and cannot ask, but they are right again from the next read.

**There is no boot race to lose.** `registerInstance` is called from inside the server's `listen`
callback, so the port is already bound the moment the entry exists — an entry whose port has no
owner cannot be a peer that is still starting. `[]` is a real answer here, not a maybe.

`liveInstances()` is unchanged and still synchronous; `server/agents/{copilot,cursor}-hooks-file.ts`
call it from sync paths and making it async is a different change.

Plan: `plans/fix-2090-registry-identity.md`.

## Items to Confirm / Review

- **`null` keeps the entry — the OPPOSITE of `stop.js`'s `confirmInstance`, on purpose.**
  `portOwners` distinguishes "could not ask" (null) from "nobody is there" (`[]`). For `stop` the
  cost of guessing is signalling a stranger's process, so it fails closed; here the cost is losing
  the #1061 second-instance warning on a machine with no `lsof`, so it fails open. Two policies for
  one question — please confirm that reads right, and that the comments make it hard to "fix" later
  by unifying them. A mutation collapsing the two turns the spec red.
- **`LOOKUP_TIMEOUT_MS` is raised for every caller of `portOwners`, including `stop`.** The cap was
  sized for `lsof`. On GitHub's Windows runners — which disable Defender's realtime scanning before
  testing — the real-socket spec's single PowerShell call used most of the old cap on every recent
  run (`gh run view <id> --log | grep port-owner.spec`), and a user's machine scans. Hitting the cap
  returns `null`, which this PR reads as "cannot disprove" and **keeps the stale entry** — the
  reported Windows case unfixed — and which `stop` reports as unconfirmed. Raised in `port-owner.js`
  rather than at this call site because both symptoms share the cause. The cost is paid only while
  an entry exists, and a stale one is deleted after its first successful lookup.
- **The real-socket spec now fails on `null` on win32.** It skipped silently on `null` everywhere,
  so a green Windows CI never showed that PowerShell answered. PowerShell always exists there, so
  `null` is a timeout or a failure — the thing that would leave an entry standing.
- **A registry file whose name disagrees with the pid inside it is no longer reported** (Codex,
  round 1). `servingInstances` deletes `<pid>.json`, so disproving a mismatched `111.json` that names
  a live server's pid deleted **that server's own entry** — reproduced with real files. No writer
  produces a mismatch, so `liveInstances` now skips such a file (and leaves it on disk) rather than
  let any by-pid reader act on it — `stop` signals by pid too. Side effect worth knowing: the
  report's own repro recipe (hand-editing a leftover file's `pid`) now simply stops warning.
- **`portOwners` returns `null` for every lookup that did not answer, not just a missing binary**
  (Claude, round 1). Its comment already said spawn failures are told apart by their error code, but
  the code checked only `ENOENT`: an `EACCES` — a `powershell.exe` policy forbids running — came back
  `[]`, which this PR deletes a live entry over (reproduced with a non-executable `lsof`). Now a string
  errno, a kill, a signal, or a non-zero exit that wrote to stderr is `null`; lsof's silent exit 1 is
  still `[]`. `stop` is unaffected (it treats both as unconfirmed). Cost: a host whose `lsof` warns on
  stderr even with no match keeps a stale entry, which is what `main` does today.
- **A real-OS spec now asks for a port nobody holds and expects `[]`**, failing on `null` on win32 —
  the only evidence that PowerShell's no-match is an answer, which the deletion depends on.
- **Why not `canBind`**, which would need no subprocess: it probes `BIND_HOST` only, so a peer
  started with a different `MULMOTERMINAL_HOST` would read as dead and #1061 would come back.
  `portOwners` asks by port number and is address-agnostic.
- **Cost**: one `lsof` / `Get-NetTCPConnection` per entry, and only once the registry is non-empty.
  The usual case is zero entries and zero lookups. Windows pays a PowerShell spawn (capped by
  `LOOKUP_TIMEOUT_MS` in `port-owner.js`) on the path that is about to spawn a whole server.
- **Not covered inside a single boot**: `server/index.ts`'s prune cutoff and the two hook posters
  still read unconfirmed entries during the boot they run in — they are healed by the file being
  removed, not by their own check. Making the `listen` callback async to await a confirmation there
  would reorder a long boot sequence for a much smaller payoff. A `yarn dev`-only developer never
  goes through the launcher, so that one case is not reached at all.
- **`forgetFile` takes the path, not the pid.** `liveInstances` removes the file it actually read,
  which need not be `<pid>.json`; deriving the name from the pid would have quietly changed that.

## Verification

- `yarn format` / `yarn lint` / `yarn typecheck` clean. Full `yarn test` green (890 files).
- Five new specs in `test/bin/instances.spec.ts`: owner matches, a different pid owns it, nobody
  owns it, the kernel could not be asked, and an entry with no port — each asserting whether the
  file survived, not just what was returned.
- Round-1 fixes break-verified: removing the name check reds two specs (the mismatch is not reported;
  the live entry survives servingInstances), restoring the `ENOENT`-only check reds the `EACCES` spec,
  and ignoring stderr reds the failed-lookup spec. Both original repros rerun against real files and
  a real non-executable `lsof`: the live entry survives, and the lookup is `null`.
- **Mutation-checked**, because a spec that asserts nothing is the failure mode here: making
  `servingInstances` believe every entry (the pre-fix behaviour) turns it red, and so does
  collapsing `null` into `[]`. The source was restored and diffed against a pristine copy after.
- Live, against a real `instances/<pid>.json` and a real listener, all three directions:
  - an entry naming a reused pid (a live process that does not own the port) — no "already
    running", **and the file is gone**;
  - an entry naming the pid that genuinely owns the port — still reported, file kept (#1061 intact);
  - the same, with `lsof` removed from `PATH` — still reported, file kept (fail-open).
- Not exercised live: a real bind afterwards. Letting the launcher reach `runServer` boots a second
  server whose start-up idle sweep runs against the live `mt-*` tmux sessions on this machine.

## User Prompt

- GitHub issue receptron/mulmoterminal#2090 を調査してほしい。
- （調査結果を受けて）直してほしい。
- D1（PID 再利用で幽霊エントリが残る件）の直し方は「ポート所有者で確認」を採用。
- PR は2本に分ける。

work in mulmoterminal4

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved detection of active terminal instances when process IDs are reused.
  * Prevented stale or mismatched registry entries from being treated as running instances.
  * Improved port-owner checks across Windows and slower environments.
  * Distinguished unavailable port checks from ports confirmed to have no owner.

* **Tests**
  * Added coverage for stale registry entries, reused process IDs, and port lookup failures.


isamu:
## Codex cross-review — tier C

**Tier C**, because the change deletes registry files (a destructive operation whose one forbidden outcome is erasing a live peer), puts a subprocess on the launcher's start-up path, and chooses a fail-open policy that is deliberately the opposite of `stop.js`.

Axes named per round in addition to the default six: can a live peer's entry be deleted (including a launcher racing a starting server); fail-open vs fail-closed on `null`; subprocess cost and timeout on start-up; cross-platform `portOwners` (win32 PowerShell, missing `lsof`, Linux); what the synchronous readers still see within one boot.

Stacked on #2091 — reviewed after #2091's loop, in the same checkout, for the same load reason stated there.

isamu:
- P2 should fix: `servingInstances()` can delete a live server's canonical registry file when the stale file it is disproving has JSON whose `pid` does not match its filename.

Finding details: Site: `bin/instances.js:144` is the only occurrence in this PR (`rg "entryFile\\(|forgetFile\\(" bin/instances.js`). `liveInstances()` reads an arbitrary `*.json` file and returns only the parsed `{ pid, port, startedAt }`; `servingInstances()` then deletes `entryFile(instance.pid)` instead of the file that produced that entry. A stale or malformed `111.json` containing `{ "pid": 222, "port": 34567 }` can pass `isProcessAlive(222)`, fail the port-owner check for `34567`, and then remove `222.json`, which may be the live server's real entry. It also leaves the actual bad file behind, so the next launcher can repeat the damage. Smallest fix: keep the reader-held path through the confirmation step and delete that exact path, or reject/ignore entries whose basename does not match the parsed pid before they reach `servingInstances()`. Add a regression test with mismatched filename/content plus a real canonical file for the same pid. What would change my mind: a hard invariant, enforced before this code runs and across older/newer launchers, that every readable registry filename always equals its JSON pid and cannot be stale/corrupt/malicious.

| Axis | Result |
|---|---|
| correctness and edge cases | Finding above: mismatched filename/content can delete the wrong entry. Other checked edge cases are covered: matching owner kept, different owner/empty owner dropped, null kept, no port kept. |
| security | Finding above is also the security/destructive concern: deletion target is derived from registry JSON content, not the path being disproved. PowerShell/lsof command construction is otherwise okay here because the port is parsed as an integer before this path calls `portOwners()`, and `execFile` avoids shell interpolation. |
| tests: does a test go red if the new behaviour breaks | Mostly yes for the intended behaviours, and local `yarn vitest run test/bin/` passed (423 tests). Missing red test for the mismatched filename/content deletion case above. |
| API / on-disk compatibility | No finding on the registry schema itself; no on-disk format change. The mismatch case is the compatibility risk because old/corrupt/manual files are still readable but can now target another entry's canonical filename. |
| consistency with the rest of the codebase | No finding besides the local inconsistency between the new `forgetFile` comment saying it deletes the reader-held path and `servingInstances()` reconstructing the path from pid. Hand-written `.d.ts` is updated. |
| accessibility and i18n lockstep | Not applicable; launcher/registry backend change only. |
| [C] DESTRUCTIVE: can a LIVE server's entry ever be deleted? | Yes, via the filename/content mismatch described above. For the requested race cases otherwise: listen-before-register makes `[]` safe for normal entries; fail-open on null avoids deleting when lsof cannot see another user's/sandboxed process; different socket owner correctly disproves the pid/port pair; extra loopback listeners and `MULMOTERMINAL_HOST` do not matter because lookup is by port ownership; recorded-port drift correctly disproves the entry, but only safe if deleting the actual file read. |
| [C] null policy | No finding. Fail-open is right for launcher warning preservation, and `portOwners()` returns null for ENOENT, timeout/killed/signal, thrown runner, and unsupported spawn failure paths; lsof no-match stays `[]`. |
| [C] start-up cost and timeout | No finding. Sequential lookup cost is bounded and only paid for existing live-by-pid entries; raising the shared timeout can make `stop` wait longer on indeterminate lookup, but that is materially preferable to silently making the Windows report unfixed. |
| [C] cross-platform | No finding from this diff. PowerShell is invoked via `execFile`; parsing numeric lines is appropriate; macOS/Linux missing `lsof` remains null/fail-open here. Local macOS bin tests/typecheck passed. Current Windows run for `b18defe3` was still in progress when I checked; the previous branch run showed `test/bin/port-owner.spec.ts` passing in 2324ms, which supports the timeout concern but does not by itself prove the latest commit. |
| [C] synchronous readers within a boot | No finding. `server/index.ts` and the hook posters still read `liveInstances()` synchronously and can see stale entries until the launcher heals them; nothing in this PR makes their same-boot behaviour worse, but the healing claim depends on deleting the actual stale file, which the finding above currently breaks for mismatched filenames. |

isamu:
## Codex exchange + triage — iteration 1

<details><summary>Prompt (verbatim)</summary>

```text
Review PR #2092 at https://github.com/receptron/mulmoterminal/pull/2092 (branch fix/2090-registry-identity, head b18defe3). It is STACKED on #2091: its base branch is fix/2090-unanswerable-prompt at 18953f5c. The working tree is checked out at b18defe3. This is ROUND 1.

Review ONLY this PR's own diff: `git diff 18953f5c...HEAD` (the merge commit f05441ef brings in #2091, which was reviewed separately and converged — do not review prompt-yes-no.js here).

Context: issue #2090, the other half. ~/.mulmoterminal/instances/<pid>.json survives a hard kill; readers drop an entry only when isProcessAlive(pid) is false, and a REUSED pid (svchost.exe, csrss.exe on the reporter's Windows machine) made a dead server look alive forever. The PR adds servingInstances() in bin/instances.js, which asks portOwners() (bin/port-owner.js: lsof / PowerShell Get-NetTCPConnection) whether the entry's pid still owns its port, deletes entries it can positively disprove, and KEEPS entries when the lookup returns null (deliberately the opposite of bin/stop.js confirmInstance, which fails closed). The launcher's confirmNoRunningInstance uses it. A pre-round commit raised LOOKUP_TIMEOUT_MS for Windows and made the real-socket spec fail on null on win32. Plan: plans/fix-2090-registry-identity.md. Read the PR body (gh pr view 2092).

Every finding you hold back costs a full extra round: a fix is reviewed for the first time in the round AFTER it is pushed. Read the WHOLE diff before you post anything. Do not stop at the first problem. Do not save anything for a follow-up review.

Post inline comments via: gh api repos/receptron/mulmoterminal/pulls/2092/comments
or top-level via: gh pr comment 2092 --repo receptron/mulmoterminal
The PR is authored by the same account you post as, so a formal request-changes review will be refused; use comments.

── ENVIRONMENT ──────────────────────────────────────────────────────────────
Pre-flight (row 0): you can run test/bin/, typecheck, bind sockets, run lsof, and reach GitHub. The literal `rm -f` is rejected by your command guard, so do NOT run a mutation sweep that needs rm to restore — name any mutation you want and it will be run for you.
The machine is heavily loaded: run `yarn vitest run test/bin/` and `yarn typecheck`, NOT the full `yarn test`. Do NOT boot the real server (`yarn dev`, or letting bin/mulmoterminal.js reach runServer on a free port): its start-up idle sweep acts on the live tmux sessions on this machine. You MAY run the launcher with HOME pointed at a scratch directory as long as the requested port is already held by a listener. There is no Windows machine here; Windows CI logs are readable with `gh run view <id> --log` (workflows windows-pr.yaml / windows-daily.yaml).

  correctness and edge cases
  security (deleting files by a path derived from registry content; command construction in portOwners)
  tests: does a test go red if the new behaviour breaks
  API / on-disk compatibility (the registry format, older/newer launchers and servers sharing one ~/.mulmoterminal)
  consistency with the rest of the codebase (bin/ conventions, hand-written .d.ts, comments, CLAUDE.md rules)
  accessibility and i18n lockstep (say whether it applies)
  [C] DESTRUCTIVE: can a LIVE server's entry ever be deleted? Consider at least: a launcher reading while a server is between listen and registerInstance or mid-shutdown; lsof not seeing another user's process or a sandboxed one; a server whose listening socket is held by a different pid; the server's extra loopback listeners (server/infra/announce-listening.ts, loopbackServers); MULMOTERMINAL_HOST binds; an entry whose port the registry recorded differently from what was bound.
  [C] null policy: is fail-open right here, and is every path that yields null (ENOENT, timeout, signal, a throwing runner, an unsupported platform) actually null and not []?
  [C] start-up cost and timeout: servingInstances runs sequentially per entry on the launcher path; the raised LOOKUP_TIMEOUT_MS also applies to `mulmoterminal stop`. Is anything made materially worse?
  [C] cross-platform: win32 PowerShell (quoting, output parsing, a timeout), macOS/Linux with and without lsof, WSL. Is the Windows CI evidence for the timeout change sound, and can the new win32-only spec assertion go red for an environment reason rather than a real defect?
  [C] synchronous readers within a boot: server/index.ts prune cutoff and server/agents/{copilot,cursor}-hooks-file.ts still read unconfirmed entries. Is the PR's "healed by deletion on the next read" claim true, and does anything get worse for them?

2. EVERY SITE. If the same mistake appears elsewhere in the repo, list every occurrence NOW — grep for it.

| 0 | 0 | sandbox pre-flight (run on the #2091 head, same checkout) | — | tests (yarn vitest run test/bin/) exit 0; yarn typecheck exit 0; socket bind exit 0; api.github.com 200; gh reaches GitHub; literal `rm -f` REJECTED by the command guard (policy) — no rm-based mutation restore |
| 1 | 0 | [Claude, before round 1] P2: portOwners' LOOKUP_TIMEOUT_MS was sized for lsof; on Windows CI (Defender scanning OFF) the one real PowerShell call used most of it on every recent run, so on a user's machine a timeout → null → servingInstances keeps the stale entry (reported case unfixed) and stop reports unconfirmed. Also the real-socket spec skipped on null everywhere, so green Windows CI never proved PowerShell answered | FIXED | b18defe3: cap raised in port-owner.js (class: launcher + stop); spec fails on null when win32. Evidence: `gh run view <id> --log \| grep port-owner.spec` on windows-daily/windows-pr runs. NEVER REVIEWED — hold to the same bar |

That table is context, NOT a list of closed topics: if a resolution is wrong, say 'REOPENING #<row>' with evidence. Row 1 was written before this round and has never been reviewed.

Decisions the user already made (challenge only with evidence): the confirmation uses portOwners (kernel port ownership), not a canBind probe and not a heartbeat; the fix is split across #2091 (prompt deadline) and this PR; liveInstances() stays synchronous.

  Then, on its own line:

Also print that same final comment to stdout at the end, verbatim, so it survives if posting fails.

```

</details>

<details><summary>Codex stdout (verbatim)</summary>

```text
- P2 should fix: `servingInstances()` can delete a live server's canonical registry file when the stale file it is disproving has JSON whose `pid` does not match its filename.

Finding details: Site: `bin/instances.js:144` is the only occurrence in this PR (`rg "entryFile\\(|forgetFile\\(" bin/instances.js`). `liveInstances()` reads an arbitrary `*.json` file and returns only the parsed `{ pid, port, startedAt }`; `servingInstances()` then deletes `entryFile(instance.pid)` instead of the file that produced that entry. A stale or malformed `111.json` containing `{ "pid": 222, "port": 34567 }` can pass `isProcessAlive(222)`, fail the port-owner check for `34567`, and then remove `222.json`, which may be the live server's real entry. It also leaves the actual bad file behind, so the next launcher can repeat the damage. Smallest fix: keep the reader-held path through the confirmation step and delete that exact path, or reject/ignore entries whose basename does not match the parsed pid before they reach `servingInstances()`. Add a regression test with mismatched filename/content plus a real canonical file for the same pid. What would change my mind: a hard invariant, enforced before this code runs and across older/newer launchers, that every readable registry filename always equals its JSON pid and cannot be stale/corrupt/malicious.

| Axis | Result |
|---|---|
| correctness and edge cases | Finding above: mismatched filename/content can delete the wrong entry. Other checked edge cases are covered: matching owner kept, different owner/empty owner dropped, null kept, no port kept. |
| security | Finding above is also the security/destructive concern: deletion target is derived from registry JSON content, not the path being disproved. PowerShell/lsof command construction is otherwise okay here because the port is parsed as an integer before this path calls `portOwners()`, and `execFile` avoids shell interpolation. |
| tests: does a test go red if the new behaviour breaks | Mostly yes for the intended behaviours, and local `yarn vitest run test/bin/` passed (423 tests). Missing red test for the mismatched filename/content deletion case above. |
| API / on-disk compatibility | No finding on the registry schema itself; no on-disk format change. The mismatch case is the compatibility risk because old/corrupt/manual files are still readable but can now target another entry's canonical filename. |
| consistency with the rest of the codebase | No finding besides the local inconsistency between the new `forgetFile` comment saying it deletes the reader-held path and `servingInstances()` reconstructing the path from pid. Hand-written `.d.ts` is updated. |
| accessibility and i18n lockstep | Not applicable; launcher/registry backend change only. |
| [C] DESTRUCTIVE: can a LIVE server's entry ever be deleted? | Yes, via the filename/content mismatch described above. For the requested race cases otherwise: listen-before-register makes `[]` safe for normal entries; fail-open on null avoids deleting when lsof cannot see another user's/sandboxed process; different socket owner correctly disproves the pid/port pair; extra loopback listeners and `MULMOTERMINAL_HOST` do not matter because lookup is by port ownership; recorded-port drift correctly disproves the entry, but only safe if deleting the actual file read. |
| [C] null policy | No finding. Fail-open is right for launcher warning preservation, and `portOwners()` returns null for ENOENT, timeout/killed/signal, thrown runner, and unsupported spawn failure paths; lsof no-match stays `[]`. |
| [C] start-up cost and timeout | No finding. Sequential lookup cost is bounded and only paid for existing live-by-pid entries; raising the shared timeout can make `stop` wait longer on indeterminate lookup, but that is materially preferable to silently making the Windows report unfixed. |
| [C] cross-platform | No finding from this diff. PowerShell is invoked via `execFile`; parsing numeric lines is appropriate; macOS/Linux missing `lsof` remains null/fail-open here. Local macOS bin tests/typecheck passed. Current Windows run for `b18defe3` was still in progress when I checked; the previous branch run showed `test/bin/port-owner.spec.ts` passing in 2324ms, which supports the timeout concern but does not by itself prove the latest commit. |
| [C] synchronous readers within a boot | No finding. `server/index.ts` and the hook posters still read `liveInstances()` synchronously and can see stale entries until the launcher heals them; nothing in this PR makes their same-boot behaviour worse, but the healing claim depends on deleting the actual stale file, which the finding above currently breaks for mismatched filenames. |

```

</details>

Verdict comment: https://github.com/receptron/mulmoterminal/pull/2092#issuecomment-5694243042

### Triage — not clean, fixes pushed in `d84d0f94`

| # | from | sev | finding | disposition |
|---|---|---|---|---|
| 1 | Codex | P2 | `servingInstances` deleted `<pid>.json` rather than the file read; a mismatched `111.json` naming a live pid deleted the live server's entry and survived itself. **Reproduced with real files** (live entry deleted). | FIXED: `liveInstances` skips — neither reports nor deletes — a file whose name is not `<its pid>.json`. No writer produces one, and every by-pid reader (`servingInstances` deleting, `stop` signalling) is kept off another entry. Rerun: live entry survives. Two specs; removing the check reds both. |
| 2 | **Claude, not Codex** | P2 | `portOwners` returned `null` only for `ENOENT`. `EACCES` — a `powershell.exe` policy forbids — and a lookup that ran and failed came back `[]`, which this PR deletes a **live** entry over. Its own comment already stated the rule the code did not implement. **Reproduced** with a non-executable `lsof` on PATH. The round's null-policy axis answered "none". | FIXED: a string errno, a kill, a signal, or a non-zero exit that wrote to stderr is `null`; lsof's silent exit 1 is still `[]`. `stop` unaffected. Cost: a host whose lsof warns on stderr with no match keeps a stale entry, as `main` does. Two specs, each red under its mutation. |
| 3 | Claude | test | Nothing proved PowerShell's no-match comes back as an answer, which the deletion depends on. | ADDED: real-OS spec asks a free port and expects `[]`, failing on `null` on win32. Only macOS verified locally — `test_windows` is the evidence. |

No declined findings, so no in-round rebuttal call. Local: `yarn lint`, `yarn typecheck`, `test/bin/` and the three `test/server/agents` specs touching the registry green. Round 2 dispatched on `d84d0f94` immediately.

isamu:
CODEX VERDICT: LGTM

| Axis | Result |
|---|---|
| correctness and edge cases | none. `liveInstances()` now filters unreadable/invalid/dead/mismatched entries before any pid-derived action, and `servingInstances()` only drops entries on a positive `number[]` answer that does not contain the registered pid. Entries with `port: null` remain fail-open. |
| security | none. Deletion is constrained to `<instancesDir>/<integer pid>.json` after `liveInstances()` has enforced filename/pid agreement for the launcher path. `portOwnerCommand()` receives an integer port from parsed registry content; no shell is used for `execFile`, and the PowerShell string interpolation is numeric-only. |
| tests: does a test go red if the new behaviour breaks | none. The new specs cover owner match, other-owner, no-owner deletion, lookup-null preservation, null port preservation, filename/pid mismatch not being reported, and the live-entry-survives regression. `port-owner.spec.ts` covers ENOENT, timeout/killed, EACCES/string errno, stderr failure, silent no-match, real no-match, and real owner. Locally: `yarn vitest run test/bin/` passed 16 files / 428 tests; `yarn typecheck` passed. |
| API / on-disk compatibility | none. Registry shape stays `{ pid, port, startedAt }`; older correctly-written `<pid>.json` files are still read. Corrupt or hand-edited filename/pid mismatches are now hidden rather than acted on, which is the safer compatibility choice for pid-based readers. |
| consistency with the rest of the codebase | none. The change keeps `liveInstances()` synchronous for existing server/hook callers, adds the hand-written `.d.ts`, follows the existing best-effort cleanup style, and keeps `stop`'s fail-closed policy separate from the launcher fail-open policy. |
| accessibility and i18n lockstep | none / not applicable. This PR changes launcher/server registry confirmation and tests; no user-visible localized UI strings or accessibility surfaces are introduced. |
| [C] DESTRUCTIVE: can a LIVE server's entry ever be deleted? | none found. The entry exists only after the primary server socket is bound; mid-shutdown is either still owned, null/unknown, or already dead. If `lsof`/PowerShell cannot answer (missing, blocked, timeout, signal, throwing runner), the entry is kept. A different pid holding the port, no listener, or a pid that no longer owns its recorded port causes deletion, which is the intended stale-entry case. Extra loopback listeners and `MULMOTERMINAL_HOST` do not make a live entry disappear because the recorded primary port is already owned by the server pid; the port-only check can keep a rare reused-pid/same-port false positive, but does not delete a live server. |
| [C] null policy | none. Matrix checked: spawn errno/string `code` including ENOENT/EACCES/EPERM -> `null`; maxBuffer/string code -> `null`; timeout/killed -> `null`; signal -> `null`; throwing runner -> `null`; unsupported/non-win platform with missing `lsof` -> `null`; non-zero exit with stderr -> `null`; non-zero silent exit -> `[]` for `lsof` no-match; exit 0 with stderr still parses stdout, which is right for warnings accompanying a successful answer. This is right for both callers: launcher keeps on `null` and deletes only on `[]`; `stop` treats both `null` and `[]` as unconfirmed. |
| [C] start-up cost and timeout | none. `servingInstances()` is sequential, but it only runs after `liveInstances()` finds live pids and only on the launcher warning path. Raising the shared cap to 10s can also lengthen `mulmoterminal stop` for unanswering lookup tools, but only per registered live-pid entry and in exchange for making PowerShell usable on the reported class of Windows machine. |
| [C] cross-platform | none. POSIX uses `lsof -nP -ti tcp:<port> -sTCP:LISTEN`; WSL follows the Linux/lsof path. Windows uses `powershell.exe -NoProfile -NonInteractive` and numeric `-LocalPort`; output parsing accepts padded CRLF pid lines. The win32 real-socket spec can fail for an environment-level PowerShell/permission/timeout problem, but that is the failure mode this PR must not silently greenwash. Current-head Windows PR was still running while I reviewed; the latest completed Windows log I checked showed `test/bin/port-owner.spec.ts` passing. |
| [C] the round-1 rules themselves | none. The filename/pid check can hide a hand-edited-but-live entry, causing a missed second-instance warning or allowing server-side prune/hook repair to ignore that peer, but no writer legitimately creates such a file and the alternative is pid-derived deletion/signalling against the wrong target. The stderr rule is fail-open for failures with diagnostics; lsof/PowerShell warnings that accompany a real zero-exit answer still parse stdout, while diagnostic non-zero failures no longer become destructive `[]`. |
| [C] synchronous readers within a boot | none. `server/index.ts` prune cutoff and the copilot/cursor hook readers still call `liveInstances()` and can see unconfirmed same-boot entries until the launcher has healed/deleted the stale file. That matches the PR's claim: they are healed by deletion on a later read, not by their own confirmation. This PR does not make those sync paths more destructive; the only new hiding is for corrupt filename/pid mismatches. |

isamu:
## Codex exchange — iteration 2

<details><summary>Prompt (verbatim)</summary>

```text
Review PR #2092 at https://github.com/receptron/mulmoterminal/pull/2092 (branch fix/2090-registry-identity, head d84d0f94). It is STACKED on #2091: its base branch is fix/2090-unanswerable-prompt at 18953f5c. The working tree is checked out at d84d0f94. This is ROUND 2.

Review ONLY this PR's own diff: `git diff 18953f5c...HEAD` (the merge commit f05441ef brings in #2091, which was reviewed separately and converged — do not review prompt-yes-no.js here).

Context: issue #2090, the other half. ~/.mulmoterminal/instances/<pid>.json survives a hard kill; readers drop an entry only when isProcessAlive(pid) is false, and a REUSED pid (svchost.exe, csrss.exe on the reporter's Windows machine) made a dead server look alive forever. The PR adds servingInstances() in bin/instances.js, which asks portOwners() (bin/port-owner.js: lsof / PowerShell Get-NetTCPConnection) whether the entry's pid still owns its port, deletes entries it can positively disprove, and KEEPS entries when the lookup returns null (deliberately the opposite of bin/stop.js confirmInstance, which fails closed). The launcher's confirmNoRunningInstance uses it. A pre-round commit raised LOOKUP_TIMEOUT_MS for Windows and made the real-socket spec fail on null on win32. Plan: plans/fix-2090-registry-identity.md. Read the PR body (gh pr view 2092).

Every finding you hold back costs a full extra round: a fix is reviewed for the first time in the round AFTER it is pushed. Read the WHOLE diff before you post anything. Do not stop at the first problem. Do not save anything for a follow-up review.

Post inline comments via: gh api repos/receptron/mulmoterminal/pulls/2092/comments
or top-level via: gh pr comment 2092 --repo receptron/mulmoterminal
The PR is authored by the same account you post as, so a formal request-changes review will be refused; use comments.

── ENVIRONMENT ──────────────────────────────────────────────────────────────
Pre-flight (row 0): you can run test/bin/, typecheck, bind sockets, run lsof, and reach GitHub. The literal `rm -f` is rejected by your command guard, so do NOT run a mutation sweep that needs rm to restore — name any mutation you want and it will be run for you.
The machine is heavily loaded: run `yarn vitest run test/bin/` and `yarn typecheck`, NOT the full `yarn test`. Do NOT boot the real server (`yarn dev`, or letting bin/mulmoterminal.js reach runServer on a free port): its start-up idle sweep acts on the live tmux sessions on this machine. You MAY run the launcher with HOME pointed at a scratch directory as long as the requested port is already held by a listener. There is no Windows machine here; Windows CI logs are readable with `gh run view <id> --log` (workflows windows-pr.yaml / windows-daily.yaml).

── MISSED LAST ROUND ─────────────────────────────────────────────────────────
Ledger row 3 is a P2 on your round-1 null-policy axis, which you answered "none": portOwners mapped
EACCES and any non-ENOENT spawn failure to [], which servingInstances deletes over. The axis asked
"is every path that yields null actually null and not []?" and the code under the question checked
only ENOENT. This round, before anything new, re-read portOwners' error handling against each way
execFile can end (spawn errno, maxBuffer, timeout kill, signal, non-zero exit with and without
stderr, exit 0 with stderr) and say what each now yields and whether that is right for BOTH callers.

  correctness and edge cases
  security (deleting files by a path derived from registry content; command construction in portOwners)
  tests: does a test go red if the new behaviour breaks
  API / on-disk compatibility (the registry format, older/newer launchers and servers sharing one ~/.mulmoterminal)
  consistency with the rest of the codebase (bin/ conventions, hand-written .d.ts, comments, CLAUDE.md rules)
  accessibility and i18n lockstep (say whether it applies)
  [C] DESTRUCTIVE: can a LIVE server's entry ever be deleted? Consider at least: a launcher reading while a server is between listen and registerInstance or mid-shutdown; lsof not seeing another user's process or a sandboxed one; a server whose listening socket is held by a different pid; the server's extra loopback listeners (server/infra/announce-listening.ts, loopbackServers); MULMOTERMINAL_HOST binds; an entry whose port the registry recorded differently from what was bound.
  [C] null policy: is fail-open right here, and is every path that yields null (ENOENT, timeout, signal, a throwing runner, an unsupported platform) actually null and not []?
  [C] start-up cost and timeout: servingInstances runs sequentially per entry on the launcher path; the raised LOOKUP_TIMEOUT_MS also applies to `mulmoterminal stop`. Is anything made materially worse?
  [C] cross-platform: win32 PowerShell (quoting, output parsing, a timeout), macOS/Linux with and without lsof, WSL. Is the Windows CI evidence for the timeout change sound, and can the new win32-only spec assertion go red for an environment reason rather than a real defect?
  [C] the round-1 rules themselves: the filename/pid check in liveInstances (older writers? Windows filename casing? anything that legitimately writes a differently named .json into instances/? what a hidden-but-live entry costs: a missed second-instance warning, a settings prune that deletes a peer's files) and the stderr rule (lsof/PowerShell writing warnings on success or no-match; a locale that changes stderr).
  [C] synchronous readers within a boot: server/index.ts prune cutoff and server/agents/{copilot,cursor}-hooks-file.ts still read unconfirmed entries. Is the PR's "healed by deletion on the next read" claim true, and does anything get worse for them?

2. EVERY SITE. If the same mistake appears elsewhere in the repo, list every occurrence NOW — grep for it.

| 0 | 0 | sandbox pre-flight (run on the #2091 head, same checkout) | — | tests (yarn vitest run test/bin/) exit 0; yarn typecheck exit 0; socket bind exit 0; api.github.com 200; gh reaches GitHub; literal `rm -f` REJECTED by the command guard (policy) — no rm-based mutation restore |
| 1 | 0 | [Claude, before round 1] P2: portOwners' LOOKUP_TIMEOUT_MS was sized for lsof; on Windows CI (Defender scanning OFF) the one real PowerShell call used most of it on every recent run, so on a user's machine a timeout → null → servingInstances keeps the stale entry (reported case unfixed) and stop reports unconfirmed. Also the real-socket spec skipped on null everywhere, so green Windows CI never proved PowerShell answered | FIXED | b18defe3: cap raised in port-owner.js (class: launcher + stop); spec fails on null when win32. Evidence: `gh run view <id> --log \| grep port-owner.spec` on windows-daily/windows-pr runs. NEVER REVIEWED — hold to the same bar |
| 2 | 1 | [Codex] P2: servingInstances deleted entryFile(pid), not the file read — a mismatched 111.json naming a live pid on an unowned port deleted the live server's <pid>.json and survived itself | FIXED | d84d0f94: liveInstances skips (does not report, does not delete) a file whose name != `${pid}.json`, so by-pid readers (servingInstances delete, stop signal) act only on the file that pid wrote. Reproduced before (live entry DELETED) and after (SURVIVED) with real files. Specs: mismatch not reported + live entry survives servingInstances; removing the check reds both |
| 3 | 1 | [Claude, not Codex] P2: portOwners returned null only for ENOENT; EACCES (a policy-blocked powershell.exe) and a failed lookup came back [], which servingInstances deletes a LIVE entry over. Codex's null-policy axis answered "none" | FIXED | d84d0f94: lookupAnswered — string errno / killed / signal / non-zero exit with stderr → null; lsof silent exit 1 → []. Repro: non-executable lsof on PATH → [] before, null after; real lsof no-match still []. Specs EACCES + stderr-failure, each reds under its mutation. Cost: lsof hosts warning on stderr with no match keep stale entries (= main). stop unaffected (null and [] both unconfirmed) |
| 4 | 1 | [Claude] real-OS spec for no-match: free port → [] (null fails on win32) — the only evidence PowerShell's no-match is an answer | ADDED | d84d0f94, test/bin/port-owner.spec.ts. Verified on macOS only; Windows evidence arrives from test_windows CI |

That table is context, NOT a list of closed topics: if a resolution is wrong, say 'REOPENING #<row>' with evidence. Rows 2-4 were written during round 1's fix and have NEVER been reviewed — hold them to the same bar as the original diff, including whether the new stderr rule or the filename check can themselves erase a live entry, hide a live peer, or keep a stale one on a platform that matters.

Decisions the user already made (challenge only with evidence): the confirmation uses portOwners (kernel port ownership), not a canBind probe and not a heartbeat; the fix is split across #2091 (prompt deadline) and this PR; liveInstances() stays synchronous.

  Then, on its own line:

Also print that same final comment to stdout at the end, verbatim, so it survives if posting fails.

```

</details>

<details><summary>Codex stdout (verbatim)</summary>

```text
CODEX VERDICT: LGTM

| Axis | Result |
|---|---|
| correctness and edge cases | none. `liveInstances()` now filters unreadable/invalid/dead/mismatched entries before any pid-derived action, and `servingInstances()` only drops entries on a positive `number[]` answer that does not contain the registered pid. Entries with `port: null` remain fail-open. |
| security | none. Deletion is constrained to `<instancesDir>/<integer pid>.json` after `liveInstances()` has enforced filename/pid agreement for the launcher path. `portOwnerCommand()` receives an integer port from parsed registry content; no shell is used for `execFile`, and the PowerShell string interpolation is numeric-only. |
| tests: does a test go red if the new behaviour breaks | none. The new specs cover owner match, other-owner, no-owner deletion, lookup-null preservation, null port preservation, filename/pid mismatch not being reported, and the live-entry-survives regression. `port-owner.spec.ts` covers ENOENT, timeout/killed, EACCES/string errno, stderr failure, silent no-match, real no-match, and real owner. Locally: `yarn vitest run test/bin/` passed 16 files / 428 tests; `yarn typecheck` passed. |
| API / on-disk compatibility | none. Registry shape stays `{ pid, port, startedAt }`; older correctly-written `<pid>.json` files are still read. Corrupt or hand-edited filename/pid mismatches are now hidden rather than acted on, which is the safer compatibility choice for pid-based readers. |
| consistency with the rest of the codebase | none. The change keeps `liveInstances()` synchronous for existing server/hook callers, adds the hand-written `.d.ts`, follows the existing best-effort cleanup style, and keeps `stop`'s fail-closed policy separate from the launcher fail-open policy. |
| accessibility and i18n lockstep | none / not applicable. This PR changes launcher/server registry confirmation and tests; no user-visible localized UI strings or accessibility surfaces are introduced. |
| [C] DESTRUCTIVE: can a LIVE server's entry ever be deleted? | none found. The entry exists only after the primary server socket is bound; mid-shutdown is either still owned, null/unknown, or already dead. If `lsof`/PowerShell cannot answer (missing, blocked, timeout, signal, throwing runner), the entry is kept. A different pid holding the port, no listener, or a pid that no longer owns its recorded port causes deletion, which is the intended stale-entry case. Extra loopback listeners and `MULMOTERMINAL_HOST` do not make a live entry disappear because the recorded primary port is already owned by the server pid; the port-only check can keep a rare reused-pid/same-port false positive, but does not delete a live server. |
| [C] null policy | none. Matrix checked: spawn errno/string `code` including ENOENT/EACCES/EPERM -> `null`; maxBuffer/string code -> `null`; timeout/killed -> `null`; signal -> `null`; throwing runner -> `null`; unsupported/non-win platform with missing `lsof` -> `null`; non-zero exit with stderr -> `null`; non-zero silent exit -> `[]` for `lsof` no-match; exit 0 with stderr still parses stdout, which is right for warnings accompanying a successful answer. This is right for both callers: launcher keeps on `null` and deletes only on `[]`; `stop` treats both `null` and `[]` as unconfirmed. |
| [C] start-up cost and timeout | none. `servingInstances()` is sequential, but it only runs after `liveInstances()` finds live pids and only on the launcher warning path. Raising the shared cap to 10s can also lengthen `mulmoterminal stop` for unanswering lookup tools, but only per registered live-pid entry and in exchange for making PowerShell usable on the reported class of Windows machine. |
| [C] cross-platform | none. POSIX uses `lsof -nP -ti tcp:<port> -sTCP:LISTEN`; WSL follows the Linux/lsof path. Windows uses `powershell.exe -NoProfile -NonInteractive` and numeric `-LocalPort`; output parsing accepts padded CRLF pid lines. The win32 real-socket spec can fail for an environment-level PowerShell/permission/timeout problem, but that is the failure mode this PR must not silently greenwash. Current-head Windows PR was still running while I reviewed; the latest completed Windows log I checked showed `test/bin/port-owner.spec.ts` passing. |
| [C] the round-1 rules themselves | none. The filename/pid check can hide a hand-edited-but-live entry, causing a missed second-instance warning or allowing server-side prune/hook repair to ignore that peer, but no writer legitimately creates such a file and the alternative is pid-derived deletion/signalling against the wrong target. The stderr rule is fail-open for failures with diagnostics; lsof/PowerShell warnings that accompany a real zero-exit answer still parse stdout, while diagnostic non-zero failures no longer become destructive `[]`. |
| [C] synchronous readers within a boot | none. `server/index.ts` prune cutoff and the copilot/cursor hook readers still call `liveInstances()` and can see unconfirmed same-boot entries until the launcher has healed/deleted the stale file. That matches the PR's claim: they are healed by deletion on a later read, not by their own confirmation. This PR does not make those sync paths more destructive; the only new hiding is for corrupt filename/pid mismatches. |

```

</details>

Verdict comment: https://github.com/receptron/mulmoterminal/pull/2092#issuecomment-5694357625

**Claude's half is held on one axis only CI can answer.** The new real-OS spec asks PowerShell about a port nobody holds and requires `[]`. Whether PowerShell's no-match exits 0, or exits non-zero with its `#< CLIXML` progress record on stderr (a known quirk when output is redirected and a module loads for the first time), decides whether the round-1 stderr rule turns the reported Windows case back into `null` — and a kept stale entry. There is no Windows machine here; `test_windows` on `d84d0f94` is the evidence, and this round is not called clean until it is read.

isamu:
## Round 2 triage — clean on `d84d0f94`

- **Codex:** `LGTM` with `FINDINGS COMPLETE`, every axis answered, including the round-1 rules themselves and a per-ending matrix of what `portOwners` yields.
- **Claude:** re-read the whole diff on the same head; no MUST-FIX. The one axis held open in the exchange comment is now answered by `test_windows` on this exact head (run 35072881952, success). On Windows, `answers [] for a real port nobody holds` passed — and that spec fails on `null` on win32 — so PowerShell's no-match comes back as an answer, not a failure, and the round-1 stderr rule does not turn the reported case back into a kept stale entry. `names the real owner of a real socket` also passed there, and the full Windows suite is green.
- **Exit:** one clean round, tier C, nothing pushed to this branch since `d84d0f94`. Two rounds in total: round 1 found two destructive P2s (one from Codex, one from Claude — both erased a live server's entry, both reproduced), round 2 reviewed their fixes.

**One more round is owed later, and why it is not run now.** CodeRabbit raised a declaration-only fix on #2091 (the base of this stack), which is going out in #2091's round 3. This branch deliberately does not merge it yet: after #2091 lands, this PR has to be retargeted to `main` and synced, and a stacked PR gets no checks from a retarget alone — so a push is coming regardless, and a round on today's merge would be spent again on that one. That post-retarget head is the one that will merge, and it gets its own Codex round.

isamu:
CODEX VERDICT: LGTM

| Axis | Result |
|---|---|
| correctness and edge cases | none. `confirmNoRunningInstance` now runs `servingInstances(liveInstances())` before any prompt, so a disproved or hidden stale entry produces no prompt; `askYesNo` on main still composes correctly for the remaining confirmed entries. |
| security (deleting files by a path derived from registry content; command construction in portOwners) | none. Deletion is constrained to `instancesDir()/<integer pid>.json` after `liveInstances` has parsed a positive integer pid and rejected filename/pid mismatches; `portOwners` uses `execFile` with argument arrays, and the interpolated PowerShell value is an integer port from registry/launcher state, not shell-expanded user text. |
| tests: does a test go red if the new behaviour breaks | none. The new tests cover matching owner, reused pid/different owner, no owner, null lookup, null port, mismatched filename survival, and real OS no-match/owner checks. Local `yarn vitest run test/bin/` passed: 16 files, 428 tests. Local `yarn typecheck` passed. |
| API / on-disk compatibility (the registry format, older/newer launchers and servers sharing one ~/.mulmoterminal) | none. Registry shape stays `{ pid, port, startedAt }`; older readers still tolerate the files, newer readers treat missing/invalid port fail-open. Mixed old/new launchers may keep the old false warning until a new launcher heals the stale file, but no format incompatibility is introduced. |
| consistency with the rest of the codebase (bin/ conventions, hand-written .d.ts, comments, CLAUDE.md rules) | none. The change follows the existing bare-Node `bin/` pattern, updates the hand-written declarations, and keeps `stop`'s fail-closed policy separate from the launcher's deliberate fail-open policy. The merge from `d84d0f94` to `54b02036` brings only `bin/prompt-yes-no.d.ts` from `c9056ce8`'s timer generic. |
| accessibility and i18n lockstep | none / not applicable. This PR changes launcher registry filtering and CLI text paths only; no UI/a11y or localized copy surface is touched. |
| [C] DESTRUCTIVE: can a LIVE server's entry ever be deleted? | none found. The destructive path requires `portOwners(port)` to answer a non-null pid list that excludes the registered pid; entries with `port: null` and lookup failures are kept. Register happens after the main socket is listening, loopback listeners are same-process extra listeners, and host binding differences still share the same local port question. Cases where the OS command cannot answer, including missing tool, timeout, signal/kill, thrown runner, non-zero stderr failure, or unsupported platform, resolve to `null` and keep. A live server whose socket is genuinely held by a different pid would be unconfirmable by this identity rule; I found no normal MulmoTerminal path that creates that ownership split. |
| [C] null policy: is fail-open right here, and is every path that yields null actually null and not []? | none. Fail-open is the right launcher tradeoff because the consequence is keeping the second-instance warning rather than deleting a live peer; it is intentionally opposite to `stop`. `ENOENT`/other string errno, timeout/killed/signal, thrown runner, unsupported platform, and non-zero stderr failures all return `null`; silent lsof exit 1 remains `[]` as the positive no-owner answer. |
| [C] start-up cost and timeout | none. The launcher pays one lookup per live registry entry, sequentially, only after `liveInstances()` reports something. The 10s cap is a worst-case failure cap for existing entries and also affects `stop`, but it prevents the Windows PowerShell path from silently timing out and leaving the reported stale entry untouched. Stale entries are deleted after the first successful no-owner lookup. |
| [C] cross-platform: win32 PowerShell, macOS/Linux with and without lsof, WSL | none. PowerShell is invoked through `execFile` and parses line-oriented owning pids; lsof paths distinguish silent no-match from lookup failures. The completed Windows PR run on `d84d0f94` passed `test/bin/port-owner.spec.ts`, including real closed-port `[]` in 1676 ms and real owner in 819 ms; current head's Windows run was still in progress when I posted, but the only post-`d84d0f94` merge delta is the #2091 declaration-only timer generic. The win32-only null assertion could go red for a genuine environment/runner failure, but that is exactly the failure mode that would make the fix ineffective on Windows. |
| [C] the round-1 rules themselves: filename/pid check and stderr rule | none. Skipping filename/pid mismatches prevents by-pid deletion/signalling from acting on a different file/process; legitimate writers name files as `<pid>.json`, and Windows filename casing does not change decimal pid names. Hidden-but-live corrupted entries can mean a missed warning or peer-protection limitation, but that is safer than deleting/signalling another pid's entry. The stderr rule keeps lookup failures fail-open; the cost is a stale entry on hosts where a no-match command warns on stderr, which is no worse than main for this bug. |
| [C] synchronous readers within a boot | none. `server/index.ts` prune cutoff and the copilot/cursor hook repairs still read `liveInstances()` only and can see unconfirmed entries during a boot that did not pass through the launcher. The PR's narrower claim is true: once the launcher deletes a positively disproved file, those synchronous readers are healed on their next read. Nothing in this PR makes their current fail-safe behaviour worse. |

isamu:
## Codex exchange + triage — iteration 3 (after #2091 merged)

<details><summary>Prompt (verbatim)</summary>

```text
Review PR #2092 at https://github.com/receptron/mulmoterminal/pull/2092 (branch fix/2090-registry-identity, head 54b02036). #2091 (the prompt-deadline half) has MERGED into main (merge commit 81840502), and this PR has been RETARGETED to main and synced: base main 81840502, head 54b02036 (a merge of origin/main). The working tree is checked out at 54b02036. This is ROUND 3.

Review this PR's diff against main: `git diff 81840502...HEAD`. prompt-yes-no.js is on main now and was reviewed in #2091; review it here only where this PR's code interacts with it (confirmNoRunningInstance).

Context: issue #2090, the other half. ~/.mulmoterminal/instances/<pid>.json survives a hard kill; readers drop an entry only when isProcessAlive(pid) is false, and a REUSED pid (svchost.exe, csrss.exe on the reporter's Windows machine) made a dead server look alive forever. The PR adds servingInstances() in bin/instances.js, which asks portOwners() (bin/port-owner.js: lsof / PowerShell Get-NetTCPConnection) whether the entry's pid still owns its port, deletes entries it can positively disprove, and KEEPS entries when the lookup returns null (deliberately the opposite of bin/stop.js confirmInstance, which fails closed). The launcher's confirmNoRunningInstance uses it. A pre-round commit raised LOOKUP_TIMEOUT_MS for Windows and made the real-socket spec fail on null on win32. Plan: plans/fix-2090-registry-identity.md. Read the PR body (gh pr view 2092).

Every finding you hold back costs a full extra round: a fix is reviewed for the first time in the round AFTER it is pushed. Read the WHOLE diff before you post anything. Do not stop at the first problem. Do not save anything for a follow-up review.

Post inline comments via: gh api repos/receptron/mulmoterminal/pulls/2092/comments
or top-level via: gh pr comment 2092 --repo receptron/mulmoterminal
The PR is authored by the same account you post as, so a formal request-changes review will be refused; use comments.

── ENVIRONMENT ──────────────────────────────────────────────────────────────
Pre-flight (row 0): you can run test/bin/, typecheck, bind sockets, run lsof, and reach GitHub. The literal `rm -f` is rejected by your command guard, so do NOT run a mutation sweep that needs rm to restore — name any mutation you want and it will be run for you.
The machine is heavily loaded: run `yarn vitest run test/bin/` and `yarn typecheck`, NOT the full `yarn test`. Do NOT boot the real server (`yarn dev`, or letting bin/mulmoterminal.js reach runServer on a free port): its start-up idle sweep acts on the live tmux sessions on this machine. You MAY run the launcher with HOME pointed at a scratch directory as long as the requested port is already held by a listener. There is no Windows machine here; Windows CI logs are readable with `gh run view <id> --log` (workflows windows-pr.yaml / windows-daily.yaml).

── WHAT CHANGED SINCE YOUR ROUND-2 LGTM ─────────────────────────────────────
Round 2 on d84d0f94 was clean (and test_windows on that head passed, including the win32-only
no-match spec). Since then: no change to this PR's own code. #2091 merged into main with one extra
commit you have not seen on this branch — c9056ce8, a declaration-only generic for askYesNo's timer
handle — and 54b02036 merges main in. This round has to cover the head that will merge. Re-read the
whole PR diff against main, confirm the merge brought nothing but c9056ce8 and #2091's merge commit,
and check that confirmNoRunningInstance on this head still composes correctly with askYesNo as it is
on main (servingInstances runs before the prompt; a hidden or deleted entry means no prompt at all).

  correctness and edge cases
  security (deleting files by a path derived from registry content; command construction in portOwners)
  tests: does a test go red if the new behaviour breaks
  API / on-disk compatibility (the registry format, older/newer launchers and servers sharing one ~/.mulmoterminal)
  consistency with the rest of the codebase (bin/ conventions, hand-written .d.ts, comments, CLAUDE.md rules)
  accessibility and i18n lockstep (say whether it applies)
  [C] DESTRUCTIVE: can a LIVE server's entry ever be deleted? Consider at least: a launcher reading while a server is between listen and registerInstance or mid-shutdown; lsof not seeing another user's process or a sandboxed one; a server whose listening socket is held by a different pid; the server's extra loopback listeners (server/infra/announce-listening.ts, loopbackServers); MULMOTERMINAL_HOST binds; an entry whose port the registry recorded differently from what was bound.
  [C] null policy: is fail-open right here, and is every path that yields null (ENOENT, timeout, signal, a throwing runner, an unsupported platform) actually null and not []?
  [C] start-up cost and timeout: servingInstances runs sequentially per entry on the launcher path; the raised LOOKUP_TIMEOUT_MS also applies to `mulmoterminal stop`. Is anything made materially worse?
  [C] cross-platform: win32 PowerShell (quoting, output parsing, a timeout), macOS/Linux with and without lsof, WSL. Is the Windows CI evidence for the timeout change sound, and can the new win32-only spec assertion go red for an environment reason rather than a real defect?
  [C] the round-1 rules themselves: the filename/pid check in liveInstances (older writers? Windows filename casing? anything that legitimately writes a differently named .json into instances/? what a hidden-but-live entry costs: a missed second-instance warning, a settings prune that deletes a peer's files) and the stderr rule (lsof/PowerShell writing warnings on success or no-match; a locale that changes stderr).
  [C] synchronous readers within a boot: server/index.ts prune cutoff and server/agents/{copilot,cursor}-hooks-file.ts still read unconfirmed entries. Is the PR's "healed by deletion on the next read" claim true, and does anything get worse for them?

2. EVERY SITE. If the same mistake appears elsewhere in the repo, list every occurrence NOW — grep for it.

| 0 | 0 | sandbox pre-flight (run on the #2091 head, same checkout) | — | tests (yarn vitest run test/bin/) exit 0; yarn typecheck exit 0; socket bind exit 0; api.github.com 200; gh reaches GitHub; literal `rm -f` REJECTED by the command guard (policy) — no rm-based mutation restore |
| 1 | 0 | [Claude, before round 1] P2: portOwners' LOOKUP_TIMEOUT_MS was sized for lsof; on Windows CI (Defender scanning OFF) the one real PowerShell call used most of it on every recent run, so on a user's machine a timeout → null → servingInstances keeps the stale entry (reported case unfixed) and stop reports unconfirmed. Also the real-socket spec skipped on null everywhere, so green Windows CI never proved PowerShell answered | FIXED | b18defe3: cap raised in port-owner.js (class: launcher + stop); spec fails on null when win32. Evidence: `gh run view <id> --log \| grep port-owner.spec` on windows-daily/windows-pr runs. NEVER REVIEWED — hold to the same bar |
| 2 | 1 | [Codex] P2: servingInstances deleted entryFile(pid), not the file read — a mismatched 111.json naming a live pid on an unowned port deleted the live server's <pid>.json and survived itself | FIXED | d84d0f94: liveInstances skips (does not report, does not delete) a file whose name != `${pid}.json`, so by-pid readers (servingInstances delete, stop signal) act only on the file that pid wrote. Reproduced before (live entry DELETED) and after (SURVIVED) with real files. Specs: mismatch not reported + live entry survives servingInstances; removing the check reds both |
| 3 | 1 | [Claude, not Codex] P2: portOwners returned null only for ENOENT; EACCES (a policy-blocked powershell.exe) and a failed lookup came back [], which servingInstances deletes a LIVE entry over. Codex's null-policy axis answered "none" | FIXED | d84d0f94: lookupAnswered — string errno / killed / signal / non-zero exit with stderr → null; lsof silent exit 1 → []. Repro: non-executable lsof on PATH → [] before, null after; real lsof no-match still []. Specs EACCES + stderr-failure, each reds under its mutation. Cost: lsof hosts warning on stderr with no match keep stale entries (= main). stop unaffected (null and [] both unconfirmed) |
| 4 | 1 | [Claude] real-OS spec for no-match: free port → [] (null fails on win32) — the only evidence PowerShell's no-match is an answer | ADDED | d84d0f94, test/bin/port-owner.spec.ts. Verified on macOS only; Windows evidence arrives from test_windows CI |

That table is context, NOT a list of closed topics: if a resolution is wrong, say 'REOPENING #<row>' with evidence. Rows 1-4 were reviewed clean in round 2.

Decisions the user already made (challenge only with evidence): the confirmation uses portOwners (kernel port ownership), not a canBind probe and not a heartbeat; the fix is split across #2091 (prompt deadline) and this PR; liveInstances() stays synchronous.

  Then, on its own line:

Also print that same final comment to stdout at the end, verbatim, so it survives if posting fails.

```

</details>

<details><summary>Codex stdout (verbatim)</summary>

```text
Posted the ROUND 3 review comment: https://github.com/receptron/mulmoterminal/pull/2092#issuecomment-5694602101

Verdict was `CODEX VERDICT: LGTM`. Local `yarn vitest run test/bin/` and `yarn typecheck` both passed; no findings held back.
```

</details>

Verdict comment: https://github.com/receptron/mulmoterminal/pull/2092#issuecomment-5694602101

### Triage — clean, loop exits on `54b02036`

- This is the round owed in the round-2 triage: #2091 merged (`81840502`), this PR was retargeted to `main`, synced, and pushed (which is also what started its full check set).
- Codex: `LGTM`, `FINDINGS COMPLETE`, on the head that will merge; it independently confirmed the sync delta is only `bin/prompt-yes-no.d.ts` from `c9056ce8`.
- Claude: no MUST-FIX. `git diff --stat d84d0f94 54b02036` shows that one declaration file, already reviewed clean in #2091 round 3; `yarn lint`, `yarn typecheck` and `test/bin/` green on the merged result.
- Rounds: 1 — two destructive P2s (Codex, Claude), both reproduced and fixed; 2 — clean, with `test_windows` confirming PowerShell's no-match is `[]`; 3 — clean on the retargeted, synced head.

PR #2094, opened 2026-09-16
feat: add persistent background bots with durable CLI questions

snakajima:
## Summary

Add persistent background Bots that frontend Claude sessions can delegate to through MCP. Bots run interactive Claude Code in tmux, keep their context between tasks, and remain hidden from terminal grids, tabs and session pickers. A bundled `bot` skill supports discovery, creation, messaging, compaction and termination.

Bots are shared across frontend terminals on the same server. Requests and replies are durable and replies wake the requesting frontend when it can accept input. Restart recovery preserves surviving tmux sessions; uncertain deliveries are held rather than replayed, and duplicate running names are rejected.

CLI questions survive a closed frontend and can be inspected and answered from another terminal. Recognized summary-resume dialogs are answered automatically by default; other menus require an explicit choice and unknown screens remain held. Answers are claimed before sending keys, revalidated against the current screen, and never automatically replayed after uncertain submission or restart.

Creation accepts optional `cwd` (an existing absolute path or `~/...`); omitting it inherits the frontend directory. The chosen directory is validated before creation and persists across restart and use from other terminals.

## Validation

- Full suite: 13,241 tests passed, 50 skipped (898 files passed, 3 skipped).
- Latest Bot/directory and bundled-skill checks: 127 tests passed across 9 files, covering explicit/tilde/inherited cwd, restart persistence, invalid paths and frontend capability checks.
- Formatting, lint, typecheck/build and diff checks passed.
- Live test server: HTTP 200; two independent MCP callers found the same two existing Bots, both ready. Bot ids, tmux sessions and requests survived the upgrade unchanged.
- User testing exercised real task delivery and MCP replies. Automated tests cover requester routing, frontend wakeups, closed-front questions, concurrent answers, changed screens and crash recovery.

## Boundaries

Currently requires Claude and tmux; sharing is within one server instance. Dialog parsing is deliberately conservative. The actual expired-cache dialog has not yet been reproduced live; its automatic response is covered with representative screen fixtures. No Bot terminal UI is exposed.

Implementation plan: `plans/feat-background-bots.md`. English/Japanese usage: `docs/background-bots.md`.

## Summary by CodeRabbit

* **New Features**
  * Added persistent background Bots with queued tasks, replies, restart recovery, and shared frontend access.
  * Added Bot management, messaging, mailbox, inspection, and CLI prompt-response tools.
  * Added optional working-directory selection when creating a Bot.
  * Added the bundled `bot` skill.

* **Bug Fixes**
  * Bot sessions are hidden from terminal lists and idle cleanup.
  * Improved coordination between Bot deliveries and user input.
  * Improved recovery and diagnostics for stalled or blocked prompts.

* **Documentation**
  * Added guidance for Bot permissions, recovery, sharing, directories, and troubleshooting.


snakajima:
Fixed the delivery/termination race in a329db3b. After a failed send, the service reloads the current Bot and only records a diagnostic while the Bot is alive and that request is still uncertain. This also avoids stale questions after a reply arrives before the send acknowledgement fails.

Added regressions for both concurrent kill and concurrent reply; all 29 service/prompt tests passed, along with lint.

@codex review

snakajima:
Addressed both CodeRabbit findings in 891a592c:

- Failed initialization now clears the published service. A regression forces persistence failure during prompt recovery and verifies that tools are unavailable, no timer starts, and saved Bots remain unchanged.
- A failed frontend wakeup is caught per recipient, preserving its unread notice while continuing to notify other frontends. A regression verifies both persisted notification states.

All 67 Bot tests, lint and typecheck/build passed locally.

Windows CI on the preceding commit failed twice only in the unchanged ShapeScript browser-rendering suite (navigation timeout, then net::ERR_ABORTED). All Bot tests passed there. Linux and macOS CI passed. The new commit will receive a fresh full CI run.

@coderabbitai review

snakajima:
Claude Code independently reviewed the full feature and returned `CLAUDE VERDICT: LGTM` for a329db3b (static review; no live CLI testing). It identified a non-blocking delayed IdlePrompt race that could lose a task reply; I addressed it in 72880e67 anyway. Idle reminders no longer complete tasks, and are ignored while delivery/work is in flight. Two regressions cover a reminder during paste and after submission. All 69 Bot tests, lint and typecheck/build passed. Claude Code is now rechecking the final delta.

The two earlier CodeRabbit findings were fixed in 891a592c. On the subsequent outside-diff comments:

- Terminal capture and submission cannot be atomic against an independently running CLI redraw. A host lease cannot freeze that process. The host already durably claims one responder, holds task dispatch while a prompt is active, rechecks the menu before each key, and has no await between final capture and Enter. I corrected the comment and documented this remaining transport limitation rather than claim an atomicity guarantee the CLI cannot provide.
- Non-Claude frontend access to list/inspect/respond/kill/read is intentional: users must be able to manage or resolve a shared Bot when its original frontend is closed. The live-Claude check is a capability requirement for starting work and automatically receiving its replies, not the authorization boundary. Bot/front-end tool separation and caller-scoped mailboxes remain enforced. The tool description now states this distinction explicitly.

All checks, including Windows, passed on 891a592c. The latest commit is running its checks again.

snakajima:
All three changes verified at 72880e67 against the code, not just the diff.

**Initialization failure** (`server/bots/host.ts:176`): `service` is cleared in the catch, so `dispatchBotTool` throws "service is not available" instead of accepting work that never dispatches. The interval was already only created after `recover()`, so no timer leaks. The new host spec pins tools disabled, zero timers, and the state file untouched.

**Frontend wakeup isolation** (`server/bots/service.ts:295-304`): a rejected `send` is logged and the loop continues to the next owner. The failed owner's notices keep `notified: false`, so they retry once that frontend is ready again. The host `send` has already called `abandonBotInput` before rethrowing, so the failing frontend stays held conservatively until its next real submit. The service spec pins both the continuation and the persisted `notified` flags.

**Delayed IdlePrompt** (`server/bots/host.ts:158-164`): the reminder is dropped outright when the input phase is `busy` or the Bot has a `sent` or `uncertain` request, so it can neither flip the phase to `ready` mid-turn nor reach `onStop`. Otherwise it only refreshes the input gate and returns. `onStop` is now driven by `Stop` alone, so an idle reminder can never record "ended its turn without returning a reply" or cause a later `replyToFrontend` to be discarded as a duplicate. The host spec runs the race at 800 ms (during the paste, phase `busy` from the lease) and 1600 ms (after the submit ack), and asserts the real reply arrives. One consequence to be aware of: a genuinely lost `Stop` after a task no longer converts to an error via the reminder. It is instead surfaced by the two-minute no-progress diagnostic in `PromptMonitor`, which is the conservative direction and keeps the request recoverable.

**Tool description** (`server/bots/tools.ts:27`): matches the code. Only `sendToBot`, `create` and `compact` call the `cwd()` validator that requires a live Claude pty, so list, inspect, respond, kill and read work from any frontend when the requester is gone. One optional wording nit: "reading replies are available to other frontend sessions" could be read by a model as access to the original requester's mailbox, while `read` is caller-scoped. "Each frontend reads its own replies" would be unambiguous.

**Docs and comment** (`docs/background-bots.md:42,62`, `server/bots/prompt-input.ts:23-24`): the wording now correctly states that capture and Enter are consecutive on the host with no await between them, and that the external CLI can still redraw. The durable `answering` claim in `BotPrompts.respond` and the `active()` hold in `nextRequest` remain as before.

No regressions found in these changes. Only the previously noted non-blocking items (synchronous tmux load per tick, no escape from `uncertain` without kill, state retention) remain open, and none of them was made worse here.

Reviewed commit: 72880e67 on feat/background-bots. Verification was read-only: no tests, lint, typecheck or live Bot runs were executed by me. I relied on the reported local pass of the 69 Bot tests and the prior CI pass.

CLAUDE VERDICT: LGTM

isamu:
Review pass (second reviewer, while the author is away). I re-checked the two heavyweight bot findings against the current head and both are fixed:

- **Codex P2 (`service.ts`, stale Bot after a concurrent kill):** the delivery catch now re-reads the store — `this.store.state.bots.find((item) => item.id === bot.id && item.alive)` — and returns unless that request is still `uncertain`, so a killed Bot no longer gets an unanswerable prompt.
- **CodeRabbit Major (`host.ts`, partial service after a failed init):** `startBots` now sets `service = undefined` in its catch, so `dispatchBotTool` cannot reach a service whose `recover()` threw.

One of CodeRabbit's Minors is real and still open:

**`prompt-monitor.ts` — the two-minute diagnostic loses its own explanation on a large pane.**
`inspect()` builds `text` as `slice(-8000)` of the stripped capture, prepends the explanation, and hands the result to `unknownPrompt()`, which does `message.slice(-8000)` again (`prompt-screen.ts:10-15`). When the capture reaches the cap, the second slice drops exactly the prefix, so the notice arrives as raw terminal text with no "do not resend or send arbitrary keys" line — the one sentence that tells the reading frontend not to act on the screen. A pane wider or taller than 8000 characters of content reaches this. Reserving the prefix's length before the append (or slicing the capture to `8000 - prefix.length`) fixes it.

Two observations, neither blocking:

- **`input-gate.ts`: a claimed lease with no following hook holds user bytes forever.** `claimBotInput` sets `pending = []`, and only a lifecycle hook or `abandonBotInput` releases it. If `send` succeeds and the agent then emits nothing, `deferDuringBotDelivery` keeps queueing writes with no timeout. The blast radius is small — these sessions are hidden, so there is usually no user typing into them — and `PromptMonitor` surfaces the stall as a diagnostic, but the gate itself has no way back.
- **Injection hygiene reads right:** the frontend notice states that replies are Bot-provided data and not new instructions, unrecognized screens never generate keys, and answers are claimed then revalidated. That is the part I looked hardest at, and I found nothing to add.

With the `prompt-monitor.ts` truncation addressed this looks good to me.

PR #2098, opened 2026-09-17
fix(theme): let a theme set the cursor colours, and stop light themes drawing black on the cursor

### LINKED ISSUE #2097: feat(theme): allow setting cursor / cursorAccent on the theme — light themes render the character under the cursor unreadable
## What happened

There is no way to set the terminal (xterm) cursor colours — `cursor` / `cursorAccent` — for all cells at once.

- A `themes` entry in `~/.mulmoterminal/config.json` may only carry the 20 UI CSS variables in `THEME_VAR_KEYS` (`common/themeVars.ts`). Of those, only `--bg-base`, `--term-fg` and `--term-selection` reach xterm. There is nowhere to write `cursor` or `cursorAccent`, so the `term.cursor` of the built-in theme named in `extends` is used as-is.
- All four built-in themes define `term.cursor` equal to `foreground`, and none defines `cursorAccent` (so xterm's default `#000000` applies). On the light themes (Daylight / Solarized, and any custom theme that `extends` them) this means a dark block cursor with black text on top of it — the character under the cursor is unreadable.
- The only place cursor colours can be set is the per-directory `.mulmoterminal.json` / `.mulmoterminal.local.json` `colors` block (`THEME_COLOR_KEYS` in `common/themeColors.ts`). To get the same cursor everywhere, the same file has to be placed in every directory in use, and for git-managed repositories that also means maintaining the `.local` variant plus `.git/info/exclude`.

## What I expected

Cursor colours should be settable on the theme (global) side. Either of these would do:

1. Allow, next to `themes[].colors`, a `themes[].term` (or `themes[].xterm`) object that takes xterm ITheme keys directly. Validating it against the same `THEME_COLOR_KEYS` used for the per-directory `colors` keeps it consistent with the existing mechanism.
   ```json
   {
     "themes": [
       {
         "id": "washi",
         "extends": "daylight",
         "colors": { "--bg-base": "#ece7dc", "--term-fg": "#2a2622", "…": "…" },
         "term": { "cursor": "#b0402a", "cursorAccent": "#fffdf8" }
       }
     ]
   }
   ```
2. Or allow a global `colors` block (same shape as the directory one) at the top level of `config.json`, with the directory block overriding it.

In addition, it would be good for the built-in light themes (Daylight / Solarized) to set `cursorAccent` explicitly, so the character under the cursor is readable out of the box.

## Steps to reproduce

1. Define a custom theme with `extends: "daylight"` in `~/.mulmoterminal/config.json` and select it in Settings (or simply select the built-in Daylight).
2. In a Claude Code or shell cell, place the cursor on a character.
3. The cursor is drawn as a dark block and the character on it is black — unreadable.
4. Writing `cursor` / `cursorAccent` into `themes[].colors` is dropped by validation (not in `THEME_VAR_KEYS`) and changes nothing.
5. Placing `{"colors":{"cursor":"#b0402a","cursorAccent":"#fffdf8"}}` in `<dir>/.mulmoterminal.json` fixes it, but only for cells in that directory.

## Environment

- mulmoterminal 4.26.0 (npx: `npm exec mulmoterminal@latest --no-open`, with tmux)
- Node v22.17.1 / Linux x86_64 (Ubuntu on AWS) / bash
- tmux 3.4 / gh 2.45.0 / Claude Code 2.1.273
- Theme in use: a custom theme with `extends: "daylight"` (`--term-fg: #2a2622`)

## Notes

- As a workaround, `tmux -L mulmoterminal set -g cursor-colour '<color>'` does change the cursor colour via OSC 12 (the client xterm advertises `ccolour` in its termfeatures). But `cursorAccent` cannot be changed that way, and the setting is lost whenever the tmux server goes away. `~/.mulmoterminal/tmux.conf` is regenerated on every start, so it cannot be persisted there either.

isamu:
Closes #2097

## Summary

xterm draws the cursor cell as a `cursor`-coloured block with the character on it painted in
`cursorAccent`. Every built-in theme set `cursor` and none set `cursorAccent`, so xterm's default
of a flat `#000000` applied — and on the light themes that is black on a dark block: the one
character you are looking at was the only unreadable one. There was also no way to set the pair
globally; the only surface that accepted `cursor` / `cursorAccent` was a per-directory
`.mulmoterminal.json`, so one cursor colour meant one such file in every project.

Three parts, in the order they matter:

1. **The built-ins state `cursorAccent`** — each its own `background`, so the cursor reads as an
   inverted cell. The dark two change from an implicit `#000000` to their own near-black
   background, which is imperceptible and makes "`cursorAccent` is the background" one rule
   everywhere rather than two.
2. **`termThemeFromVars` derives the pair** (`cursor` ← `--term-fg`, `cursorAccent` ← `--bg-base`).
   Needed on top of 1, because a custom theme inherits `cursor` from the built-in it extends — a
   light theme extending a dark one was unreadable however the built-ins were fixed.
3. **`themes[].term`** (issue option 1) lets a theme state the xterm palette outright, validated
   against `THEME_COLOR_KEYS` — the same key set a directory's `colors` block already takes. It
   also reaches the 16 ANSI colours, which no CSS variable could, so changing one red no longer
   means giving up `extends`.

```json
{
  "themes": [
    {
      "id": "washi",
      "label": "Washi",
      "extends": "daylight",
      "colors": { "--bg-base": "#ece7dc", "--term-fg": "#2a2622" },
      "term": { "cursor": "#b0402a", "cursorAccent": "#fffdf8" }
    }
  ]
}
```

Precedence, widest to narrowest: what `extends` names → what `colors` implies → `term` → the
directory's own `colors`. The directory still wins for its own cells: the existing rule extended,
not changed.

Option 2 in the issue (a global `colors` block at the top level of `config.json`) was not taken:
it would sit outside the theme, so switching themes would leave the previous theme's cursor
behind, and the same key set would stack twice with nothing saying which scope owned it.

The review loop added two things beyond the original three. `PALETTE_COLOR_RE` / `isPaletteColor`
now live in `common/themeColors.ts` beside the key list, and **both** client paths check the
colour's *shape* and not just its key — the theme's `term` block and `useDirConfig`'s
`parseColors`, which had the identical gap. A string that is not a colour does not degrade to the
theme's there: xterm throws parsing it, during the assignment of the whole theme object, so one
bad entry costs every colour in the terminal.

## Items to Confirm / Review

- **The dark built-ins changed too.** The issue only asked for the light ones. All four are set so
  that the derivation in (2) and the built-ins in (1) state one rule; Midnight/Nord go from
  `#000000` to `#1a1a2e` / `#2e3440`. Say the word and I will narrow it to the light two.
- **(2) changes existing custom themes.** A theme already in someone's config now takes its cursor
  from its own `--term-fg` rather than from the base it extends. That is the fix for the
  reporter's case, and it is a visible change to a theme they did not edit. It cannot make the
  cursor *less* legible than ordinary text — see the ledger comment — but it is a change.
- **An unknown key in `term` drops the whole theme**, matching what `colors` already does (zod's
  `partialRecord` over an enum rejects unrecognised keys). `term` and `colors` take *different*
  vocabularies, so a `--bg-base` written into `term` silently costs the entry. Documented in both
  guides and the skill; still a sharp edge worth a second opinion.
- Two doc corrections rode along, both load-bearing rather than cosmetic: the theme skill said an
  unknown key was "dropped" in place (the entry goes whole), and described `--bg-deep` as the
  terminal background (it is `--bg-base`; `--bg-deep` colours side panes, so an author following
  that line changed the wrong key and saw nothing happen).

## Verification

`yarn format` / `lint` / `typecheck` / `build` clean; `yarn test` green; CI green on every check.

**Driven in a real browser** against a built server on a scratch `HOME`, so the maintainer's own
config was never touched. Before/after captured by rebuilding with the client change stashed, and
the cursor cell's pixels histogrammed inside its own bounding box rather than eyeballed:

| capture | inside the cursor cell |
|---|---|
| built-in Daylight, **before** | contains `#000000` — xterm's default `cursorAccent`. The bug. |
| built-in Daylight, **after** | block `#1b2430`, glyph `#f4f6fb`. No `#000000` anywhere. |
| custom theme with `term`, **before** | daylight's dark navy block; the `term` block had no effect at all |
| custom theme with `term`, **after** | block exactly `#b0402a`, glyph `#fffdf8` — what was written |
| 24-bit `#111111` ANSI text, cursor on it | block plus light glyph; **`#111111` does not appear in the cell** — the glyph is redrawn in `cursorAccent`, so ANSI colour is not at risk |

The last row answers a review finding rather than the issue. Both post-fix captures were retaken
on the exact commit proposed for merge.

Every new test was mutation-checked: reverting `cursorAccent` to black turns six red across three
files, inverting the `term`/derivation spread order turns four red, and dropping the value check
turns one red.

## User Prompt

- https://github.com/receptron/mulmoterminal/issues/2097 直せる？

work in mulmoterminal3


isamu:

Findings raised so far and what happened to each. Re-flagging a `FIXED` or
`REJECTED` row is welcome **if the resolution is wrong** — say which row and what
is wrong with it.

| # | iter | bot | finding (one line) | disposition | why / where |
|---|------|-----|--------------------|-------------|-------------|
| 1 | 1 | codex | Deriving the cursor pair can make an existing custom theme *less* readable (e.g. `--term-fg` equal to `--bg-base`) | REJECTED | The cursor cell is the surrounding cell with its two colours swapped, so `contrast(cursor, cursorAccent)` **is** `contrast(foreground, background)` — same pair, same number. Pinned over 500 generated palettes in `test/common/themeVars.spec.ts`. See the note below. |
| 2 | 1 | codex | `themeTermOverrides()` checks a `term` entry's key and `typeof string`, but not the colour shape | FIXED | `PALETTE_COLOR_RE` / `isPaletteColor` moved to `common/themeColors.ts`; both client paths use it — `customThemes.ts` and `useDirConfig.ts`'s `parseColors`, which had the same gap. Commit `0de7cb8`. |
| 3 | 1 | codex | Precedence tested in pieces, never end to end; no test for the directory leg | FIXED | `test/src/components/TerminalDirFontApply.spec.ts` now asserts the object the terminal is actually handed: `term` over the derivation, a directory over `term` key by key, and a junk entry falling back to the derived colour. |
| 4 | 1 | codex | Skill's security paragraph says these values land in CSS — true for `colors`, not for `term` | FIXED | Reworded: `colors` is a CSS-injection surface, `term` is a canvas one where an unparseable colour throws during the whole-theme assignment. |
| 5 | 1 | claude | Theme skill described `--bg-deep` as "the terminal background". It is not — xterm's background is `--bg-base`; `--bg-deep` colours side panes and overlays | FIXED | A theme author following that line changes the wrong key and sees nothing happen. Not flagged by any bot. |

### On row 1, in more detail

The worry is reasonable and the answer is structural rather than a judgement call.
`cursor` is `--term-fg` and `cursorAccent` is `--bg-base` — **the same two values**
that are already `foreground` and `background`. So the cursor cell is an exact
inverse of an ordinary cell, and the contrast a reader gets on it is arithmetically
identical to the contrast they get on every other character.

Codex's example — a theme extending `midnight` that sets `--bg-base: "#e0e0e0"` and
inherits `--term-fg: "#e0e0e0"` — does produce an invisible cursor. It also produces
invisible body text, for the same reason and by the code that was already there: the
terminal's `foreground` and `background` have derived from those two variables since
#996. Before this PR the cursor was an accidental island of legibility in a terminal
you could not read. That is not a property worth preserving, and preserving it would
mean the cursor's colour stops being predictable from the theme.

There is now a test asserting the degenerate case stays degenerate, specifically so a
later "safety" fallback is not added on the belief that the cursor is a special case.

### Reviewer availability — please read before treating this as reviewed

- **Codex**: the repo's `Codex auto-review` workflow is failing with
  `401 Unauthorized` from Azure OpenAI (`invalid subscription key or wrong API
  endpoint`) and has been since **before this PR** — run `35158956238` on 2026-09-16
  failed the same way. The findings above come from the **local** `codex` CLI instead,
  so they are a real second opinion but they did **not** come from CI.
- **CodeRabbit**: reported *"Review rate limited"* on this PR and has posted nothing.

Neither of those is a sign-off. Re-triggering below.

isamu:
@coderabbitai review

isamu:

Re-flagging a `FIXED` or `REJECTED` row is welcome **if the resolution is wrong** —
say which row and what is wrong with it.

| # | iter | bot | finding (one line) | disposition | why / where |
|---|------|-----|--------------------|-------------|-------------|
| 1 | 1 | codex | Deriving the cursor pair can make an existing custom theme *less* readable | REJECTED | The cursor cell is the surrounding cell with its two colours swapped, so its contrast **is** the contrast of ordinary text. Pinned over 500 generated palettes. |
| 2 | 1 | codex | Client checked a `term` entry's key but not its colour shape | FIXED | `isPaletteColor` in `common/themeColors.ts`, used by `customThemes.ts` and by `useDirConfig.ts`'s `parseColors`, which had the same gap. `0de7cb8` |
| 3 | 1 | codex | Precedence tested in pieces, never end to end | FIXED | `TerminalDirFontApply.spec.ts` asserts the object the terminal is handed. `0de7cb8`, refined in `32e88ab` — see row 8. |
| 4 | 1 | codex | Skill said these values land in CSS — true for `colors`, not `term` | FIXED | Reworded: `colors` is a CSS-injection surface, `term` a canvas one. |
| 5 | 1 | claude | Skill described `--bg-deep` as the terminal background; it is `--bg-base` | FIXED | Not flagged by any bot. |
| 6 | 2 | codex | ANSI-coloured text goes unreadable under the derived cursor (`#111111` text vs a black cursor block ≈ 1.1:1) | REJECTED | **Measured** — see below. The glyph does not keep its ANSI colour on the cursor block. |
| 7 | 2 | codex | `isPaletteColor` tested `value.trim()` but returned the untrimmed value, so `" #fff "` passed and was forwarded verbatim | FIXED | Now strict — whitespace is a reject. A predicate guarding a value about to be forwarded must judge *that* value. New spec `test/common/themeColors.spec.ts`. `32e88ab` |
| 8 | 2 | codex | `appliedTheme()` fell back between two seams, so a broken `setTheme()` could pass on `attach()`'s correctness | FIXED — **and it was hiding a real defect in the test** | Split into `builtWith()` / `repaintedTo()`. That immediately failed: a directory's `colors` arrive on the **repaint**, not at attach, because `/api/dir-config` is async. The spec claimed "dir beats term" while asserting a seam the dir config had not reached. Corrected, plus a late-resolution case. `32e88ab` |

### Row 6 — measured, not argued

The claim rests on the glyph keeping its ANSI colour on the cursor block. It does
not: for a block cursor xterm draws the cell in `cursor` and re-draws the character
in `cursorAccent`, whatever the cell's own colour is.

Driven against a real server and browser, with a prompt in 24-bit `#111111` on
built-in Daylight, cursor parked on a character. Histogram of the pixels **inside
the cursor cell's bounding box**:

| capture | inside the cursor cell |
|---|---|
| this PR, ANSI `#111111` text | `#1b2430` (block) and `#f4f6fb` / `#f3f5fa` / `#f0f2f7` (the glyph, antialiased) — **no `#111111` anywhere in the cell** |
| **before** this PR, same theme | contains `#000000` — xterm's default `cursorAccent`, i.e. the bug |

`#000000` appears in the pre-fix capture and in no post-fix capture, and the only
thing that changed is `cursorAccent`. So the derived pair does not put ANSI-coloured
text at risk, and it is what removes the black-on-dark glyph in the first place.

The selection variant has the same answer for the same reason — the cursor cell is
always the `cursor`/`cursorAccent` pair, so its contrast is always the theme's
foreground against its background.

### Reviewer availability

Unchanged and still worth stating: the repo's **`Codex auto-review` workflow is
401-ing against Azure OpenAI** and has been since before this PR (run
`35158956238`, 2026-09-16, same failure). The findings above come from the **local**
`codex` CLI. **CodeRabbit** reported *"Review rate limited"* and has posted nothing
on either head, including after an explicit `@coderabbitai review`.

Neither silence is a sign-off, and I will not count either as a clean round.

isamu:
## Iteration 3 — status

- **Codex (local CLI): `CODEX VERDICT: LGTM`** on `32e88ab`, with no new commits made in
  this round. It checked xterm's renderer directly and **accepted the row 6 rejection**,
  adding a useful detail worth recording here: bar / underline / outline cursors keep the
  glyph's *own* colour and never consult `cursorAccent`, so the derived pair cannot hurt
  them either — and this repo configures no alternate cursor style or inactive style.
- It also confirmed the reworked precedence specs are non-vacuous, and that making
  `isPaletteColor` strict breaks no caller (both client callers receive server-normalised
  values).
- **Clean-round count: 1 of 2.** A single clean round is a round that happened to find
  nothing; the loop wants a second one on a head the first has already seen.
- **CodeRabbit's rate limit has cleared** and it is reviewing now (run
  `2819d7a8-216a-456f-97ed-897d0bbdfc10`, against `0de7cb8`). Its findings will be
  iteration 4. Until they land, nothing here is converged.
- CI: green on every settled check.

isamu:
## Review ledger — final (after iteration 4)

Both reviewers have now cleared **the same head, `32e88ab`**, with no commit between
the two rounds.

| # | iter | bot | finding (one line) | disposition | why / where |
|---|------|-----|--------------------|-------------|-------------|
| 1 | 1 | codex | Deriving the cursor pair can make an existing custom theme *less* readable | REJECTED | The cursor cell is the surrounding cell with its two colours swapped, so its contrast **is** the contrast of ordinary text. Pinned over 500 generated palettes. Codex accepted this in iteration 3. |
| 2 | 1 | codex | Client checked a `term` entry's key but not its colour shape | FIXED | `isPaletteColor` in `common/themeColors.ts`, used by `customThemes.ts` **and** `useDirConfig.ts`'s `parseColors`, which had the same gap. `0de7cb8` |
| 3 | 1 | codex | Precedence tested in pieces, never end to end | FIXED | `TerminalDirFontApply.spec.ts` asserts the object the terminal is handed. `0de7cb8`, refined in `32e88ab` — see row 8. |
| 4 | 1 | codex | Skill said these values land in CSS — true for `colors`, not `term` | FIXED | Reworded: `colors` is a CSS-injection surface, `term` a canvas one. |
| 5 | 1 | claude | Skill described `--bg-deep` as the terminal background; it is `--bg-base` | FIXED | An author following that line changes the wrong key and sees nothing happen. Not flagged by any bot. |
| 6 | 2 | codex | ANSI-coloured text goes unreadable under the derived cursor | REJECTED | Measured: `#111111` does not appear inside the cursor cell — the glyph is redrawn in `cursorAccent`. Codex accepted in iteration 3 and added that bar/underline/outline cursors keep the glyph's own colour and never consult `cursorAccent`, and this repo configures no alternate cursor style. |
| 7 | 2 | codex | `isPaletteColor` tested `value.trim()` but returned the untrimmed value | FIXED | Now strict — a predicate guarding a value about to be forwarded must judge *that* value. `test/common/themeColors.spec.ts`. `32e88ab` |
| 8 | 2 | codex | A two-seam fallback helper could let a broken `setTheme()` pass on `attach()`'s correctness | FIXED — **and it was hiding a real defect in the test** | Split into `builtWith()` / `repaintedTo()`, which immediately went red: a directory's `colors` arrive on the **repaint**, not at attach, because `/api/dir-config` is async. The spec claimed "dir beats term" while asserting a seam the dir config had not reached. Corrected, plus a late-resolution case. `32e88ab` |

### Rounds

| iter | Codex | CodeRabbit | outcome |
|---|---|---|---|
| 1 | CHANGES REQUESTED (4) | rate limited — no review | not clean |
| 2 | CHANGES REQUESTED (3) | rate limited — no review | not clean |
| 3 | **LGTM** on `32e88ab` | still reviewing | **clean (1 of 2)** |
| 4 | — (no commit since) | **no actionable comments** on `32e88ab`, merge risk minimal | **clean (2 of 2)** |

### Reviewer caveat, recorded rather than glossed

The repo's **`Codex auto-review` workflow is broken** — `401 Unauthorized` from Azure
OpenAI, on this PR and on run `35158956238` from 2026-09-16, so it predates this branch
and is not something this PR caused. Every Codex finding above came from the **local**
`codex` CLI instead. That is a genuine independent review, but it did not come from CI,
and the workflow will keep failing on other PRs until the Azure key is renewed. Worth its
own issue.

CodeRabbit was rate-limited for the first two rounds and reviewed on the third attempt;
its clean result is on the current head.

### Verification on the merge head

Re-run after the last commit, not carried over from an earlier one. Cursor-cell pixels,
histogrammed inside the block's own bounding box:

- custom theme with `term`: block exactly `#b0402a`, glyph `#fffdf8`, **no `#000000`**
- built-in Daylight: glyph `#f4f6fb`, **no `#000000`** (the same capture before this PR has it)

CI green on every check.

PR #2102, opened 2026-09-17
feat: open a file by name in the Files pane (#2099)

### LINKED ISSUE #2099: feat: ファイルペインで、ファイル名を入力して目的のファイルを一度で開きたい（Visual Studio CodeのCommand+Pのような検索）
MulmoTerminalを日々便利に使わせていただいています🙇

ターミナルの横でファイルを開いて確認・編集できるファイルペインも、よく使っています。
そのファイルペインで、ファイル名の一部を入力して目的のファイルへ一度でたどり着けると嬉しいです。

## 概要

エージェントの作業内容を確認するときや、自分で少し手を入れたいときに
Visual Studio CodeのCommand+Pのように、ファイル名の一部を入力して候補から選び、そのままファイルペインで開きたいです。

## 現状課題
ファイルペインのツリーは、ディレクトリを1階層ずつ展開していく方式です。深い場所にあるファイルや、どのディレクトリにあるかをはっきり覚えていないファイルを開くには、ツリーを順番にたどるしかなく、ファイル名が分かっていても時間がかかります。

## 希望する対応

ファイルペインに、ファイル名で探して開く機能がほしいです。

- キーボードショートカットで検索欄が開く。
- ファイル名やパスの一部を入力すると、プロジェクト全体から候補が絞り込まれる。多少あいまいな入力でも一致してほしいです。
- 上下キーとEnterで選ぶと、そのファイルがファイルペインで開く。ツリー側でもそのファイルの位置が分かると、周辺のファイルへ移りやすくなります。
- `.gitignore`で除外されているもの（`node_modules`など）は候補に出ない。
- ファイルペインを開いていない状態でショートカットを押した場合は、いま見ているセルの作業ディレクトリを対象にファイルペインが開いて、そのまま検索できる。

ショートカットのキーについては、ブラウザではCommand+Pが印刷に割り当てられているため、既定のキーにはこだわりません。
既存の`keymap`で自分で割り当てられれば十分です。

## 関連

- #1859: ファイルペインのツリーから、パスをターミナルへ挿入する機能

isamu:
## Summary

Closes #2099. The Files pane's tree opens one directory at a time, so a file whose **name** you know but whose **directory** you do not takes as long to reach as one you are browsing for.

A **search button** in the pane's header now opens a finder over it. Type part of a name or a path, pick from the list, and it opens in the editor **with the tree expanded down to it** — picking is a reveal, not just an open, because the tree is how you get to the neighbouring files.

- **Fuzzy, and ranked on purpose.** The match is a subsequence, scored by dynamic programming rather than a greedy left-to-right scan. Greedy spends the `t` of `test` inside `componenTs` and loses the run at the file's own name, which is the alignment a reader sees. The scoring prefers an unbroken run, a word boundary (`/ - _ .` and camelCase), and the **basename** over the directories above it. Ties go to the shorter path, then code-unit order, so the list never depends on input order.
- **`.gitignore` is honoured by asking git**, never by parsing it: `git ls-files --cached --others --exclude-standard -z`, run with `-C` so the paths come back relative to the pane's own root. A nested ignore file, `.git/info/exclude` and the user's global excludes all count, which is why `server/backends/collectionSelfContainment.ts` already made this same call.
- **A directory that is not a repository is walked instead — and the panel says so**, rather than letting a reader conclude their ignore file was ignored. A project past the cap is cut, and the panel says that too. Silence about a partial list is the one wrong answer a finder can give.
- **One request per opening, filtering in the browser.** No round trip per keystroke, and the ranking becomes a pure function that can be tested over generated input — which is where this feature's whole value sits.
- **`files-find` joins the keymap**, unbound by default (`Cmd+P` is the browser's Print, and no other key is anyone's habit). It opens the pane first when it is closed, which is the last item in the issue. It needs an enlarged cell, because the pane only exists in the zoomed row.

## Items to Confirm / Review

1. **The ranking is a judgement, not a fact.** `BONUS_CONSECUTIVE` is larger than `BONUS_BOUNDARY` deliberately: with the two equal, `s-e-t-t-i-n-g-s.ts` outranks `settings.ts` for `settings`, because every separator hands the next character a boundary bonus worth more than the gap it costs. The constants are pinned by tests that state the preference rather than the number. **Please try it on a repo you know and say if an ordering feels wrong.**
2. **`/api/files/browse/index` ignores `?path=`**, unlike every other browse route. The finder hands what it picks to the tree and the editor, which both resolve against the pane's **root**, so a list relative to some subdirectory would open the wrong file at every depth. A spec pins it.
3. **The non-git walk skips a named list of directories** (`.git`, `node_modules`, `.venv`, `__pycache__`, …). That list is a guess, and every name on it hides real files from someone. It is deliberately short — `dist` and `build` are **not** on it, because a user may well want to open what is in them.
4. **The cap is 20,000 paths** and the walk's own budget is 200,000 entries. Both are round numbers chosen to bound one request; neither has been measured against a repository that reaches them. This repository lists ~2,900.
5. **`FilesPane.vue` was at its `max-lines` limit**, so the three functions in it that touch **no component state** — the conditional write, the backup, the OS opener — moved to `src/components/filesPaneApi.ts` and gained direct tests. They were reachable only through a mounted editor before, and they hold the outcomes that matter most: a save that loses the version race (409), a store that refuses a backup, a host with no file manager to call. **Behaviour preservation is evidenced by the pre-existing suites run unchanged** (`FilesPane.spec.ts`, `filesRowMenu.spec.ts`, `FilesOverlay.spec.ts`, `gridOpenFiles.spec.ts`), and the two error messages differ by design (revealing names the absolute path, opening names the relative one), so `failure` is passed in and no string changed.
6. **Hovering a row moves the keyboard selection**, as it does in an editor's own "go to file". Say if that is unwanted.
7. **The route's exposure is the same class as the existing ones, but amplified.** `resolveBase` already lets `/api/files/browse/list` list *any* absolute existing directory — the documented trusted-local-user / loopback posture. This adds no new capability, but it turns "one level per request" into "the whole tree in one request". Worth a second opinion on whether that changes the calculation.

## How it was verified

`yarn format` / `lint` / `typecheck` / `build` / `test` all clean — the full suite is green (no file skipped, nothing excluded).

Beyond the suite, it was **run**: `yarn dev` on isolated ports with a scratch `HOME`, so the real `~/.mulmoterminal/config.json` was never touched.

- `GET /api/files/browse/index?cwd=<this repo>` answered `source: "git"`, with zero `node_modules` and zero `.git` entries, and included files that were not yet committed.
- `/files?cwd=<this repo>` in a real browser: the header button opened the finder, `fpane` put `src/components/FilesPane.vue` first with the matched characters highlighted, and Enter opened it **with the tree expanded to `src/components` and scrolled to the row**.
- With `{"keymap": {"zoom-toggle": "F8", "files-find": "F9"}}` in the scratch config: a cell enlarged with F8, then **F9 with no pane open** brought the pane up and put the finder over it. Typing and Enter opened the file.
- That cell's directory was not a repository, so the **"Not a git repository…" note appeared for real** — the walk branch, announcing itself.

**What was not exercised locally:** Windows and Linux (the walk's symlink handling and `git ls-files` quoting are the parts that could differ; CI runs the matrix), and a repository large enough to hit either cap.

## User Prompt

> https://github.com/receptron/mulmoterminal/issues/2099
> これほしいね！

Asked during the work:

- Entry point: `keymap` action **plus** a search button in the pane, so the feature is not invisible to anyone who has not written a keymap.
- Filtering: the server returns the list once and the browser filters, rather than a request per keystroke.

work in mulmoterminal5

## Summary by CodeRabbit

- **New Features**
  - Added fuzzy file-name and path search to the Files pane, with highlighted matches, keyboard navigation, and automatic tree expansion when a file is selected.
  - Added the optional `files-find` keyboard shortcut; it opens the Files pane when needed and has no default binding.
  - Search uses Git-tracked files when available and indicates when directory scanning is used.
  - Displays notices for incomplete or truncated results.

- **Documentation**
  - Updated English and Japanese guides and the changelog with search behavior and shortcut details.


isamu:

Findings raised so far and what happened to each. Re-flagging a `FIXED` or `REJECTED` row is welcome **if the resolution is wrong** — say which row and what is wrong with it.

| # | iter | reviewer | finding | disposition | why / where |
|---|------|----------|---------|-------------|-------------|
| 1 | 1 | CodeRabbit | Gitlink (submodule) entries are offered; the path is a directory and `/text` answers 400 | FIXED | `d149d97` — the cached half now comes from `ls-files --stage`, so mode `160000` is told from a file. `--others` needs no check: an untracked directory is not listed without `--directory`. Spec writes the gitlink straight into the index with `update-index --cacheinfo`, so no remote is needed |
| 2 | 1 | CodeRabbit | The walk's entry budget can stop enumeration below the path cap, and `truncated` reports false | FIXED | `d149d97` — both sources now report whether they reached the end; the cap is one of two ways the answer can be partial. Budget is injectable so a spec can exhaust it |
| 3 | 1 | CodeRabbit | `highlightParts` iterates code points, `matchPath` indexes by code unit | FIXED | `d149d97` — `split("")`. Spec: `😀a.ts` matched on `a` highlights `a`, and the emoji stays one character |
| 4 | 1 | CodeRabbit | A superseded `revealPath` can replace the user's later pick | FIXED | `d149d97` — a reveal generation, bumped by each pick and by `teardown()`. The spec holds the `src/deep` listing open so the second pick overtakes the first; **verified to fail without the guard** (it opened `src/deep/buried.ts` over `README.md`) |
| 5 | 1 | CodeRabbit | README's `#keyboard-shortcuts` fragment has no target | FIXED | `d149d97` — points at the guide's `#keymap`, which is where README links the keymap everywhere else. There is no such heading in README to anchor |
| 6 | 1 | CodeRabbit | The `.gitignore` guarantee is stated flatly; the walk fallback does not apply it | FIXED | `d149d97` — qualified in README, `guide/{en,ja}/basics.md` and both keymap tables. The panel already said it at runtime; the prose did not |
| 7 | 1 | Claude (not flagged by any bot) | The browser gave up before the server could answer: `fetchWithTimeout` defaults to 8s, the server's `git ls-files` cap is 10s, and a repo that misses it is then walked | FIXED | `d149d97` — uses the slow-command deadline, since the route shells out |
| 8 | 1 | Claude (not flagged by any bot) | A symlinked directory, and a broken link, were offered — the same "opens nothing" shape as row 1 | FIXED | `d149d97` — the walk stats a symlink and offers it only when it resolves to a file. Three specs: directory loop, file alias, dangling link |

**Not yet reviewed by anyone:** Codex. Its workflow in this repo is `workflow_dispatch` only, so it has to be asked for; I am triggering it against `d149d973` now. Sourcery is not configured here.

Gates on `d149d973`: `format` / `lint` / `typecheck` / `build` clean, full suite green (896 files, 13318 tests, none skipped by this PR).

isamu:
## Review ledger — after iteration 8

Re-flagging a `FIXED` or `REJECTED` row is welcome **if the resolution is wrong** — say which row and what is wrong with it.

### Codex could not run in CI

`Codex auto-review` here is `workflow_dispatch`-only and **401s against Azure OpenAI** — the two runs on `main` before this PR failed identically, so it is a credential that needs renewing and not something this branch caused. The rounds below were run with the `codex` CLI locally over the same diff instead, which is the same reviewer without the CI wiring. Sourcery is not configured on this repo.

| # | iter | reviewer | finding | disposition | why / where |
|---|------|----------|---------|-------------|-------------|
| 1–6 | 1 | CodeRabbit | gitlink rows; walk-budget truncation; code-point vs code-unit highlights; superseded reveal; dead README anchor; flat `.gitignore` claim | FIXED | `d149d97` — see the iteration-1 ledger above |
| 7–8 | 1 | Claude | client deadline shorter than the server's; symlinked directories and broken links offered | FIXED | `d149d97` |
| 9 | 1 | Claude | Splitting the listing into two `ls-files` calls created a state the single call did not have: tracked half answers, untracked half does not | FIXED | `3be3666` — keeps git's answer marked partial instead of falling back to the walk, which would show `node_modules` in a repo that plainly has a `.gitignore`. Both branches reachable only by mocking the runner |
| 10 | 2 | Codex | Git results were not guaranteed openable: tracked symlinks, deleted-but-tracked files, untracked symlinks | FIXED | `c9effd7` then rewritten in `803c12e` — see row 15 |
| 11 | 2 | Codex | The walk reported complete when a subtree could not be READ | FIXED | `c9effd7` — `readDirSafely` answers null |
| 12 | 2 | Codex | `teardown()` bumped only the reveal generation, so an in-flight `loadFile` could adopt the old project's content | FIXED | `c9effd7` — all three generations move |
| 13 | 2 | Codex | The finder's index request had no unmount cancellation, on the SLOW deadline | FIXED | `c9effd7` — AbortController; an abort is not reported as a failure |
| 14 | 3 | Codex | `complete` read the counter, so a tree holding exactly the budget was called truncated | FIXED | `f08108e` — a flag set where the walk actually gives up. Three boundary cases |
| 15 | 4 | Codex | A symlink can resolve to a real file OUTSIDE the project; the route answers 403 | FIXED | `96eca23` — candidates go through `resolveContained`, the gate the route itself applies, so the two cannot disagree |
| 16 | 4 | Codex | The panel asserted "this project **has** more files" where the flag only means "may not be everything" | FIXED | `96eca23` — the copy now matches what the flag means |
| 17 | 3 | Codex | A budget of zero, and a budget stopping AT a directory that turns out to be empty, report truncated | **REJECTED** | `413375b0` — the intended side of the trade, and Codex accepted the reasoning next round. Telling an empty directory from an omitted one means READING it, which is the syscall the budget exists to prevent. Over-reporting costs one line of UI; under-reporting tells someone a file is absent when the finder never looked. Both cases now carry a comment and a test so the choice is on the record |
| 18 | 5 | Codex | The reveal could race the tree load — `files-find` mounts the pane and opens the finder in the same breath, so a pick can land while `roots` is empty and the tree stays collapsed | FIXED | `5b90ff5` — the reveal awaits the startup it races. This was the half of #2099 the issue actually asked for |
| 19 | 5 | Codex | The docs claimed "nothing filters it" for a non-git directory, which the skip list contradicts | FIXED | `5b90ff5` — README, both guides' basics, both keymap tables and the panel now name the skip list |
| 20 | 5 | Codex | The git listing is collected, sorted and de-duplicated before the cap, so a very large repo allocates more than the cap suggests | **REJECTED** | Capping earlier moves nothing: the whole `ls-files` stdout is already one string in memory and is larger than the array derived from it. Bounding it genuinely means streaming the child's stdout, which is a different design than this PR is making. Codex accepted this |
| 21 | 6 | Codex | A tracked entry with a regular-file mode is not necessarily a regular file NOW — replaced by a directory, or left out by a sparse checkout | FIXED | `803c12e`, and it made the route **faster**. Measured on this repository, `ls-files --deleted` costs more than an `lstat` of every candidate *and misses the case*: against a tracked file the worktree turned into a directory it reports nothing while `git status` says `AD`. So the third git call is gone and the filesystem is asked instead — one `lstat` in the ordinary case |
| 22 | 7 | Codex | A test comment claimed "the route cannot be pointed at an arbitrary directory by query tampering", which `resolveBase` does not do | FIXED | `992670b` — the test was right, the sentence above it was not, and it was wrong in the direction that invents a guard |
| 23 | 8 | Windows CI | `chmod 000` does not make a directory unreadable on Windows, so the unreadable-subtree case failed there | FIXED | `796a3c3` — POSIX-only, next to the root skip. The behaviour is platform-independent; only the way to provoke it is not |

**Codex verdict on `992670b`: `CODEX VERDICT: LGTM`.** `796a3c3` is a test-only skip on top of it.

@coderabbitai full review

Gates: `format` / `lint` / `typecheck` / `build` clean, full suite green locally, and Windows CI was green on everything except row 23.

PR #2120, opened 2026-09-17
feat(keymap): focus-prev / focus-next — walk the cursor across the tiled grid (#2106)

### LINKED ISSUE #2106: feat(keymap): focus-prev / focus-next — move between cells while the grid is not zoomed
## What I want

A keymap action that moves keyboard focus to the previous / next cell **while the grid is not zoomed**, the same way `zoom-prev` / `zoom-next` walk the enlarged view.

Something like:

```json
{ "keymap": { "focus-prev": "Cmd+ArrowUp", "focus-next": "Cmd+ArrowDown" } }
```

## What happens today

- `zoom-prev` / `zoom-next` are dropped when nothing is zoomed (`Qjr` in the bundle skips the zoom-only action list), so the same keys do nothing in the grid.
- `next-attention` works in the grid, but it cycles blocked → done → idle and skips working cells, so it is a "go to whoever called" key rather than a way to walk the cells in order.
- The only way to move to a neighbouring cell in the grid is the mouse.

## Why this is not covered yet

#829 chose to reuse `zoomedUid` as "the current cell" so no new selection state was needed, and left "actions that need a selection while not zoomed" as a later item. The closing comment says to open a new issue when picking it up, so here it is.

## Proposal (answering the open points in #829)

- **Selection = the terminal that has DOM focus.** The grid already scrolls to and focuses a cell in `next-attention` (`term.focus()`), so "current cell" can be derived from which xterm holds focus instead of adding grid state. With no focused terminal, fall back to the first cell on the current page.
- **`focus-next` / `focus-prev`** move ±1 along the visible order on the current page, call `term.focus()` on the target, and scroll it into view. Whether to wrap at the page edge or step to the neighbouring page can follow whatever `zoom-next` does today.
- **Visible marker:** a focus ring on the focused cell's border (the header colour already exists per project, so an outline in the accent colour is probably enough). This is what makes the keyboard position readable at a glance in a 3×3 grid.
- Zoomed view unchanged: `zoom-prev` / `zoom-next` keep their current meaning. It would be natural to allow binding the same key to both pairs and pick by state, but two separate actions are fine too.

## Environment

- mulmoterminal 4.26.0 (npx, `mulmoterminal --no-open` inside tmux)
- Node v22.17.1 / Linux x86_64 host, browser on macOS (Cmd-based keymap)
- keymap in use: `zoom-toggle: Cmd+Enter`, `zoom-prev: Cmd+ArrowUp`, `zoom-next: Cmd+ArrowDown`

## Related

- #829 keymap design (closed; "selection while not zoomed" deferred)

isamu:
## Summary

Adds two keymap actions, `focus-next` / `focus-prev`, that move the **keyboard cursor** one cell
along the on-screen order while the grid is **not** zoomed — the un-zoomed counterpart of
`zoom-next` / `zoom-prev`, which move the enlargement. Until now the only way to reach a
neighbouring cell in the tiled grid was the mouse.

Closes #2106.

Most of what the issue proposes already existed under other names, so the change is small:

- **Selection** is `focusedCellUid`, which `GridView.vue` already tracks and which ZOOM INVARIANT 4
  in `gridTabs.ts` already declares ("the enlarged cell while zoomed, the focused cell otherwise").
- **Visible marker** is the existing focused-cell lift (`TerminalGrid.vue` scales the cell and
  drops a shadow under it, un-zoomed only). No new CSS.
- **Page crossing** reuses `revealCell`, the same way `next-attention` brings a cell's page on
  screen un-zoomed.

## Items to Confirm / Review

1. **No state-dependent dispatch — this is the decision the PR rests on.** The issue floats binding
   ONE key to both pairs and picking by zoom state. We are not doing that: `actionForKey` returns
   the lowest-ranked bound action and stops, and `validateKeymap`'s collision reporting is built on
   that (`fallthroughWinner` states only a `send` can be the claim that fires when a conditional
   action stands aside, and a spec pins it). So a key bound to both `zoom-prev` and `focus-prev`
   still resolves to `zoom-prev` alone, and validation warns that `focus-prev` never fires.
   `gridShortcut.spec.ts` pins that as behaviour. Doing it the other way is a separate change to
   the shared resolver — reasoning in `plans/feat-2106-focus-prev-next.md`.
2. **`focus-*` declines while a cell is enlarged**, rather than doing something. That is the mirror
   of `NEEDS_A_CURRENT_TERMINAL`, and it means a `send` bound to the same keystroke fires in that
   state. Worth a look at whether "declines" or "behaves like `zoom-next`" is the better answer —
   the latter is the state-dependent behaviour we rejected above, so this is deliberate.
3. **Empty launch cells are skipped in the walk.** `conn.focus` is a no-op on a cell with no xterm,
   so stepping onto one would read as a dead key. `moveZoom` does NOT skip them (zooming a launch
   form is useful); the asymmetry is intentional but is a judgement call.
4. **With no origin the walk lands on the first terminal of the current page** rather than stepping
   past it. Reachable when nothing has taken the cursor yet, or when the focused cell has closed.
5. The wording in `docs/guide/{en,ja}/basics.md` changed: the old text said an un-zoomed grid has no
   "current terminal", which is now misleading — that is the sentence a reader looking for this
   feature would have stopped at.
6. **`switchTo` now drops the selection when the cell holding it leaves the screen**, and that is a
   behaviour change to three PRE-EXISTING actions — `zoom-toggle`, `next-attention` and
   `terminal-new-here` all read `focusedCellUid`. It came out of the review: a page-tab click leaves
   no cell holding the cursor, so the retained uid named a terminal nobody could see and the walk
   went straight back to the page just left. The condition is **visibility after the switch**, not
   that a tab was clicked — `switchPage` returns the state unchanged for the page already shown, and
   dropping the selection there would break all three for a click that changed nothing. Both halves
   have a spec. This is the part of the diff most worth a human eye, because it changes shipped
   behaviour that the issue never asked about.

## User Prompt

- 状態依存のキー操作だけど実装できる？ (issue #2106 のリンク付き) — #2106 の「同じキーを状態で
  出し分ける」案が実装可能か、という問い。
- 状態で出し分けはしたくないね — 3案（別アクション＋別キー / 同一キーを状態で出し分け /
  既存 `zoom-*` を状態依存化）のうち、状態による出し分けは採らないという判断。

## Approach and key decisions

- `common/keymap.ts` — `NEEDS_NOTHING_ENLARGED`, the mirror of `NEEDS_A_CURRENT_TERMINAL`.
  `standsAside` learns it, so the collision reporting stays honest in both directions: a `send` on
  a `focus-*` key is now correctly named as the claim that fires while a cell IS enlarged.
- `src/composables/gridShortcut.ts` — the one-line ternary became two explicit checks, because
  there are now two opposite conditions and the ternary could not say which one an action failed.
- `src/components/gridTabs.ts` — `moveFocus` / `moveFocusUid`, shaped like the existing
  `nextAttention` / `nextAttentionUid` pair (the caller needs both the new state and the uid to put
  the cursor in). Both refuse while zoomed, holding ZOOM INVARIANT 3 where the arithmetic is rather
  than relying on the caller's guard alone.
- `src/components/GridView.vue` — one branch plus a small `moveGridFocus`, keeping `runShortcut`
  readable.
- Docs: `server/skills/mulmoterminal-keys/SKILL.md` owns `keymap`, so the action table and the
  "don't put both pairs on one key" warning live there; the guide's reference tables in both
  languages follow.

## Verification

- `yarn format` / `yarn lint` / `yarn typecheck` / `yarn build` clean; `yarn test` green across the
  whole suite.
- New specs in `gridTabs.spec.ts`, `gridShortcut.spec.ts`, `keymapSend.spec.ts` and
  `GridView.spec.ts`.
- **Driven in a real browser**, because the grid's key handler is on the path of every keystroke the
  grid can hear. A dev server on a scratch `HOME` (so no real config was touched), ten shells started
  through the launcher and `terminal-new-adjacent`, then the walk driven with Playwright:
  - `focus-next` / `focus-prev` move the focused cell one step each way, and the focus lift follows.
  - Stepping off the last cell of page 1 **switches to page 2 and focuses the cell there** — the part
    a green suite does not prove, since the spec's `conn.focus` is a stub and the real target cell has
    only just mounted.
  - At the far end the key does nothing, which also shows the trailing empty launch cell is skipped.
  - While a cell is enlarged, the key changes nothing.
  - No console errors throughout.

`origin/main` moved during the work, so it is merged in here and the gates were re-run on the
merged result rather than on the branch alone.

work in mulmoterminal3

## Summary by CodeRabbit

- **New Features**
  - Added keyboard shortcuts to move focus to the next or previous terminal in tiled grid view.
  - Navigation crosses pages, skips empty launch cells, and stops at either end.
  - Focus navigation is unavailable while a terminal is enlarged; enlargement navigation remains separate.
  - Added English and Japanese shortcut labels.
  - Preserved the selected terminal when switching to the active page.

- **Documentation**
  - Updated keyboard shortcut guides with usage details and key-conflict warnings.


isamu:
## Review ledger — after CodeRabbit iteration 1

Findings raised so far and what happened to each. Re-flagging a `FIXED` or `REJECTED`
row is welcome **if the resolution is wrong** — say which row and what is wrong with it.

| 1 | CR-1 | CodeRabbit | `switchTo` leaves `focusedCellUid` naming a cell on the page just left, so `focus-next` walks back to that page | FIXED | Real. Reproduced with a failing spec **before** fixing (`cell-1` where `cell-9` was expected), then fixed at the source in `switchTo` rather than inside the walk — `zoom-toggle`, `next-attention` and `terminal-new-here` read the same value and were reading the same stale cell. Commit `139c88a6` |
| 2 | CR-1 | CodeRabbit | basics guide calls the losing binding silent; config guide says startup reports the collision | FIXED | The two guides disagreed and `enforceKeymap` does warn, so basics was the wrong one. Corrected in both languages. Commit `139c88a6` |

### Note on the fix for row 1 — it changes three pre-existing actions

Clearing `focusedCellUid` on a hand page switch is not scoped to the new actions.
`zoom-toggle`, `next-attention` and `terminal-new-here` all read that value, and all
three were reading a cell that is no longer on screen after a page-tab click. They now
fall back to their documented no-selection behaviour instead. That is the correct
reading of ZOOM INVARIANT 4 — the focused cell *is* the un-zoomed selection, and a
selection nobody can see is not one — but it is a behaviour change to shipped actions
and worth a human eye.

### Test evidence

The suite is green; the full run under load produced timeouts in ten unrelated
`test/server/**` worktree/git specs, and every one of them passed on a standalone
re-run (seven as a group, the remaining three one file at a time). None of them is
reachable from this diff.

src/components/GridView.vue
```diff
@@ -402,7 +404,15 @@ const onRunSpare = (uid: number, command: RunCommand) => (state.value = runScrip
...
 const toggleSortMode = () => (state.value = setSortMode(state.value, nextSortMode(state.value.sortMode)));
-const switchTo = (page: number) => (state.value = switchPage(state.value, page));
+// Switching page BY HAND leaves no cell holding the cursor: the page's cells unmount, so nothing
+// emits focus-cell and a retained uid names a terminal that is no longer on screen. INVARIANT 4 makes
+// the focused cell the un-zoomed selection, and a selection nobody can see is not one — walking from
+// it sent the user straight back to the page they had just left (CodeRabbit on #2120). Every other
+// page change moves the cursor with it, so this is the one place the selection is dropped instead.
+const switchTo = (page: number) => {
+  state.value = switchPage(state.value, page);
+  focusedCellUid.value = null;
```
isamu: P2: This clears the grid selection even when the user clicks the already-active tab. `switchPage` intentionally returns the same state for that case (see `gridTabs.ts` and its no-op spec), so the page cells did not unmount and the focused cell is still visible. Clearing it here changes the pre-existing focused-cell actions anyway: `zoom-toggle` now expands the first cell on the page instead of the focused cell, `next-attention` restarts from the beginning instead of from the focused cell, and `terminal-new-here` falls back to the default workspace instead of the focused terminal cwd. I grepped for the pattern and this wrapper is the only occurrence (`switchTo` / `focusedCellUid.value = null`).

Smallest fix: only clear when the page actually changes, e.g. capture `const oldPage = state.value.page`, call `switchPage`, and set `focusedCellUid.value = null` only when `page !== oldPage` (or early-return for the active page). Add a GridView spec that focuses a cell, clicks the current tab, then verifies `zoom-toggle` (or `terminal-new-here`) still uses that focused cell.

What would change my mind: if clicking the active tab is intentionally meant to reset keyboard selection and that behavior is documented/pinned for the three focused-cell consumers above.

isamu:

- P2: `switchTo` clears `focusedCellUid` even when the already-active tab is clicked, changing `zoom-toggle`, `next-attention`, and `terminal-new-here` despite no page switch/unmount happening. Inline comment posted on `src/components/GridView.vue:414`.

| Axis | Result |
|---|---|
| correctness and edge cases | One finding: active-tab clicks now drop the visible focused cell. Smallest fix is to clear focus only when the page actually changes; what would change my mind is an explicit, tested decision that active-tab clicks are a keyboard-selection reset. |
| security (XSS / SSRF / path traversal / injection / secrets) | none |
| tests: does a test go red if the new behaviour breaks | Mostly covered: focused Vitest slice passed (`keymap`, `gridShortcut`, `gridTabs`, `GridView`, keymap labels/settings), and `yarn typecheck` passed. Gap: no GridView test for clicking the already-active tab and then using `zoom-toggle` / `next-attention` / `terminal-new-here`; that is why the P2 survives. |
| API / schema / wire / on-disk compatibility | none. Existing configs remain accepted; older binaries still warn on unknown `focus-*` actions rather than refusing startup, and this branch treats the new actions as known. |
| consistency with the rest of the codebase | none beyond the P2: `switchPage` has a no-op invariant for the already-active page, but the wrapper currently violates the same spirit by clearing focus anyway. Other additions follow the existing keymap/i18n/test patterns. |
| accessibility and i18n lockstep (this repo ships en + ja dictionaries) | none. `keymapLabels.ts`, `en.ts`, and `ja.ts` are updated; existing label tests iterate `KEYMAP_ACTIONS` and passed. |
| TIER-C: the shared dispatch — does adding NEEDS_NOTHING_ENLARGED to gridShortcutFor change what ANY pre-existing action does in ANY state? The old line was `return zoomed || !NEEDS_A_CURRENT_TERMINAL.includes(action) ? action : null;` | none. For pre-existing actions, the new branches are equivalent: `NEEDS_A_CURRENT_TERMINAL` still gates on zoomed, terminal-scoped actions still return null before the gate, and all other existing grid actions still return action in both states. The only new state gate applies to `focus-next` / `focus-prev`. |
| TIER-C: config compatibility — a config.json written for an older version, and the server's refuse-to-start path (server/config/keymap-check.ts). Does validateKeymap still classify every case the way it did, including the collision reporting that #1901 / #1906 established? | none. I checked `validateKeymap`, `checkKeymap`, and the #1901/#1906 collision tests. Fatal vs warning classification is unchanged for old entries; same-key action/send reporting is extended to `focus-*` without promoting a second action into the fallthrough state. |
| TIER-C: the `switchTo` focus reset (src/components/GridView.vue) clears focusedCellUid on a hand page switch. That ALSO changes zoom-toggle, next-attention and terminal-new-here, which read the same value. Is each of those changes correct, and is any of them a regression? | Finding: clearing is correct for a real page switch, but a regression for clicking the already-active tab because the focused terminal remains visible. That changes all three consumers listed above. |
| TIER-C: ZOOM INVARIANTS 1-5 in src/components/gridTabs.ts. Does moveFocus / moveFocusUid hold all five? Invariant 3 in particular ("page is decided at exactly ONE moment") — the new function writes `page` un-zoomed, as nextAttention already does. | none. `moveFocusUid` refuses while zoomed, `moveFocus` never enters/leaves zoom or mutates cells, and its un-zoomed `revealCell` page write matches the existing `nextAttention` shape rather than adding a second zoom-time page decision. |

Verification run: `yarn vitest run test/common/keymap.spec.ts test/common/keymapSend.spec.ts test/src/composables/gridShortcut.spec.ts test/src/components/gridTabs.spec.ts test/src/components/GridView.spec.ts test/src/components/keymapLabels.spec.ts test/src/components/settings/keyboardShortcutsSection.spec.ts` passed (7 files, 361 tests). `yarn typecheck` passed. I did not run a mutation sweep.

isamu:
## Codex cross-review — round 1

<details><summary>Prompt sent (round 1)</summary>

```text
Review PR #2120 at https://github.com/receptron/mulmoterminal/pull/2120.

TIER: C (complex). Reason: the change lands on the grid's global capture-phase keydown
handler and on common/keymap.ts — a module BOTH the Express server (startup validation,
which refuses to boot on a bad keymap) and the Vue UI read. Every keystroke the grid can
hear passes through this dispatch. It also changes the behaviour of THREE pre-existing
actions as a side effect (see the axis list).

  gh api repos/receptron/mulmoterminal/pulls/2120/comments
or top-level comments via:
  gh pr comment 2120

Do NOT run a mutation sweep: `rm -f` is guard-rejected in your sandbox, so a sweep's
restore can fail and every count after that measures the harness rather than the code.
Name any mutation you want run and I will run it and report the result.

  correctness and edge cases
  security (XSS / SSRF / path traversal / injection / secrets)
  tests: does a test go red if the new behaviour breaks
  API / schema / wire / on-disk compatibility
  accessibility and i18n lockstep (this repo ships en + ja dictionaries)
  TIER-C: the shared dispatch — does adding NEEDS_NOTHING_ENLARGED to gridShortcutFor
    change what ANY pre-existing action does in ANY state? The old line was
    `return zoomed || !NEEDS_A_CURRENT_TERMINAL.includes(action) ? action : null;`
  TIER-C: config compatibility — a config.json written for an older version, and the
    server's refuse-to-start path (server/config/keymap-check.ts). Does validateKeymap
    still classify every case the way it did, including the collision reporting that
    #1901 / #1906 established?
  TIER-C: the `switchTo` focus reset (src/components/GridView.vue) clears
    focusedCellUid on a hand page switch. That ALSO changes zoom-toggle,
    next-attention and terminal-new-here, which read the same value. Is each of those
    changes correct, and is any of them a regression?
  TIER-C: ZOOM INVARIANTS 1-5 in src/components/gridTabs.ts. Does moveFocus /
    moveFocusUid hold all five? Invariant 3 in particular ("page is decided at exactly
    ONE moment") — the new function writes `page` un-zoomed, as nextAttention already
    does.

   occurrence NOW — grep for it.

── CONTEXT YOU NEED, WHICH THE DIFF DOES NOT SHOW ───────────────────────────
The user's explicit instruction shaped this design: they do NOT want one keystroke
dispatching to different actions depending on zoom state. So binding one key to both
`zoom-prev` and `focus-prev` deliberately still resolves to ONE action (actionForKey
stops at the first bound action in KEYMAP_ACTIONS order) and validateKeymap warns that
the other never fires. A finding that says "these should pick by state" is out of
scope — but a finding that the CODE or the DOCS misdescribe what actually happens is
exactly what this round is for. plans/feat-2106-focus-prev-next.md records the reasoning.

| 0 | setup | Codex sandbox pre-flight | N/A | vitest exit 0, `yarn typecheck` exit 0, socket bind exit 0, api.github.com 200. Deletions permitted (`rm` works; `rm -f` is guard-rejected) — so NO mutation sweep is to be run by Codex; name any mutation and I will run it |
| 1 | CR-1 | `switchTo` leaves `focusedCellUid` pointing at a cell on the page just left, so `focus-next` walks back to that page | FIXED | CodeRabbit, reproduced by a failing spec first (`cell-1` where `cell-9` was expected). Fixed at the source in `switchTo`, not inside the walk, because `zoom-toggle` / `next-attention` / `terminal-new-here` read the same value |
| 2 | CR-1 | basics guide called the losing binding silent; config guide said startup reports it | FIXED | The two guides disagreed; `enforceKeymap` does warn, so basics was the wrong one. Both languages corrected |

That table is context so you do not spend this round re-deriving what earlier rounds
already decided. It is NOT a list of closed topics: if a resolution is wrong, or a FIXED
row's fix does not do what it claims, say so and mark it 'REOPENING #<row>' with the
specific evidence.

  Then, on its own line:
  not reach and why.

```

</details>

<details><summary>Follow-up call (step C-bis) — prompt and reply, verbatim</summary>

```text
Follow-up on PR #2120, round 1. Do NOT re-review the PR and do NOT post anything.

I ACCEPTED your P2 on src/components/GridView.vue:414, and reproduced it first with a
failing spec: focus cell 2, click the already-active tab, press the `zoom-toggle` key —
it enlarged cell 0 instead of cell 2. So the finding is confirmed, not disputed.

I did NOT use the remedy you proposed (compare the old and new page number). I used the
condition the rule is actually about — whether the focused cell is still ON SCREEN after
the switch:

    const switchTo = (page: number) => {
      state.value = switchPage(state.value, page);
      if (!displayCells.value.some((c) => c.uid === focusedCellUid.value)) focusedCellUid.value = null;
    };

`displayCells` is `cellsToDisplay(orderedCells, state.page, zoomed, …)` — un-zoomed it is
the current page's slice, and `switchPage` clears `expanded`, so after this call it is
always the page slice. It is a computed over `state`, so reading it right after the
assignment sees the new state.

Answer only these three, as plain text:

1. Does the visibility condition cover your finding COMPLETELY, or is there a case where
   the page number is unchanged but a previously-focused cell is no longer displayed (or
   the reverse) that the page-number comparison would handle and this one would not?
2. `switchPage` also drops a trailing launch cell when the page changes. Can that make
   `displayCells` exclude a cell that IS still on screen, or include one that is not?
3. Is there any OTHER caller in this repo that leaves `focusedCellUid` naming a cell that
   is no longer displayed? I fixed `switchTo` only; you said it was the sole occurrence of
   this wrapper, but I am asking about the invariant, not the pattern — for example
   `closeCell`'s page clamp, or `parseGridState` on reload.

If you dispute any answer, say which single change would settle it.

--- reply ---

1. Yes, for the confirmed finding, the visibility condition covers it completely and is stricter than comparing page numbers. A page-number comparison would wrongly clear focus on “page changed but the same focused cell is still rendered” cases; this condition keeps it only if the cell is actually still in `displayCells`. The one caveat is that `displayCells` can include a collection-claimed off-page cell, so it means “rendered by the grid,” not necessarily “in the page slice.”

2. No. Dropping a trailing launch cell cannot make `displayCells` exclude an occupied cell that is still on screen, or include an occupied cell that is not on screen. It can only remove the abandoned launcher from the ordered list before the new page slice is computed.

3. Yes. `onClose` can leave `focusedCellUid` naming a cell that no longer exists/displayed when the focused cell is closed. `parseGridState` on reload does not, because `focusedCellUid` is a fresh component ref and is not restored from persisted grid state. `closeCell`’s page clamp is not the problem by itself; closing the focused uid is. The single change that would settle it is clearing `focusedCellUid` in `onClose` when `uid === focusedCellUid.value`, or more generally applying the same post-mutation displayed-cell check after close.
```

</details>

Codex posted its own findings and verdict directly (inline on `GridView.vue:414`, plus the axis table). Its full stdout for round 1 is short and is reproduced here:

```text
Posted the review.

Found one P2: `switchTo` clears `focusedCellUid` even when clicking the already-active tab, which changes `zoom-toggle`, `next-attention`, and `terminal-new-here` despite no page change. I left an inline comment on `src/components/GridView.vue:414` and a top-level `CODEX VERDICT: CHANGES REQUESTED` with the required axis table.

Verification run:
`yarn vitest run ...` passed: 7 files, 361 tests.
`yarn typecheck` passed.
```

### Triage

| # | finding | disposition | why |
|---|---------|-------------|-----|
| 3 | P2: `switchTo` cleared the selection for the **already-active** tab too, where `switchPage` is a documented no-op and nothing unmounts | FIXED | Confirmed with a failing spec before fixing — `zoom-toggle` enlarged cell 0 with cell 2 focused. Fixed by keying on **visibility** (`displayCells`) rather than the page-number comparison Codex proposed; in the follow-up call Codex confirmed visibility covers the finding completely and is the stricter of the two. `085b1be5` |
| 4 | `onClose` also leaves `focusedCellUid` naming a cell that no longer exists (raised in the follow-up) | DEFERRED | **Verified inert, not deferred on cost.** Every consumer already degrades exactly as clearing would: `moveFocusUid` treats it as no origin (tested), `toggleZoom` / `nextAttention` get `indexOf` −1 and fall back to the page start, `terminal-new-here` falls through `preferredLaunchDir` and `insertCellAfter` appends on an unknown uid. The uid can never name a **different** cell either — `nextUid` only increases within a session and a reload makes a fresh ref. `onClose` predates this PR. No issue filed: there is no observable behaviour to fix, and filing one would ask a reader to chase a symptom that cannot occur. |

isamu:
CODEX VERDICT: LGTM

| Axis | Result |
|---|---|
| correctness and edge cases | none. I re-read the current head including `085b1be5`; the visibility-based `switchTo` reset keeps the active-tab selection and clears only when the focused uid is no longer rendered after the switch. `moveFocusUid` covers null/stale origins, launch-cell skipping, page-edge movement, ends, and zoom refusal. |
| security (XSS / SSRF / path traversal / injection / secrets) | none. The change is local key dispatch, grid state, docs, labels, and tests; it does not introduce new URL/file/path handling, HTML injection, secret exposure, or server-side request surface. |
| tests: does a test go red if the new behaviour breaks | none. The focused suite now includes the round-2 active-tab regression plus page-switch/stale-origin/focus-walk/dispatch/collision/i18n-label coverage. I ran `yarn vitest run test/common/keymap.spec.ts test/common/keymapSend.spec.ts test/src/composables/gridShortcut.spec.ts test/src/components/gridTabs.spec.ts test/src/components/GridView.spec.ts test/src/components/keymapLabels.spec.ts test/src/components/settings/keyboardShortcutsSection.spec.ts`: 7 files, 362 tests passed. `yarn typecheck` passed. I did not run a mutation sweep. |
| API / schema / wire / on-disk compatibility | none. Existing config keys retain their prior fatal/warning classification; unknown actions remain non-fatal for downgrade/newer-config compatibility; adding `focus-next` / `focus-prev` only makes those names known to this version. No wire or persisted grid schema changes. |
| consistency with the rest of the codebase | none. The implementation follows existing `zoom-*` / `next-attention` patterns: state gates live in `common/keymap.ts`, dispatch stays in `gridShortcutFor`, page revelation uses `revealCell`, and UI labels are backed by the full `Record<KeymapAction, string>` pattern. |
| accessibility and i18n lockstep (this repo ships en + ja dictionaries) | none. `keymapLabels.ts`, `src/i18n/en.ts`, and `src/i18n/ja.ts` are updated together, and the docs/guide additions are present in both English and Japanese. Existing label/settings tests cover the new actions. |
| TIER-C: the shared dispatch — does adding NEEDS_NOTHING_ENLARGED to gridShortcutFor change what ANY pre-existing action does in ANY state? The old line was `return zoomed || !NEEDS_A_CURRENT_TERMINAL.includes(action) ? action : null;` | none. For every pre-existing action, the new branches are equivalent to the old expression: terminal-scoped actions still return null first, `NEEDS_A_CURRENT_TERMINAL` actions still require `zoomed`, and all other existing grid actions still return in both states. The only new state gate applies to `focus-next` / `focus-prev`. |
| TIER-C: config compatibility — a config.json written for an older version, and the server's refuse-to-start path (server/config/keymap-check.ts). Does validateKeymap still classify every case the way it did, including the collision reporting that #1901 / #1906 established? | none. `validateKeymap` still reports malformed bindings/send entries as fatal, unknown actions as non-fatal, and duplicate/collision losers as warnings. The #1901/#1906 model is preserved: conditional action + send names the send as the other-state winner, while a second action on the same key is not promoted into that gap. The new mirror condition is covered for `focus-*`. |
| TIER-C: the `switchTo` focus reset (src/components/GridView.vue) clears focusedCellUid on a hand page switch. That ALSO changes zoom-toggle, next-attention and terminal-new-here, which read the same value. Is each of those changes correct, and is any of them a regression? | none. For a real page switch, clearing is correct because the selected cell is off-screen/unmounted, so `zoom-toggle`, `next-attention`, `terminal-new-here`, and the new focus walk should use their no-visible-selection behaviour. For clicking the already-active tab, the round-2 visibility check retains the selection, so the regression from round 1 is fixed. |
| TIER-C: ZOOM INVARIANTS 1-5 in src/components/gridTabs.ts. Does moveFocus / moveFocusUid hold all five? Invariant 3 in particular ("page is decided at exactly ONE moment") — the new function writes `page` un-zoomed, as nextAttention already does. | none. `moveFocusUid` refuses while zoomed, so it does not create a second zoom-time page decision. `moveFocus` never enters/leaves zoom, never mutates cells, uses the supplied full order with `revealCell`, and only writes `page` in the un-zoomed case where `nextAttention` already does the same. |

isamu:
## Codex cross-review — round 2 (LGTM) and the browser check

Round 2 reviewed `085b1be5` — the head carrying the round-1 fix — and returned **`CODEX VERDICT: LGTM`** with every axis answered and `FINDINGS COMPLETE`. Its verdict comment is above; nothing was pushed during that round.

<details><summary>Round 2 prompt sent (verbatim)</summary>

```text
Review PR #2120 at https://github.com/receptron/mulmoterminal/pull/2120. This is ROUND 2.

Since round 1 the head moved from 139c88a6 to 085b1be5. Round 1's P2 is fixed, with the
remedy you accepted in the follow-up call (visibility via `displayCells`, not a page-number
comparison). Review the CURRENT head — including that fix, which has not been reviewed by
anyone yet.

TIER: C (complex). Reason: the change lands on the grid's global capture-phase keydown
handler and on common/keymap.ts — a module BOTH the Express server (startup validation,
which refuses to boot on a bad keymap) and the Vue UI read. Every keystroke the grid can
hear passes through this dispatch. It also changes the behaviour of THREE pre-existing
actions as a side effect (see the axis list).

  gh api repos/receptron/mulmoterminal/pulls/2120/comments
or top-level comments via:
  gh pr comment 2120

Do NOT run a mutation sweep: `rm -f` is guard-rejected in your sandbox, so a sweep's
restore can fail and every count after that measures the harness rather than the code.
Name any mutation you want run and I will run it and report the result.

  correctness and edge cases
  security (XSS / SSRF / path traversal / injection / secrets)
  tests: does a test go red if the new behaviour breaks
  API / schema / wire / on-disk compatibility
  accessibility and i18n lockstep (this repo ships en + ja dictionaries)
  TIER-C: the shared dispatch — does adding NEEDS_NOTHING_ENLARGED to gridShortcutFor
    change what ANY pre-existing action does in ANY state? The old line was
    `return zoomed || !NEEDS_A_CURRENT_TERMINAL.includes(action) ? action : null;`
  TIER-C: config compatibility — a config.json written for an older version, and the
    server's refuse-to-start path (server/config/keymap-check.ts). Does validateKeymap
    still classify every case the way it did, including the collision reporting that
    #1901 / #1906 established?
  TIER-C: the `switchTo` focus reset (src/components/GridView.vue) clears
    focusedCellUid on a hand page switch. That ALSO changes zoom-toggle,
    next-attention and terminal-new-here, which read the same value. Is each of those
    changes correct, and is any of them a regression?
  TIER-C: ZOOM INVARIANTS 1-5 in src/components/gridTabs.ts. Does moveFocus /
    moveFocusUid hold all five? Invariant 3 in particular ("page is decided at exactly
    ONE moment") — the new function writes `page` un-zoomed, as nextAttention already
    does.

   occurrence NOW — grep for it.

── CONTEXT YOU NEED, WHICH THE DIFF DOES NOT SHOW ───────────────────────────
The user's explicit instruction shaped this design: they do NOT want one keystroke
dispatching to different actions depending on zoom state. So binding one key to both
`zoom-prev` and `focus-prev` deliberately still resolves to ONE action (actionForKey
stops at the first bound action in KEYMAP_ACTIONS order) and validateKeymap warns that
the other never fires. A finding that says "these should pick by state" is out of
scope — but a finding that the CODE or the DOCS misdescribe what actually happens is
exactly what this round is for. plans/feat-2106-focus-prev-next.md records the reasoning.

| 0 | setup | Codex sandbox pre-flight | N/A | vitest exit 0, `yarn typecheck` exit 0, socket bind exit 0, api.github.com 200. Deletions permitted (`rm` works; `rm -f` is guard-rejected) — so NO mutation sweep is to be run by Codex; name any mutation and I will run it |
| 1 | CR-1 | `switchTo` leaves `focusedCellUid` pointing at a cell on the page just left, so `focus-next` walks back to that page | FIXED | CodeRabbit, reproduced by a failing spec first (`cell-1` where `cell-9` was expected). Fixed at the source in `switchTo`, not inside the walk, because `zoom-toggle` / `next-attention` / `terminal-new-here` read the same value |
| 2 | CR-1 | basics guide called the losing binding silent; config guide said startup reports it | FIXED | The two guides disagreed; `enforceKeymap` does warn, so basics was the wrong one. Both languages corrected |
| 3 | 1 | Codex P2: `switchTo` cleared the selection for the ALREADY-ACTIVE tab too, where `switchPage` is a no-op and nothing unmounted | FIXED | Confirmed by a failing spec first (`zoom-toggle` enlarged cell 0 with cell 2 focused). Fixed by keying on VISIBILITY (`displayCells`) rather than the page number Codex proposed; Codex confirmed in C-bis that visibility covers the finding completely and is the stricter of the two. Commit `085b1be5` |
| 4 | 1 | Codex (C-bis q3): `onClose` also leaves `focusedCellUid` naming a cell that no longer exists | DEFERRED | Verified inert, not deferred on cost. Every consumer already degrades exactly as clearing would: `moveFocusUid` treats it as no origin (tested), `toggleZoom`/`nextAttention` get `indexOf` -1 and fall back to the page start, `terminal-new-here` falls through `preferredLaunchDir` and `insertCellAfter` appends on an unknown uid. And the uid can never name a DIFFERENT cell — `nextUid` only increases within a session and a reload makes a fresh ref. `onClose` predates this PR and nothing in it is reached differently because of this change. No issue filed: there is no observable behaviour to fix |

That table is context so you do not spend this round re-deriving what earlier rounds
already decided. It is NOT a list of closed topics: if a resolution is wrong, or a FIXED
row's fix does not do what it claims, say so and mark it 'REOPENING #<row>' with the
specific evidence.

  Then, on its own line:
  not reach and why.

```

</details>

<details><summary>Round 2 stdout (verbatim)</summary>

```text
Posted the round-2 review comment: LGTM, with every requested axis answered and the completeness line included.

Verification I ran:
`yarn vitest run ...` passed: 7 files, 362 tests.
`yarn typecheck` passed.

Comment URL: https://github.com/receptron/mulmoterminal/pull/2120#issuecomment-5723351922
```

</details>

### Driven in a real browser

The page-tab rule changes a control every user clicks, so both branches were exercised against a running server (scratch `HOME`, ten shells, Playwright):

```
A. focused before clicking the active tab: {"page":"1","focused":"2"}
A. focused after  clicking the active tab: {"page":"1","focused":"2"}
A. zoom-toggle enlarges (must be 2, not 0): {"enlarged":"2"}
B. focused before switching page:          {"page":"1","focused":"2"}
B. after switching to page 2:              {"page":"2","focused":"9"}
B. focus-next lands on:                    {"page":"2","focused":"9"}
console errors: []
```

A is the finding both bots raised, gone. B is the original defect, gone — the walk stays on the page the user switched to.

### One more round is owed

The plan file has been brought up to date with the page-tab rule and pushed as `29de2fcb`. That is a doc-only commit, but an unreviewed push is unreviewed, so round 3 is running against it. The LGTM above covers `085b1be5`, not the current head.

isamu:
CODEX VERDICT: LGTM

| Axis | Result |
|---|---|
| correctness and edge cases | none — current head is `29de2fcb`; the only change since my round-2 LGTM at `085b1be5` is `plans/feat-2106-focus-prev-next.md`, and the new plan text matches the implemented visible-after-switch rule. |
| security (XSS / SSRF / path traversal / injection / secrets) | none — round-3 delta is plan text only, and the reviewed code path adds no new parsing of untrusted HTML/URLs/paths, no secret handling, and no network/file access. |
| tests: does a test go red if the new behaviour breaks | none — the plan accurately names the red tests that cover the behavior: `GridView.spec.ts` for focus walk/page-tab reset/active-tab survival, `gridTabs.spec.ts` for focus arithmetic and zoom refusal, `gridShortcut.spec.ts` for dispatch gating/state-independent same-key behavior, and `keymapSend.spec.ts` for validation collision wording. I did not rerun the suite in round 3 because source and tests did not move after `085b1be5`. |
| API / schema / wire / on-disk compatibility | none — adding `focus-next` / `focus-prev` as recognized keymap actions is backward-compatible for older configs; unknown-action downgrade behavior remains warning-only, malformed known bindings remain fatal, and no persisted grid/session schema or wire shape changes. |
| consistency with the rest of the codebase | none — the plan now reflects the codebase pattern: shared precondition lists live in `common/keymap.ts`, grid dispatch declines by state, page changes use `displayCells`, and focus movement mirrors existing `nextAttention` / `revealCell` behavior. |
| accessibility and i18n lockstep (this repo ships en + ja dictionaries) | none — Settings labels were added in both `src/i18n/en.ts` and `src/i18n/ja.ts`, `keymapLabels.ts` references both new actions, and guide/config documentation is updated in both English and Japanese. |
| TIER-C: the shared dispatch — does adding NEEDS_NOTHING_ENLARGED to gridShortcutFor change what ANY pre-existing action does in ANY state? | none — the old rule still holds for all pre-existing actions: `NEEDS_A_CURRENT_TERMINAL` actions only fire zoomed, terminal-scoped actions still fall through, and all other existing actions still return `action`. `NEEDS_NOTHING_ENLARGED` only gates the two new focus actions. Same-key `zoom-prev` + `focus-prev` still resolves to one action, not state-dependent dispatch. |
| TIER-C: config compatibility — a config.json written for an older version, and the server's refuse-to-start path (server/config/keymap-check.ts). Does validateKeymap still classify every case the way it did, including the collision reporting that #1901 / #1906 established? | none — `server/config/keymap-check.ts` still maps `validateKeymap` fatal problems to refusal and nonfatal problems to warnings. Existing classifications remain intact; the new `standsAside` case extends the #1901/#1906 conditional-collision model to `focus-*` without promoting action-to-action fallthrough. |
| TIER-C: the `switchTo` focus reset (src/components/GridView.vue) clears focusedCellUid on a hand page switch. That ALSO changes zoom-toggle, next-attention and terminal-new-here, which read the same value. Is each of those changes correct, and is any of them a regression? | none — the plan now correctly records that `switchTo` clears only when the previously focused uid is not visible after the switch. Clearing stale off-page selection is correct for `zoom-toggle`, `next-attention`, and `terminal-new-here`; preserving selection when clicking the active tab avoids the regression found in round 1. |
| TIER-C: ZOOM INVARIANTS 1-5 in src/components/gridTabs.ts. Does moveFocus / moveFocusUid hold all five? Invariant 3 in particular ("page is decided at exactly ONE moment") — the new function writes `page` un-zoomed, as nextAttention already does. | none — `moveFocusUid` refuses while zoomed, so it does not write `page` during the zoomed state; un-zoomed movement uses `revealCell`, never enters/leaves zoom, does not mutate cells, and keeps the single current-terminal notion as live focus. The plan wording matches that implementation. |

isamu:
## Codex cross-review — round 3 (LGTM) and the final ledger

Round 3 reviewed `29de2fcb` — the head that will merge — and returned **`CODEX VERDICT: LGTM`** with `FINDINGS COMPLETE`. Nothing was pushed during it, so this is the clean round the loop exits on.

<details><summary>Round 3 prompt sent (verbatim)</summary>

```text
Review PR #2120 at https://github.com/receptron/mulmoterminal/pull/2120. This is ROUND 3.

You gave LGTM on 085b1be5 in round 2. The head has since moved to 29de2fcb, which adds
ONE commit and it touches only plans/feat-2106-focus-prev-next.md — recording the page-tab
selection rule that came out of your round-1 finding. No source, test, or doc-site file
changed. An unreviewed push is unreviewed however small, which is why this round exists.

Review the CURRENT head. Concentrate on whether the plan file now describes what the code
actually does — a plan that misdescribes the change is a claim a future reader inherits —
and confirm nothing else moved. Still answer every axis.

TIER: C (complex). Reason: the change lands on the grid's global capture-phase keydown
handler and on common/keymap.ts — a module BOTH the Express server (startup validation,
which refuses to boot on a bad keymap) and the Vue UI read. Every keystroke the grid can
hear passes through this dispatch. It also changes the behaviour of THREE pre-existing
actions as a side effect (see the axis list).

  gh api repos/receptron/mulmoterminal/pulls/2120/comments
or top-level comments via:
  gh pr comment 2120

Do NOT run a mutation sweep: `rm -f` is guard-rejected in your sandbox, so a sweep's
restore can fail and every count after that measures the harness rather than the code.
Name any mutation you want run and I will run it and report the result.

  correctness and edge cases
  security (XSS / SSRF / path traversal / injection / secrets)
  tests: does a test go red if the new behaviour breaks
  API / schema / wire / on-disk compatibility
  accessibility and i18n lockstep (this repo ships en + ja dictionaries)
  TIER-C: the shared dispatch — does adding NEEDS_NOTHING_ENLARGED to gridShortcutFor
    change what ANY pre-existing action does in ANY state? The old line was
    `return zoomed || !NEEDS_A_CURRENT_TERMINAL.includes(action) ? action : null;`
  TIER-C: config compatibility — a config.json written for an older version, and the
    server's refuse-to-start path (server/config/keymap-check.ts). Does validateKeymap
    still classify every case the way it did, including the collision reporting that
    #1901 / #1906 established?
  TIER-C: the `switchTo` focus reset (src/components/GridView.vue) clears
    focusedCellUid on a hand page switch. That ALSO changes zoom-toggle,
    next-attention and terminal-new-here, which read the same value. Is each of those
    changes correct, and is any of them a regression?
  TIER-C: ZOOM INVARIANTS 1-5 in src/components/gridTabs.ts. Does moveFocus /
    moveFocusUid hold all five? Invariant 3 in particular ("page is decided at exactly
    ONE moment") — the new function writes `page` un-zoomed, as nextAttention already
    does.

   occurrence NOW — grep for it.

── CONTEXT YOU NEED, WHICH THE DIFF DOES NOT SHOW ───────────────────────────
The user's explicit instruction shaped this design: they do NOT want one keystroke
dispatching to different actions depending on zoom state. So binding one key to both
`zoom-prev` and `focus-prev` deliberately still resolves to ONE action (actionForKey
stops at the first bound action in KEYMAP_ACTIONS order) and validateKeymap warns that
the other never fires. A finding that says "these should pick by state" is out of
scope — but a finding that the CODE or the DOCS misdescribe what actually happens is
exactly what this round is for. plans/feat-2106-focus-prev-next.md records the reasoning.

| 0 | setup | Codex sandbox pre-flight | N/A | vitest exit 0, `yarn typecheck` exit 0, socket bind exit 0, api.github.com 200. Deletions permitted (`rm` works; `rm -f` is guard-rejected) — so NO mutation sweep is to be run by Codex; name any mutation and I will run it |
| 1 | CR-1 | `switchTo` leaves `focusedCellUid` pointing at a cell on the page just left, so `focus-next` walks back to that page | FIXED | CodeRabbit, reproduced by a failing spec first (`cell-1` where `cell-9` was expected). Fixed at the source in `switchTo`, not inside the walk, because `zoom-toggle` / `next-attention` / `terminal-new-here` read the same value |
| 2 | CR-1 | basics guide called the losing binding silent; config guide said startup reports it | FIXED | The two guides disagreed; `enforceKeymap` does warn, so basics was the wrong one. Both languages corrected |
| 3 | 1 | Codex P2: `switchTo` cleared the selection for the ALREADY-ACTIVE tab too, where `switchPage` is a no-op and nothing unmounted | FIXED | Confirmed by a failing spec first (`zoom-toggle` enlarged cell 0 with cell 2 focused). Fixed by keying on VISIBILITY (`displayCells`) rather than the page number Codex proposed; Codex confirmed in C-bis that visibility covers the finding completely and is the stricter of the two. Commit `085b1be5` |
| 4 | 1 | Codex (C-bis q3): `onClose` also leaves `focusedCellUid` naming a cell that no longer exists | DEFERRED | Verified inert, not deferred on cost. Every consumer already degrades exactly as clearing would: `moveFocusUid` treats it as no origin (tested), `toggleZoom`/`nextAttention` get `indexOf` -1 and fall back to the page start, `terminal-new-here` falls through `preferredLaunchDir` and `insertCellAfter` appends on an unknown uid. And the uid can never name a DIFFERENT cell — `nextUid` only increases within a session and a reload makes a fresh ref. `onClose` predates this PR and nothing in it is reached differently because of this change. No issue filed: there is no observable behaviour to fix |

That table is context so you do not spend this round re-deriving what earlier rounds
already decided. It is NOT a list of closed topics: if a resolution is wrong, or a FIXED
row's fix does not do what it claims, say so and mark it 'REOPENING #<row>' with the
specific evidence.

  Then, on its own line:
  not reach and why.

```

</details>

<details><summary>Round 3 stdout (verbatim)</summary>

```text
Posted the round-3 review comment: LGTM.

Comment: https://github.com/receptron/mulmoterminal/pull/2120#issuecomment-5723396159

I verified the incremental diff from `085b1be5` to `29de2fcb` is only `plans/feat-2106-focus-prev-next.md`, then re-read the split PR diff by file group before posting.
```

</details>

### Final ledger

| 0 | setup | Codex sandbox pre-flight | N/A | vitest exit 0, `yarn typecheck` exit 0, socket bind exit 0, api.github.com 200. Deletions permitted (`rm` works; `rm -f` is guard-rejected) — so NO mutation sweep is to be run by Codex; name any mutation and I will run it |
| 1 | CR-1 | `switchTo` leaves `focusedCellUid` pointing at a cell on the page just left, so `focus-next` walks back to that page | FIXED | CodeRabbit, reproduced by a failing spec first (`cell-1` where `cell-9` was expected). Fixed at the source in `switchTo`, not inside the walk, because `zoom-toggle` / `next-attention` / `terminal-new-here` read the same value |
| 2 | CR-1 | basics guide called the losing binding silent; config guide said startup reports it | FIXED | The two guides disagreed; `enforceKeymap` does warn, so basics was the wrong one. Both languages corrected |
| 3 | 1 | Codex P2: `switchTo` cleared the selection for the ALREADY-ACTIVE tab too, where `switchPage` is a no-op and nothing unmounted | FIXED | Confirmed by a failing spec first (`zoom-toggle` enlarged cell 0 with cell 2 focused). Fixed by keying on VISIBILITY (`displayCells`) rather than the page number Codex proposed; Codex confirmed in C-bis that visibility covers the finding completely and is the stricter of the two. Commit `085b1be5` |
| 4 | 1 | Codex (C-bis q3): `onClose` also leaves `focusedCellUid` naming a cell that no longer exists | DEFERRED | Verified inert, not deferred on cost. Every consumer already degrades exactly as clearing would: `moveFocusUid` treats it as no origin (tested), `toggleZoom`/`nextAttention` get `indexOf` -1 and fall back to the page start, `terminal-new-here` falls through `preferredLaunchDir` and `insertCellAfter` appends on an unknown uid. And the uid can never name a DIFFERENT cell — `nextUid` only increases within a session and a reload makes a fresh ref. `onClose` predates this PR and nothing in it is reached differently because of this change. No issue filed: there is no observable behaviour to fix |
| 5 | 2 | Round 2 on `085b1be5` — no findings on any axis | CLEAN | `CODEX VERDICT: LGTM`, every axis answered, `FINDINGS COMPLETE`. Nothing pushed during the round |
| 6 | 3 | Round 3 on `29de2fcb` (plan file brought up to date) — no findings | CLEAN | `CODEX VERDICT: LGTM` with `FINDINGS COMPLETE`. Codex confirmed the incremental diff is the plan file alone before re-reading the whole PR |


### What the two reviewers cost, and what they bought

One real defect, found twice independently. CodeRabbit found the stale selection after a page switch; Codex found that the first fix for it was too broad and dropped a **visible** selection when the tab already shown was clicked. Neither was reachable from the feature the issue asked for — both live in `switchTo`, a control this PR was never about — and both change three pre-existing actions.

Tier **C** throughout, for the grid dispatch and for `common/keymap.ts` being read by the server as well as the browser. Every axis was answered on every round, including the four tier-C ones.

Deferred, with evidence: `onClose` leaves `focusedCellUid` naming a closed cell. Verified inert rather than deferred on cost — every consumer already degrades exactly as clearing would, and a uid cannot be reused within a session. It predates this PR and no issue was filed, because there is no observable behaviour to fix.

PR #2122, opened 2026-09-18
fix(roster): read a cell's prompt and reply from its own agent's log (#2121)

### LINKED ISSUE #2121: bug: codex セルの cockpit roster が summary / prompt / reply を出さない（/api/session/:id が claude のトランスクリプトしか読まない）
## What happened

グリッドの cockpit roster 行は、セルごとに `summary` / `prompt` / `reply` の3行を出す。
**codex で起動したセルでは、この3行がすべて空のまま**になる（claude のセルでは埋まる）。

読んだ限り、原因はロースターが読むルートが agent を見ていないこと:

- `server/routes/session-routes.ts:107` — `?agent=` が効くのは2つのバッジ（モデル名・トークン）だけで、
  コメントにも "nothing else on this route" と書かれている。
- `server/session/session-reads.ts:186` — `readSessionSummary()` は agent を問わず
  `projectSessionsDir(cwd)/<id>.jsonl`（= `~/.claude/projects/<encoded-cwd>/<id>.jsonl`）だけを開く。
  codex は自分で id を採番して `~/.codex/sessions/**/rollout-*.jsonl` に書くので、このパスにファイルは無く、
  `EMPTY_SUMMARY` が返る → `lastPrompt` / `lastResponse` が null。
- `server/session/session-title.ts:73` — AI タイトル（`summary` 行）も同じパスを読むので同様に空。
- `server/routes/hook-routes.ts:152` — メモリ側の `lastPrompts` は Claude Code の `UserPromptSubmit`
  フックでしか埋まらない。codex はこのフックを撃たないので、フォールバックも効かない。

codex の rollout から直近のユーザー発話を取り出す実装はすでにある
（`server/agents/codex-user-turn.ts` の `codexUserTurn` / `codexUserPrompt`）。
履歴一覧（`codex-sessions.ts`）、プロンプト履歴ペイン（`prompt-history.ts`）、
`/api/transcript/last-turn`（`last-turn.ts`）、transcript ビュー（`transcript-view-codex.ts`）が使っている。
つまりデータも読み口も揃っていて、`/api/session/:id` だけがそこを見ていない、という形に見える。

#1465 でバッジは Codex / Antigravity / Grok に広がったが、roster の3行は claude のままになっている。

## What I expected

codex のセルでも、claude のセルと同じように roster 行に直近のプロンプトと返答（と要約）が出ること。
少なくとも `prompt` 行は、`codexUserPrompt` がすでに読めているので出せるはず。

## Steps to reproduce

1. セルを codex で起動し、1ターン以上やり取りする。
2. グリッドをズームして cockpit roster を開く。
3. codex のセルの行を見る → `summary` / `prompt` / `reply` が空。同じ画面の claude のセルは埋まっている。

補足（このマシンで確認した裏付け）:

- codex の rollout は `~/.codex/sessions/2026/09/02/rollout-...-<session-id>.jsonl` にあり、
  `session_meta.session_id` が codex の採番した id。
- 同じ id の `~/.claude/projects/*/<session-id>.jsonl` は存在しない
  （`ls` が no matches）。`readSessionSummary()` が開こうとしているのはこのパス。

## Environment

- mulmoterminal 5.0.1 / Node v24.18.1 / macOS (Darwin arm64) / zsh
- tmux 3.7b / gh 2.92.0 / claude 2.1.275 / codex-cli 0.155.0
- 起動: ローカルの開発チェックアウト（`~/Developer/mulmoterminal`）

## Attachments

なし（症状はロースター行が空であること）。

isamu:
## Summary

`GET /api/session/:id` took `?agent=` and let it decide ONE thing: where the two header badges are
read from (#1465). `lastPrompt` and `lastResponse` still came out of `readSessionSummary`, which
opens claude's per-project transcript and nothing else. A codex session has no file there — codex
mints its own id and writes `$CODEX_HOME/sessions/**/rollout-*-<id>.jsonl` — so the route answered
nulls, and the cockpit roster's `prompt` and `reply` lines were blank for every codex cell while the
claude cell beside it was filled (#2121).

The rule the badges already follow now covers the exchange too: **a field read from a session's
conversation is read from THAT AGENT's log.** The route asks `sessionLastTurn`, which has been
agent-branched all along — it reads codex's rollout and cursor's transcript, and answers the agents
whose logs have no reader yet with the empty turn, which is what this route was already showing
them. An eighth agent — seven are hosted today — gains these lines by teaching that one function,
not this route.

The other half was on the client: the roster's own poll sent no `?agent=` at all, so a fixed route
would have kept answering it as claude. Three call sites decide this and two had it wrong:

| call site | before | now |
|---|---|---|
| `server/routes/session-routes.ts` | claude's transcript, whatever `?agent=` said | the agent's own log |
| `src/components/GridView.vue` (`seedMeta`) | no `?agent=` | the cell's agent |
| `useSessionSummary` + `CollectionChatPane.vue` | no `?agent=` | the held chat's agent |

`TerminalCell` has sent it since #1465 and is unchanged.

## Items to Confirm / Review

- **The `summary` line stays empty for a codex cell, deliberately.** It is claude's `ai-title`,
  which Claude Code writes into its own transcript; codex writes no equivalent anywhere, so there is
  nothing to read. Filling it would mean generating a title (the `headless` source, which is off by
  default even for claude) or promoting codex's first user prompt — a different value wearing the
  same label. Filed as #2123 with the options costed; #2121 asks for all three lines and this
  delivers two.
- **Per-poll cost, raised by Codex as a P2 and fixed.** Sending `?agent=` from the roster is what
  made this matter: before, the roster did no codex lookups at all, and after, a codex cell resolved
  its rollout twice per poll — `agentBadges` and `sessionLastTurn` each call `codexRolloutPath`,
  which walks every day directory under `$CODEX_HOME/sessions` with a synchronous `readdirSync`.
  The walk, not the tail read, is the dominant cost, and it repeats for the same id for as long as
  the cell lives. So the lookup is now memoized at process level rather than reused per request:
  a hit is re-checked with one `existsSync` (codex prunes), a miss is never remembered (a cell whose
  rollout codex has not written yet must not read as absent until restart), and the map is capped
  because nothing prunes it. The before/after bench and the command that produced it are in the
  round-1 thread below. **Worth a look:** whether a process-lifetime memo is the right call here
  versus its own PR — Codex judged it in scope and sound, and I would rather a human agreed.

- **`publishActivity`'s `refreshLastResponse`** (`server/session/lifecycle.ts`) reads claude's
  transcript whatever the agent. I wrote here that this means "the pub/sub row carries no reply for
  a codex session" and **that was wrong** — checked afterwards, it is unreachable on four
  independent grounds: it is gated on `activity.waiting`, which codex never sets; a failed read
  writes nothing; `ActivityPush` does not accept `lastResponse` at all, so no client reads it off
  that channel; and the roster takes these fields from this route regardless. Recorded as #2124 as
  a latent trap rather than a defect, with the evidence, so nobody redoes the analysis.

- **One incidental fix rides along, and it is pinned:** an agent whose log has no reader yet
  (grok, muse, agy, copilot) was being handed CLAUDE's transcript whenever the two ids collided.
  It now answers nulls.

## Verification

- New route-level spec `test/server/routes/session-detail-agent.spec.ts`: a codex rollout on disk
  with no claude transcript for the id. **Two of its five cases go red without the route change**,
  including the collision above.
- The client specs pin that the agent reaches the query string, and each goes red without its own
  half of the fix.
- **Driven against the real stores on this machine**, through the real Express route mounted as the
  app mounts it: a real codex session answered `?agent=codex` with the prompt and reply that are in
  its rollout; the same id with no `?agent=` still answered nulls; a real claude session answered
  identically with and without the param, which is the behaviour-preservation half.
- `yarn format` / `lint` / `typecheck` / `build` / `test` all clean locally, the full suite with
  nothing failing and nothing skipped beyond what this repo already skips. An earlier full run on
  this machine timed out two untouched spec files while the load average was above 50; both pass
  standalone and both pass in the clean run.
- **Not exercised: the browser rendering a codex roster row.** Port 34567 is held by a running
  MulmoTerminal on this machine, and booting a second instance runs the boot-time idle-tmux sweep
  against the same tmux server and shares `~/.mulmoterminal` — so it was not run. `rosterRow` and
  the roster template are untouched and already render these two fields for any row.

## User Prompt

- 調べて直して https://github.com/receptron/mulmoterminal/issues/2121
- （実機確認について）ルート検証で十分

Closes #2121

work in mulmoterminal4

## Summary by CodeRabbit

- **Bug Fixes**
  - Session details, roster metadata, and collection summaries now retrieve prompts and replies from the selected agent’s conversation history.
  - Claude-specific transcript data is no longer used for sessions belonging to other agents.
  - Session summaries refresh correctly when the selected agent changes.
  - Repeated Codex session lookups are faster while still detecting removed or newly restored records.
- **Tests**
  - Added coverage for agent-specific retrieval, fallback behavior, cache limits, pruning, and client requests.


isamu:
## Cross-review: tier C

Tiered **C** before round 1, on two of the tier-C triggers:

- **Everything passes through the changed line.** `sessionDetail` is a route entry point, and
  `/api/session/:id` is polled by every grid cell every few seconds plus once per cell mount.
- **The PR claims behaviour preservation** on the claude path ("identical with and without the
  param"), and that class is proved by running both, never by reading.

So the round names its axes and each gets answered per axis, rather than "review this PR":
correctness, security, tests-go-red, API/wire compatibility, codebase consistency, i18n/a11y,
plus the C-list ones this diff touches — **blast radius** on the shared route, **async ordering**
(both changed clients carry seed-token races over overlapping polls, and `useSessionSummary`'s
watcher source changed), and the **behaviour-preservation proof** for claude.

Exit bar is one clean round on a head nothing was pushed to. Any P1/P2 finding promotes nothing
further — C is already the top tier — but it does buy the round after the fix, as always.

server/routes/session-routes.ts
```diff
@@ -138,6 +138,16 @@ async function sessionDetail(req: Request<{ id: string }>, res: Response, freshe
...
+  // The last exchange, from whichever log this agent keeps. Claude's already came out of the fold
+  // above — re-reading its transcript to answer the same two fields would double the cost of the
+  // busiest route in the app, which is the reason agentBadges leaves claude to the caller too.
+  //
+  // Everything else was answered from claude's transcript whatever `?agent=` said, so a codex cell
+  // — which has no file there at all — left the cockpit roster's `prompt` and `reply` lines blank
+  // while the claude cell beside it was filled (#2121). `sessionLastTurn` is the branch that was
+  // missing: it reads codex's rollout and cursor's transcript, and answers the agents whose logs
+  // have no reader yet with the empty turn, which is what this route was already showing them.
+  const exchange = agent === "claude" ? { prompt: transcriptPrompt, reply: transcriptResponse } : await sessionLastTurn(cwd, id, agent);
```
isamu: **P2 — duplicate Codex rollout directory scans on every roster poll.**\n\nFor `agent=codex`, this route first awaits `agentBadges(cwd, id, agent)` (which calls `codexRolloutPath`), then awaits `sessionLastTurn(...)` (whose `codexLastTurn` calls `codexRolloutPath` again). Each lookup synchronously readdir-walks every day directory under `$CODEX_HOME/sessions`; `/api/session/:id` is polled once per grid cell every ~4s, so N Codex cells now do 2N full directory walks per poll in addition to the tail reads. This is the only changed route site of this duplication; the underlying readers are `server/session/agent-badges.ts:241` and `server/session/session-reads.ts:302`, and other `sessionLastTurn` callers do not combine it with `agentBadges` in one request.\n\nSmallest fix: make the Codex path lookup request-local and pass/reuse the resolved rollout path (or add a single helper returning both badges and last turn) so the route performs one `codexRolloutPath` scan per Codex request, without changing each agent badge reader.\n\nI would withdraw this if `codexRolloutPath` were proven to be O(1) via a shared indexed/cache result for the lifetime of the request or if the route were no longer polled per cell; neither is true in this diff.

isamu:
- P2: Codex roster polling resolves the same rollout path twice per request, causing duplicate synchronous directory walks; inline comment posted on `server/routes/session-routes.ts:150`.

| Axis | Result |
|---|---|
| 1. Correctness and edge cases | none |
| 2. Security | none. The changed id/cwd/agent inputs remain behind the existing UUID, workspace, and exact-agent normalization; no new traversal, injection, or cross-session selection was found. |
| 3. Tests | none. Targeted specs passed (75/75); mutating the route branch made 2 route assertions fail, and mutating the composable watcher made the same-session agent-switch assertion fail. |
| 4. API / wire compatibility | none. Missing `?agent=` still normalizes to Claude and the response shape is unchanged. |
| 5. Codebase consistency | none. The clients now use the same agent contract as TerminalCell/context, and routing uses `normalizeAgent`/`sessionLastTurn`. |
| 6. Accessibility and i18n | none. No user-facing labels, bilingual docs, or accessible markup changed. |
| 7. Blast radius | finding: P2 duplicate Codex rollout directory scans per polled cell; see inline comment. |
| 8. Async ordering | none. The existing per-id seed/latest guards still reject stale responses; the composable watcher covers both session and agent and its mutation test went red. |
| 9. Claude behavior preservation | none. The Claude exchange remains the pre-existing summary pair for absent, malformed/array, mis-cased, and `shell` inputs because exact normalization returns Claude; normalization tests passed. |
| 10. `/clear` precedence | none. The exchange still enters `sessionDetailView` through `??`, so the empty-string sentinel wins; the Codex lack of the Claude hook is an existing agent asymmetry, not introduced routing behavior. |
| 11. Muse self-heal | none. The self-heal is for Claude-labelled Muse badge discovery; the exchange is independently selected from the normalized agent log and does not need that badge fallback. |

The deliberately omitted lifecycle `refreshLastResponse` is a separate existing pub/sub reader; I found no third changed-site of the same Claude-transcript blindness.

isamu:
## Codex exchange — round 1

<details><summary>Prompt sent (step A)</summary>

```text
Review PR #2122 at https://github.com/receptron/mulmoterminal/pull/2122 (repo receptron/mulmoterminal, branch fix/2121-roster-agent-exchange, base main at 7661a41d76308a5dfd656accefe06651e6553986). Work in /Users/isamu/ss/llm/mulmoterminal4.

anything for "a follow-up review".

  gh api repos/receptron/mulmoterminal/pulls/2122/comments
or top-level comments via:
  gh pr comment 2122

GET /api/session/:id took `?agent=` and let it decide ONE thing: where the two header
badges are read from. `lastPrompt` / `lastResponse` still came from readSessionSummary,
which opens claude's per-project transcript only — so a codex session (which has no file
there; codex mints its own id and writes $CODEX_HOME/sessions/**/rollout-*-<id>.jsonl)
answered nulls and the cockpit roster's prompt/reply lines were blank. The route now asks
`sessionLastTurn(cwd, id, agent)` for non-claude agents. Two clients that were sending no
`?agent=` now send it.

── TIER C — ANSWER EVERY AXIS, EVEN WHEN THE ANSWER IS "no findings" ────────
This PR is tier C on two triggers: it changes a ROUTE ENTRY POINT that every grid cell
polls, and it CLAIMS BEHAVIOUR PRESERVATION on the claude path. Answer every line below;
an axis with no findings is a RESULT and has to be stated. "No findings" on an axis you
did not look at is the one answer that makes this loop longer.

  1. correctness and edge cases
  2. security (path traversal via the id/cwd/agent params, injection, secrets, reading
     another user's or another session's log)
  3. tests: does a test go red if the new behaviour breaks — VERIFY THIS BY MUTATING,
     not by reading. See the mutation note below.
  4. API / wire compatibility: the `?agent=` contract and the response shape, for a client
     that does NOT send the param (an older build, a cached bundle, the phone)
  5. consistency with the rest of the codebase — especially with how agentBadges and
     TerminalCell.vue already handle `?agent=`, and with common/sessionAgent.ts
  6. accessibility and i18n lockstep (this repo has bilingual docs under docs/guide/{en,ja})
  7. BLAST RADIUS (tier C): /api/session/:id is polled by every grid cell every ~4s while
     the cockpit roster is open, plus per cell mount. What does the new read cost per poll
     for each agent? codexLastTurn calls codexRolloutPath, which readdirs day directories —
     and agentBadges ALREADY calls codexRolloutPath separately on the same request, so a
     codex cell now locates its rollout twice per poll. Is that acceptable, and is there a
     shape that avoids it without touching every agent's badge reader?
  8. ASYNC ORDERING (tier C): both changed clients carry a seed/sequence token over
     overlapping polls (latestMetaSeed in GridView.vue, `latest` in useSessionSummary.ts).
     useSessionSummary's watcher SOURCE changed from `sessionId` to
     `[sessionId, () => agent?.value ?? "claude"]`. Can a stale answer now be applied, can a
     poll be left running under the wrong agent, can the interval leak, can the ref-vs-getter
     array source misfire on an unchanged value?
  9. BEHAVIOUR-PRESERVATION PROOF (tier C): the PR claims the claude path is byte-identical.
     PROVE OR DISPROVE IT by running, not by reading. Concretely: is
     `{ prompt: transcriptPrompt, reply: transcriptResponse }` reached on exactly the same
     inputs as before, including when `?agent=` is absent, malformed, an array
     (`?agent=a&agent=b`), mis-cased ("CODEX"), or names "shell" (a session kind that is not
     a TerminalAgent)?
 10. The interaction with sessionDetailView's /clear precedence: `lastPrompts`/`lastResponses`
     hold "" as a CLEAR sentinel and `??` must let it win. Does routing the transcript pair
     through `exchange` change anything about that, for any agent? Note codex never fires the
     clear hook — is the resulting asymmetry a bug?
 11. The muse self-heal block immediately above the new line: it re-reads badges when
     `agent === "claude"` and the read came back empty. Does the new `exchange` line need the
     same self-heal, and if not, why not?

2. EVERY SITE. If the same mistake appears elsewhere in the repo, list every occurrence
   NOW — grep for it. Specifically: are there OTHER callers of /api/session/:id, or other
   readers that assume claude's transcript regardless of agent? (I know of one I
   deliberately left out — see the ledger note — say if there are more.)

── MUTATION / SANDBOX NOTES ─────────────────────────────────────────────────
Your sandbox was probed before this round: vitest, `npx vue-tsc -b`, socket bind and
api.github.com all work. `rm` is refused by shell-safety policy but `unlink` is allowed —
so if you mutate a file to check a test goes red, RESTORE WITH `git checkout -- <path>`,
never with rm, and verify `git status --porcelain` is clean between mutations. Mutating is
encouraged for axis 3; a guard that asserts a value is absent passes just as well when the
matcher is broken, so pin it in both directions.

The relevant specs:
  test/server/routes/session-detail-agent.spec.ts   (new, route level)
  test/src/components/GridView.spec.ts              (the roster fetch, at the end)
  test/src/composables/useSessionSummary.spec.ts
  test/src/components/CollectionChatPane.spec.ts
Running the whole suite takes ~3.5 min: `npx vitest run`. The machine is loaded, so a
timeout in an UNTOUCHED file is the load, not the change — re-run that file standalone with
`--testTimeout=120000` before reporting it.

| 0 | setup | Codex sandbox pre-flight | N/A | vitest 0, vue-tsc 0, socket 0, github 200; `rm` blocked, `unlink` works |

_No review findings yet — this is round 1._

Two things I decided BEFORE this review, deliberately, and stated in the PR body. Judge
whether the reasoning holds; if it does not, that is a finding:

  (a) The roster's `summary` line stays EMPTY for codex. It is claude's own `ai-title`,
      which Claude Code writes into its transcript and codex writes nowhere. Filling it
      would mean generating a title (the `headless` source, off by default even for claude)
      or promoting codex's first user prompt (a different value under the same label).
  (b) `refreshLastResponse` in server/session/lifecycle.ts still reads claude's transcript
      whatever the agent, so the pub/sub `sessions` row carries no reply for codex. Left
      out because the roster reads these fields from the ROUTE (GridView's own comment says
      the route is their single source of truth) and because publishActivity is synchronous
      where the agent-branched read is not. Say if you think that reasoning is wrong, or if
      there is a THIRD site of the same blindness I have not named.

  Then the axis table: one row per axis 1-11 above, findings or 'none'.
  Then, on its own line:
  Omit that last line if it is not true, and say which part of the diff you did not reach
  and why.

```

</details>

<details><summary>Codex reply (verbatim)</summary>

```text
Review complete and posted.

- P2 finding: duplicate Codex rollout directory scans per `/api/session/:id` poll. [Inline comment](https://github.com/receptron/mulmoterminal/pull/2122#discussion_r4042826877)
- Top-level axis-by-axis review posted: [review comment](https://github.com/receptron/mulmoterminal/pull/2122#issuecomment-5723471953)

Targeted tests passed: 87 tests total. Mutation checks confirmed the new route and async watcher tests fail under their regressions. No tracked workspace changes remain.
exit=0
```

</details>

<details><summary>Follow-up prompt (step C-bis — rebuttals settled before the push)</summary>

```text
Follow-up on PR #2122, round 1. Do NOT re-review the PR and do NOT post anything. Reply as plain text.
Repo /Users/isamu/ss/llm/mulmoterminal4, branch fix/2121-roster-agent-exchange. The changes below are
in the WORKING TREE, not yet pushed — read them with `git diff`.

Three things. Answer each ACCEPTED or DISPUTED, with the specific evidence if DISPUTED.

── 1. YOUR P2, and a fix that is broader than the one you proposed ──────────
You asked for the codex rollout path lookup to be request-local so the route resolves it once.
I measured it first: on this machine $CODEX_HOME/sessions has 84 day directories and one warm
`codexRolloutPath` costs 11.05 ms of SYNCHRONOUS readdirSync. Two per poll per codex cell, nine
cells, every 4 s = 281 ms of blocking work per cycle.

Request-local reuse halves that. It does not touch the fact that the SAME id is re-resolved from
scratch every 4 seconds forever, which is where almost all of the cost is. So I memoized the
lookup at process level instead (server/agents/codex-sessions.ts). After: warm locate 0.08 ms,
the nine-cell cycle 281 ms -> 90 ms, and what is left is the tail reads, which existed before
this PR anyway.

Your stated withdrawal condition was "if codexRolloutPath were proven to be O(1) via a shared
indexed/cache result for the lifetime of the request". Mine is process lifetime, which is
stronger and also riskier, so judge the staleness argument specifically:

  - Soundness: the file's own comment above `metaCache` already asserts a path's answer never
    goes STALE, because the filename carries the creation timestamp and the id and both are
    written once. Is that argument actually true for `codexRolloutPath`, or does it only hold for
    what `metaCache` stores?
  - Absence: codex prunes rollouts and day directories, so a remembered path is re-checked with
    one existsSync and a vanished one falls back to the full scan. Is one stat enough, or is there
    a case where the path exists but is the WRONG file?
  - A MISS is deliberately never remembered — a cell that has just started has no rollout yet, and
    remembering "absent" would hide it until restart (the failure `rolloutMeta` records twice in
    the same file). Confirm that is the right half to leave uncached.
  - Bound: nothing prunes this map (pruneMetaCache works off a full listing scan, which this
    lookup never does), so it is capped at PATH_CACHE_MAX = 512 with oldest-inserted evicted.
    Is a cap the right answer, or does eviction introduce a problem I have not seen?
  - Is a process-level cache the wrong CALL here even if sound — i.e. is it scope creep on a PR
    about which log gets read, and should it be its own PR with the request-local fix here?

Pinned by test/server/agents/codex-rollout-path-cache.spec.ts (7 cases). I break-verified three
mutations: caching a miss -> 2 red, dropping the existsSync revalidation -> 3 red, removing the
cap -> 1 red. Tell me if a property is pinned in a way that would pass while broken.

── 2. A FINDING YOU MISSED, and I want your judgement on my fix ─────────────
You wrote: "I found no third changed-site of the same Claude-transcript blindness." That is true
of CHANGED sites and I think it made both of us stop one step short. The route reads claude's
transcript for THREE things, and my original diff fixed only the exchange:

  a. lastPrompt / lastResponse  -> fixed in the original diff
  b. workPhase                  -> still claude's, for every agent
  c. userTurns + aiTitle, handed to freshenRosterTitle -> still claude's, for every agent

Measured, not reasoned: I mounted the route over a fixture where a claude transcript exists at the
id and asked as `agent=grok`. Before my new fix it answered `workPhase: "implementing"` and called
freshenRosterTitle with claude's turn count and claude's own `ai-title` — which storeKnownTitle
writes into `aiTitles` and publishes, i.e. a grok roster row wearing claude's title.

My fix: read claude's transcript ONLY for claude
(`const summary = agent === "claude" ? await readSessionSummary(cwd, id) : EMPTY_SUMMARY`) and
gate `freshenRosterTitle` behind `agent === "claude"`.

  - Is gating freshenRosterTitle correct, or does a NON-claude path depend on it being called?
    Check the muse self-heal above it, and check server/session/session-title.ts.
  - Is EMPTY_SUMMARY the right value for every field a non-claude agent then reports, workPhase
    included? Should workPhase instead be omitted rather than null?
  - Is there a FOURTH thing on this route, or elsewhere, still reading claude's store regardless of
    agent? I already know about lifecycle.ts refreshLastResponse and am leaving it (you accepted
    that). Look for others — this is the second time the enumeration was short.
  - How real is the id collision? If you think it cannot happen, say so — the fix still removes a
    stat and a fold per poll per non-claude cell, but I want the severity right in the PR.

── 3. A PROTOCOL GAP ────────────────────────────────────────────────────────
Your round-1 comment gave the full axis table but did NOT include the line
Which parts of the diff did you not reach, and is your round-1 set every finding you had?
```

</details>

<details><summary>Codex follow-up reply (verbatim)</summary>

```text
1. ACCEPTED

The process-level cache is sound for Codex’s filename invariant: the filename includes the immutable rollout ID and creation timestamp. A cached existing path cannot become a different valid rollout under normal Codex pruning; pruning only makes it disappear, which the existsSync check handles.

Not caching misses is correct. Eviction is also safe: it only causes a future directory scan. The 512-entry cap is appropriate because this lookup has no full-store pruning pass.

The tests cover the important properties. They do not cover an external process replacing a cached path with unrelated contents, or root-path aliases, but those are outside the stated Codex storage invariant. This is not scope creep severe enough to require a separate PR; it directly fixes the measured hot path and is bounded and revalidated.

2. ACCEPTED

Gating freshenRosterTitle to Claude is correct. session-title.ts reads Claude’s transcript and stores the result in the shared aiTitles map; invoking it for another agent can publish Claude’s title on that agent’s roster row. Muse’s self-heal is independent and remains above the gate.

EMPTY_SUMMARY is correct. Its workPhase is null, matching the response schema and existing client semantics; omitting workPhase would be an API behavior change.

I found no additional unconditional Claude-store read in this route. sessionLastTurn, sessionPrompts, and agentBadges all dispatch explicitly by agent. lifecycle.refreshLastResponse remains the separate accepted Claude-only path you identified.

The collision is possible as a stale or deliberately mismatched agent query, but an actual UUID collision between independently created server sessions is extremely unlikely. The severity is therefore low, though the fix is still correct and removes unnecessary work.

The focused tests pass: 14 tests in 2 files.

3. ACCEPTED

The protocol gap is real. I did not reach the two non-exchange consumers of the summary in sessionDetail: workPhase and userTurns/aiTitle passed to freshenRosterTitle. I also did not perform a complete audit of every agent-dispatched reader.

The round-1 comments represented every finding I had identified at that time, but they were not every finding present in the diff.
exit=0
```

</details>

isamu:
## Round 1 triage — every finding and its disposition

Tier C. Two findings, one from each reviewer; both fixed in `56c26e5a`, both break-verified.

| 0 | setup | Codex sandbox pre-flight | N/A | vitest exit 0, `vue-tsc -b` exit 0, socket bind exit 0, api.github.com HTTP 200. `rm` refused by shell-safety policy; `unlink` works, so restores use `git checkout --`. Nothing in this loop is UNVERIFIED for sandbox reasons. |
| 1 | 1 | P2 (Codex): `/api/session/:id` resolves the codex rollout path TWICE per poll — `agentBadges` and `sessionLastTurn` each call `codexRolloutPath`, which is synchronous `readdirSync` per day directory | FIXED | Measured first: 84 day dirs here, **11.05 ms warm per locate**, so nine codex cells cost 281 ms of blocking work per 4 s cycle. This PR is what introduced it — before, the roster sent no `?agent=` so it did ZERO codex scans. Fixed broader than proposed: memoized `codexRolloutPath` at process level (`server/agents/codex-sessions.ts`) rather than per request, because the dominant cost is the SAME id re-resolved every 4 s forever, which request-local reuse does not touch. Warm locate 11.05 → 0.08 ms; cycle 281 → 90 ms (the remainder is tail reads that pre-date the PR). Only a HIT is remembered (a miss is a cell whose rollout codex has not written yet); a hit is re-checked with one `existsSync` because codex prunes; capped at 512 since nothing prunes this map. Pinned by `test/server/agents/codex-rollout-path-cache.spec.ts` (7 cases); break-verified by three mutations — caching a miss → 2 red, dropping revalidation → 3 red, removing the cap → 1 red. Codex ACCEPTED in C-bis, including that it is not scope creep. |
| 2 | 1 | **Claude-side, not flagged by Codex:** the route read claude's transcript for THREE things and the original diff fixed only one — `workPhase` and the `userTurns`/`aiTitle` handed to `freshenRosterTitle` were still claude's for every agent | FIXED | Measured on a fixture, not reasoned: mounting the route over a claude transcript at the same id and asking as `agent=grok` answered `workPhase: "implementing"` and called `freshenRosterTitle` with claude's turn count and claude's own `ai-title` — which `storeKnownTitle` writes into `aiTitles` and publishes, i.e. a grok roster row wearing claude's title. Fix: `readSessionSummary` only for claude (`EMPTY_SUMMARY` otherwise) and `freshenRosterTitle` gated behind `agent === "claude"`. Also removes a stat and a fold per poll for every non-claude cell. Pinned by two cases in `session-detail-agent.spec.ts` (one gate, one claude control); break-verified — reverting the fix turns the gate case red. Codex ACCEPTED in C-bis and confirmed there is no FOURTH unconditional claude read on this route. **Severity: low** — Codex's judgement, which I agree with: a real UUID collision between independently created sessions is extremely unlikely, so this is mostly a correctness-of-rule and a cost fix, not a live bug. |
| 3 | 1 | My pre-stated decision (a): the roster `summary` line stays empty for codex | REJECTED (upheld) | Codex raised nothing against it. It is claude's own `ai-title`, which codex writes nowhere; filling it means generating a title or relabelling codex's first prompt. |
| 4 | 1 | My pre-stated decision (b): `lifecycle.refreshLastResponse` stays claude-only | REJECTED (upheld) | Codex, round 1: "a separate existing pub/sub reader; I found no third changed-site of the same Claude-transcript blindness", and again in C-bis: "remains the separate accepted Claude-only path you identified". |
| 5 | 1 | **Protocol:** round 1 omitted the `FINDINGS COMPLETE` line | NOTED | Asked in C-bis rather than spending a round. Codex: "The protocol gap is real. I did not reach the two non-exchange consumers of the summary in sessionDetail... The round-1 comments represented every finding I had identified at that time, but they were not every finding present in the diff." Row 2 is exactly what it did not reach. Round 2's prompt names this as a late-finding note. |

### What the round cost, and why

Row 2 is the expensive one to notice: the original diff fixed **one of three** places the route
read claude's transcript regardless of agent. Codex reviewed the changed lines and did not reach
the other two consumers of the same value, and said so when asked directly. That is the
"fix the CLASS, not the site" failure from the review protocol, caught inside round 1 rather
than arriving as round 3.

Both rebuttals were settled in the same round (step C-bis) rather than pushed and re-reviewed —
Codex accepted the process-level cache (including that it is not scope creep), accepted the
transcript gate, and confirmed there is no fourth unconditional claude read on this route.

### Gates

`yarn format` / `lint` / `typecheck` / `build` clean. Full `vitest run` green on the merged
result after `origin/main` (#2120) was merged in. Round 2 is already dispatched against
`56c26e5a` — CI gates the merge, not the review.

isamu:
## Codex exchange — round 2

```text
Review PR #2122 at https://github.com/receptron/mulmoterminal/pull/2122 — ROUND 2, tier C.
Repo receptron/mulmoterminal, branch fix/2121-roster-agent-exchange, in /Users/isamu/ss/llm/mulmoterminal4.
The head is 56c26e5a. `origin/main` was merged in this round, so the diff also carries #2120's changes
from the merge — review MY branch's changes, `git diff $(git merge-base origin/main HEAD)...HEAD`.

A fix is reviewed for the first time in the round AFTER it is pushed, so a finding held back is
not a later finding — it is a repeat of this whole exchange. Read the WHOLE diff before posting.

── LATE FINDING — read this before looking for anything new ─────────────────
Ledger row 2 was raised by ME, not by you, and it was in the code you reviewed in round 1: the
route read claude's transcript for THREE things (the exchange, `workPhase`, and the
userTurns/aiTitle handed to `freshenRosterTitle`) and round 1's diff fixed only the exchange. You
confirmed this yourself in the follow-up: "I did not reach the two non-exchange consumers of the
summary in sessionDetail... The round-1 comments represented every finding I had identified at
that time, but they were not every finding present in the diff." You also omitted the
FINDINGS COMPLETE line, which is what would have said so.

So: before looking for anything new, re-read the parts of round 1's diff you did not reach —
every CONSUMER of a value the route computes, not just the lines that changed.

── WHAT CHANGED THIS ROUND ──────────────────────────────────────────────────
1. `server/agents/codex-sessions.ts` — `codexRolloutPath` now memoizes resolved paths at process
   level (hit re-checked with existsSync, misses never cached, capped at 512, oldest-inserted
   evicted). New spec `test/server/agents/codex-rollout-path-cache.spec.ts`.
2. `server/routes/session-routes.ts` — claude's transcript is read ONLY for claude
   (`EMPTY_SUMMARY` otherwise), and `freshenRosterTitle` is gated behind `agent === "claude"`.
   `workPhase` now comes from that gated summary.
3. `test/server/routes/session-detail-agent.spec.ts` — two cases for the gate + its claude control.

── TIER C — ANSWER EVERY AXIS, EVEN WHEN THE ANSWER IS "no findings" ────────
  1. correctness and edge cases
  2. security
  3. tests: does a test go red if the new behaviour breaks — VERIFY BY MUTATING. In particular,
     can the new cache spec pass while the cache is broken, and can the route spec's
     `expect(freshenRosterTitle).not.toHaveBeenCalled()` pass vacuously?
  4. API / wire compatibility — `workPhase` is now null for every non-claude agent where it
     previously could be a phase. Is that a client-visible regression anywhere? Check
     src/components/rosterPhase.ts and what renders workPhase.
  5. consistency with the rest of the codebase
  6. accessibility and i18n lockstep
  7. BLAST RADIUS: the memo is process-global and shared by EVERY caller of `codexRolloutPath`,
     not just this route — list them and say whether a remembered path is safe for each. Is
     there a caller for which a stale-but-existing path is worse than a rescan?
  8. ASYNC ORDERING / CONCURRENCY: the memo is a plain Map mutated from async request handlers.
     Node is single-threaded per process, but two concurrent requests can interleave around the
     `await`s. Can eviction race a read? Can the cap be exceeded? Is there a torn state?
  9. BEHAVIOUR PRESERVATION for claude: the claude path now flows through `summary.*` rather than
     destructured locals, and `freshenRosterTitle` is gated. PROVE claude is unchanged by running,
     not reading.
 10. THE GATE ITSELF: is `agent === "claude"` the right predicate in both new places, given the
     muse self-heal directly above it uses the same test for a DIFFERENT purpose (a cell whose
     persisted agent is claude but is really muse)? Does that cell now lose anything?
 11. The merge from origin/main this round — does anything in #2120 interact with this change?

1. SEVERITY (P1/P2/P3). 2. EVERY SITE — grep, do not guess. 3. THE SMALLEST CHANGE THAT RESOLVES
IT. 4. WHAT WOULD CHANGE YOUR MIND.

── MUTATION / SANDBOX NOTES ─────────────────────────────────────────────────
Your sandbox works (vitest, vue-tsc, socket, network all exit 0). `rm` is refused by policy —
RESTORE WITH `git checkout -- <path>`, and check `git status --porcelain` between mutations.
Full suite `npx vitest run` takes ~3.5 min and is currently green (906 files, 13502 tests). The
machine is loaded; a timeout in an UNTOUCHED file is the load — re-run it standalone with
`--testTimeout=120000` before reporting it.

# Ledger — PR #2122 (cross-review), tier C

| 0 | setup | Codex sandbox pre-flight | N/A | vitest exit 0, `vue-tsc -b` exit 0, socket bind exit 0, api.github.com HTTP 200. `rm` refused by shell-safety policy; `unlink` works, so restores use `git checkout --`. Nothing in this loop is UNVERIFIED for sandbox reasons. |
| 1 | 1 | P2 (Codex): `/api/session/:id` resolves the codex rollout path TWICE per poll — `agentBadges` and `sessionLastTurn` each call `codexRolloutPath`, which is synchronous `readdirSync` per day directory | FIXED | Measured first: 84 day dirs here, **11.05 ms warm per locate**, so nine codex cells cost 281 ms of blocking work per 4 s cycle. This PR is what introduced it — before, the roster sent no `?agent=` so it did ZERO codex scans. Fixed broader than proposed: memoized `codexRolloutPath` at process level (`server/agents/codex-sessions.ts`) rather than per request, because the dominant cost is the SAME id re-resolved every 4 s forever, which request-local reuse does not touch. Warm locate 11.05 → 0.08 ms; cycle 281 → 90 ms (the remainder is tail reads that pre-date the PR). Only a HIT is remembered (a miss is a cell whose rollout codex has not written yet); a hit is re-checked with one `existsSync` because codex prunes; capped at 512 since nothing prunes this map. Pinned by `test/server/agents/codex-rollout-path-cache.spec.ts` (7 cases); break-verified by three mutations — caching a miss → 2 red, dropping revalidation → 3 red, removing the cap → 1 red. Codex ACCEPTED in C-bis, including that it is not scope creep. |
| 2 | 1 | **Claude-side, not flagged by Codex:** the route read claude's transcript for THREE things and the original diff fixed only one — `workPhase` and the `userTurns`/`aiTitle` handed to `freshenRosterTitle` were still claude's for every agent | FIXED | Measured on a fixture, not reasoned: mounting the route over a claude transcript at the same id and asking as `agent=grok` answered `workPhase: "implementing"` and called `freshenRosterTitle` with claude's turn count and claude's own `ai-title` — which `storeKnownTitle` writes into `aiTitles` and publishes, i.e. a grok roster row wearing claude's title. Fix: `readSessionSummary` only for claude (`EMPTY_SUMMARY` otherwise) and `freshenRosterTitle` gated behind `agent === "claude"`. Also removes a stat and a fold per poll for every non-claude cell. Pinned by two cases in `session-detail-agent.spec.ts` (one gate, one claude control); break-verified — reverting the fix turns the gate case red. Codex ACCEPTED in C-bis and confirmed there is no FOURTH unconditional claude read on this route. **Severity: low** — Codex's judgement, which I agree with: a real UUID collision between independently created sessions is extremely unlikely, so this is mostly a correctness-of-rule and a cost fix, not a live bug. |
| 3 | 1 | My pre-stated decision (a): the roster `summary` line stays empty for codex | REJECTED (upheld) | Codex raised nothing against it. It is claude's own `ai-title`, which codex writes nowhere; filling it means generating a title or relabelling codex's first prompt. |
| 4 | 1 | My pre-stated decision (b): `lifecycle.refreshLastResponse` stays claude-only | REJECTED (upheld) | Codex, round 1: "a separate existing pub/sub reader; I found no third changed-site of the same Claude-transcript blindness", and again in C-bis: "remains the separate accepted Claude-only path you identified". |
| 5 | 1 | **Protocol:** round 1 omitted the `FINDINGS COMPLETE` line | NOTED | Asked in C-bis rather than spending a round. Codex: "The protocol gap is real. I did not reach the two non-exchange consumers of the summary in sessionDetail... The round-1 comments represented every finding I had identified at that time, but they were not every finding present in the diff." Row 2 is exactly what it did not reach. Round 2's prompt names this as a late-finding note. |

That table is context, not a list of closed topics. If a FIXED row's fix does not do what it
claims, say so and mark it 'REOPENING #<row>' with specific evidence. Rows 3 and 4 are my
declined decisions and you accepted both — do not re-raise them unless you have new evidence.

  Then the axis table: one row per axis 1-11, findings or 'none'.
  Then, on its own line:
  Omit it if untrue and say which part you did not reach. You omitted it last round; do not
  omit it silently this round.

```

</details>

<details><summary>Codex reply (verbatim)</summary>

```text
CODEX VERDICT: LGTM

- No findings.
- Targeted tests: 5 files, 89 tests passed.
- `vue-tsc -b` passed.
- Mutation checks confirmed cache reuse and non-vacuous route gating.

| Axis | Result |
|---|---|
| 1. Correctness / edge cases | None |
| 2. Security | None |
| 3. Tests | Adequate; mutations went red as expected |
| 4. API / wire compatibility | No client regression found; `workPhase` is correctly agent-scoped |
| 5. Codebase consistency | Consistent |
| 6. Accessibility / i18n | No impact |
| 7. Cache blast radius | Safe for `agent-badges`, `session-reads` (2 sites), `spawn-codex` (2), `transcript-view-read`, and `codexRolloutExists`; no stale-existing-path hazard absent impossible duplicate Codex IDs |
| 8. Async ordering / concurrency | Safe; lookup and Map mutation are synchronous and atomic per event-loop turn |
| 9. Claude preservation | Proven by passing Claude route-control test |
| 10. Gate predicate | Correct; Muse self-heal behavior is preserved |
| 11. #2120 merge interaction | No interaction; requested merge base is #2120’s merge commit |

exit=0
```

</details>

### Triage

Codex: **LGTM**, all eleven axes answered, `FINDINGS COMPLETE` present, working tree restored clean.

Not an exit, because my own pass on the same head found two things and I pushed them (`8d1538da`):

- `summary` did not say **whose**. The value is read only for claude now, and the name is where
  that invariant is cheapest to see. Identifier-only rename; I grepped the tests for it as a
  string literal first, since a rename is behaviour-preserving only if nothing reads the name as data.
- A **claims sweep**: this PR body claimed the duplicate per-poll read was accepted and left out —
  which the round-1 fix made false — and it carried a figure that was simply wrong, taken from an
  early probe on a recent rollout rather than across the store. Body corrected, plan file updated.

Round 3 is dispatched against `8d1538da`. An LGTM is stamped on a commit, not on a PR.

isamu:
## Codex exchange — round 3

```text
Review PR #2122 at https://github.com/receptron/mulmoterminal/pull/2122 — ROUND 3, tier C.
Repo receptron/mulmoterminal, branch fix/2121-roster-agent-exchange, in /Users/isamu/ss/llm/mulmoterminal4.
The head is 8d1538da. `origin/main` was merged in this round, so the diff also carries #2120's changes
from the merge — review MY branch's changes, `git diff $(git merge-base origin/main HEAD)...HEAD`.

A fix is reviewed for the first time in the round AFTER it is pushed, so a finding held back is
not a later finding — it is a repeat of this whole exchange. Read the WHOLE diff before posting.

── NO LATE FINDING THIS ROUND ───────────────────────────────────────────────
Round 2 was complete: every axis answered, FINDINGS COMPLETE present, and nothing has come up since
that you should have reached. This round is short by design.

── WHAT CHANGED THIS ROUND — prose and ONE rename, nothing behavioural ──────
You LGTM'd `56c26e5a`. I then pushed `8d1538da`, so your sign-off no longer covers the head and
this round exists to put that right. It contains exactly two things, both raised by ME in my own
pass, neither from a finding of yours:

1. `server/routes/session-routes.ts` — the local `summary` renamed to `claudeSummary`. Identifier
   only. The value is read only for claude now, and the name is where that invariant is cheapest to
   see. I grepped the tests for the identifier as a STRING LITERAL before renaming (no matches), so
   nothing reads the name as data.
2. `plans/fix-2121-roster-agent-exchange.md` — it described the change as the route plus two
   clients, which your P2 fix and the transcript gate made false.

The PR BODY was also corrected (not a commit): it claimed the duplicate per-poll read was accepted
and left out, and carried a figure that was wrong — "single-digit ms warm", taken from an early
probe on a recent rollout rather than across the store.

So the question for this round is narrow but not rhetorical: **is the rename genuinely
behaviour-preserving, and is the prose now TRUE?** Read the plan file and the PR body against the
code, not just the diff. If any claim in either is still wrong, that is a finding.

── TIER C — ANSWER EVERY AXIS, EVEN WHEN THE ANSWER IS "no findings" ────────
  1. correctness and edge cases
  2. security
  3. tests: does a test go red if the new behaviour breaks — VERIFY BY MUTATING. In particular,
     can the new cache spec pass while the cache is broken, and can the route spec's
     `expect(freshenRosterTitle).not.toHaveBeenCalled()` pass vacuously?
  4. API / wire compatibility — `workPhase` is now null for every non-claude agent where it
     previously could be a phase. Is that a client-visible regression anywhere? Check
     src/components/rosterPhase.ts and what renders workPhase.
  5. consistency with the rest of the codebase
  6. accessibility and i18n lockstep
  7. BLAST RADIUS: the memo is process-global and shared by EVERY caller of `codexRolloutPath`,
     not just this route — list them and say whether a remembered path is safe for each. Is
     there a caller for which a stale-but-existing path is worse than a rescan?
  8. ASYNC ORDERING / CONCURRENCY: the memo is a plain Map mutated from async request handlers.
     Node is single-threaded per process, but two concurrent requests can interleave around the
     `await`s. Can eviction race a read? Can the cap be exceeded? Is there a torn state?
  9. BEHAVIOUR PRESERVATION for claude: the claude path now flows through `summary.*` rather than
     destructured locals, and `freshenRosterTitle` is gated. PROVE claude is unchanged by running,
     not reading.
 10. THE GATE ITSELF: is `agent === "claude"` the right predicate in both new places, given the
     muse self-heal directly above it uses the same test for a DIFFERENT purpose (a cell whose
     persisted agent is claude but is really muse)? Does that cell now lose anything?
 11. The merge from origin/main this round — does anything in #2120 interact with this change?

1. SEVERITY (P1/P2/P3). 2. EVERY SITE — grep, do not guess. 3. THE SMALLEST CHANGE THAT RESOLVES
IT. 4. WHAT WOULD CHANGE YOUR MIND.

── MUTATION / SANDBOX NOTES ─────────────────────────────────────────────────
Your sandbox works (vitest, vue-tsc, socket, network all exit 0). `rm` is refused by policy —
RESTORE WITH `git checkout -- <path>`, and check `git status --porcelain` between mutations.
Full suite `npx vitest run` takes ~3.5 min and is currently green (906 files, 13502 tests). The
machine is loaded; a timeout in an UNTOUCHED file is the load — re-run it standalone with
`--testTimeout=120000` before reporting it.

# Ledger — PR #2122 (cross-review), tier C

| 0 | setup | Codex sandbox pre-flight | N/A | vitest exit 0, `vue-tsc -b` exit 0, socket bind exit 0, api.github.com HTTP 200. `rm` refused by shell-safety policy; `unlink` works, so restores use `git checkout --`. Nothing in this loop is UNVERIFIED for sandbox reasons. |
| 1 | 1 | P2 (Codex): `/api/session/:id` resolves the codex rollout path TWICE per poll — `agentBadges` and `sessionLastTurn` each call `codexRolloutPath`, which is synchronous `readdirSync` per day directory | FIXED | Measured first: 84 day dirs here, **11.05 ms warm per locate**, so nine codex cells cost 281 ms of blocking work per 4 s cycle. This PR is what introduced it — before, the roster sent no `?agent=` so it did ZERO codex scans. Fixed broader than proposed: memoized `codexRolloutPath` at process level (`server/agents/codex-sessions.ts`) rather than per request, because the dominant cost is the SAME id re-resolved every 4 s forever, which request-local reuse does not touch. Warm locate 11.05 → 0.08 ms; cycle 281 → 90 ms (the remainder is tail reads that pre-date the PR). Only a HIT is remembered (a miss is a cell whose rollout codex has not written yet); a hit is re-checked with one `existsSync` because codex prunes; capped at 512 since nothing prunes this map. Pinned by `test/server/agents/codex-rollout-path-cache.spec.ts` (7 cases); break-verified by three mutations — caching a miss → 2 red, dropping revalidation → 3 red, removing the cap → 1 red. Codex ACCEPTED in C-bis, including that it is not scope creep. |
| 2 | 1 | **Claude-side, not flagged by Codex:** the route read claude's transcript for THREE things and the original diff fixed only one — `workPhase` and the `userTurns`/`aiTitle` handed to `freshenRosterTitle` were still claude's for every agent | FIXED | Measured on a fixture, not reasoned: mounting the route over a claude transcript at the same id and asking as `agent=grok` answered `workPhase: "implementing"` and called `freshenRosterTitle` with claude's turn count and claude's own `ai-title` — which `storeKnownTitle` writes into `aiTitles` and publishes, i.e. a grok roster row wearing claude's title. Fix: `readSessionSummary` only for claude (`EMPTY_SUMMARY` otherwise) and `freshenRosterTitle` gated behind `agent === "claude"`. Also removes a stat and a fold per poll for every non-claude cell. Pinned by two cases in `session-detail-agent.spec.ts` (one gate, one claude control); break-verified — reverting the fix turns the gate case red. Codex ACCEPTED in C-bis and confirmed there is no FOURTH unconditional claude read on this route. **Severity: low** — Codex's judgement, which I agree with: a real UUID collision between independently created sessions is extremely unlikely, so this is mostly a correctness-of-rule and a cost fix, not a live bug. |
| 3 | 1 | My pre-stated decision (a): the roster `summary` line stays empty for codex | REJECTED (upheld) | Codex raised nothing against it. It is claude's own `ai-title`, which codex writes nowhere; filling it means generating a title or relabelling codex's first prompt. |
| 4 | 1 | My pre-stated decision (b): `lifecycle.refreshLastResponse` stays claude-only | REJECTED (upheld) | Codex, round 1: "a separate existing pub/sub reader; I found no third changed-site of the same Claude-transcript blindness", and again in C-bis: "remains the separate accepted Claude-only path you identified". |
| 5 | 1 | **Protocol:** round 1 omitted the `FINDINGS COMPLETE` line | NOTED | Asked in C-bis rather than spending a round. Codex: "The protocol gap is real. I did not reach the two non-exchange consumers of the summary in sessionDetail... The round-1 comments represented every finding I had identified at that time, but they were not every finding present in the diff." Row 2 is exactly what it did not reach. Round 2's prompt names this as a late-finding note. |

That table is context, not a list of closed topics. If a FIXED row's fix does not do what it
claims, say so and mark it 'REOPENING #<row>' with specific evidence. Rows 3 and 4 are my
declined decisions and you accepted both — do not re-raise them unless you have new evidence.

  Then the axis table: one row per axis 1-11, findings or 'none'.
  Then, on its own line:
  Omit it if untrue and say which part you did not reach. You omitted it last round; do not
  omit it silently this round.

```

</details>

<details><summary>Codex reply (verbatim)</summary>

```text

- P3 documentation finding: the plan’s line 40 says neither review-loop change concerns which log is read, but the Claude gate explicitly changes whether Claude’s transcript is read for non-Claude requests.
- Rename is behavior-preserving. Targeted route tests passed, including the Claude control; mutation made that control fail.
- Cache tests also fail when caching is bypassed.
- Only unrelated pre-existing untracked files remain.

| Axis | Result |
|---|---|
| 1. Correctness / edge cases | One P3 prose finding above; code otherwise none. |
| 2. Security | None. |
| 3. Tests | None. Cache and route gates were mutation-verified. |
| 4. API / wire compatibility | `workPhase: null` for non-Claude is client-visible but renders safely as plain “running”; no valid non-Claude phase is lost. |
| 5. Codebase consistency | P3 plan inconsistency above. |
| 6. Accessibility / i18n | None. |
| 7. Blast radius | Callers: `agent-badges.ts`, `session-reads.ts` twice, `transcript-view-read.ts`, `spawn-codex.ts` twice, plus `codexRolloutExists` from `resumable-sessions.ts`, `survivor-agent-guard.ts`, and `ws-routes.ts`. Remembered paths are safe for all; no caller is worse than rescanning because the path is existence-checked and the filename/id invariant is stable. |
| 8. Async ordering / concurrency | None. Lookup and Map mutation are synchronous; awaits occur outside the mutation, so no torn state or cap overshoot. |
| 9. Claude preservation | None. `test/server/routes/session-detail-agent.spec.ts`: 7/7 passed, including Claude phase/title behavior. |
| 10. Gate predicate | None. Muse’s persisted-`claude` compatibility path retains its prior behavior and loses nothing. |
| 11. #2120 merge interaction | None found. |

Finding details: `plans/fix-2121-roster-agent-exchange.md:40`. Smallest fix: replace the sentence with wording that distinguishes the memo’s per-poll lookup optimization from the Claude-transcript gate. What would change my mind: if “which log is read” is explicitly intended to exclude the decision to read or skip Claude’s transcript, though the adjacent bullet describes precisely that behavior.

exit=0
```

</details>

### Triage

Codex: **CHANGES REQUESTED**, one P3, and it was right — the plan said neither review-loop change
was about which log is read, which is false of the claude gate. The rename was confirmed
behaviour-preserving by mutation, and the cache callers were enumerated across eight sites.

Fixing that made me re-read the rest of the prose against the source, which turned up **a second
wrong claim neither of us had flagged**: "a seventh agent", in the plan *and* in the PR body.
`TERMINAL_AGENTS` has seven entries today, so the next is the eighth — which this repo's own
CLAUDE.md says, and which my round-1 commit message said while the plan contradicted it. Counting
the array found it; re-reading the sentence would not have. Both surfaces corrected in `3e63d99b`.

Round 4 is dispatched against `3e63d99b` — prose-only diff, and the question put to it is whether
every claim in the plan and the body is now true of the code.

isamu:
## Codex exchange — round 4 (exit round)

```text
Review PR #2122 at https://github.com/receptron/mulmoterminal/pull/2122 — ROUND 4, tier C.
Repo receptron/mulmoterminal, branch fix/2121-roster-agent-exchange, in /Users/isamu/ss/llm/mulmoterminal4.
The head is 3e63d99b. `origin/main` was merged in this round, so the diff also carries #2120's changes
from the merge — review MY branch's changes, `git diff $(git merge-base origin/main HEAD)...HEAD`.

A fix is reviewed for the first time in the round AFTER it is pushed, so a finding held back is
not a later finding — it is a repeat of this whole exchange. Read the WHOLE diff before posting.

── NO LATE FINDING THIS ROUND ───────────────────────────────────────────────
Round 2 was complete: every axis answered, FINDINGS COMPLETE present, and nothing has come up since
that you should have reached. This round is short by design.

── WHAT CHANGED THIS ROUND — prose only, no code at all ─────────────────────
You raised one P3 in round 3: the plan's "What the review loop added" lead claimed neither
review-loop change was about which log is read, which is false of the claude gate. Fixed in
`3e63d99b`, reframed as what each change actually is.

Reading the rest of the prose against the code for that fix, I found a second wrong claim you did
NOT flag, and I want it treated as a late finding against BOTH of us: "a seventh agent" appeared in
the plan AND in the PR body. `TERMINAL_AGENTS` has seven entries today, so the next one is the
eighth — which is what this repo's own CLAUDE.md says, and what my round-1 commit message said
while the plan contradicted it. Counting the array is what found it; re-reading the sentence would
not have. I also dropped a "three fields" that counted the roster's `summary` line among the ones
read from the agent's log — it is not one of them, which is the entire reason it stays empty for
codex.

`git diff 8d1538da..3e63d99b` is one markdown file. No source changed this round.

So this round is narrow: **is every claim in the plan file and the PR body now TRUE of the code?**
Read them BOTH against the source, not against the diff. Check every count and every named symbol
against the thing it names. If a claim is still wrong, that is a finding; if you cannot verify one
from the code, say which.

── TIER C — ANSWER EVERY AXIS, EVEN WHEN THE ANSWER IS "no findings" ────────
  1. correctness and edge cases
  2. security
  3. tests: does a test go red if the new behaviour breaks — VERIFY BY MUTATING. In particular,
     can the new cache spec pass while the cache is broken, and can the route spec's
     `expect(freshenRosterTitle).not.toHaveBeenCalled()` pass vacuously?
  4. API / wire compatibility — `workPhase` is now null for every non-claude agent where it
     previously could be a phase. Is that a client-visible regression anywhere? Check
     src/components/rosterPhase.ts and what renders workPhase.
  5. consistency with the rest of the codebase
  6. accessibility and i18n lockstep
  7. BLAST RADIUS: the memo is process-global and shared by EVERY caller of `codexRolloutPath`,
     not just this route — list them and say whether a remembered path is safe for each. Is
     there a caller for which a stale-but-existing path is worse than a rescan?
  8. ASYNC ORDERING / CONCURRENCY: the memo is a plain Map mutated from async request handlers.
     Node is single-threaded per process, but two concurrent requests can interleave around the
     `await`s. Can eviction race a read? Can the cap be exceeded? Is there a torn state?
  9. BEHAVIOUR PRESERVATION for claude: the claude path now flows through `summary.*` rather than
     destructured locals, and `freshenRosterTitle` is gated. PROVE claude is unchanged by running,
     not reading.
 10. THE GATE ITSELF: is `agent === "claude"` the right predicate in both new places, given the
     muse self-heal directly above it uses the same test for a DIFFERENT purpose (a cell whose
     persisted agent is claude but is really muse)? Does that cell now lose anything?
 11. The merge from origin/main this round — does anything in #2120 interact with this change?

1. SEVERITY (P1/P2/P3). 2. EVERY SITE — grep, do not guess. 3. THE SMALLEST CHANGE THAT RESOLVES
IT. 4. WHAT WOULD CHANGE YOUR MIND.

── MUTATION / SANDBOX NOTES ─────────────────────────────────────────────────
Your sandbox works (vitest, vue-tsc, socket, network all exit 0). `rm` is refused by policy —
RESTORE WITH `git checkout -- <path>`, and check `git status --porcelain` between mutations.
Full suite `npx vitest run` takes ~3.5 min and is currently green (906 files, 13502 tests). The
machine is loaded; a timeout in an UNTOUCHED file is the load — re-run it standalone with
`--testTimeout=120000` before reporting it.

# Ledger — PR #2122 (cross-review), tier C

| 0 | setup | Codex sandbox pre-flight | N/A | vitest exit 0, `vue-tsc -b` exit 0, socket bind exit 0, api.github.com HTTP 200. `rm` refused by shell-safety policy; `unlink` works, so restores use `git checkout --`. Nothing in this loop is UNVERIFIED for sandbox reasons. |
| 1 | 1 | P2 (Codex): `/api/session/:id` resolves the codex rollout path TWICE per poll — `agentBadges` and `sessionLastTurn` each call `codexRolloutPath`, which is synchronous `readdirSync` per day directory | FIXED | Measured first: 84 day dirs here, **11.05 ms warm per locate**, so nine codex cells cost 281 ms of blocking work per 4 s cycle. This PR is what introduced it — before, the roster sent no `?agent=` so it did ZERO codex scans. Fixed broader than proposed: memoized `codexRolloutPath` at process level (`server/agents/codex-sessions.ts`) rather than per request, because the dominant cost is the SAME id re-resolved every 4 s forever, which request-local reuse does not touch. Warm locate 11.05 → 0.08 ms; cycle 281 → 90 ms (the remainder is tail reads that pre-date the PR). Only a HIT is remembered (a miss is a cell whose rollout codex has not written yet); a hit is re-checked with one `existsSync` because codex prunes; capped at 512 since nothing prunes this map. Pinned by `test/server/agents/codex-rollout-path-cache.spec.ts` (7 cases); break-verified by three mutations — caching a miss → 2 red, dropping revalidation → 3 red, removing the cap → 1 red. Codex ACCEPTED in C-bis, including that it is not scope creep. |
| 2 | 1 | **Claude-side, not flagged by Codex:** the route read claude's transcript for THREE things and the original diff fixed only one — `workPhase` and the `userTurns`/`aiTitle` handed to `freshenRosterTitle` were still claude's for every agent | FIXED | Measured on a fixture, not reasoned: mounting the route over a claude transcript at the same id and asking as `agent=grok` answered `workPhase: "implementing"` and called `freshenRosterTitle` with claude's turn count and claude's own `ai-title` — which `storeKnownTitle` writes into `aiTitles` and publishes, i.e. a grok roster row wearing claude's title. Fix: `readSessionSummary` only for claude (`EMPTY_SUMMARY` otherwise) and `freshenRosterTitle` gated behind `agent === "claude"`. Also removes a stat and a fold per poll for every non-claude cell. Pinned by two cases in `session-detail-agent.spec.ts` (one gate, one claude control); break-verified — reverting the fix turns the gate case red. Codex ACCEPTED in C-bis and confirmed there is no FOURTH unconditional claude read on this route. **Severity: low** — Codex's judgement, which I agree with: a real UUID collision between independently created sessions is extremely unlikely, so this is mostly a correctness-of-rule and a cost fix, not a live bug. |
| 3 | 1 | My pre-stated decision (a): the roster `summary` line stays empty for codex | REJECTED (upheld) | Codex raised nothing against it. It is claude's own `ai-title`, which codex writes nowhere; filling it means generating a title or relabelling codex's first prompt. |
| 4 | 1 | My pre-stated decision (b): `lifecycle.refreshLastResponse` stays claude-only | REJECTED (upheld) | Codex, round 1: "a separate existing pub/sub reader; I found no third changed-site of the same Claude-transcript blindness", and again in C-bis: "remains the separate accepted Claude-only path you identified". |
| 5 | 1 | **Protocol:** round 1 omitted the `FINDINGS COMPLETE` line | NOTED | Asked in C-bis rather than spending a round. Codex: "The protocol gap is real. I did not reach the two non-exchange consumers of the summary in sessionDetail... The round-1 comments represented every finding I had identified at that time, but they were not every finding present in the diff." Row 2 is exactly what it did not reach. Round 2's prompt names this as a late-finding note. |

That table is context, not a list of closed topics. If a FIXED row's fix does not do what it
claims, say so and mark it 'REOPENING #<row>' with specific evidence. Rows 3 and 4 are my
declined decisions and you accepted both — do not re-raise them unless you have new evidence.

  Then the axis table: one row per axis 1-11, findings or 'none'.
  Then, on its own line:
  Omit it if untrue and say which part you did not reach. You omitted it last round; do not
  omit it silently this round.

```

</details>

<details><summary>Codex reply (verbatim)</summary>

```text
CODEX VERDICT: LGTM

- No findings.
- Focused tests: 89 passed.
- Mutation checks confirmed the cache and Claude gate tests fail when broken.
- No source changes made.

| Axis | Result |
|---|---|
| 1. Correctness / edge cases | None. Empirical claims about six rollouts and benchmark timings are not source-verifiable, but no contradiction found. |
| 2. Security | None. |
| 3. Tests | None. Cache tests catch disabled caching; route test catches vacuous title gating. |
| 4. API / wire compatibility | None. `workPhase: null` is rendered as the normal plain status; non-Claude phases previously came only from incorrect Claude-transcript reads. |
| 5. Codebase consistency | None. |
| 6. Accessibility / i18n | None. |
| 7. Blast radius | Safe. Direct callers are `transcript-view-read`, both `session-reads` paths, `agent-badges`, and both `spawn-codex` paths; `codexRolloutExists` additionally serves resumable/survivor and WebSocket checks. Existing paths are safe under Codex’s unique immutable rollout IDs; resume callers specifically benefit from retaining the selected rollout. |
| 8. Async ordering / concurrency | None. Map mutations occur synchronously with no awaits; cap enforcement cannot interleave or produce torn state. |
| 9. Claude preservation | Verified by the passing Claude control test; mutation of the gate makes the non-Claude test fail. |
| 10. Gate predicate | Correct. Muse’s persisted-`claude` fallback remains unchanged and loses nothing. |
| 11. Merge from #2120 | No interaction found. #2120’s GridView changes are separate from this branch’s metadata-fetch changes. |

exit=0
```

</details>

isamu:
## Cross-review closed — 4 rounds, tier C

Exit bar met: Codex **LGTM** on `3e63d99b` with every axis answered and `FINDINGS COMPLETE`
present, nothing pushed during that round, and no MUST-FIX left on my side.

### Full ledger

| 0 | setup | Codex sandbox pre-flight | N/A | vitest exit 0, `vue-tsc -b` exit 0, socket bind exit 0, api.github.com HTTP 200. `rm` refused by shell-safety policy; `unlink` works, so restores use `git checkout --`. Nothing in this loop is UNVERIFIED for sandbox reasons. |
| 1 | 1 | P2 (Codex): `/api/session/:id` resolves the codex rollout path TWICE per poll — `agentBadges` and `sessionLastTurn` each call `codexRolloutPath`, which is synchronous `readdirSync` per day directory | FIXED | Measured first: 84 day dirs here, **11.05 ms warm per locate**, so nine codex cells cost 281 ms of blocking work per 4 s cycle. This PR is what introduced it — before, the roster sent no `?agent=` so it did ZERO codex scans. Fixed broader than proposed: memoized `codexRolloutPath` at process level (`server/agents/codex-sessions.ts`) rather than per request, because the dominant cost is the SAME id re-resolved every 4 s forever, which request-local reuse does not touch. Warm locate 11.05 → 0.08 ms; cycle 281 → 90 ms (the remainder is tail reads that pre-date the PR). Only a HIT is remembered (a miss is a cell whose rollout codex has not written yet); a hit is re-checked with one `existsSync` because codex prunes; capped at 512 since nothing prunes this map. Pinned by `test/server/agents/codex-rollout-path-cache.spec.ts` (7 cases); break-verified by three mutations — caching a miss → 2 red, dropping revalidation → 3 red, removing the cap → 1 red. Codex ACCEPTED in C-bis, including that it is not scope creep. |
| 2 | 1 | **Claude-side, not flagged by Codex:** the route read claude's transcript for THREE things and the original diff fixed only one — `workPhase` and the `userTurns`/`aiTitle` handed to `freshenRosterTitle` were still claude's for every agent | FIXED | Measured on a fixture, not reasoned: mounting the route over a claude transcript at the same id and asking as `agent=grok` answered `workPhase: "implementing"` and called `freshenRosterTitle` with claude's turn count and claude's own `ai-title` — which `storeKnownTitle` writes into `aiTitles` and publishes, i.e. a grok roster row wearing claude's title. Fix: `readSessionSummary` only for claude (`EMPTY_SUMMARY` otherwise) and `freshenRosterTitle` gated behind `agent === "claude"`. Also removes a stat and a fold per poll for every non-claude cell. Pinned by two cases in `session-detail-agent.spec.ts` (one gate, one claude control); break-verified — reverting the fix turns the gate case red. Codex ACCEPTED in C-bis and confirmed there is no FOURTH unconditional claude read on this route. **Severity: low** — Codex's judgement, which I agree with: a real UUID collision between independently created sessions is extremely unlikely, so this is mostly a correctness-of-rule and a cost fix, not a live bug. |
| 3 | 1 | My pre-stated decision (a): the roster `summary` line stays empty for codex | REJECTED (upheld) | Codex raised nothing against it. It is claude's own `ai-title`, which codex writes nowhere; filling it means generating a title or relabelling codex's first prompt. |
| 4 | 1 | My pre-stated decision (b): `lifecycle.refreshLastResponse` stays claude-only | REJECTED (upheld) | Codex, round 1: "a separate existing pub/sub reader; I found no third changed-site of the same Claude-transcript blindness", and again in C-bis: "remains the separate accepted Claude-only path you identified". |
| 5 | 1 | **Protocol:** round 1 omitted the `FINDINGS COMPLETE` line | NOTED | Asked in C-bis rather than spending a round. Codex: "The protocol gap is real. I did not reach the two non-exchange consumers of the summary in sessionDetail... The round-1 comments represented every finding I had identified at that time, but they were not every finding present in the diff." Row 2 is exactly what it did not reach. Round 2's prompt names this as a late-finding note. |
| 6 | 2 | Codex: LGTM, all 11 axes answered, `FINDINGS COMPLETE` present | N/A | Clean round on `56c26e5a`. Not an exit only because of rows 7-8 below, which I then pushed. |
| 7 | 2 | **Claude-side:** `summary` did not say WHOSE — the value is now read only for claude, and the name is where that invariant is cheapest to see | FIXED | Renamed `claudeSummary` in `8d1538da`. Identifier only; grepped the tests for it as a string literal first (no matches), so nothing reads the name as data. |
| 8 | 2 | **Claims sweep:** the PR body claimed the duplicate per-poll read was accepted and left out, which row 1 made false — and carried a figure ("single-digit ms warm") that was wrong, taken from an early probe on a RECENT rollout rather than across the store | FIXED | PR body corrected (no commit); plan file updated in `8d1538da` — it still described the change as the route plus two clients. The corrected body states the direction and the mechanism and points at the thread for the bench, rather than planting another magnitude in prose. |
| 9 | 3 | P3 (Codex): the plan's "What the review loop added" lead said neither change was about which log is read — false of the claude gate, which is precisely that decision | FIXED | `3e63d99b`. Reframed as what each change is: one per-poll cost, one extending the agent's-own-log rule to the two fields the first commit left behind. |
| 10 | 3 | **Claude-side claims sweep:** "a seventh agent" was wrong in BOTH the plan and the PR body — seven terminal agents are hosted today, so the next is the eighth, which this repo's own CLAUDE.md says and which the round-1 commit message said while the plan disagreed | FIXED | `3e63d99b` + PR body. Found by counting `TERMINAL_AGENTS` in the source rather than trusting the sentence. Also dropped a "three fields" that counted the `summary` line among those read from the agent's log — it is not one, which is the whole reason it stays empty for codex. |
| 11 | 4 | Codex: LGTM, all 11 axes, `FINDINGS COMPLETE` present, nothing pushed during the round | N/A | **Exit.** Clean round on `3e63d99b`, which is the head that will merge. Codex noted the plan's empirical measurements are not source-verifiable — true, and why the magnitudes were moved out of the prose and left in the thread with the command that produced them. |

### What each round found

| round | Codex | Claude | pushed |
|---|---|---|---|
| 1 | P2: the rollout path resolved twice per poll | the claude-transcript class fixed at 1 site of 3 | `56c26e5a` |
| 2 | LGTM | `summary` did not say whose; PR body carried a stale + wrong figure | `8d1538da` |
| 3 | P3: plan lead contradicted its own bullet | "a seventh agent" wrong on both surfaces | `3e63d99b` |
| 4 | LGTM | nothing | — |

### Why it took four rounds

Only round 1 was about the original diff, and it found a real defect on each side. The other
three were spent on what the loop itself produced — a rename, and prose that the fixes made
false. That is the documented shape: past the point where the diff has been read properly, a
round mostly reviews the review.

Two things are worth keeping, though, because neither would have been found by reading:

- **The round-1 class fix.** Codex reviewed the changed lines and did not reach the other two
  consumers of the same value, and said so when asked directly rather than leaving it to surface
  in round 3. Fixing one site of three is the failure mode the protocol names, and it was mine.
- **The wrong count.** "A seventh agent" was wrong in the plan *and* the PR body, while the
  commit message next to them said "eighth" correctly. Counting `TERMINAL_AGENTS` in the source
  found it; re-reading the sentence never would have.

### Verification

Every fix break-verified by mutation, not by reading: three mutations of the path memo go red
(caching a miss, dropping the revalidation, removing the cap), and reverting the transcript gate
turns its own case red. `format` / `lint` / `typecheck` / `build` / full `vitest run` clean
locally on the exit head. `knip` locally reports nothing this branch added.

**Still not exercised:** a browser rendering a codex roster row. Port 34567 is held by a running
MulmoTerminal on this machine and a second instance runs the boot-time idle-tmux sweep against
the same tmux server, so it was not started. The route was driven against the real on-disk stores
instead; `rosterRow` and the roster template are untouched.

isamu:
Follow-ups filed, and one correction to this PR's body.

- **#2123** — the roster `summary` line is blank for every agent but claude. Real, and the part of
  #2121 this PR did not deliver. Three options costed there; the honest one may be showing the
  agent's first prompt *labelled as such* rather than passing it off as a summary.
- **#2124** — `refreshLastResponse` reading claude's transcript for every agent. **This PR's body
  claimed that leaves the pub/sub row without a reply for codex cells. That was wrong**, and I only
  found out by checking it before filing: it is unreachable on four independent grounds (gated on
  `waiting`, which codex never sets; a failed read writes nothing; `ActivityPush` has no
  `lastResponse` field so nothing consumes it off that channel; and the roster reads these fields
  from this route anyway). Filed as a latent trap with that evidence rather than as a defect, and
  the body above is corrected.

The second one is the more useful outcome: the follow-up I was most confident about was the one
that did not survive being looked at.

PR #2129, opened 2026-09-18
feat(roster): drag-and-drop reorder in the cockpit roster (manual sort)

### LINKED ISSUE #2126: feat(roster): drag-and-drop reorder in the cockpit roster (manual sort)
## 概要

グリッドをズームしたときに左へ出る cockpit roster（`data-testid="cockpit"`）の各行を、**ドラッグ＆ドロップで並べ替え**できるようにしたい。

現状は #707 で入った各行右端の **⋮ → 上へ移動 / 下へ移動**（およびグリッドセル側の ◀▶）で隣と入れ替えるだけ。行数が多いと目的位置まで何度も押すことになり、roster を主戦場にしているときの手数が重い。

## なぜ設計とぶつからないか

並びの本体はフラットな `cells[]` ひとつで、グリッドと roster は同じ順序を読む（#720）。手で並べられるのは **`sortMode === "manual"` のときだけ**（`reorderable`）。auto / priority では並べても次の再計算で戻るので、DnD も **manual 限定**にすれば既存の三モード設計と両立する（#876 が priority を第三モードにした理由と同じ）。

#707 の ⋮ は「DnD を拒否した」のではなく、ズーム中に並べ替えできない穴を既存の `moveCell(uid, ±1)` で埋めたもの。DnD はその延長の別 UI。

## 提案

- **manual のときだけ** roster 行をドラッグして任意位置へドロップできる
- 行本体クリック（＝拡大するセルの切替）とドラッグを混ぜない（#707 が ⋮ に `@click.stop` を置いた理由と同じ）
- 末尾の空ランチャーは越えられない（いまの `canMoveCell` と同じガード）
- ⋮ メニューとグリッドの ◀▶ は残す（置き換えではなく追加）
- 実処理は既存の並び替え経路に流す。任意位置ドロップなら隣同士 swap の繰り返しではなく、`moveCellTo(uid, index)` のような操作を足すのが自然そう

## スコープ外（この issue ではやらない）

- auto / priority モードでの DnD
- グリッドのタイル面そのものの DnD（必要なら別 issue）
- `orderPriority` の自動書き込み

## 受け入れ条件

- [ ] manual 時、roster 行をドラッグして別位置へ落とすと並びが変わり、グリッド側の表示順も同じになる
- [ ] auto / priority 時はドラッグ開始しない（または no-op）
- [ ] ドラッグ中・ドロップ後に、意図せず拡大セルが切り替わらない
- [ ] 末尾の空ランチャーの後ろへは落とせない
- [ ] ⋮ の上へ／下へ、およびキーボード操作の既存経路が壊れていない

isamu:
## Summary

The cockpit roster could only be reordered a step at a time. The ⋮ menu on each row (#707) and the
◀▶ buttons on a grid tile both drive `moveCell(uid, ±1)` — a swap with the neighbour — so moving a
row across a long list means pressing the same item over and over. #2126 asks for a drag.

Each roster row now carries a **drag handle**, and a row can be dropped at any position:

- **Manual sort only.** The handle rides on the existing `reorderable` prop, which is already
  `sortMode === "manual"`. `auto` and `priority` recompute the order on every status change
  (`orderCells`), so a hand-placed row would be undone by the next recomputation; there is nothing
  new on those paths.
- **The ⋮ menu and the tiles' ◀▶ stay.** This is an added gesture, not a replacement — and the ⋮
  remains the keyboard route, which a drag cannot be. The handle is `aria-hidden` for that reason.
- **The order is still one flat `cells[]`**, so the tiled grid re-orders with the roster (#720).

### The state transform

`moveCell(state, uid, dir)` cannot express "drop it here", so there is a second one beside it:

```ts
canMoveCellBefore(cells, uid, beforeUid: number | null): boolean
moveCellBefore(state, uid, beforeUid: number | null): GridState   // null = the end of the list
```

**The destination is a uid, not an index.** `TerminalGrid` is handed `displayCells`; the reducer
holds `state.cells`. Those are the same array while zoomed in manual sort — the only state a drag
exists in — but that is an invariant held in two files, and an index is the payload that would
silently mean something else if it ever stopped holding. `null` for "past the last row" also names
the one case the guard has to refuse, instead of hiding it in an off-by-one.

The guard mirrors `canMoveCell`: an unknown cell on either side is refused; the no-op drops are
refused (onto yourself, and in front of your immediate successor), so a drag ending where it started
returns the same state object; and the trailing launch cell keeps the last slot, which here is
exactly "`beforeUid === null` is refused while the list ends in a launch cell that is not the one
being dragged". Dragging that launcher itself is allowed, as `canMoveCell(launchUid, -1)` already
allows.

### The gesture, and what it shows

- `draggable` is on the HANDLE, not the row: the row body's click swaps which terminal is enlarged,
  so making the row draggable would put a reorder and a navigation on one press. `@click.stop` on
  the handle covers a press that never became a drag — the separation #707 made for the ⋮.
- **What the drag shows is the list itself, re-ordered live and animated into place.** An insertion
  bar was built first and is gone: the roster's chrome already spends its border, its ring and its
  background on status and on "you are here" (`rosterAlertClasses.ts`), so a fourth mark competed
  with three that were already saying something — and it could not answer "where does that leave the
  others" at all.
- The preview applies `reorderBefore` to `listRows` while the reducer applies it to the cells, so the
  preview cannot describe a move the drop does not make. `<TransitionGroup>` with a `move-class`
  FLIPs the re-order; no `tag`, so it renders a fragment and the rows stay direct children of the
  aside's flex column. The duration rides in on a CSS variable because Tailwind generates utilities
  from literal source text — the same reason the zoom's FLIP passes `--flip-ms`.
- Which row the pointer names is a measurement over the rows' boxes (`dropSlot`), not an event
  target — it has to be, because the preview moves the rows around under the pointer.
- A slot the cell may not occupy, and the row's own slot, both LEAVE THE PREVIEW ALONE rather than
  collapsing it: a refusal that snapped the list back would make holding the pointer over the
  forbidden gap past the trailing launcher flicker the whole roster.

## Three things only running it could say

All three were found by driving Chrome against a throwaway server, and each one shipped green
through the unit tests first.

**1. `drop` does not always arrive, so `dragend` commits too.** A drop only fires on the element the
browser calls the current target, re-hit-tested as the drag moves. Re-ordering the list puts a
DIFFERENT row under a pointer that has not moved, so the row that accepted the last `dragover` is
not the one released on: the browser leaves it (a `dragleave` naming no `relatedTarget`) and ends
with **no `drop` at all**. The preview snapped back and the reorder was lost — a gesture that
visibly worked and did nothing, with nothing erroring anywhere. `dragend` always arrives, so
whichever comes first commits; the first clears the state the second reads.

That makes Escape the one ending that must commit nothing, and `dragend` reports a cancel exactly as
it reports a drop the browser declined — same type, same `dropEffect: "none"` — so the key is
watched for the duration of the drag.

Making the rows `pointer-events-none`, so the aside stays the drop target, was tried first and is
worse: the drag SOURCE loses hit-testing mid-gesture, which stalled the drag outright.

**2. The slot is measured from `offsetTop`, not `getBoundingClientRect`.** The FLIP moves a row with
a TRANSFORM, which the rect reports and the offsets do not. Measured with the rect, a pointer held
still during the slide was tested against boxes that were still moving — the roster showed one order
and the drop committed another. The offsets are the settled layout, which is where the rows are
going and what the drop will mean. The aside carries `relative` so it is the rows' offsetParent.

**3. A `dragleave` naming no element is not necessarily a leave.** The browser reports
`relatedTarget` as null for the release itself; reading that as "gone" wipes the target a beat
before the commit reads it. It reports null for a WINDOW EXIT too, though, and that one is a leave —
the pointer's position separates them (`pointerInside`). The second half of that came out of the
cross-review, not out of the browser run.

**4. The move animation is live only while a drag is.** The roster re-orders for a second reason —
an `auto` / `priority` re-sort on each status change — and `rosterAlertClasses.ts` rations motion in
this list to the one blinking state on purpose, because a working row already animates a spinner.
Animating the re-sort was an unrequested change on a frequent path.

## Items to Confirm / Review

- **`dragend` as a commit path is the part to look at hardest.** It is what Sortable-style lists do
  and it is why the gesture works at all, but it means the commit no longer hangs off an event that
  says "the user dropped here" — it hangs off "the drag ended and the roster still holds a target".
  The endings that must commit nothing are handled explicitly: Escape (watched), and the pointer
  leaving the roster (`dragleave` with a real `relatedTarget` clears the target). If there is a
  fourth ending, I have not thought of it.
- **Two guards where there was one.** `canDropCellBefore` ("may it occupy this slot") and
  `canMoveCellBefore` ("and would that change anything"). They had to separate once the preview
  existed — hovering the slot a row started in is a legal thing to do mid-drag. Worth checking the
  split reads as necessary rather than as duplication.
- **`GridView.vue` sat one line under `max-lines`**, so the lines this adds needed room. It came out
  of a real duplication rather than a shortened comment: `orderedCells.value.map(c => c.uid)` was
  written out in two places and is now the `orderUids` computed both read. That is the one change in
  this PR that is not the feature.
- **The handle's `title` is Japanese** (`ドラッグして並べ替え`), matching `CockpitRowMenu`'s
  `並べ替え` / `上へ移動` right next to it. These components use no i18n at all today; adding one
  English string beside two Japanese ones seemed worse than matching.
- **Touch is not covered.** HTML5 drag-and-drop does not fire on touch, so this is a pointer
  gesture; the phone's roster is a different component and untouched. The ⋮ remains the way to
  reorder without a mouse, and it is why the handle is `aria-hidden`.
- **`docs/facts.json` is left alone on purpose.** `yarn format` wants to reformat it (it is
  committed in a shape Prettier disagrees with), which is unrelated to this change and reverted
  here — but the next person running `yarn format` will hit the same thing.

## Verified by running it, not only by the suite

A second server on a throwaway `HOME` (port 34599, `dist/` built from this branch), three shell
cells in `~/ws/{a,b,c}`, one enlarged, driven through Chrome with Playwright:

- the handle appears on every row and the drag starts from it;
- mid-drag the roster already shows the order the drop will leave, animated into place (screenshot
  taken during the slide);
- the order the roster is showing at release is the order that is committed (`a,b,c` → `c,a,b`), and
  the enlarged terminal stays where it was;
- no uncaught console errors.

That run is the whole reason the three findings above are in the code rather than in the bug tracker.

`yarn format` / `lint` / `typecheck` / `build` are clean, and the component and common specs pass
(3568). The server-side failures in a full local `yarn test` are timeouts under a load average near
30 — single files taking six minutes — on files this diff does not touch; CI was green on
ubuntu / macOS / Windows for the first commit.

Coverage was mutation-checked rather than assumed, in both rounds: removing the slot guard, the
Escape cancel, the `dragend` commit, the lower-half split, the no-op refusal, the drag image, the
pointer-in-roster test, the per-dragover measurement, the drag-start reset or the move-class switch
each turns a test red.

## Cross-review (`/codex-cross-review`), tier C

Tiered C because the change is a state machine over browser drag events whose ordering decides
whether a reorder commits at all — and because `drop` is known not to arrive in the case this PR
exists to support. Round 1's prompt named the C-list axes and Codex answered them line by line with
`FINDINGS COMPLETE`.

**Round 1: `CHANGES REQUESTED`, two P2s, both real and both fixed.**

1. **A still pointer over a scrolling or resized roster committed the stale slot.** The dragover
   cache keyed only on `clientY`, and my comment justifying it reasoned only about re-orders. The
   rows also move for reasons the pointer knows nothing about — Chrome auto-scrolls an overflowing
   roster while you hold still near its edge, the splitter resizes it, and a row's own height
   changes as its summary and reply lines do. The cache is gone rather than re-keyed: no key over
   the pointer determines the geometry.
2. **Releasing outside the browser window committed the last in-roster slot.** A `dragleave` naming
   no element was always read as the release; a window exit reports the same null. The pointer's
   position separates them.

Two more are mine rather than Codex's — the move class animating every re-sort, and a header comment
still calling the row the drop target — and one note Codex explicitly declined to raise (a second
drag starting before the first ends) was acted on anyway, because `dragend` is the only event in the
gesture that arrives at an element.

The four choices in those fixes went back to Codex **before** the push and all four came back
ACCEPTED, including the one worth checking: `transition-none` as a `move-class` makes Vue 3.5 skip
the FLIP through `hasCSSTransform` rather than throwing or falling back to `v-move`.

**One edge is agreed rather than fixed:** where the roster's bottom edge is the viewport's, a drag
leaving the window downwards reports a `clientY` still inside the roster, so it reads as the release
and commits the slot on screen. Exits left, right and upward clear correctly. Codex: *"there is no
reliable extra distinction… committing the visible slot is defensible; Escape remains the explicit
cancel."*

Every prompt and reply is on the PR verbatim.

## User Prompt

- https://github.com/receptron/mulmoterminal/issues/2126 — "マニュアルの時だと出来そうだね。できる？"
  (manual sort looks like the case where this works; can you do it?)

Closes #2126

work in mulmoterminal2


isamu:
## Cross-review tier: **C**

The diff is small and its risk is not. The change is a **state machine over browser drag events**
whose ordering decides whether a reorder commits at all — `dragstart` / `dragover` / `dragleave` /
`drop` / `dragend` / `keydown`, with `drop` known not to arrive in the case this PR exists to
support. Async ordering is on the tier-C list, and the surface is `TerminalGrid.vue`, the one
component every enlarged session renders.

So round 1's prompt names the C-list axes this PR touches and asks for an answer per axis, not a
verdict over whatever the reviewer happened to look at:

- event ordering between `drop`, `dragend`, `dragleave` and the Escape keydown — every ending, and
  whether any of them can commit twice or commit nothing
- the preview / commit agreement (one `reorderBefore`, two callers)
- blast radius on `TerminalGrid.vue`: what the roster's new `relative`, the `TransitionGroup`
  fragment and the `@dragover` on the aside do to the modes that are NOT the roster
  (`docs/grid-view-modes.md` — tiled grid, filmstrip, the teleported cell, the collection pane)
- listener lifetime: the window `keydown` added per drag
- plus the default six (correctness, security, tests-go-red, wire/format compatibility,
  codebase consistency, a11y / i18n)

Exit bar is one clean round on a head nothing was pushed to.

isamu:
- P2: a still pointer over a scrolling/resized roster can commit the stale slot because the dragover throttle keys only on `clientY`.
- P2: dragging out of the browser window can still commit the last in-roster slot because `relatedTarget === null` is always treated as release, not leave.

Findings:

1. P2 should fix — Stale slot can commit after the roster scrolls or resizes while `clientY` is unchanged. Site: `src/components/TerminalGrid.vue:1365` (`if (event.clientY === lastDragClientY) return`). I grepped for `lastDragClientY`/`dropSlot`/`rosterRowBoxes`; this is the only occurrence of this throttle. The comment above `rosterRowBoxes()` correctly says scroll/resize moves the boxes, but the early return prevents re-measuring exactly when Chrome auto-scrolls the overflow roster under a stationary pointer, or when a splitter resize changes row geometry without a new Y. Repro shape: start a drag, hover near the bottom so a valid `rosterDrop` is set, let the roster scroll while the pointer stays at the same viewport Y, then release; `dragend` commits the old `beforeUid`, not the slot now under the pointer. Smallest fix: either remove this throttle, or make the cached key include the roster geometry that changes independently of the pointer, at least `scrollTop` and probably row count / client size. What would change my mind: evidence from a real Chrome/Safari/Firefox run that the roster cannot scroll or resize during native row drag without a different `clientY` dragover before release.

2. P2 should fix — Releasing outside the browser window can commit even though the pointer left the roster. Site: `src/components/TerminalGrid.vue:1419-1420` in `onRosterDragLeave`. I grepped for `relatedTarget`/`onRosterDragLeave`; this is the only roster leave handler. The code ignores every non-Node `relatedTarget`, which preserves the release-on-row case the PR found, but viewport/window exits also report `relatedTarget === null` in browsers. In that path `rosterDrop` remains set, no `drop` fires on the aside, and the handle’s `dragend` commits the last previewed slot. Smallest fix: for a null `relatedTarget`, only ignore it when the dragleave coordinates are still inside the roster rect; otherwise clear `rosterDrop` and `lastDragClientY` just like a normal leave. What would change my mind: a browser-matrix run showing that native HTML drag from this handle never emits a null-relatedTarget `dragleave` when exiting the viewport/window, or that the intended UX is explicitly to commit the last in-roster preview when the user releases outside the window.

Axis table:

| Axis | Result |
| --- | --- |
| 1. EVENT ORDERING | Findings above cover two endings. Released on a row: `drop` commits, then `dragend` sees cleared state and does not double-commit. Released on a different row after live reorder with no `drop`: `dragend` commits once. Released outside the roster to an in-document target: `dragleave` clears, then drop/dragend commit nothing. Released outside the window: P2, null `relatedTarget` can leave stale state and `dragend` commits once. Escape: expected no commit because keydown clears `rosterDrop` before dragend; no double path found. Component unmount mid-drag: `onBeforeUnmount(endRosterDrag)` removes state/listener, no commit. Second native drag before first ended: not reachable in the browser; a synthetic second `dragstart` would overwrite `dragUid` but not duplicate the listener because the callback/options are identical. |
| 2. PREVIEW vs COMMIT agreement | No finding on the normal invariant: zoomed + manual means `cellsToDisplay([...ordered], zoomed=true)` returns the whole ordered list and manual `orderCells` returns `state.cells`, so `props.cells` and `listRows` have the same uid order. The stale-geometry finding is the case where the preview target can stop matching the actual pointer slot. |
| 3. BLAST RADIUS | No finding for non-roster modes: the aside exists only under `zoomed && listMode`; filmstrip/tiled/collection teleport are not rendered through this aside. `relative` is scoped to the aside, and fragment `TransitionGroup` keeps rows as flex children. Zero rows produce no slot. Scrolled/resized roster has the P2 throttle finding above. |
| 4. LISTENER LIFETIME | No leak found in normal paths: `dragstart` adds one window keydown listener, `drop`/`dragend`/unmount remove it, and duplicate add/remove of the same listener is harmless. A second real drag cannot start before the first ends; if that assumption is wrong, `onRowDragStart` should call `endRosterDrag()` first, but I’m not raising it as a real-browser finding. |
| 5. GUARD SPLIT | No finding. `canDropCellBefore` is used for preview occupancy, `canMoveCellBefore` gates the reducer/no-op cases, and `reorderBefore` is intentionally unguarded. Unguarded `reorderBefore` is reachable only after `canDropCellBefore` in the component; if parent state changes before commit, `moveCellBefore` refuses. |
| 6. Correctness and edge cases | Findings: stale slot on scroll/resize and outside-window release. No finding for zero rows (`dropSlot` returns null), negative tops, or differing row heights; the pure helper handles those by row box. |
| 7. Security | None. No XSS/SSRF/path traversal/injection/secrets surface added; Vue interpolation/title binding and numeric uids are used, no HTML injection or network/file path entry point. |
| 8. Tests | Existing new tests cover handle visibility, drag image/dataTransfer, live preview, drop commit, dragend-without-drop commit, no double commit, Escape cancel, TransitionGroup, lower-half slot, trailing launcher refusal, drop at end without launcher, repeated-position throttle, own-row hover hold, click swallowing, foreign drag ignore, dragleave null vs body, guard/reducer helpers, and `dropSlot` geometry. Missing coverage for the two findings: scroll/resize at same `clientY`, and null-relatedTarget leave caused by viewport/window exit. |
| 9. API / schema / wire / on-disk compatibility | None. No schema or wire change. `GridState` persists the reordered `cells[]` exactly as existing manual moves do; `parseGridState` already rebuilds cells and sort mode without a new field. |
| 10. Consistency with codebase / CLAUDE.md | No blocking finding. No new `any`, eslint-disable, or casts in production code; no new `<style>` block. The production change does add `let lastDragClientY` and numeric constants `180`/`16`, but both are local gesture constants with comments and consistent with existing Vue code style. Docs still contain the pre-existing robot emoji in the PR body only, not repo files. |
| 11. Accessibility / i18n | None. The handle is intentionally `aria-hidden` and the adjacent ⋮ menu remains the keyboard route. User-facing strings are inline in this repo; the added handle title is Japanese, matching the neighboring row-menu strings, and there are no locale files to keep in step. |

isamu:
## Codex exchange — round 1

**Tier C.** Sandbox pre-flight (ledger row 0): vitest exit 0, `yarn typecheck` exit 0, socket bind exit 0, `api.github.com` reachable; `rm -f` refused by the command guard but `unlink` works, and Codex was told not to run a mutation sweep — it names them, I run them.

<details><summary>Round 1 — review prompt sent</summary>

```text
Review PR #2129 at https://github.com/receptron/mulmoterminal/pull/2129.

The diff is `git diff origin/main...HEAD` on branch feat/2126-roster-drag-reorder
(base commit 7d774709). Read the repo's CLAUDE.md and docs/grid-view-modes.md —
both are binding context here.

DO NOT run a mutation sweep: `rm` may be refused by the sandbox and a failed
restore silently mutates on top of a mutation. Name any mutation you want run and
I will run it.

Use the gh CLI to inspect the PR, then post inline review comments on specific lines via:
  gh api repos/receptron/mulmoterminal/pulls/2129/comments
or top-level comments via:
  gh pr comment 2129

The cockpit roster (the text list beside an enlarged terminal) could only be
reordered one step at a time. This adds a drag handle per row, active only in
manual sort. What the drag SHOWS is the list itself, re-ordered live and animated
with a Vue <TransitionGroup> move class; the preview and the commit both run
through `reorderBefore`.

Three behaviours in it were found by driving Chrome, not by reading:
  - `drop` does not always arrive (re-ordering puts a different row under a still
    pointer, so the row that accepted the last dragover is not the one released
    on), so `dragend` commits too, and Escape is watched because dragend reports
    a cancel exactly as it reports a declined drop;
  - the slot is measured from `offsetTop`/`offsetHeight`, not
    `getBoundingClientRect`, because the FLIP moves rows with a transform;
  - a `dragleave` whose `relatedTarget` is null is the release itself, not a leave.

This PR is tier C. The first five are the C-list axes it touches; answer each one
by name.

  1. EVENT ORDERING across the whole drag: dragstart / dragover / dragleave /
     drop / dragend / the window keydown. Enumerate the ENDINGS (released on a
     row, released outside the roster, released outside the window, Escape,
     the component unmounting mid-drag, a second drag started before the first
     ended) and for each say whether it commits once, twice, or not at all.
     `commitRosterDrag` is called from both `@drop` on the aside and `@dragend`
     on the handle — show me a path where that double-commits or where neither
     fires.
  2. PREVIEW vs COMMIT agreement. `rosterRows` renders
     `reorderBefore(props.listRows, dragUid, rosterDrop.beforeUid)` and the
     reducer applies `moveCellBefore(state, uid, beforeUid)`. Can they disagree —
     including when `props.listRows` and `props.cells` are not the same list
     (they are equal only while zoomed in manual sort; `displayCells.ts` and
     `orderCells` are where that equality comes from)?
  3. BLAST RADIUS on TerminalGrid.vue, which every enlarged session renders.
     docs/grid-view-modes.md describes three view modes plus a teleport into a
     collection pane. What do these do to the modes that are NOT the roster:
     the new `relative` on the aside, the `<TransitionGroup>` rendering a
     fragment, `@dragover`/`@drop` on the aside, and `offsetTop` measured against
     the aside as offsetParent (scrolled roster, resized roster, zero rows)?
  4. LISTENER LIFETIME. A window `keydown` listener is added in `onRowDragStart`
     and removed in `endRosterDrag` and `onBeforeUnmount`. Find a path that leaks
     it or removes it twice, including a dragstart that fires while a previous
     drag is still in flight.
  5. GUARD SPLIT. `canDropCellBefore` ("may it occupy this slot") vs
     `canMoveCellBefore` ("and would that change anything") vs `reorderBefore`
     (order only, no guard). Is any caller using the wrong one? Is
     `reorderBefore`'s ungurded contract reachable with inputs that produce an
     order `moveCellBefore` would refuse?
  6. correctness and edge cases (including `dropSlot` over zero rows, negative
     tops, a scrolled roster, and rows of differing heights)
  7. security (XSS / SSRF / path traversal / injection / secrets)
  8. tests: does a test go red if the new behaviour breaks — name the test for
     each new behaviour, and name any new behaviour that has none
  9. API / schema / wire / on-disk compatibility (the grid state is persisted to
     localStorage; see parseGridState)
 10. consistency with the rest of the codebase and with CLAUDE.md (no `any`, no
     `as` casts, no eslint-disable, functions under 20 lines, Tailwind utilities
     only and never a <style> block, no emojis, const over let, map/filter over
     for loops, no magic numbers)
 11. accessibility (the handle is an aria-hidden span; the ⋮ menu beside it is
     the keyboard route) and i18n lockstep — note this repo has NO locale files,
     user-facing strings are inline, so there is no dictionary to keep in step

   occurrence NOW — grep for it.

_no findings yet — this is round 1._

  Then, on its own line:
  not reach and why.

```

</details>

<details><summary>Round 1 — Codex stdout (the review itself was posted separately, above)</summary>

```text
Posted the full top-level review comment on PR #2129:

https://github.com/receptron/mulmoterminal/pull/2129#issuecomment-5725716081

I raised two P2 findings: stale slot commits when the roster scrolls/resizes at the same `clientY`, and stale commit when leaving/releasing outside the browser window with `relatedTarget === null`.

Verification run: `yarn vitest run test/src/components/rosterDrag.spec.ts test/src/components/gridTabs.spec.ts test/src/components/TerminalGrid.spec.ts` passed, 284 tests.
```

</details>

<details><summary>Round 1 C-bis — four choices put back to Codex BEFORE pushing (prompt)</summary>

```text
Follow-up on PR #2129, round 1. Do NOT re-review the PR and do NOT post anything.

I accepted both of your P2 findings. I am asking you to judge FOUR choices I made in the fix, not
to find anything new. For each, say ACCEPTED (my reasoning holds) or DISPUTED (with the specific
evidence that refutes it), and if DISPUTED, the single change that would settle it.

The working tree at /Users/isamu/ss/llm/mulmoterminal2 has the fixes applied (uncommitted). Read
src/components/TerminalGrid.vue and src/components/rosterDrag.ts as they now stand.

(1) YOUR FINDING 1 — I removed the throttle entirely rather than keying the cache on scrollTop and
    geometry. `lastDragClientY` is gone; `rosterRowBoxes()` runs on every dragover. My reasoning: a
    cache whose key does not determine the value is the bug, and the second key you proposed
    (scrollTop + row count + client size) is still not the whole of what moves a row — a row's own
    height changes when its summary/prompt/reply lines change, which happens while an agent runs.
    The cost is one getBoundingClientRect plus one offsetTop/offsetHeight per row, inside a single
    event, which the browser resolves in one layout pass; the roster is tens of rows, not thousands.
    Is removing it right, or is there a measurement that says otherwise?

(2) YOUR FINDING 2 — I used the pointer-in-rect test you proposed. There is an edge I want judged:
    when the roster's bottom edge IS the viewport's bottom edge (it usually is — the aside fills the
    stage), a pointer exiting the window downwards reports a clientY inside the rect, so that exit
    is read as the release and the preview commits. Exits left, right and upward are outside the
    rect and clear correctly. I decided committing is the defensible answer there, because the
    roster was showing that slot and Escape is the explicit cancel. Do you agree, or is there a
    signal that separates "released inside the roster" from "left through the bottom edge"?

(3) MY OWN FINDING, not yours — the TransitionGroup move class is now live only while a drag is
    (`:move-class="dragUid !== null ? ROSTER_MOVE_CLASS : ROSTER_STILL_CLASS"`, the latter being
    `transition-none`). Before, every re-order of this list animated, including the `auto` and
    `priority` sort modes re-ranking rows on every status change. src/components/rosterAlertClasses.ts
    lines 30-32 state that motion in this roster is deliberately rationed to one blinking state
    because a working row already animates a spinner. Is scoping it to the drag right, or does
    `transition-none` as a move class do something to Vue's TransitionGroup I have not anticipated
    (does it still FLIP, does it throw, is there a better way to say "do not animate this move")?

(4) YOUR AXIS-4 NOTE — you said a second real drag cannot start before the first ends and did not
    raise it. I added `endRosterDrag()` as the first line of `onRowDragStart` anyway, because
    `dragend` is the only event in the gesture that arrives at an ELEMENT rather than at the roster,
    so a drag whose handle is unmounted mid-gesture (a sort-mode flip removes every handle) ends
    without one and leaves `dragUid` set. Is that path real, or am I defending against nothing?

Reply as plain text.
```

</details>

<details><summary>Round 1 C-bis — Codex reply, verbatim</summary>

```text
(1) ACCEPTED. The current code measures `rosterRowBoxes()` on every `dragover`, and your reasoning holds: the removed cache key did not determine the measured value. The local code also uses `offsetTop` / `offsetHeight` specifically to avoid transform-in-flight rect drift, so a stale cache would undercut that fix. I see no measurement here that refutes the cost argument.

(2) ACCEPTED. In this handler, `relatedTarget === null` plus coordinates is the only signal being used, and when the roster bottom equals the viewport bottom there is no reliable extra distinction between “released at the bottom edge while still over the roster” and “left through that same bottom edge.” Committing the visible slot is defensible; Escape remains the explicit cancel.

(3) ACCEPTED. Vue 3.5’s `TransitionGroup` checks `hasCSSTransform(..., moveClass)` before applying move FLIP. With `moveClass="transition-none"`, the cloned element has no transform transition, so Vue returns early and does not apply the move translation/classes. It does not throw, and this is a reasonable way to say “no move animation” while avoiding the default `v-move`.

(4) ACCEPTED. That path is real enough to defend against. The only `@dragend` handler is on the draggable handle span; if sort-mode changes unmount that handle mid-gesture, the component can miss the cleanup path and leave `dragUid` / `rosterDrop` set. Calling `endRosterDrag()` first in `onRowDragStart` is a low-risk stale-state guard, not defending against nothing.
```

</details>

isamu:
## Round 1 triage — every finding and what happened to it

Codex verdict: `CHANGES REQUESTED`, 2 × P2, axis table answered line by line, `FINDINGS COMPLETE` present. Both accepted. Two more are mine, one was considered and rejected, and one is a note Codex explicitly declined to raise that I acted on anyway.

| 0 | 0 | sandbox pre-flight | N/A | vitest exit 0, `yarn typecheck` exit 0, socket bind exit 0, api.github.com reachable. `rm -f` refused by the command guard; `unlink` exits 0. Codex was told NOT to run a mutation sweep — it names them, I run them. |
| 0b | 0 | pre-existing CI baseline | N/A | All checks green on 4df15f8b (jscpd, knip, triage, Socket, CodeRabbit, package-smoke, lint-and-build ubuntu+macOS, test_windows). No pre-existing red to inherit. |
| 1 | 1 | P2 (codex): a still pointer over a scrolling/resized roster commits the stale slot — the dragover cache keys only on `clientY` | FIXED | the cache is gone; `rosterRowBoxes()` runs on every dragover. Test "re-answers the same pointer position when the roster itself moved under it" (scrollTop moves two rows under a still pointer). A cache keyed on scrollTop would still miss a row whose own height changed. |
| 2 | 1 | P2 (codex): releasing outside the window commits the last in-roster slot — `relatedTarget === null` is always read as the release | FIXED | `pointerInside(rosterRect, clientX, clientY)` separates them (rosterDrag.ts, own spec). Test "puts the rows back when the drag leaves the window, and commits nothing". Known edge: an exit through the viewport's BOTTOM edge, where the roster's bottom is the viewport's, still reads as the release — raised with codex in round 1's C-bis. |
| 3 | 1 | mine, not codex: the TransitionGroup move class animated EVERY roster re-order, including the auto/priority re-sort on each status change | FIXED | scoped to `dragUid !== null`; the other branch is `transition-none`. rosterAlertClasses.ts:30-32 rations motion in this list to one blinking state on purpose. Test "animates the re-order while dragging, and not otherwise". |
| 4 | 1 | mine, not codex: the header comment still said "The DROP TARGET is the whole row", which the aside-level handlers made false | FIXED | rewritten to point at commitRosterDrag's explanation. |
| 5 | 1 | mine, considered and NOT fixed: a row vanishing from `listRows` mid-drag could strand `dragUid` | REJECTED | not reachable — the grid state is per-browser localStorage and a cell is removed only by a user action the pointer is busy for. The `endRosterDrag()` now opening `onRowDragStart` makes any stranded state self-heal on the next drag anyway (row 6). |
| 6 | 1 | codex axis 4 noted a second drag cannot start before the first ends, and declined to raise it | FIXED anyway | `onRowDragStart` opens with `endRosterDrag()`. `dragend` is the only event in the gesture that arrives at an ELEMENT, so a handle unmounted mid-gesture ends the drag without one. Test "starts each drag from a clean slate". |

**The four choices in the fixes went back to Codex before the push** (C-bis, transcript above) and all four came back ACCEPTED — including the one I most wanted checked: that `transition-none` as a `move-class` makes Vue 3.5 skip the FLIP via `hasCSSTransform` rather than throwing or falling back to `v-move`.

**Known edge, agreed rather than fixed:** when the roster's bottom edge is the viewport's, a drag leaving the window downwards reports a `clientY` still inside the roster, so it reads as the release and commits the slot the roster was showing. Exits left, right and upward clear correctly. Codex: "there is no reliable extra distinction… committing the visible slot is defensible; Escape remains the explicit cancel."

isamu:
- P3: the new `pointerInside` spec introduces an avoidable `as DOMRect` cast; use the real DOMRect constructor instead.

Finding:

1. P3 nit — avoid adding a new `as DOMRect` cast in the new spec. Site: `test/src/components/rosterDrag.spec.ts:76`. I grepped for `as DOMRect`; this is the only occurrence of this exact cast. The repo style axis for this review explicitly calls out no `as` casts, and this one is easy to avoid because the test wants an actual DOM rectangle with left/top/right/bottom. Smallest fix: replace it with `const box = new DOMRect(10, 50, 200, 600);` (or `new window.DOMRect(...)` if the test environment wants the window constructor), which gives left 10, top 50, right 210, bottom 650 without a type assertion. What would change my mind: evidence that this Vitest/jsdom environment does not provide `DOMRect`, or that the constructor produces different boundary values than the object literal in this test.

Axis table:

| Axis | Result |
|---|---|
| 1. EVENT ORDERING | No behavioral finding. Released on a row commits once: `drop` and `dragend` both call `commitRosterDrag`, but the first call clears `dragUid`/`rosterDrop`, so the second is a no-op. Released outside the roster but inside the window clears `rosterDrop` on `dragleave` and commits zero. Released outside the window clears zero/one depending on the already-agreed bottom-edge case; left/right/up exits clear and commit zero. Escape sets `rosterDrop` null, so later `dragend` commits zero. Component unmount calls `endRosterDrag` and commits zero. A second drag start first calls `endRosterDrag`, so stale state from the first drag commits zero and the new drag starts clean. I do not see a valid in-roster release path where both commit twice or neither fires; when `drop` is missing, `dragend` covers it. |
| 2. PREVIEW vs COMMIT agreement | No finding. During reachable drags, `reorderable` means manual sort, `orderCells` returns `state.cells`, and zoomed `displayCells` is the full ordered list, so `props.cells`, `props.listRows`, and the reducer’s `state.cells` name the same uid order. Preview and reducer both use `reorderBefore`; `canDropCellBefore` gates preview slots and `moveCellBefore` refuses no-op commits. |
| 3. BLAST RADIUS on `TerminalGrid.vue` | No finding. The new `relative` is scoped to the roster aside and is needed for `offsetTop`’s offset parent. `TransitionGroup` renders a fragment, leaving rows direct flex children. `dragover`/`drop` on the aside are inert unless `dragUid !== null`. Tiled grid, filmstrip, zoom teleport, and collection-pane teleport do not render this aside. Scrolled roster is handled by `rect.top - scrollTop`; resized roster is remeasured each `dragover`; zero rows return no slot. |
| 4. LISTENER LIFETIME | No finding. `onRowDragStart` opens with `endRosterDrag`, then adds one window `keydown` listener. `commitRosterDrag`, unmount, and the next drag start all remove it; duplicate removes are harmless. A missing element-level `dragend` is self-healed by unmount or the next drag start. |
| 5. GUARD SPLIT | No finding. `canDropCellBefore` is used for “may occupy this slot” preview gating, `canMoveCellBefore` is used by `moveCellBefore` for “would this change state,” and `reorderBefore` remains intentionally unguarded/order-only. I do not see a reachable drag input where the unguarded preview can show an order the reducer would refuse, because preview updates only after `canDropCellBefore` accepts the slot. |
| 6. correctness and edge cases | No finding. `dropSlot` covers empty lists, negative tops, below/above-list coordinates, zero-height rows, and differing row heights via per-row top/height. Scrolled and resized rosters are remeasured every `dragover`. The round-1 cache and null-relatedTarget fixes do what they claim; I am not reopening #1 or #2. |
| 7. security | None. This is local drag state and DOM geometry; no new XSS/SSRF/path traversal/injection/secrets surface. Vue interpolation still escapes roster text. |
| 8. tests | No behavioral finding. The new tests go red for the named behaviors: `re-orders the rows live...`, `commits on dragend when no drop event arrives at all`, `commits once when the drop arrives and dragend follows it`, `commits nothing when the drag is cancelled with Escape`, `starts each drag from a clean slate...`, `re-answers the same pointer position when the roster itself moved under it`, `puts the rows back when the drag leaves the window...`, `animates the re-order while dragging, and not otherwise`, plus the `gridTabs` guard/reducer specs and `rosterDrag` pure helper specs. Native browser DnD delivery itself is still represented by simulated component events, not a new e2e test, which matches the PR’s stated jsdom limitation. |
| 9. API / schema / wire / on-disk compatibility | None. No persisted shape changes. `moveCellBefore` reorders existing `cells[]`; `parseGridState` and localStorage format remain compatible. |
| 10. consistency with codebase / CLAUDE.md | P3 finding above: one new avoidable `as DOMRect` cast in `test/src/components/rosterDrag.spec.ts:76`. Otherwise no new `any`, eslint-disable, schema drift, or non-Tailwind styling issue found in the drag change; the pre-existing scoped style block in `TerminalGrid.vue` is outside this PR’s new surface. |
| 11. accessibility and i18n lockstep | No finding. The drag handle is `aria-hidden` and the adjacent menu remains the keyboard route. The Japanese title matches the existing inline Japanese reorder strings; this repo has no locale dictionary to update. |

Verification I ran: `yarn vitest run test/src/components/rosterDrag.spec.ts test/src/components/gridTabs.spec.ts test/src/components/TerminalGrid.spec.ts` passed (291 tests), and `git diff --check origin/main...HEAD` passed.

isamu:
## Codex exchange — round 2

Verdict: `CODEX VERDICT: CHANGES REQUESTED` for one P3 (an `as DOMRect` cast I added in round 1's own fix). Every behavioural axis came back clean and neither round-1 fix was reopened.

<details><summary>Round 2 — review prompt sent (ledger compacted, REJECTED rows in full)</summary>

```text
Review PR #2129 at https://github.com/receptron/mulmoterminal/pull/2129.

This is ROUND 2. Round 1's two P2 findings are both fixed and pushed; the head to review
is 3d4ae77f. The ledger below says what happened to every finding, including three of
mine and one you declined to raise. Your four ACCEPTED answers in round 1's follow-up
are settled and do not need revisiting — but if a FIXED row's fix does not do what it
claims, say so and mark it 'REOPENING #<row>'.

The diff is `git diff origin/main...HEAD` on branch feat/2126-roster-drag-reorder
(base commit 7d774709). Read the repo's CLAUDE.md and docs/grid-view-modes.md —
both are binding context here.

DO NOT run a mutation sweep: `rm` may be refused by the sandbox and a failed
restore silently mutates on top of a mutation. Name any mutation you want run and
I will run it.

Use the gh CLI to inspect the PR, then post inline review comments on specific lines via:
  gh api repos/receptron/mulmoterminal/pulls/2129/comments
or top-level comments via:
  gh pr comment 2129

The cockpit roster (the text list beside an enlarged terminal) could only be
reordered one step at a time. This adds a drag handle per row, active only in
manual sort. What the drag SHOWS is the list itself, re-ordered live and animated
with a Vue <TransitionGroup> move class; the preview and the commit both run
through `reorderBefore`.

Three behaviours in it were found by driving Chrome, not by reading:
  - `drop` does not always arrive (re-ordering puts a different row under a still
    pointer, so the row that accepted the last dragover is not the one released
    on), so `dragend` commits too, and Escape is watched because dragend reports
    a cancel exactly as it reports a declined drop;
  - the slot is measured from `offsetTop`/`offsetHeight`, not
    `getBoundingClientRect`, because the FLIP moves rows with a transform;
  - a `dragleave` whose `relatedTarget` is null is the release itself, not a leave.

── ROUND 2 ALSO ASKS ───────────────────────────────────────────────────────
The fixes are themselves new code and have had one review fewer than the diff
they sit in. Read them as first-class:
  - `rosterRowBoxes()` now runs on every dragover with no cache. Is there a
    dragover path where it measures something misleading, or a cost you can
    measure that the removal makes unacceptable?
  - `pointerInside` and the two-branch `onRosterDragLeave`. Enumerate the
    leaves again against the new code.
  - `endRosterDrag()` opening `onRowDragStart`, which now also runs on the very
    first drag of a session.
  - the move class switching on `dragUid`, which changes what the roster does
    in `auto` and `priority` sort as well as during a drag.
  - the new tests: does each one go red for the reason its name gives?

This PR is tier C. The first five are the C-list axes it touches; answer each one
by name.

  1. EVENT ORDERING across the whole drag: dragstart / dragover / dragleave /
     drop / dragend / the window keydown. Enumerate the ENDINGS (released on a
     row, released outside the roster, released outside the window, Escape,
     the component unmounting mid-drag, a second drag started before the first
     ended) and for each say whether it commits once, twice, or not at all.
     `commitRosterDrag` is called from both `@drop` on the aside and `@dragend`
     on the handle — show me a path where that double-commits or where neither
     fires.
  2. PREVIEW vs COMMIT agreement. `rosterRows` renders
     `reorderBefore(props.listRows, dragUid, rosterDrop.beforeUid)` and the
     reducer applies `moveCellBefore(state, uid, beforeUid)`. Can they disagree —
     including when `props.listRows` and `props.cells` are not the same list
     (they are equal only while zoomed in manual sort; `displayCells.ts` and
     `orderCells` are where that equality comes from)?
  3. BLAST RADIUS on TerminalGrid.vue, which every enlarged session renders.
     docs/grid-view-modes.md describes three view modes plus a teleport into a
     collection pane. What do these do to the modes that are NOT the roster:
     the new `relative` on the aside, the `<TransitionGroup>` rendering a
     fragment, `@dragover`/`@drop` on the aside, and `offsetTop` measured against
     the aside as offsetParent (scrolled roster, resized roster, zero rows)?
  4. LISTENER LIFETIME. A window `keydown` listener is added in `onRowDragStart`
     and removed in `endRosterDrag` and `onBeforeUnmount`. Find a path that leaks
     it or removes it twice, including a dragstart that fires while a previous
     drag is still in flight.
  5. GUARD SPLIT. `canDropCellBefore` ("may it occupy this slot") vs
     `canMoveCellBefore` ("and would that change anything") vs `reorderBefore`
     (order only, no guard). Is any caller using the wrong one? Is
     `reorderBefore`'s ungurded contract reachable with inputs that produce an
     order `moveCellBefore` would refuse?
  6. correctness and edge cases (including `dropSlot` over zero rows, negative
     tops, a scrolled roster, and rows of differing heights)
  7. security (XSS / SSRF / path traversal / injection / secrets)
  8. tests: does a test go red if the new behaviour breaks — name the test for
     each new behaviour, and name any new behaviour that has none
  9. API / schema / wire / on-disk compatibility (the grid state is persisted to
     localStorage; see parseGridState)
 10. consistency with the rest of the codebase and with CLAUDE.md (no `any`, no
     `as` casts, no eslint-disable, functions under 20 lines, Tailwind utilities
     only and never a <style> block, no emojis, const over let, map/filter over
     for loops, no magic numbers)
 11. accessibility (the handle is an aria-hidden span; the ⋮ menu beside it is
     the keyboard route) and i18n lockstep — note this repo has NO locale files,
     user-facing strings are inline, so there is no dictionary to keep in step

   occurrence NOW — grep for it.

|---|---|---|---|---|
| 0 | 0 | sandbox pre-flight | N/A | vitest exit 0, `yarn typecheck` exit 0, socket bind exit 0, api.github.com reachable. `rm -f` refused by the command ... |
| 0b | 0 | pre-existing CI baseline | N/A | All checks green on 4df15f8b (jscpd, knip, triage, Socket, CodeRabbit, package-smoke, lint-and-build ubuntu+macOS, te... |
| 1 | 1 | P2 (codex): a still pointer over a scrolling/resized roster commits the stale slot — the dragover cache keys only on `clientY` | FIXED | the cache is gone; `rosterRowBoxes()` runs on every dragover. Test "re-answers the same pointer position when the ros... |
| 2 | 1 | P2 (codex): releasing outside the window commits the last in-roster slot — `relatedTarget === null` is always read as the release | FIXED | `pointerInside(rosterRect, clientX, clientY)` separates them (rosterDrag.ts, own spec). Test "puts the rows back when... |
| 3 | 1 | mine, not codex: the TransitionGroup move class animated EVERY roster re-order, including the auto/priority re-sort on each status change | FIXED | scoped to `dragUid !== null`; the other branch is `transition-none`. rosterAlertClasses.ts:30-32 rations motion in th... |
| 4 | 1 | mine, not codex: the header comment still said "The DROP TARGET is the whole row", which the aside-level handlers made false | FIXED | rewritten to point at commitRosterDrag's explanation. |
| 5 | 1 | mine, considered and NOT fixed: a row vanishing from `listRows` mid-drag could strand `dragUid` | REJECTED | not reachable — the grid state is per-browser localStorage and a cell is removed only by a user action the pointer is busy for. The `endRosterDrag()` now opening `onRowDragStart` makes any stranded state self-heal on the next drag anyway (row 6). |
| 6 | 1 | codex axis 4 noted a second drag cannot start before the first ends, and declined to raise it | FIXED anyway | `onRowDragStart` opens with `endRosterDrag()`. `dragend` is the only event in the gesture that arrives at an ELEMENT,... |

  Then, on its own line:
  not reach and why.

```

</details>

<details><summary>Round 2 — Codex stdout (the review itself is posted above)</summary>

```text
Posted the round 2 top-level review:

https://github.com/receptron/mulmoterminal/pull/2129#issuecomment-5725842479

Verdict: `CODEX VERDICT: CHANGES REQUESTED` for one P3 nit: the new `rosterDrag.spec.ts` uses an avoidable `as DOMRect` cast. Behavioral axes were otherwise clean; no reopenings for the two round-1 fixes.

Verification run passed: `yarn vitest run test/src/components/rosterDrag.spec.ts test/src/components/gridTabs.spec.ts test/src/components/TerminalGrid.spec.ts` with 291 tests, plus `git diff --check origin/main...HEAD`.
```

</details>

<details><summary>Round 2 C-bis — the fix was made differently from the one proposed; put back to Codex BEFORE pushing (prompt)</summary>

```text
Follow-up on PR #2129, round 2. Do NOT re-review the PR and do NOT post anything.

Your one P3 is fixed, but not the way you proposed, and I want you to judge the difference.

You suggested replacing `{ ... } as DOMRect` in test/src/components/rosterDrag.spec.ts with
`new DOMRect(10, 50, 200, 600)`. I instead removed the reason the cast existed: `pointerInside`
now takes a `PointerBox` — an exported interface naming the four edges it actually reads — rather
than a whole `DOMRect`. `rosterRoot.value?.getBoundingClientRect()` satisfies it structurally, so
the call site is unchanged, and the test states a box with no assertion of any kind. It matches
`RowBox` in the same file, which was already the precedent for "name the fields the rule reads".

Say ACCEPTED or DISPUTED. Specifically:
  - is there a reason to prefer the real `DOMRect` constructor here that I am missing (something
    about `DOMRect`'s derived `left/right/top/bottom` for negative widths, say)?
  - does widening the accepted type from `DOMRect` to a four-field structural interface let any
    caller pass something that is NOT a viewport-coordinate box, in a way that matters?

The working tree at /Users/isamu/ss/llm/mulmoterminal2 has it applied (uncommitted): read
src/components/rosterDrag.ts and test/src/components/rosterDrag.spec.ts as they now stand.

Reply as plain text.
```

</details>

<details><summary>Round 2 C-bis — Codex reply, verbatim</summary>

```text
ACCEPTED.

No, I don’t see a reason to prefer `new DOMRect(10, 50, 200, 600)` here. The function does not read `x`, `y`, `width`, or `height`, so DOMRect’s derived-edge behavior for negative dimensions is irrelevant unless the production code were constructing boxes from dimensions. It is not: `getBoundingClientRect()` already gives the edge fields the rule consumes.

The structural `PointerBox` widening also seems fine. Yes, it lets a caller pass any `{ left, right, top, bottom }`, including a non-viewport-coordinate box or an inconsistent one. But `DOMRect` was never a meaningful coordinate-provenance guarantee either, especially in TypeScript where the safety was mostly “has lots of DOMRect-shaped fields,” not “came from this roster in viewport space.” The important contract is semantic, and the exported interface plus comment names it accurately.

So the changed version is at least as good as the proposed DOMRect constructor fix, and arguably better because the test now states the rule’s real dependency instead of manufacturing a broader object.
```

</details>

### Triage

Accepted, fixed at the cause rather than at the site: `pointerInside` now takes a `PointerBox` — an exported interface naming the four edges the rule reads — so the test states a box with no assertion and `getBoundingClientRect()` still satisfies it structurally. That matches `RowBox` in the same file. Codex judged the substitution ACCEPTED and "arguably better, because the test now states the rule's real dependency instead of manufacturing a broader object."

| 0 | 0 | sandbox pre-flight | N/A | vitest exit 0, `yarn typecheck` exit 0, socket bind exit 0, api.github.com reachable. `rm -f` refused by the command guard; `unlink` exits 0. Codex was told NOT to run a mutation sweep — it names them, I run them. |
| 0b | 0 | pre-existing CI baseline | N/A | All checks green on 4df15f8b (jscpd, knip, triage, Socket, CodeRabbit, package-smoke, lint-and-build ubuntu+macOS, test_windows). No pre-existing red to inherit. |
| 1 | 1 | P2 (codex): a still pointer over a scrolling/resized roster commits the stale slot — the dragover cache keys only on `clientY` | FIXED | the cache is gone; `rosterRowBoxes()` runs on every dragover. Test "re-answers the same pointer position when the roster itself moved under it" (scrollTop moves two rows under a still pointer). A cache keyed on scrollTop would still miss a row whose own height changed. |
| 2 | 1 | P2 (codex): releasing outside the window commits the last in-roster slot — `relatedTarget === null` is always read as the release | FIXED | `pointerInside(rosterRect, clientX, clientY)` separates them (rosterDrag.ts, own spec). Test "puts the rows back when the drag leaves the window, and commits nothing". Known edge: an exit through the viewport's BOTTOM edge, where the roster's bottom is the viewport's, still reads as the release — raised with codex in round 1's C-bis. |
| 3 | 1 | mine, not codex: the TransitionGroup move class animated EVERY roster re-order, including the auto/priority re-sort on each status change | FIXED | scoped to `dragUid !== null`; the other branch is `transition-none`. rosterAlertClasses.ts:30-32 rations motion in this list to one blinking state on purpose. Test "animates the re-order while dragging, and not otherwise". |
| 4 | 1 | mine, not codex: the header comment still said "The DROP TARGET is the whole row", which the aside-level handlers made false | FIXED | rewritten to point at commitRosterDrag's explanation. |
| 5 | 1 | mine, considered and NOT fixed: a row vanishing from `listRows` mid-drag could strand `dragUid` | REJECTED | not reachable — the grid state is per-browser localStorage and a cell is removed only by a user action the pointer is busy for. The `endRosterDrag()` now opening `onRowDragStart` makes any stranded state self-heal on the next drag anyway (row 6). |
| 6 | 1 | codex axis 4 noted a second drag cannot start before the first ends, and declined to raise it | FIXED anyway | `onRowDragStart` opens with `endRosterDrag()`. `dragend` is the only event in the gesture that arrives at an ELEMENT, so a handle unmounted mid-gesture ends the drag without one. Test "starts each drag from a clean slate". |
| 7 | 2 | P3 (codex): the new `pointerInside` spec introduces an avoidable `as DOMRect` cast, which CLAUDE.md forbids | FIXED | fixed at the cause, not the site: `pointerInside` takes a `PointerBox` — the four edges it reads — so the test needs no assertion and the caller's `getBoundingClientRect()` still satisfies it structurally. Matches `RowBox`, the same file's precedent. |

isamu:
CODEX VERDICT: LGTM

| Axis | Result |
| --- | --- |
| 1. EVENT ORDERING | none. Endings checked: release on a row commits once via `drop`, then `dragend` sees cleared state; release on a row where the browser never sends `drop` commits once via `dragend`; release outside the roster but inside the window clears `rosterDrop` on `dragleave` and commits nothing; release outside the window also clears via `pointerInside(...) === false` and commits nothing; Escape clears only the target, so following `dragend` commits nothing; component unmount calls `endRosterDrag()` and commits nothing; a second real browser drag cannot start before the first drag operation ends, and the defensive `endRosterDrag()` at dragstart prevents stale state from carrying into a later gesture. I do not see a path where `drop` plus `dragend` double-commit, because `commitRosterDrag()` clears before emitting; I also do not see a normal in-roster release path where neither fires, because `dragend` covers the missing-`drop` case. |
| 2. PREVIEW vs COMMIT agreement | none. `rosterRows` previews with `reorderBefore(props.listRows, dragUid, beforeUid)`, and the reducer commits with `moveCellBefore` -> `reorderBefore(state.cells, uid, beforeUid)`. The only rendered drag handle is gated by manual sort; in that state `orderCells` returns `state.cells`, and while zoomed `cellsToDisplay` returns the whole ordered list, so `props.cells` and `props.listRows` are the same uid order. In auto/priority, the handle is absent. |
| 3. BLAST RADIUS | none. The `relative` on the aside is local and supplies the intended offset parent; `TransitionGroup` without a tag keeps rows as direct flex children; aside-level `dragover`/`drop` only prevent defaults after this component has started a roster drag; `offsetTop` plus `getBoundingClientRect().top - scrollTop` handles scrolled and resized rosters, differing row heights, and zero rows by returning no slot. Non-roster modes are behind `zoomed && listMode`, so tiled grid, filmstrip, and collection teleport paths are not affected beyond receiving unchanged `cells`. |
| 4. LISTENER LIFETIME | none. `onRowDragStart` begins by calling `endRosterDrag()`, then adds one window `keydown` listener; all normal endings call `commitRosterDrag()` or `endRosterDrag()`, and `onBeforeUnmount(endRosterDrag)` covers teardown mid-drag. Repeated `removeEventListener` is harmless, and I do not see a leak path in the reachable gesture lifecycle. |
| 5. GUARD SPLIT | none. The callers use the intended split: preview asks `canDropCellBefore` so legal no-op slots can keep showing the row in place, commit asks `canMoveCellBefore` via `moveCellBefore`, and `reorderBefore` remains the order-only primitive. Its unguarded contract is only reachable in production after `canDropCellBefore` has accepted the slot, and the reducer still refuses destinations that would not change state. |
| 6. correctness and edge cases | none. `dropSlot` has coverage for empty rows, negative tops, above-list, below-list, channels between rows, exact midpoints, zero-height rows, and scrolled geometry; TerminalGrid coverage adds same-pointer re-answering, scroll/geometry changes, own-row hover, trailing launcher refusal, and end-of-list drops without a launcher. |
| 7. security | none. This is local pointer/DOM event handling and array reordering. No new XSS, SSRF, path traversal, command injection, secret exposure, or unsafe HTML sink appears in the diff. |
| 8. tests | none. New behavior is covered by: `re-orders the rows live while dragging, and commits the slot the preview was showing`; `commits on dragend when no drop event arrives at all`; `commits once when the drop arrives and dragend follows it`; `commits nothing when the drag is cancelled with Escape`; `puts the rows back when the drag leaves the window, and commits nothing`; `re-answers the same pointer position when the roster itself moved under it`; `answers a pointer position it has already answered the same way`; `refuses the slot after the trailing launcher`; `drops a row at the end of a list that does not end in a launcher`; plus `gridTabs.spec.ts` for the guards/reducer and `rosterDrag.spec.ts` for pure slot and `pointerInside` rules. I also ran `yarn vitest run test/src/components/rosterDrag.spec.ts test/src/components/gridTabs.spec.ts test/src/components/TerminalGrid.spec.ts`: 3 files / 291 tests passed. |
| 9. API / schema / wire / on-disk compatibility | none. No wire/API schema changes. Persisted grid state remains the existing `cells[]` order plus `sortMode`; `moveCellBefore` only changes the in-memory order that already persists through the existing `parseGridState` path. |
| 10. consistency with codebase and CLAUDE.md | none. The round-2 fix removed the `as DOMRect` cast by narrowing the helper contract to `PointerBox`; I found no new `any`, `as` cast, eslint-disable, style block, or non-Tailwind style block in the changed source. The new functions are small, use existing Composition API patterns, and the targeted tests pass. |
| 11. accessibility and i18n lockstep | none. The drag handle is intentionally `aria-hidden` and not a keyboard control; the adjacent `CockpitRowMenu` remains the keyboard-accessible reorder path. The only new user-facing handle string is inline (`title="ドラッグして並べ替え"`), and this repo has no locale dictionary to keep in sync. |

Round-2 fix check: none. `PointerBox` is structural, but that does not make the rule read a caller wrongly in a way `DOMRect` would have caught: the function reads only `left/right/top/bottom`, TypeScript still requires those numeric properties, and `DOMRect` would not have enforced stronger invariants such as normalized edges. The spec still tests the rule directly with a plain `PointerBox`: inside, inclusive edges, outside each edge, negative viewport-exit coordinates, and missing box.

isamu:
## Codex exchange — round 3

Verdict: `CODEX VERDICT: LGTM`, every axis clean, `FINDINGS COMPLETE`, no reopenings. The review comment itself is above.

<details><summary>Round 3 — review prompt sent</summary>

```text
Review PR #2129 at https://github.com/receptron/mulmoterminal/pull/2129.

This is ROUND 3. The head to review is 7bc12d51. Round 2's one P3 is fixed — not the way
you proposed, and you judged the substitution ACCEPTED in round 2's follow-up, so that is
settled. Nothing else changed since the head you reviewed in round 2.

Round 2 answered every axis clean and reopened nothing. So the question this round is
narrow and you should say so if the answer is nothing: does the round-2 fix itself hold
(`PointerBox` widening `pointerInside`'s parameter, and the spec that no longer asserts a
DOMRect into existence), and is there anything in the diff you have still not reached?

The diff is `git diff origin/main...HEAD` on branch feat/2126-roster-drag-reorder
(base commit 7d774709). Read the repo's CLAUDE.md and docs/grid-view-modes.md —
both are binding context here.

DO NOT run a mutation sweep: `rm` may be refused by the sandbox and a failed
restore silently mutates on top of a mutation. Name any mutation you want run and
I will run it.

Use the gh CLI to inspect the PR, then post inline review comments on specific lines via:
  gh api repos/receptron/mulmoterminal/pulls/2129/comments
or top-level comments via:
  gh pr comment 2129

The cockpit roster (the text list beside an enlarged terminal) could only be
reordered one step at a time. This adds a drag handle per row, active only in
manual sort. What the drag SHOWS is the list itself, re-ordered live and animated
with a Vue <TransitionGroup> move class; the preview and the commit both run
through `reorderBefore`.

Three behaviours in it were found by driving Chrome, not by reading:
  - `drop` does not always arrive (re-ordering puts a different row under a still
    pointer, so the row that accepted the last dragover is not the one released
    on), so `dragend` commits too, and Escape is watched because dragend reports
    a cancel exactly as it reports a declined drop;
  - the slot is measured from `offsetTop`/`offsetHeight`, not
    `getBoundingClientRect`, because the FLIP moves rows with a transform;
  - a `dragleave` whose `relatedTarget` is null is the release itself, not a leave.

── ROUND 3 ALSO ASKS ───────────────────────────────────────────────────────
  - `PointerBox` is exported and structural. Can a caller now pass something the
    rule reads wrongly, in a way `DOMRect` would have caught?
  - Is `pointerInside`'s spec still testing the rule, now that the box is a
    plain object rather than a DOMRect?

This PR is tier C. The first five are the C-list axes it touches; answer each one
by name.

  1. EVENT ORDERING across the whole drag: dragstart / dragover / dragleave /
     drop / dragend / the window keydown. Enumerate the ENDINGS (released on a
     row, released outside the roster, released outside the window, Escape,
     the component unmounting mid-drag, a second drag started before the first
     ended) and for each say whether it commits once, twice, or not at all.
     `commitRosterDrag` is called from both `@drop` on the aside and `@dragend`
     on the handle — show me a path where that double-commits or where neither
     fires.
  2. PREVIEW vs COMMIT agreement. `rosterRows` renders
     `reorderBefore(props.listRows, dragUid, rosterDrop.beforeUid)` and the
     reducer applies `moveCellBefore(state, uid, beforeUid)`. Can they disagree —
     including when `props.listRows` and `props.cells` are not the same list
     (they are equal only while zoomed in manual sort; `displayCells.ts` and
     `orderCells` are where that equality comes from)?
  3. BLAST RADIUS on TerminalGrid.vue, which every enlarged session renders.
     docs/grid-view-modes.md describes three view modes plus a teleport into a
     collection pane. What do these do to the modes that are NOT the roster:
     the new `relative` on the aside, the `<TransitionGroup>` rendering a
     fragment, `@dragover`/`@drop` on the aside, and `offsetTop` measured against
     the aside as offsetParent (scrolled roster, resized roster, zero rows)?
  4. LISTENER LIFETIME. A window `keydown` listener is added in `onRowDragStart`
     and removed in `endRosterDrag` and `onBeforeUnmount`. Find a path that leaks
     it or removes it twice, including a dragstart that fires while a previous
     drag is still in flight.
  5. GUARD SPLIT. `canDropCellBefore` ("may it occupy this slot") vs
     `canMoveCellBefore` ("and would that change anything") vs `reorderBefore`
     (order only, no guard). Is any caller using the wrong one? Is
     `reorderBefore`'s ungurded contract reachable with inputs that produce an
     order `moveCellBefore` would refuse?
  6. correctness and edge cases (including `dropSlot` over zero rows, negative
     tops, a scrolled roster, and rows of differing heights)
  7. security (XSS / SSRF / path traversal / injection / secrets)
  8. tests: does a test go red if the new behaviour breaks — name the test for
     each new behaviour, and name any new behaviour that has none
  9. API / schema / wire / on-disk compatibility (the grid state is persisted to
     localStorage; see parseGridState)
 10. consistency with the rest of the codebase and with CLAUDE.md (no `any`, no
     `as` casts, no eslint-disable, functions under 20 lines, Tailwind utilities
     only and never a <style> block, no emojis, const over let, map/filter over
     for loops, no magic numbers)
 11. accessibility (the handle is an aria-hidden span; the ⋮ menu beside it is
     the keyboard route) and i18n lockstep — note this repo has NO locale files,
     user-facing strings are inline, so there is no dictionary to keep in step

   occurrence NOW — grep for it.

|---|---|---|---|---|
| 0 | 0 | sandbox pre-flight | N/A | vitest exit 0, `yarn typecheck` exit 0, socket bind exit 0, api.github.com reachable. `rm -f` refused by the command ... |
| 0b | 0 | pre-existing CI baseline | N/A | All checks green on 4df15f8b (jscpd, knip, triage, Socket, CodeRabbit, package-smoke, lint-and-build ubuntu+macOS, te... |
| 1 | 1 | P2 (codex): a still pointer over a scrolling/resized roster commits the stale slot — the dragover cache keys only on `clientY` | FIXED | the cache is gone; `rosterRowBoxes()` runs on every dragover. Test "re-answers the same pointer position when the ros... |
| 2 | 1 | P2 (codex): releasing outside the window commits the last in-roster slot — `relatedTarget === null` is always read as the release | FIXED | `pointerInside(rosterRect, clientX, clientY)` separates them (rosterDrag.ts, own spec). Test "puts the rows back when... |
| 3 | 1 | mine, not codex: the TransitionGroup move class animated EVERY roster re-order, including the auto/priority re-sort on each status change | FIXED | scoped to `dragUid !== null`; the other branch is `transition-none`. rosterAlertClasses.ts:30-32 rations motion in th... |
| 4 | 1 | mine, not codex: the header comment still said "The DROP TARGET is the whole row", which the aside-level handlers made false | FIXED | rewritten to point at commitRosterDrag's explanation. |
| 5 | 1 | mine, considered and NOT fixed: a row vanishing from `listRows` mid-drag could strand `dragUid` | REJECTED | not reachable — the grid state is per-browser localStorage and a cell is removed only by a user action the pointer is busy for. The `endRosterDrag()` now opening `onRowDragStart` makes any stranded state self-heal on the next drag anyway (row 6). |
| 6 | 1 | codex axis 4 noted a second drag cannot start before the first ends, and declined to raise it | FIXED anyway | `onRowDragStart` opens with `endRosterDrag()`. `dragend` is the only event in the gesture that arrives at an ELEMENT,... |
| 7 | 2 | P3 (codex): the new `pointerInside` spec introduces an avoidable `as DOMRect` cast, which CLAUDE.md forbids | FIXED | fixed at the cause, not the site: `pointerInside` takes a `PointerBox` — the four edges it reads — so the test needs ... |

That table is context so you do not spend this round re-deriving what earlier rounds
already decided. It is NOT a list of closed topics: if a resolution is wrong, or a FIXED
row's fix does not do what it claims, say so and mark it 'REOPENING #<row>' with the
specific evidence.

  Then, on its own line:
  not reach and why.

```

</details>

<details><summary>Round 3 — Codex stdout</summary>

```text
Posted the round-3 review comment: CODEX VERDICT: LGTM.

I read the full `origin/main...HEAD` diff, checked the round-2 `PointerBox` fix specifically, covered every requested axis, and ran the targeted specs:

`yarn vitest run test/src/components/rosterDrag.spec.ts test/src/components/gridTabs.spec.ts test/src/components/TerminalGrid.spec.ts`

Result: 3 files / 291 tests passed. Comment is here: https://github.com/receptron/mulmoterminal/pull/2129#issuecomment-5725913737
```

</details>

isamu:
## Cross-review closed — one clean round on the head that will merge

**Tier C, three rounds, exit on round 3: `CODEX VERDICT: LGTM`, every axis answered, `FINDINGS COMPLETE`, nothing pushed during that round.** The head it covers is `7bc12d51`, which is the head that will merge.

### What each round found

| round | verdict | what it found |
|---|---|---|
| 1 | CHANGES REQUESTED | two P2s, both real: a stale slot committed when the roster scrolls or resizes under a still pointer, and a release outside the browser window committing the last in-roster slot. Plus two of mine (the move class animating every auto/priority re-sort; a header comment the aside-level handlers had made false) and one note Codex declined to raise that I acted on anyway. |
| 2 | CHANGES REQUESTED | one P3, and it was in round 1's own fix: an `as DOMRect` cast CLAUDE.md forbids. Every behavioural axis clean, no reopenings. |
| 3 | **LGTM** | nothing. |

### Why it took three rounds

Round 2 exists because round 1's fix introduced the finding round 2 raised — the loop converging on its own output, which is the documented hazard. Round 3 is the one nobody can save: a fix is reviewed for the first time in the round after it is pushed. Nothing was trickled: both rounds answered every axis by name and both carried `FINDINGS COMPLETE`, and the four judgement calls in round 1's fixes plus the one in round 2's went back to Codex **before** each push rather than becoming the next round's findings — five disagreements settled in-round, all five ACCEPTED.

### Final ledger

| 0 | 0 | sandbox pre-flight | N/A | vitest exit 0, `yarn typecheck` exit 0, socket bind exit 0, api.github.com reachable. `rm -f` refused by the command guard; `unlink` exits 0. Codex was told NOT to run a mutation sweep — it names them, I run them. |
| 0b | 0 | pre-existing CI baseline | N/A | All checks green on 4df15f8b (jscpd, knip, triage, Socket, CodeRabbit, package-smoke, lint-and-build ubuntu+macOS, test_windows). No pre-existing red to inherit. |
| 1 | 1 | P2 (codex): a still pointer over a scrolling/resized roster commits the stale slot — the dragover cache keys only on `clientY` | FIXED | the cache is gone; `rosterRowBoxes()` runs on every dragover. Test "re-answers the same pointer position when the roster itself moved under it" (scrollTop moves two rows under a still pointer). A cache keyed on scrollTop would still miss a row whose own height changed. |
| 2 | 1 | P2 (codex): releasing outside the window commits the last in-roster slot — `relatedTarget === null` is always read as the release | FIXED | `pointerInside(rosterRect, clientX, clientY)` separates them (rosterDrag.ts, own spec). Test "puts the rows back when the drag leaves the window, and commits nothing". Known edge: an exit through the viewport's BOTTOM edge, where the roster's bottom is the viewport's, still reads as the release — raised with codex in round 1's C-bis. |
| 3 | 1 | mine, not codex: the TransitionGroup move class animated EVERY roster re-order, including the auto/priority re-sort on each status change | FIXED | scoped to `dragUid !== null`; the other branch is `transition-none`. rosterAlertClasses.ts:30-32 rations motion in this list to one blinking state on purpose. Test "animates the re-order while dragging, and not otherwise". |
| 4 | 1 | mine, not codex: the header comment still said "The DROP TARGET is the whole row", which the aside-level handlers made false | FIXED | rewritten to point at commitRosterDrag's explanation. |
| 5 | 1 | mine, considered and NOT fixed: a row vanishing from `listRows` mid-drag could strand `dragUid` | REJECTED | not reachable — the grid state is per-browser localStorage and a cell is removed only by a user action the pointer is busy for. The `endRosterDrag()` now opening `onRowDragStart` makes any stranded state self-heal on the next drag anyway (row 6). |
| 6 | 1 | codex axis 4 noted a second drag cannot start before the first ends, and declined to raise it | FIXED anyway | `onRowDragStart` opens with `endRosterDrag()`. `dragend` is the only event in the gesture that arrives at an ELEMENT, so a handle unmounted mid-gesture ends the drag without one. Test "starts each drag from a clean slate". |
| 7 | 2 | P3 (codex): the new `pointerInside` spec introduces an avoidable `as DOMRect` cast, which CLAUDE.md forbids | FIXED | fixed at the cause, not the site: `pointerInside` takes a `PointerBox` — the four edges it reads — so the test needs no assertion and the caller's `getBoundingClientRect()` still satisfies it structurally. Matches `RowBox`, the same file's precedent. |

### Claude's own sign-off

- `yarn format` / `lint` / `typecheck` / `build` clean; component and common specs 3583 passing.
- Mutation-checked in both rounds: the slot guard, the Escape cancel, the `dragend` commit, the lower-half split, the no-op refusal, the drag image, the pointer-in-roster test, the per-dragover measurement, the drag-start reset and the move-class switch each turn a test red when removed.
- **Driven in a real browser** (throwaway HOME, Chrome via Playwright) after round 1's batch: the order the roster is showing at release is the order committed, the enlarged terminal is unchanged, no console errors. The only change since that run is a parameter's type annotation and a spec file, neither of which reaches the bundle's behaviour.
- One edge is **agreed rather than fixed**, and is the thing to disagree with if anything: where the roster's bottom edge is the viewport's, a drag leaving the window downwards reads as the release and commits the slot on screen. Escape is the explicit cancel.

PR #2127, opened 2026-09-18
fix(keymap): say that Cmd+Shift+<uppercase letter> can never fire on macOS (#2125)

isamu:
## Summary

`"files-find": "Cmd+Shift+P"` loads, shows in Settings as bound, and never fires. While Cmd is held a
macOS browser puts the **unshifted** character in `KeyboardEvent.key`, so the keystroke arrives as
`"p"` and the binding waits for a `"P"` that never comes. `"Cmd+Shift+p"` fires on the identical
keys — the difference is the spelling in the file, not what the user presses.

Matching is case-sensitive by design (`"a"` and `"A"` are different keystrokes, pinned by
`test/common/keymap.spec.ts`), so this change does **not** touch matching. It closes the two gaps
#2125 names: nothing said so, and the thing that writes `keymap` did not know.

- **`validateKeymap` now warns** on a `meta` + `shift` + uppercase-ASCII-letter binding, naming the
  lowercase spelling — for an action and for a `send` entry alike, since both go through the same
  parser and matcher. **Non-fatal**, because the identical entry is correct for a Windows or Linux
  browser (`Meta+Shift+P` really does report `"P"` there) and the server cannot know which browser
  will connect — a Mac browser against a Linux host and the phone remote view are both normal here.
- **`enforceKeymap`'s warning headline** said *"ignoring unknown keymap entries"*. That was already
  wrong for a duplicate-keystroke warning and would be a flat lie here: the entry is kept, it just
  never fires. It now says these entries will not do what they say, true of all three warning kinds.
- **`mulmoterminal-keys`** gains the rule beside the `Option`+letter trap it is a twin of. Settings'
  Keyboard shortcuts section is read-only, so the skill is the **writer** of `keymap` — and it is
  what wrote the dead binding in #2125. Its `files-find` advice now offers `"Cmd+Shift+p"`.
- **`docs/guide/{en,ja}/config.md`**, in the macOS section beside `F1`–`F12` and `Option`+letter.

## Items to Confirm / Review

- **The platform fact is second-hand, and deliberately so.** This session could not post synthetic
  OS-level keystrokes (macOS denied `osascript` the keystroke permission), and a Playwright/CDP
  keypress cannot reproduce it — CDP supplies `key` itself instead of going through the macOS
  translation. So the evidence is the reporter's capture in #2125
  (`key: "p" code: "KeyP" metaKey: true shiftKey: true`) plus
[...]

PR #2131, opened 2026-09-18
test(docs): ship no keymap sample nobody validated (#2130)

### LINKED ISSUE #2130: test(docs): ship no keymap sample nobody validated — run the guide's and the skill's samples through validateKeymap
## 何が問題か

**出荷しているドキュメントの `keymap` サンプルが正しいかどうかを、誰も機械的に確かめていません。**

#2125 の PR (#2127) で、CodeRabbit が `docs/guide/en/config.md` の iTerm2 風サンプルが
`"next-attention": "Cmd+Shift+A"` を配っていることを見つけました。これは同じページの下の節が
「macOS では発火しない」と説明しているまさにその綴りです。ja 版にも同じものがありました。
つまり **ガイドが、そのページ自身が罠だと書いている binding を読者にコピーさせていた**ことになります。

見つけたのは bot で、しかも「diff の外」なのでインラインコメントすら付けられませんでした。
次に同じことが起きたときに気づく仕組みは、今もありません。

リリース手順（`CLAUDE.md`）には既に

> **Verify before committing**: every internal link resolves to a real page *and anchor*, and any
> config sample is run through its real validator

と書いてあります。**規則はあるが、実行するものが無い**——という状態です。

## 今どうなっているか

- `common/keymap.ts` の `validateKeymap()` は純粋関数で、fatal / 非 fatal の問題を配列で返します。
  サーバ起動時に `server/config/keymap-check.ts` がこれを使っています。
- テストから同じ関数を呼ぶのに必要なものは何もありません（DOM も fs も不要）。
- 先例あり: `test/docs/claude-requirement-claims.spec.ts` が、docs を読んで規則を守らせる spec です。

現状の棚卸し（このリポジトリの `validateKeymap` に実際に通した結果）:

| 置き場所 | keymap サンプル数 | 今の結果 |
|---|---|---|
| `docs/guide/{en,ja}/config.md` | 各 8 | 全部 clean |
| `docs/guide/{en,ja}/basics.md` | 各 1 | clean |
| `docs/guide/{en,ja}/v*.md`（日付入りリリースページ） | 計 12 | 全部 clean |
| `server/skills/mulmoterminal-keys/SKILL.md` | 3（うち 1 は `…` を含むスケッチ） | 残り 2 は clean |

**つまり今この瞬間は全部 green** です。#2127 で直したばかりなので。
この issue は「壊れているものを直す」ではなく「**次に壊れたときに気づく**」ための網です。

## 設計の選択肢

**A. `docs/guide/**` + bundled skill の `json` ブロックを全部通す（推奨）**
`json` フェンスを取り出し、`> ` の引用接頭辞を剥がし、`"keymap"` を含むものだけを
`validateKeymap` にかけて `[]` を要求する。fatal も警告もゼロであることを求めます。

- 利点: サンプルが死ぬ経路を実際に塞ぐ。実行コストはゼロに近い
- 注意点: `…` を含むブロックは**スケッチ**として明示的に飛ばす必要あり
  （`{ "keymap": { "send": [ … ] } }` は partial merge の罠を説明するための図であって、サンプルではない）

**B. A に加えて、対象を `keymap` 以外の設定サンプルにも広げる**
テーマ・ヘッダ・ランチャーなども同じ「コピーされる JSON」です。ただしそれぞれ別のバリデータが要り、
公開されていないものもあります。**この issue では見送り**、必要になったら別 issue で。

**C. 何もせず、リリース時の手作業チェックに任せる**
現状これです。#2127 で通り抜けました。

## 決めてほしいこと

- [ ] **日付入りリリースページ（`v*.md`）を対象に含めるか。**
      `CLAUDE.md` には「古いページを新しい挙動に合わせて書き換えるな、スナップショットだ」とあります。
      いっぽうリリース直後の読者が手順を写すのはまさにそのページで、`json` サンプルは「挙動の説明」ではなく
      「コピーされるもの」です。**推奨は含める**（今は全部 green なのでコストゼロ）。
      将来ルールが厳しくなって古いサンプルが引っかかったら、そのとき「直す / そのページを除外する」を
      意図的に決める——という注意書きを spec に残す想定です。
- [ ] **スケッチの飛ばし方**を `…` の有無で決めてよいか（allowlist を育てるより規則のほうが腐りにくい）。
      推奨は `…` 規則。
- [ ] 警告（非 fatal）も失敗扱いにしてよいか。**推奨は yes**——#2125 の綴り問題はまさに非 fatal です。

## 空振りしない担保

この手の spec は「正規表現が何にもマッチしなくなって、黙って全部 pass する」のが最大の失敗です。
そうならないよう:

- 抽出した件数の**下限をコードの定数**として持ち、下回ったら落とす
- `config.md`（en/ja 両方）が最低 1 件を出すことを要求する
- 抽出＋検証を**純粋関数**に切り出し、インラインの markdown 断片で**両方向**をテストする
  （`"Cmd+Shift+P"` を含む偽サンプルを渡したら必ず報告されること）

## 関連

- #2125 / #2127 — この穴が見つかった経緯

isamu:
## Summary

Nothing checked that a `keymap` sample this repo ships actually works. `CLAUDE.md` has required it
since the release checklist was written — *"any config sample is run through its real validator"* —
but nothing executes a checklist, so it held only while someone remembered. In #2127 nobody did:
the guide was handing readers `"next-attention": "Cmd+Shift+A"` in **both** languages, the exact
spelling the section below it describes as dead on macOS. CodeRabbit caught it, from **outside the
diff**, which is where a reviewer is least likely to look.

Every fenced `json` block under `docs/guide/**` and in a bundled `SKILL.md` that carries a `keymap`
is now parsed and passed through `validateKeymap` — the same function the server runs at startup.

- **Warnings fail as well as errors.** `Cmd+Shift+A` parses, loads and shows in Settings as bound;
  a warning is the only thing that ever says otherwise. Treating warnings as noise would leave
  exactly the hole this closes.
- **Files are enumerated from the directory**, never typed out — a hand-typed list silently drops a
  page, and the check written from the same list agrees with it.
- **Test-only.** No runtime file changes; `common/keymap.ts` is imported, not touched.

Nothing is broken today: all 18 samples pass, including every dated release page. This is the net
for the next one.

### Not vacuous, which is the whole risk

A doc-scanning guard fails **silently** the day its extractor stops matching — zero blocks found,
zero failures, green forever. Three guards, and I checked each by breaking the code:

1. The extraction and validation is a pure function in `test/support/keymapSamples.ts` with its own
   spec feeding it markdown in **both directions** — a clean sample and a `Cmd+Shift+A` one — so the
   reporting path is pinned rather than assumed.
2. The number of samples found is held against a floor in a constant.
3. Each of the four **living** guide pages (`config.md`, `basics.md`, both languages) must still
   yield at least one sample — a floor alone would pass if one page dropped out while another grew.

## Items to Confirm / Review

Three judgement calls, each reversible in a line. The first is the one I would most like a second
opinion on, because it touches a rule I invoked an hour ago in #2127:

- **Dated release pages (`v*.md`) are INCLUDED.** `CLAUDE.md` says never to edit an old dated page
  to match new behaviour, and in #2127 I rejected a CodeRabbit finding on exactly that ground. My
  reading: the snapshot rule is about **prose describing what a release did**, and a `json` block is
  not a description — it is a thing a reader copies, and the day a release ships its page is what
  they copy from. Every dated sample passes today, so including them costs nothing now. The spec
  says, where the decision is made, that a future conflict is to be decided deliberately (fix the
  sample, or exclude that page with a reason) rather than by dropping warnings. **Say if you would
  rather exclude `v*.md` outright.**
- **A block containing `…` is a SKETCH and is skipped.** The keys skill's
  `{ "keymap": { "send": [ … ] } }` illustrates the partial-merge trap and was never valid JSON. A
  rule rather than an allowlist, because an allowlist grows one entry at a time until it *is* the
  check. The risk it accepts: an ellipsis inside an otherwise-real sample would skip it silently —
  guard 2 above catches a wholesale loss, not a single one.
- **Scope is `keymap` only.** Themes, header buttons and launchers are copied out of the guide too
  and have their own validators. Four validators in one spec is a spec nobody reads; if that is
  wanted, it is its own issue.

## Verification

- `yarn format`, `yarn lint`, `yarn typecheck`, `yarn build`, `yarn knip` all clean (knip reports
  only the same pre-existing unused exports as on main).
- **Re-introduced the real regression**: putting `"Cmd+Shift+A"` back into
  `docs/guide/en/config.md` turns the file red with
  `docs/guide/en/config.md json block #26: keymap.next-attention "Cmd+Shift+A" — never fires in a
  macOS browser … write the key lowercase ("a")`. Restored afterwards and compared against a
  pristine copy.
- **Mutation-checked the guards**: breaking the fence regex fails the floor and all four
  living-page checks; removing the blockquote strip fails the two `config.md` pages (a sample sits
  inside a `{: .note }` callout). Pre-flight compared the file to a pristine copy before each
  mutation and after each restore.
- The full local suite is not a clean signal on this machine — load average sat near 30 and the
  I/O-heavy server specs time out, as `vitest.config.ts` documents. **CI is the verification for
  those**; these two spec files run in well under a second.

## User Prompt

Across this session, after #2127 merged:

> サンプル検証の spec、別 PR でやって

The idea itself came out of #2127's review loop — I proposed it there as a follow-up and left it out
of that PR because it is a different invariant.

Closes #2130

work in mulmoterminal3

## Summary by CodeRabbit

- **Tests**
  - Added automated validation for keymap examples in documentation and bundled skills.
  - Checks now cover JSON validity, supported actions, shortcut conflicts, platform compatibility, and validation warnings.
  - Coverage checks ensure required guide pages retain representative action and message-binding examples.
  - Reports identify the affected page and sample when issues are found.

- **Documentation**
  - Added a maintenance plan for validating keymap examples across current and dated documentation pages.


isamu:

| # | bot | finding | disposition | why / where |
|---|-----|---------|-------------|-------------|
| 1 | codex (local) | the ellipsis rule is too broad — a valid sample carrying `…` in a binding or in `send.bytes` is skipped in silence | FIXED | `102bc4f6`. A sketch is now **an ellipsis AND unparseable**, which is what a sketch structurally is. A spec pins the other direction: `{ "send": [{ "key": "Cmd+Shift+A", "bytes": "…" }] }` parses, so it is a sample, and it is reported |
| 2 | codex (local) | the vacuity guards are all aggregate — adding `!text.includes('"send"')` to the filter keeps the floor satisfied and the living pages covered while unguarding every `send` sample | FIXED | `102bc4f6`. The two shapes a keymap takes are now named. Mutation-checked with codex's own mutation: excluding every `send` sample fails the new guard |
| 3 | codex (local) | dated pages: the inclusion is defensible, but the maintenance policy should be **operational** rather than described | FIXED, partly | `102bc4f6`. A dated page that trips now says at the point of failure what the decision is. I did **not** add an empty exclusion list — an allowlist that exists is an allowlist that grows, and the conflict is hypothetical today (every dated sample passes) |
| 4 | codex (local) | indented fences are missed — the real `keymap.send` samples in `docs/guide/{en,ja}/v2.5.3.md` are skipped | **REJECTED** | Checked against the file rather than the claim: the fence regex is deliberately **unanchored**, so indentation never reaches it, and `JSON.parse` ignores leading whitespace. `jsonBlocks` finds 2 blocks in that page and `keymapSamples` returns the indented `send` sample as block #1 — it was already being validated before this PR's first commit. Anchoring the fence to the line start, as the finding implies, would have **broken** the blockquoted samples in `config.md`, which is the opposite of the fix |
| 5 | codex (local) | a mid-line ```json occurrence would be picked up | ACCEPTED, not changed | True, and it follows from the same unanchoredness that makes row 4 wrong. The error direction is safe: a false positive costs a reported problem, a false negative costs a dead sample shipped. Row 4 is why I would rather have the first |

Mutation results after the fixes, each with the file compared to a pristine copy before and after:
breaking the fence regex fails six checks, removing the blockquote strip fails the two `config.md`
pages, and excluding every `send` sample fails the shape guard.

**Note for the bots:** this repo runs Codex review from the local CLI — the CI `Codex auto-review`
workflow is switched off deliberately, so its absence is not a missing signal.

isamu:

| # | bot | finding | disposition | why / where |
|---|-----|---------|-------------|-------------|
| 6 | codex (local) | the shape guard matched action names as **text**, so weakening it to a pattern anything satisfies would keep it green while it proved nothing | FIXED | `ba5d152b`. `sampleShape` parses the sample and asks whether a `KEYMAP_ACTIONS` key holds a string and whether `send` is a non-empty array. Its own spec pins the case the docs guard leans on — a send-only sample must **not** count as an action binding — plus the both / empty-send / unknown-action / non-object cases |
| 4 | codex (local) | *(round 1)* indented fences are missed, the `v2.5.3` send samples are skipped | **WITHDRAWN by codex** | Round 2, after checking the file: *"Indented and blockquoted fences are correctly found; v2.5.3 send samples are not missed."* |

Codex round 2 also confirmed rows 1–3 resolved, and found no Windows path or line-ending problem.

**What the mutation runs actually show**, stated precisely because it is easy to overclaim here:
mutating `sampleShape` to always report an action turns the support spec red, so the helper's
behaviour is pinned. Editing the docs spec's own assertion to something meaningless (`shape !== null`)
still passes — no suite tests its own assertions. What changed is that gutting this guard now requires
writing something visibly empty, instead of swapping one regex for a slightly broader one that still
reads like a check.

Everything else this round: `yarn format`, `lint`, `typecheck`, `build` and `knip` clean, and
`knip` says nothing about the new files.

**CodeRabbit has not seen this PR at all** — it is rate-limited from the `full review` I asked for on
#2127, and its check here reads `Review rate limited`. That is a reviewer that never looked, not a
sign-off, and it is recorded rather than counted.

isamu:
## Review ledger — after iteration 3

**Correction to the previous ledger:** I wrote that CodeRabbit had not seen this PR. It had —
the rate limit cleared and it posted two inline comments. Both were real.

| # | bot | finding | disposition | why / where |
|---|-----|---------|-------------|-------------|
| 7 | CodeRabbit | `guidePages()` walks one level deep, so a page in a subdirectory is skipped | FIXED | `0e9a6337`. The walk is recursive, and it now covers **every** markdown file under `server/skills` rather than `SKILL.md` alone — a template or reference page beside it is copied the same way. Separators are normalised because `recursive` returns `\\` on Windows |
| 8 | CodeRabbit | the sample filter requires `"keymap"` **with quotes**, so a sample whose keys are unquoted is skipped — and unquoted keys are malformed JSON, the thing most worth reporting | FIXED | `0e9a6337`. Matches the key quoted or not; a spec pins that the unquoted sample is now reported as malformed rather than passed over. A prose mention inside another sample can match too — that block then parses, carries no keymap, and validates clean, which is the safe direction to be wrong in |
| 9 | codex (local) | round 3, reading the diff fresh with the thread set aside | **LGTM** | *"I would approve it… I found no remaining plausible mutation of test/support/keymapSamples.ts that both preserves its support-spec coverage and makes the production guard vacuous."* |

Neither of rows 7 and 8 describes content this repo has today: there is no nested guide page, and
every keymap mention in a json block is quoted. I fixed them anyway, because both are **a silent
skip inside a guard whose entire purpose is to stop silent skips** — the failure this PR exists to
prevent, committed in the guard itself.

Codex's one note for a human, unresolved by design: the `isSketch` exemption means a **malformed**
sample that also contains `…` is skipped. That is the accepted cost of not keeping an allowlist,
and it is stated where the rule is written.

Gates after these commits: `format`, `lint`, `typecheck`, `build`, `knip` clean.

isamu:
Merging. Final head `0e9a6337`, reviewed by both:

- **codex** (local CLI) — LGTM on this head, asked specifically about the last commit's matching
  logic: the recursive walk is correct on Windows and posix, and the broader key regex adds only a
  clean candidate rather than suppressing a malformed one.
- **CodeRabbit** — **No actionable comments** on the same range, after its two findings were fixed.
- All checks green. `test_windows` failed once on `codex-rollout-path-cache.spec.ts` and
  `shell-spawn-win.spec.ts` — neither imports anything this PR touches, this PR's own spec passed
  there (23 tests, 19ms), the previous head passed with the same two specs present, and another
  branch hit a different Windows-only failure in the same window. Re-run: green.

PR #2147, opened 2026-09-18
feat(files): follow the file on disk in an open Markdown view (#2136)

### LINKED ISSUE #2136: feat(markdown): 開いている Markdown ビューを、ファイルの変更に追随させたい（プレビュー / Canvas カード）
## やりたいこと

開いている Markdown の表示を、**ディスク上のファイルの変更に追随させたい**。とくに、別のセルで動いているエージェントが `.md` を書き換えたときに、開いたまま見ているものが更新されてほしい。

いまはブラウザをフルリロードすることだけが更新の契機になっています。

## なぜ

エージェントに設計メモや plan、README を書かせながら、隣のペインでそれを読む、という使い方をすると、書き換わるたびにリロードが必要になります。エージェントが書く側・人間が読む側という分担ではこれが常時発生するので、リロードを挟む回数がそのまま摩擦になります。

## いまの挙動

3つの表示面があり、状態が違います。

| 表示面 | 外部の書き込みへの追随 |
|---|---|
| Files ペインの **Preview** トグル（サーバーが描く iframe） | **しない**（リロードのみ） |
| Canvas の **presentDocument** カード | **しない**（アプリ自身の保存時のみ更新） |
| Files ペインの **エディタ** | する（ただし最大30秒のポーリング待ち） |

確認手順:

1. Files ペインで `.md` を開き、**Preview** に切り替える
2. 別のセルのエージェント（または外部エディタ）でその `.md` を書き換える
3. プレビューは変わらない。30秒以上待っても変わらない
4. ブラウザをリロードすると新しい内容が出る

Canvas 側: `presentDocument` でカードを出した `.md` を同じように外部から書き換えても、カードは更新されません。

## 足りていないもの（5.2.0 / main `426b08e9` のソースを読んだ結果）

**独立して2つあります。**

### 1. プレビューの iframe は URL が変わらないので再取得されない

```js
// src/components/FilesPane.vue:110
const previewSrc = computed(() => (openPath.value ? `/api/files/browse/md?${qs(openPath.value)}` : ""));
```

```html
<!-- src/components/FilesPane.vue:809 -->
<iframe :src="previewSrc" sandbox="" title="Markdown preview" />
```

`previewSrc` は `openPath` だけから計算されます。ファイルがディスク上で変わっても URL が1バイトも変わらないため、ブラウザは再取得せず、サーバーが前回レンダリングした HTML を出し続けます。

一方、**エディタ側はすでに外部編集に追随しています**。`checkForExternalChange`（同 515-529 行）が `/api/files/browse/version` を `EXTERNAL_CHECK_MS = 30_000`（30秒）ごとにポーリングし、加えて `FILE_WRITE_CHANNEL`（`common/fileWriteChannel.ts`）を購読。バッファが clean なら `loadFile(pathRel, true)` で中身を差し替えて `baseVersion` を更新し、dirty なら競合バナーを出します。

つまり**検出の仕組みはすでにあって、プレビューだけがそこに繋がっていない**状態です。

### 2. アプリ外の書き込みを拾うトリガーが無い

- Canvas の `presentDocument` カード（`@mulmoclaude/markdown-plugin`）は `plugin:markdown:file:<path>` を購読して live-refresh します（`src/plugins-registry.ts:130-145` の登録コメント）。配信経路そのものは完成しています。
- publish 側は `server/backends/fileChange.ts` の共有パブリッシャですが、**呼び出し元はアプリ自身の書き込みルートだけ** — `server/backends/markdown.ts` / `html.ts` / `shapescript.ts` の save 経路です。`FILE_WRITE_CHANNEL` も同様にアプリ側の書き込み（`server/routes/app-routes.ts:347`）でしか publish されません。
- **fs 監視はどこにもありません。** `server/` 全体で `fs.watch` を使っているのは `server/backends/collectionWatchers.ts`（collection の data dir、`server/index.ts:861` でマウント）だけで、markdown ファイルは監視対象外です。

結果として、**セル内のエージェントが `.md` を Edit したとき、どのチャネルにも publish されません。** Canvas カードは一切更新されず、Files ペインは30秒ポーリングだけが頼りで、そのうえプレビューは 1 の理由で更新されません。

## 提案（二段階）

**段階1 — プレビューをキャッシュバストする。** `previewSrc` に version を混ぜる。`/api/files/browse/md` は `cwd` と `path` しか読まない（`server/files/files-browse.ts:218` → `containedFor`）ので `&v=<version>` のような余分なクエリは無害に無視されます。`baseVersion` は既に外部編集の検出時に更新されるので、繋ぐだけでリロードが不要になり、最大30秒で追随します。

**段階2 — 開いているファイルに fs 監視を付け、既存の `publishFileChange` を呼ぶ。** 購読は `plugin:markdown:file:<path>` という **path 単位**のチャネルなので、監視対象は「いま開いている文書」だけで足ります。pubsub は socket.io の room で、`subscribe` / `unsubscribe` ハンドラと `subscriberCount` が `server/infra/pubsub.ts`（78行）にあるので、購読が付いたら監視を張り、最後の購読者が消えたら畳む形が自然に書けます。ワークスペース全体の再帰監視は不要です。

段階1 だけでも「リロードしなくても追随する」までは到達するので、分けて入れられます。

## 実装上の注意

- **アトミック書き込み（temp → rename）はパス指定の `fs.watch` を殺します。** ディレクトリを監視して basename で絞るか、mtime ポーリングにする必要があります。エージェントの編集がまさにこの形なので、ここを外すと「時々効かない」という一番厄介な壊れ方をします。
- **Windows**: `docs/windows-gotchas.md` の "Filesystem watching"（8.3 短縮パスに対する `fs.watch` がプロセスを落とす）。`server/session/codex-activity-watch.ts` は近い理由で**意図的にポーリングを選んでいる**前例です。
- チャネル名に含まれるのは client 由来の文字列なので、それを path として扱う前に既存の `containedFor` 相当の封じ込め検証を通すこと。
- **共有パッケージの規約（CLAUDE.md）**: `publishFileChange` は `@mulmoclaude/core/file-change` で MulmoClaude と共有です。`../mulmoclaude/server/events/file-change.ts` を読み、`fs.watch` / `watchFile` / `chokidar` を grep して確認したところ、**MulmoClaude 側にも markdown の fs 監視はありません**（両ホストとも「アプリ自身が保存したときだけ publish」）。監視を**ホスト側**に置いて既存の publisher を呼ぶ形ならチャネル名・ペイロードは不変で乖離になりません。core に入れると MulmoClaude の挙動を変えるので単独で決めない話になります。なお `fileChange.ts` は既に意図的な乖離を1件コメント宣言しています（markdown スコープを MulmoClaude の `artifacts/markdowns/**` より広げている）。

## 決めておくべきこと

**未保存の編集があるとき、プレビューは何を映すか。** プレビューは元々ディスクの内容をレンダリングしており、保存前はすでにエディタと違うものを見せています。現状の設計は「clean なら自動で取り込む、dirty なら競合バナーを出してユーザーに選ばせる」なので、「プレビューはディスクに従う」で一貫はしますが、実装中に発見するのではなく先に決めたい部分です。

## 関連

- **#1603** — `mulmoterminal watch`（CLI から Canvas の同じスロットを描き直す）。あちらの検討事項「Canvas 側が更新されたと分かる必要がある（開きっぱなしのビューが自動で追随するか）」がこの issue と重なります。あちらは CLI と再実行ループ、こちらは既に開いているビューがディスクに追随するかの話で、この issue は #1603 の前提条件のひとつを満たす形になります。
- **#2137** — 同じ Preview をめぐる別の要望。こちらは「開いたまま内容が変わったときに追随しない」、あちらは「開き直したときに Preview / Edit のモードが戻る」。
- **#184** — 全画面ファイルエクスプローラー＋Markdown 編集ビュー（このプレビューの出自）

## 環境

- mulmoterminal 5.2.0（npm、`npx mulmoterminal@latest` で起動）
- Node v26.8.2 / macOS 26.6.2 / Darwin arm64 / zsh
- Chrome 153.0.8010.53
- tmux 3.7c / gh 2.101.0 / claude 2.1.276 / codex-cli 0.155.0
- リポジトリ: main = `426b08e9`（mulmoterminal@5.2.0）、clean

isamu:
## Summary

Closes #2136.

An open Markdown view now follows the file on disk. Three things were missing and all three are here:

1. **The preview's URL never changed**, so a rewritten file left the browser serving the rendering it already had. The src now carries the version the pane knows is on disk.
2. **Every load dropped the Preview toggle** — including the forced re-read the external-change poll uses to adopt an edit. Following the file would have thrown the reader into the editor at the moment the preview finally had something new to show. A load keeps the toggle when it re-reads the file already open. (Not in the issue; found while wiring 1, and without it 1 does not visibly work.)
3. **Nothing in the server watched the filesystem.** The file-change publisher only ever heard this app's own saves, so an agent editing a `.md` in a cell published on no channel at all — the Canvas `presentDocument` card never moved, and the Files pane had only its periodic version check. A View's channel names one file, so a subscription now starts a watch on it and the last unsubscribe stops it.

## Items to Confirm / Review

- **Preview follows disk while the buffer is dirty.** This is the issue's open question, decided that way because the preview is the server's rendering of the file — it already showed something other than the buffer before the first keystroke. The buffer's own handling is unchanged (clean adopts, dirty raises the conflict banner).
- **Deliberately out of scope:** the version poll stops while a conflict banner is up, so a second external write during an open conflict does not move the preview. Changing that means touching the conflict machinery, which is not what this PR is about. Called out in the plan file.
- **Containment for a watched path.** A relative channel path resolves under the workspace and must stay there; an absolute one is taken as given, because `presentDocument` already opens any `.md` on disk and deliberately has no containment root (`server/backends/openPath.ts`). So a watcher can only stat what a tool call could already read. Worth a second opinion, since the channel name comes from a browser subscribe.
- **A new hook on pubsub.** `onSubscriptionChange` reads socket.io's room lifecycle rather than the subscribe/unsubscribe handlers, because a tab that closes leaves its rooms without ever sending `unsubscribe` — a listener built on the handlers would start work nothing ever stops.
- **A log line per external edit of an absolute-path document.** The announcement goes through the shared publisher, which joins its argument onto the workspace and so cannot stat an absolute path; it falls back to `Date.now()` and warns. Pre-existing for absolute saves (`server/backends/markdown.ts` documents it) and cosmetic — the channel is still right and `mtimeMs` only cache-busts. Fixing it properly is a change to `@mulmoclaude/core`, i.e. MulmoClaude's call too.
- **Which documents are watched, and which are not.** The workspace, and the directory of any live session — the same roots the raw file route serves from. A Markdown file outside all of those still opens and still edits, and the pane's periodic check still catches changes to it; what it does not get is the immediate refresh, and a Canvas card on it waits for the next thing that does announce. That is narrower than `presentDocument`, which opens any `.md` with no containment root, and the difference is deliberate: a channel name is a string a **browser** chose, while a tool-call path came from an agent the user launched. Closing the gap means registering the paths those tool calls opened — a trusted-path list with its own persistence and lifecycle, i.e. a separate change. Both reviewers agreed with narrowing the claim rather than widening the gate.
- **`FilesPane.vue` crossed `max-lines`,** so the pane's three ways of hearing about a change moved into `src/composables/externalFileChanges.ts`. Behaviour-preserving, and re-verified by running the app after the extraction (below).

## Approach and key decisions

- **Polling, not `fs.watch`.** An agent's edit lands as a temp file renamed over the target, which a path-level watch does not survive, and on Windows a watch opened on an 8.3 short path aborts the process (`docs/windows-gotchas.md`). `server/session/codex-activity-watch.ts` chose polling for the same reason; the loop follows its shape, with every dependency injected so it runs against fakes.
- **The comparison is mtime AND size.** mtime alone misses a rewrite inside one tick of the filesystem's timestamp resolution, which is exactly the rapid case here. A spurious announcement costs one refetch; a missed one is the feature not working.
- **A channel is announced on its OWN spelling of the path, never a normalised one.** `publishFileChange` builds the channel from the string it is handed, and two Views legitimately name one file differently — the pane holds an absolute path, a card may hold a workspace-relative one. Keying on the channel avoids inventing a normalisation scheme inside a shared package's semantics; each subscriber hears the channel it asked for.
- **The watch is only armed for a scope that would be published to.** A scope nothing forwards to, or a file that scope does not match, gets no watcher — it could only stat a file and announce into the void.
- **The channel format is shared, not restated by accident.** `common/fileChannel.ts` holds it because both sides decide it, and the UI cannot import `@mulmoclaude/core/file-change` (that entry pulls in `node:fs`). A spec asserts our spelling equals core's, so a drift fails a test rather than producing a subscription nobody publishes to.
- **The watcher lives in the host, not in core.** MulmoClaude has no markdown fs watching either; putting it in the shared package would change MulmoClaude's behaviour. Host-side, the channel and payload are unchanged, so nothing diverges.

## Verification

`yarn format` / `lint` / `typecheck` / `build` are clean; `yarn test` is green (two failures during the run were a missing `@vitest-environment node` docblock, now added, and one spec that timed out under heavy load and passes standalone — the machine was at a very high load average throughout).

Run against a real server in an isolated workspace (Puppeteer for the browser, a socket.io client for the wire), with the document rewritten as temp-file + rename, the shape that kills a path-level watch:

- **Files pane preview** followed the rewrite without a reload and stayed in Preview. The control — the same script against `origin/main` — reproduced both failures: content unchanged and the toggle flipped back to Edit.
- **The wire the Canvas card uses**: a subscriber on the workspace-relative channel and one on the absolute channel each received an announcement on their own channel.
- **Speed**: with the watcher on, the preview followed well inside the poll interval; with only the watcher disabled, the same short wait was not enough. So the new path, not the old poll, is what delivers it.
- **Lifecycle**: a subscriber that disconnects without unsubscribing releases the watch, and a fresh subscriber re-arms it and hears the next change.
- Re-ran the browser and wire checks after the composable extraction.

Not exercised locally: Windows, and the Canvas card's own rendering (the wire it listens on was driven directly instead).

## User Prompt

- GitHub issue #2136（開いている Markdown ビューをファイルの変更に追随させたい）を指して「これできる？」
- スコープの確認に対して「段階1 + 段階2」を選択（プレビューのキャッシュバストと、サーバー側の監視の両方）
- 未保存の編集があるときのプレビューの挙動は「ディスクに従う」を選択
- PR は「1本にまとめて」

## Plan

`plans/feat-2136-markdown-live-refresh.md`

work in mulmoterminal6

## Summary by CodeRabbit

- **New Features**
  - Markdown previews automatically refresh when files change externally.
  - Open documents detect external edits without requiring a page reload.
  - Preview updates preserve the current view and use the latest on-disk version.
  - Unsaved local edits trigger a conflict prompt, allowing users to choose which copy to keep.

- **Documentation**
  - Added guidance on live file tracking, automatic updates, and conflict handling.


isamu:
@coderabbitai review

isamu:
@coderabbitai review

isamu:
## Codex cross-review — round 1 (tier C)

**Tier C**, decided before round 1 and stated here for the record: the change adds a server-side watcher armed by a channel name a browser chooses (a path-containment surface), touches `pubsub`, which every subscriber shares, and claims behaviour preservation for a composable extraction. Any one of those forces C.

**Reviewers this round:** Codex (local CLI) and CodeRabbit. The repo's **Codex auto-review workflow is broken and did not review** — it fails with `401 Unauthorized ... invalid subscription key or wrong API endpoint` against its Azure endpoint, and has failed the same way on every run since 2026-09-16, including on `main`. That is not a sign-off and is unrelated to this PR.

### Ledger

| # | iter | reviewer | finding (one line) | disposition | why / where |
|---|------|----------|--------------------|-------------|-------------|
| 0 | 1 | — | Codex sandbox pre-flight | READ-ONLY | it could read the diff but NOT run the suite (`EPERM` on `node_modules/.vite-temp`). Every test-related claim from Codex this round is UNVERIFIED by Codex and was run by me instead. |
| 0b | 1 | — | pre-existing red: `test_windows (1)` on `test/bin/port-owner.spec.ts` | NOT OURS | PowerShell asked the real OS about a port and answered null under a loaded runner. No file in this diff is reachable from `bin/port-owner`. Re-run of the same job: success. |
| 1 | 1 | codex | High: client-chosen channel can watch arbitrary paths; `/etc/passwd` given as the example | PARTLY REJECTED, PARTLY FIXED | the EXAMPLE is refuted — the scope matcher (`/\.md$/i`) refuses `/etc/passwd` before containment, pinned by a test. The substance was real for two other reasons (any absolute `.md`; a relative climb through a symlink) and is fixed below. Codex ACCEPTED this in-round. |
| 2 | 1 | codex | High (same finding): containment was lexical only, so a symlink climb escaped | FIXED | the channel path now goes through the repo's own `resolveContained` (lexical AND symlink) against the workspace + live session dirs — the root set `authorizedServingBase` gives the raw file route. `containForWatching` in `server/files/pathContainment.ts`, tested there incl. the symlink case. Break-verified: reverting to the lexical check turns that test red. |
| 3 | 1 | codex | Medium: unbounded watchers — one client can start arbitrarily many poll loops | FIXED | `MAX_WATCHED_DOCUMENTS` in `server/files/documentWatch.ts`; past it `start` refuses and warns. Reproduced first (500 watchers from one client), then break-verified. |
| 4 | 1 | coderabbit | Major (DoS, CWE-400): bound file subscriptions PER SOCKET before `socket.join` | FIXED DIFFERENTLY | same defect as #3, different remedy: a GLOBAL watcher ceiling, because the resource is the poll loop and a per-socket limit lets N tabs multiply it by N — and pubsub stays a transport that knows nothing about files. Cost accepted: one page can occupy the ceiling. Codex ACCEPTED this reasoning in-round. |
| 5 | 1 | codex | Medium: a watcher restarted after the last subscriber left re-baselines and never announces the change | FIXED | the last stamp is kept per channel after the watcher stops and passed back as `startFrom`; a restart that finds something different announces at once. Reproduced, then break-verified. |
| 6 | 1 | coderabbit | Minor: mtime+size can miss a same-size replacement inside one timestamp tick; suggests BigInt `ino` | FIXED | accepted as suggested. The agent edit this feature exists for is a RENAME, which always changes `ino`. `stat(path, {bigint: true})`, `mtimeMs:size:ino`. Test creates the exact collision (same size, forced identical mtime, renamed over) — red without `ino`. |


Re-flagging a `FIXED` or `REJECTED` row is welcome **if the resolution is wrong** — say which row and what is wrong with it.


### Every finding was reproduced before it was accepted

A throwaway spec ran each claim against the real modules first. It confirmed three of them, and
refuted the headline example of the first:

- `plugin:markdown:file:/etc/passwd` → **refused** (the scope matcher runs first). The example was wrong.
- `plugin:markdown:file:/anywhere/x.md` → accepted. Real.
- a relative path through a symlink inside the root → **escaped**. Real, and the one a lexical check cannot catch.
- 500 watchers from one client → all started. Real.
- stop, change the file, start again → **no announcement**. Real.
- two files, same size, forced identical mtime → **identical stamp**; their `ino` differed. Real.

Then each fix was **break-verified**: the fix was reverted, the guard went red, the file was
restored and diffed against a backup. All four.

### Verified against a running server, not just the suite

An isolated workspace, a real server, the document rewritten as temp-file + rename (the shape that
kills a path-level `fs.watch`):

- a document **outside** every served root is refused a watcher — rewriting it announces nothing — while a document inside still live-refreshes. That is the security fix's positive case on the wire.
- both a workspace-relative subscriber and an absolute-path subscriber still receive their own channel.
- the Files pane preview still follows the file and stays in Preview, inside the watcher's window rather than the poll's.

isamu:
## Codex exchange — round 1

Two invocations this round: the review, and the in-round rebuttal call (which reviews no head and
produces no verdict, so it is not a round of its own).


<details><summary>Prompt sent (review)</summary>

```text
Review the diff of the current branch against origin/main (git diff origin/main...HEAD) in this repository.

Context: this change makes an open Markdown view in the Files pane, and the Canvas presentDocument
card, follow the file on disk when something OUTSIDE the app rewrites it (an agent in another
terminal cell, an editor, a git checkout). Three parts: a cache-busting version on the preview
iframe's URL, keeping the Preview/Edit toggle across a forced re-read of the open file, and a
server-side mtime poller armed by pubsub room subscriptions that calls the existing
publishFileChange.

Focus on correctness, edge cases, security (path traversal in particular — the watched path comes
from a channel name a browser client chose), resource leaks in the watcher lifecycle, race
conditions, Windows path handling, and test coverage of boundary cases. Read the repo's CLAUDE.md
and honour its conventions. Skip stylistic nits unless they hide a real bug.

Do not apply any fixes. Report findings only, most severe first, each with file:line and a concrete
failure scenario. End with a line that is exactly 'CODEX VERDICT: LGTM' or
'CODEX VERDICT: CHANGES REQUESTED'.

```

</details>


**Codex reply (review).** The run produced a 311 KB transcript, almost all of it the tool calls it made while reading. Posting it in parts is what trips GitHub's comment rate limit, so the findings are given verbatim below and the remainder was deliberately not posted; it is in `/tmp/codex-cross-review-2147/` on the author's machine. Note the run could NOT execute the suite — `EPERM` on `node_modules/.vite-temp` — so every test-related claim in it is unverified by Codex.


<details><summary>Codex findings (verbatim)</summary>

```text
- Medium — [server/files/documentWatch.ts:111-117](/Users/isamu/ss/llm/mulmoterminal6/server/files/documentWatch.ts:111): the unsubscribe race is not fully closed. After `keepGoing()` passes, `announce()` starts an asynchronous `publishFileChange`; the final subscriber can leave before that publish completes. A reconnect then establishes a new baseline after the change and receives no event, leaving a Canvas card stale until another change or refresh.

tokens used
97,016
- High — [server/files/documentWatch.ts:69-80](/Users/isamu/ss/llm/mulmoterminal6/server/files/documentWatch.ts:69): client-controlled channels can watch arbitrary filesystem paths. A browser can subscribe to `plugin:markdown:file:/etc/passwd`, which is accepted and polled. Relative paths can also escape through workspace symlinks because containment is lexical only. This exposes filesystem-change metadata and allows monitoring outside the intended workspace.

- Medium — [server/files/documentWatch.ts:97-117](/Users/isamu/ss/llm/mulmoterminal6/server/files/documentWatch.ts:97): each unique attacker-chosen valid channel creates an independent polling loop. A client can subscribe to thousands of distinct paths, causing unbounded timers and filesystem `stat` calls until disconnect; no limit or authorization bounds this resource usage.

- Medium — [server/files/documentWatch.ts:111-117](/Users/isamu/ss/llm/mulmoterminal6/server/files/documentWatch.ts:111): the unsubscribe race is not fully closed. After `keepGoing()` passes, `announce()` starts an asynchronous `publishFileChange`; the final subscriber can leave before that publish completes. A reconnect then establishes a new baseline after the change and receives no event, leaving a Canvas card stale until another change or refresh.

```

</details>


<details><summary>Prompt sent (rebuttals)</summary>

```text
Follow-up on PR #2147, round 1. Do NOT re-review the PR and do NOT post anything.

You reviewed this branch and raised three findings. I accepted all three in substance and pushed
fixes (not yet on the remote at the time of writing). Two points I did NOT do the way they were
asked. For each, say ACCEPTED (my reasoning holds) or DISPUTED (with specific evidence that
refutes it). Answer only these two.

(1) Your High finding's stated example was `plugin:markdown:file:/etc/passwd` being "accepted and
polled". I believe that specific example is wrong: resolveWatchableDocument applies the SCOPE
matcher before anything else, and the markdown scope is /\.md$/i, so `/etc/passwd` is refused at
that gate. I verified this with a test that asserts it returns null. The rest of the finding I
accepted: an absolute `/anywhere/x.md` WAS accepted, and a relative path COULD climb out through a
symlink because my containment was lexical. Both are now fixed by routing the channel path through
the repo's existing `resolveContained` (lexical AND symlink containment) against the workspace plus
the live sessions' directories — the same root set `authorizedServingBase` gives the raw file route.
A document outside every root now gets no watcher at all.
My reasoning: the example was a false positive; the finding was right for two other reasons.

(2) CodeRabbit asked for a PER-SOCKET limit on file-channel subscriptions, enforced before
`socket.join`. I implemented a GLOBAL ceiling on concurrently watched documents instead
(MAX_WATCHED_DOCUMENTS in server/files/documentWatch.ts), refusing past it with a warn.
My reasoning: the resource being consumed is the poll loop, not the room — a room costs nothing.
A global ceiling bounds the machine's stat rate however many pages are open, whereas a per-socket
limit lets N tabs multiply the cost by N. It also keeps server/infra/pubsub.ts a channel transport
that knows nothing about file semantics; a per-socket file-channel limit would have to parse
channel names inside the subscribe handler.
The cost I accept: one page can occupy the whole ceiling and another page's document then does not
live-refresh (it still works, it just polls at the old interval).

Reply as plain text. If you DISPUTE either, say which single change would settle it.

```

</details>


<details><summary>Codex reply (rebuttals, verbatim)</summary>

```text
(2) ACCEPTED — the consumed resource is the polling loop, and the global ceiling bounds total polling regardless of socket count. The watcher limit is checked before starting a poller; `socket.join` itself does not create polling work.
tokens used
26,230
(1) ACCEPTED — the SCOPE matcher rejects `/etc/passwd` before containment; the remaining absolute-path and symlink concerns were valid and are now covered by `resolveContained`.

(2) ACCEPTED — the consumed resource is the polling loop, and the global ceiling bounds total polling regardless of socket count. The watcher limit is checked before starting a poller; `socket.join` itself does not create polling work.

```

</details>

isamu:
@coderabbitai review

isamu:
## Codex cross-review — round 2 (tier C)

Codex could RUN things this round (round 1 could not), so its claims here are executed rather than read.

**Verdict: CHANGES REQUESTED** — one P2 reopening row #5, one P3 about missing tests. Both fixed and pushed.

### What it found, and what it means

`pollDocument` asked whether anyone was still watching only after the **sleep**. Reading the file
is an await too, and that window is the one that matters: the last subscriber can leave inside it,
and the announcement then goes to a room nobody is in — or, if the room has been re-created
meanwhile, arrives a **second time** beside the new watcher's own. The carried-over announcement
from a previous watcher had the same hole, since it fires before the loop is entered.

`keepGoing()` is now asked after **every** await. `alive` per `start` was already a generation
token; asking it after each await is what makes it behave as one, so no separate token is needed.

Two tests that were missing: a subscriber leaving while the stat is in flight, and the socket.io
room wiring itself — which is read off the **adapter** precisely because a closing tab never sends
`unsubscribe`, so only a real server and a real client can show it works.

**One of those tests was wrong first.** The in-flight version I wrote passed *without* the fix: it
tripped the guard before the loop rather than the one inside it. Break-verifying is what caught it
— the version that shipped goes red when the in-loop guard is removed, and only that one goes red.

### Axis table (Codex, verbatim in the transcript below)

| Axis | Result |
|---|---|
| security: path containment | none — `containForWatching` correctly uses `resolveContained` for workspace and live-session roots |
| resource exhaustion | none — watchers and remembered stamps are globally bounded |
| concurrency and async ordering | the P2 above |
| reconnect / replay idempotence | the P2 above |
| wire compatibility | none — the spec pins the local channel formatter to core's |
| behaviour preservation (`externalFileChanges.ts`) | none found |
| merge with #2137 | none found |
| tests | the P3 above |
| consistency with CLAUDE.md | none found |
| locale lockstep | none — no locale files in this repo |

It ran 132 targeted tests green; its typecheck hung in the sandbox and was interrupted, so that
one is UNVERIFIED by Codex and was run here instead.

### CI

`test_windows (1)` has gone red twice, both times on `test/bin/port-owner.spec.ts`, where a
PowerShell lookup hits its own `LOOKUP_TIMEOUT_MS = 10_000` (measured at 10050ms and 10322ms on
the runner). Nothing in this diff is reachable from `bin/port-owner`, and a re-run of the first
occurrence passed. Recorded as not-ours rather than chased.

### Also merged in during the round

`origin/main` moved twice: #2137 (`filesPreviewMode.ts`) and #2148 (files tree loading state), both
in `FilesPane.vue`. Both merged in, gates re-run, and the pane re-verified against a running server
after each.


<details><summary>Prompt sent (round 2)</summary>

```text
Review PR #2147 at https://github.com/receptron/mulmoterminal/pull/2147, ROUND 2. This is tier C.

Every finding you hold back costs a full extra round: a fix is reviewed for the first time in the
round AFTER it is pushed, so a finding raised late is not a later finding, it is a repeat of this
entire exchange. Read the WHOLE diff (git diff origin/main...HEAD) before you say anything. Do not
stop at the first problem. Do not save anything for a follow-up review.

Do NOT post anything to GitHub and do NOT modify any files. Reply as plain text only.

  correctness and edge cases
  security: path containment for a channel name a BROWSER chose (server/files/pathContainment.ts
    containForWatching, server/files/documentWatch.ts resolveWatchableDocument)
  resource exhaustion: the watcher ceiling and the remembered-stamp map
  concurrency and async ordering: watcher start/stop against socket.io room create/delete,
    and the poll loop's own lifecycle
  reconnect / replay idempotence: the `startFrom` memory across a watcher restart
  wire compatibility: common/fileChannel.ts restates @mulmoclaude/core's channel format for a
    client that cannot import it; a spec pins them equal
  behaviour preservation: src/composables/externalFileChanges.ts was extracted out of
    FilesPane.vue and is claimed to behave identically
  the merge with origin/main, which landed #2137 (filesPreviewMode.ts) in the SAME file while
    this branch was in review: my own duplicate of `keepsPreview` was deleted in favour of it,
    my preview module was renamed filesPanePreview.ts -> filesPreviewSrc.ts to stop colliding
    with #2137's spec of that name, and FilesPane.vue was taken from main and my two changes
    re-applied on top. Check that nothing of #2137's was dropped in that resolution.
  tests: does a test go red if each new behaviour breaks
  consistency with the rest of the codebase (read CLAUDE.md)
  this repo has NO locale files — user-facing strings are inline, so there is no i18n lockstep

(rows compacted; ask for any row in full and I will paste it)

| # | iter | reviewer | finding | disposition | why |
|---|------|----------|---------|-------------|-----|
| 0 | 1 | — | sandbox pre-flight | READ-ONLY | Codex could not run the suite (EPERM on node_modules/.vite-temp); test claims were run by me |
| 0b | 1 | — | test_windows (1) red on test/bin/port-owner.spec.ts | NOT OURS | PowerShell/real-OS port query under load; nothing in this diff reaches bin/port-owner; re-run of the same job: success |
| 1 | 1 | codex | High: `/etc/passwd` example for arbitrary-path watching | REJECTED | the scope matcher (/\.md$/i) refuses it before containment — pinned by a test. Codex ACCEPTED this rebuttal in round 1. The finding's OTHER two reasons were real; see #2 |
| 2 | 1 | codex | containment was lexical, so a relative path climbed out through a symlink | FIXED | now routed through the repo's `resolveContained` against workspace + live session dirs (`containForWatching`); break-verified |
| 3 | 1 | codex | Medium: unbounded watchers | FIXED | MAX_WATCHED_DOCUMENTS ceiling; break-verified |
| 4 | 1 | coderabbit | Major DoS: bound subscriptions PER SOCKET | FIXED DIFFERENTLY | global watcher ceiling instead — the resource is the poll loop, and per-socket lets N tabs multiply it. Codex ACCEPTED this reasoning in round 1 |
| 5 | 1 | codex | Medium: a restarted watcher re-baselines and loses the change | FIXED | last stamp kept per channel and passed back as `startFrom`; break-verified |
| 6 | 1 | coderabbit | Minor: mtime+size misses a same-size replacement; use BigInt ino | FIXED | stat({bigint:true}), `mtimeMs:size:ino`; break-verified with a forced same-size same-mtime rename |


That table is context so you do not re-derive what round 1 decided. It is NOT a list of closed
topics: if a resolution is wrong, or a FIXED row's fix does not do what it claims, say so and mark
it 'REOPENING #<row>' with specific evidence.

Note: you may not be able to run the test suite in this sandbox (round 1 hit EPERM on
node_modules/.vite-temp). If so, say which claims you could not execute rather than reporting the
tooling as a finding, and I will run anything you name.

── END WITH THIS, IN THIS ORDER ─────────────────────────────────────────────
  Line 1, on its own line: 'CODEX VERDICT: LGTM' or 'CODEX VERDICT: CHANGES REQUESTED' followed by
  a bulleted summary.
  Then, on its own line:

```

</details>


<details><summary>Codex reply (round 2, verbatim)</summary>

```text

- P2 REOPENING #5: stale poll iterations can announce after unsubscribe and duplicate announcements after restart.
- P3: lifecycle integration lacks a regression test for that async race.

| Axis | Result |
|---|---|
| correctness and edge cases | P2 above. |
| security: path containment | none. `containForWatching` correctly uses `resolveContained` for workspace and live-session roots. |
| resource exhaustion | none. Watchers and remembered stamps are globally bounded. |
| concurrency and async ordering | P2: `pollDocument` checks `keepGoing()` after `sleep()` but not after awaited `stamp()`, and the `startFrom` announcement is unconditional. Site: `server/files/documentWatch.ts`, affecting its caller in `server/backends/fileChange.ts`. Smallest fix: check `keepGoing()` immediately after every awaited stamp and before the `startFrom` announcement; ideally use a generation token so an old poll cannot announce after restart. This is non-issue only if room recreation is guaranteed to wait for the prior poll’s async work, which it is not. |
| reconnect / replay idempotence | P2 REOPENING #5: normal restart detection works, but an old in-flight watcher can emit alongside the restarted watcher. |
| wire compatibility | none. The spec pins the local channel formatter to core’s implementation. |
| behaviour preservation: `externalFileChanges.ts` | none found. The timer, write-channel filter, cleanup, and recheck behavior are preserved. |
| merge with origin/main / #2137 | none found. `filesPreviewMode.ts`, preview-state restoration, and the FilesPane reapplication are present. |
| tests | P3: no test drives unsubscribe/restart while `stamp()` is pending, and no test covers the real Socket.IO room-event wiring. Sites: `test/server/files/documentWatch.spec.ts` and the untested adapter path in `server/infra/pubsub.ts`. Add one pending-stamp race test. |
| consistency with CLAUDE.md | none found. |
| locale lockstep | none. User-facing strings remain inline; this repo has no locale files. |

Targeted tests passed: 132 tests across the changed areas. Full typecheck was not completed because it hung in the sandbox and was interrupted.

```

</details>

isamu:
@coderabbitai review

isamu:
@coderabbitai review

isamu:
## Codex cross-review — round 3 (tier C)

**Verdict: CHANGES REQUESTED** — one P2, reopening row #7. It is a defect the loop itself
introduced in round 2, which is the honest version of what these rounds are for.

### The finding

Guarding the announcement (round 2) opened a worse hole than it closed. The stamp was recorded as
each read happened, so a change **read but never announced** — the last subscriber left between
the two — was remembered as *delivered*. The next watcher then started from it, found nothing to
say, and the change was lost **for good** rather than merely late. Exactly the failure row #5 was
supposed to close, re-created by the fix for row #7.

What is remembered is now what subscribers have been brought **up to date with**: the baseline they
opened on, and every stamp that was actually announced. It goes through its own `onObserved`
callback rather than riding along inside the read, because the difference between "read" and
"delivered" *is* the defect.

### Two guards, and why there are two

Reverting the record-after-announce line left **every test green**, so it had none. That half is a
different rule — a change already delivered must not be delivered again when the document is
re-subscribed — and it now has its own test. Both were break-verified:

- remembering on every read again → "does not forget a change it read but never announced" goes red.
- dropping the record after an announcement → "does not re-announce a change the subscribers already had" goes red.

The first version of the reproduction test also had to be fixed: its fake stopped the watcher on
*every* read of the changed stamp, including the reconnecting one, so the test was sabotaging the
restart it was checking.

### Axis table (Codex)

Security, resource exhaustion, wire compatibility, behaviour preservation, the #2137 and #2148
merges, CLAUDE.md consistency and locale lockstep: **none**. Correctness, concurrency and
reconnect idempotence: the P2 above. It ran 138 targeted tests green and typecheck green this
round; it stopped the full suite on unrelated pre-existing failures.

### CI

`test_windows (1)` has now gone red three times, always on tests that shell out to PowerShell on
the runner — the two in `test/bin/port-owner.spec.ts` (which hit their own `LOOKUP_TIMEOUT_MS =
10_000`, measured at 10050ms and 10322ms) and once `test/server/files/win-picker-encoding.spec.ts`.
This diff touches none of those files, and a re-run of the first occurrence passed. Recorded as
not-ours rather than chased, and called out here so it is not mistaken for a sign-off.

`origin/main` moved again during the round (#2148 tree loading state, and a files-row-menu
composable refactor). Merged in, gates re-run, pane re-verified against a running server.

isamu:
## Checkpoint — round 5, my half

Re-read the whole diff against the PR body, as if it had arrived today with none of the thread attached.

**1. Does it still do what the PR says?** Yes, and that has been checked against a running server
after every push, not inferred: the pane's preview follows a temp-file-plus-rename edit without a
reload and stays in Preview; a workspace-relative subscriber and an absolute one each hear their
own channel; a document outside every served root gets no watcher; a subscriber that disconnects
and comes back is told what it missed.

**2. Has the conversation drifted?** It has narrowed. Four of the five rounds are about ONE
mechanism — the watcher's stamp memory and the async windows around it — and each finding was a
real way to lose a change. That is the feature's own promise rather than a tangent, but it is
worth naming: the reviewers have spent most of their attention on ~15 lines.

**3. Is anything here only because a reviewer asked?** One thing, and it is the thing those four
rounds were about. The cross-restart memory exists entirely because of round 1's row #5. Without
it, a change that lands while the socket is down is adopted silently by the next watcher.

The honest case against it: the Files pane has a periodic version check that would catch that
case anyway, a few tens of seconds later. The case for it: the **Canvas card has no such
fallback** — for the card, a change missed during a reconnect is missed until the next change,
which is exactly the "it sometimes doesn't update" this PR exists to remove. And a dropped socket
is not exotic; a server restart during development produces one every time.

I keep it, and I have put the opposite case to Codex in this round's prompt rather than deciding
it alone.

**4. Split, redirect or close?** No. The one structural regret is that the preview cache-bust and
the server-side watcher could have been two PRs — they are independently revertible — but that was
decided with the author before round 1 and re-cutting it now would cost more than it buys.

**5. Largest remaining risk.** The watcher's behaviour under rapid, repeated reconnects against
the real socket.io room lifecycle. The unit tests drive that sequence with fakes and one live
check drives it once for real; what neither covers is many documents reconnecting at once while
the ceiling is full.

**Shape of the diff.** Of ~1,480 added lines, roughly 900 are tests and about 160 are the plan
file. The production code is the smaller half, and most of what the loop added is coverage rather
than machinery.

isamu:
## Codex cross-review — rounds 4 and 5 (tier C)

### Round 4 — CHANGES REQUESTED, two P2s, both the same mechanism one layer further out

**Delivery is itself async.** The shared publisher stats the file before it emits, and the poll was
recording the stamp without waiting for that. A subscriber who left inside the await got nothing
while the stamp said it had been told — so its reconnect stayed quiet. The poll now waits for the
announcement and asks again afterwards whether anyone is still there.

**Remembered stamps were evicted by age alone**, so a long-lived watcher's baseline could be spent
on documents churning through the ceiling beside it — losing exactly the comparison that tells its
next watcher whether the file moved while the socket was down. Only a channel nobody is watching
may be evicted; every watched channel records its baseline, so the watched ones never outnumber the
ceiling and there is always room for them.

**Both of my first tests passed without their fix.** The delivery one and the eviction one, for
different reasons — one tripped an earlier guard, the other never changed the stamp it was
supposedly detecting a change in. Break-verifying is the only reason either was caught. The
versions that shipped go red.

### Round 5 — CHANGES REQUESTED, one P2, and it was a scope claim rather than a defect

Absolute Markdown documents that `presentDocument` can open, but which lie outside the workspace
and outside every live session directory, are never watched — so the README's "the open view
follows the file on disk" claimed more than the code delivers.

That exclusion is the round-1 containment fix working as intended: a channel name is a string a
**browser** chose, while a tool-call path came from an agent the user launched. Widening the gate
is the direction round 1's High finding warned against. Closing the gap properly means registering
the paths those tool calls opened — a trusted-path list with its own persistence, lifecycle and
restart semantics.

So the **claim** was narrowed rather than the gate widened, in all four places a claim lives: the
README, the containment gate's own comment, the PR body and the plan file. Put to Codex in-round:
**ACCEPTED** — *"your narrowing is the correct resolution … trusted-path registration would be a
separate feature … not required to merge this fix."*

### Checkpoint (round 5)

Both halves are posted above. Codex: no material drift; the PR should not be split or closed; the
stamp memory is justified only if missed changes across watcher gaps are a product requirement —
which for the Canvas card, with no periodic fallback of its own, they are. Largest remaining risk
by its reckoning was the absolute-document gap, now stated rather than implied.

### CI — the Windows red is not this PR

`test_windows (1)` has failed on every run of this branch, and the evidence says it is repo-wide:
`main`'s own daily Windows run failed in the same window, branch `feat/2140` failed too, and
**issue #2154 / PR #2155 is already fixing exactly this** — PowerShell exceeding the port lookup's
ten-second budget on a CI runner. Every failing spec across those runs shells out to a subprocess
(`port-owner`, `win-picker-encoding`, `headlessPreview`, `worktree-diff`); this diff touches none of
them. Not chased here, and not counted as a sign-off either.

isamu:
@coderabbitai review

isamu:
## Codex cross-review — round 6, and CodeRabbit's second pass

**Codex round 6: `CODEX VERDICT: LGTM`** — no findings on any axis, `FINDINGS COMPLETE`, 156
targeted tests and typecheck green in its sandbox. CI was fully green on that head, Windows
included.

That would have been the exit, except CodeRabbit had reviewed an earlier head in the meantime and
raised two things:

**1. Record the stamp only after the publish has landed.** This is the same defect Codex raised as
its round-4 P2, and it was already fixed. Two reviewers reaching the same finding independently is
worth more than either of them raising it alone, and it is recorded that way rather than waved off
as a duplicate.

**2. The pubsub spec's two negative assertions were not denying anything.** They slept for a fixed
interval and then checked that nothing had been reported — which a join Socket.IO had not applied
yet satisfies just as well as one it deliberately refused to report. A test that goes green either
way.

Each now waits for `subscriberCount` to show the very subscription it is judging — the same room
the events come from — and only then asserts. Break-verified: making the listener report *every*
join rather than only the first turns the second-subscriber test red, which it did not do before.

That fix is a push, so round 6's LGTM no longer covers the head that would merge, and another round
is owed. That is the rule working rather than an accident: an unreviewed fix is unreviewed, however
small.

**CodeRabbit has not seen the final head** — it is rate-limited on this repository's free tier and
its last two triggers were refused. Its findings above are from the round-3 head. Said plainly
rather than counted as a sign-off.

isamu:
## Codex cross-review — round 7: clean, on the head that will merge

**`CODEX VERDICT: LGTM`**, no findings on any of the twelve axes, `FINDINGS COMPLETE`, 110 targeted
tests and typecheck green in its sandbox. **Nothing was pushed during this round**, so the verdict
covers the commit that would merge — which is the whole of the exit bar, and the reason round 6's
LGTM did not count.

### What the seven rounds actually produced

Every round but the last found something real, and four of the six findings were in code the loop
itself had added:

| round | finding | whose |
|---|---|---|
| 1 | containment was lexical, so a relative path climbed out through a symlink | the original diff |
| 1 | one client could start unbounded watchers (both reviewers, independently) | the original diff |
| 1 | mtime+size misses a same-size rename | the original diff |
| 1 | a restarted watcher re-baselined and lost the change | the original diff |
| 2 | the liveness check covered the sleep but not the file read | round 1's fix |
| 3 | a change read but never announced was remembered as delivered | round 2's fix |
| 4 | delivery is async, so the stamp was recorded before it landed (both reviewers, independently) | round 3's fix |
| 4 | remembered stamps were evicted by age, spending a watched channel's baseline | round 1's fix |
| 5 | the README claimed more than the watcher covers | the prose |
| 6 | two negative assertions in a test denied nothing | round 2's tests |

### What made it seven rounds rather than three

One mechanism — the watcher's stamp memory — has an await at every step, and each round closed one
window and revealed the next one outward: sleep, then read, then delivery. That is a fix enumerating
cases, and the skill's own advice is to invert such a rule rather than keep patching it. The
inversion here was to stop asking *"has the subscriber left?"* at particular points and instead
record only what was **demonstrably delivered** — which is what rounds 3 and 4 arrived at, and
nothing has been found in it since.

**Three of my tests passed without their fix**, caught only by break-verifying: one tripped an
earlier guard, one never changed the stamp it claimed to detect a change in, and two slept instead
of waiting for the thing they denied. Every fix in this PR has now been reverted and watched to go
red.

### Reviewer coverage, stated plainly

- **Codex (local CLI)** reviewed every round including this one, and could execute tests from round 2 on.
- **CodeRabbit** reviewed rounds 1 and 3 and found two real things; it is rate-limited on this repository's free tier and **has not seen the final head**. That is not a sign-off.
- **The repo's Codex auto-review workflow reviewed nothing at all** — it fails with a 401 against its Azure endpoint and has done so on every run since 2026-09-16, `main` included. Unrelated to this PR, and worth fixing separately.

### CI

Green except `test_windows (1)`, which is the repo-wide flake already tracked as **#2154** and being
fixed in **#2155**: PowerShell exceeding the port lookup's ten-second budget on a CI runner. `main`'s
own daily Windows run failed in the same window, as did another branch. Re-run requested.

isamu:
## Taking in #2155, ahead of merging

`test_windows (1)` was red on every run of this branch, on the two `test/bin/port-owner.spec.ts`
cases where PowerShell exceeded the lookup's ten-second budget on a CI runner. That is **#2154**,
and **#2155 fixed it and has merged**, so `origin/main` is merged in here to pick it up.

The merge brings in **exactly one file** — `test/bin/port-owner.spec.ts`, someone else's test fix.
No file of this PR's changed, which `git diff --name-only HEAD~1 HEAD` shows.

Stated plainly, because it matters for what the sign-off covers: **Codex's round-7 LGTM was on
`31f8ab7d`, and the head that will merge is now `396e9d18`.** The difference between them is that
one test file and nothing else. Gates re-run locally and green; the affected specs plus the
port-owner spec pass together (123 tests).

PR #2144, opened 2026-09-18
feat: the Files pane comes back in the view mode it was left in (#2137)

isamu:
## Summary

The Files pane restored the open file and the expanded tree, but not **which of its two views was
up** — so reading a `.md` in Preview ended at every reload and every walk of the zoom to another
cell (#2137). It now remembers the mode along the two routes that already carry the file: the
uid-keyed map in `TerminalGrid.vue` during a session, and `files_pane_state` in localStorage across
a reload.

Two things held it, and either alone changes nothing:

- `FilesPaneState` carried `openPath` and `expanded` only, so neither route had anywhere to put it.
- `loadFile` reset `showPreview` **unconditionally**, and restoring goes through `loadFile`.

**That reset was doing real work and could not simply go.** The preview iframe is shown on
`openPath && !unpreviewable && showPreview` — it does not ask whether the file is Markdown, while
the *toggle back to the editor* does (`v-if="openPath && isMarkdown"`). A preview left up while a
non-Markdown file is opened is therefore an iframe over the new file with no way back. The rule it
was a blunt version of: **the view mode belongs to the file it was turned on for.**

That rule is now two pure predicates in `src/components/filesPreviewMode.ts`:

- `keepsPreview(openPath, nextPath)` — opening ANOTHER file drops the mode; re-reading the SAME one
  keeps it (the conflict banner's Reload, and the external-change refresh).
- `restoresPreview(remembered, reopened)` — a remembered Preview comes back only over the very path
  it was remembered for, and only while that path still holds Markdown the server served as text.
  A `.md` since replaced by a binary opens in the editor rather than as a blank iframe.

The second is consulted **inside `loadFile`**, beside the adopt and under the same
`id === fileReqId` guard, rather than by `restore()` after its await: the pane re-roots by
`reload()` as the zoom walks between cells, so a slow read from the cell being left can outlive the
new cell's whole restore, and a caller applying the mode afterwards hands the abandoned pane's mode
to the one that replaced it. Codex found that in review; a spec pins it.

`FilesPaneState.showPreview` is optional, and the storage layer treats a value of any other shape as
absent rather than dropping the entry: an entry already in someone's browser must cost them the mode
at most, never the open file they came back for.

## Items to Confirm / Review

- **An external change to the file you are reading now KEEPS Preview** instead of dropping you into
  the editor. That follows from `keepsPreview` (the same file re-read is not the pane leaving it),
  and it is the behaviour I believe a reader wants. The catch: the preview iframe's URL carries no
[...]

PR #2150, opened 2026-09-18
fix: the Files pane no longer calls a directory empty before reading it (#2148)

isamu:
## Summary

The Files pane said **"Empty directory."** while it was still reading the directory (#2148). `roots`
carried two different facts in one value:

```ts
const roots = ref<Node[]>([]);   // mount: nothing read yet
```
```html
<p v-else-if="roots.length === 0">Empty directory.</p>
```

An empty array meant both *"no listing has come back"* and *"the listing came back with nothing in
it"*, and the template rendered the second over the first — so for the whole root `/list` round trip
the pane asserted a fact it had not learned. `teardown()` set `roots` back to `[]`, which reopened
the same window on **every re-root**, which is why it shows up while walking the zoom between cells
and not only at mount.

`roots` is now `Node[] | null`, and `null` is the missing state. The empty array then means one
thing only.

- `loadRoot()` needs no new state: assigning the listing **is** the transition out of "not read".
- `teardown()` sets `null` rather than `[]`.
- The header's **Reload tree** button does not go through `teardown()`, so a manual refresh keeps
  the tree on screen and swaps the result in — it does not flash "Loading…" over a correct tree.
- The four readers take `roots.value ?? []`. That is not a null-check apology: an unread tree really
  does hold no rows, no node at a path and no expanded paths.

`Loading…` is the string `PromptsPane`, `TranscriptPane`, `WikiBrowseOverlay` and `GithubPane`
already use, in the same muted style.

This is stage 1 of #2148. Stage 2 — showing the previous tree from a cache while the read is in
flight — is a separate PR, and it needs this one first: a directory with no cache still has to say
something honest.

## Items to Confirm / Review

- **`FilesPane.vue` is at the repo's 600-line `max-lines` cap.** A first attempt used a
  `treeLoaded` boolean beside `roots` and pushed the file over it; this version is smaller *and* the
  better model, but the file is full either way and the next change to it has to take something out.
  The candidate is the row menu (`openRowMenu` / `runRowAction` / `onMenuNav`, the click-away and
  focus handling, and its `<Teleport>` markup) — about 125 lines, a boundary that already half
  exists in `filesRowActions.ts`, and `filesRowMenu.spec.ts` drives it through the pane, so the lift
  would be provable rather than argued. Deliberately not done here: it is a refactor and this is a
  bug fix.
- **The Reload button's behaviour is a choice**, not a side effect: it keeps the current tree
  visible while refetching. Say so if you would rather it showed "Loading…" too.
- The two new `data-testid`s (`files-tree-loading`, `files-tree-empty`) exist so a test can tell the
  two states apart; the empty one had none before.

## How it was verified

- `test/src/components/filesPaneTreeLoading.spec.ts` — its own file rather than growing
  `FilesPane.spec.ts`, which is near the same cap. Five cases: in flight says Loading and **not**
  Empty; an empty listing says Empty and **not** Loading; a listing with entries says neither; a
  failed listing reports the error and neither; and `reload()` goes back to Loading rather than to
  Empty.
- Break-verified in both directions. Removing the loading branch turns the file red; changing
  `teardown()` back to `roots.value = []` turns exactly the re-root case red and nothing else, which
  is the half of the bug that shows while walking the zoom.
- `yarn format`, `yarn lint`, `yarn typecheck` clean. Every Files-pane spec run together
  (`FilesPane`, `filesPanePreview`, `FilesOverlay`, `gridOpenFiles`, `filesFinderInPane`,
  `filesRowMenu`, `filesTreeState`, `filesPaneStore`): green.
- Not exercised in a browser: the change is one render branch with unit cover on all three states.

## User Prompt

- 「ファイルペイン、load時に、Empty directory.ってなる。時間かかる場合あるから、前開いたときのををデータでキャッシュしておいて表示して、読み終わったら更新するのってダメかな？」
- 対応範囲として「①ローディング状態 ＋ ②ツリーのキャッシュ表示」を選択。これは①。

## Related

- #2148 — this, stage 1. Stage 2 (the cache) follows.
- #2149 — the other half of "coming back looks the way you left it": the scroll and cursor
  positions.
- #2137 — the view mode, merged in #2144.

work in mulmoterminal5

## Summary by CodeRabbit

- **Bug Fixes**
  - The Files pane now displays a loading indicator while directory contents are being retrieved.
  - Empty-directory messaging appears only after an empty result is confirmed.
  - Re-rooting the file tree correctly returns it to a loading state during the next listing.
  - Loading and empty states no longer appear incorrectly when requests fail or entries are available.

- **Tests**
  - Added coverage for loading, empty, populated, error, and re-rooting scenarios.


src/components/FilesPane.vue
```diff
@@ -569,7 +573,10 @@ function teardown(): void {
   closeFinder();
   editor?.destroy();
   editor = null;
-  roots.value = [];
+  // Not `[]`: the root is changing and nothing has been read for the new one. The header's Reload
+  // button deliberately does NOT come through here — that tree is still this root's, and swapping
+  // the result in beats replacing a correct tree with "Loading…".
+  roots.value = null;
```
isamu, line 579: P2 — stale error masks the new loading state during re-root.

Every occurrence of this mistake in the PR: src/components/FilesPane.vue:579; teardown() sets roots.value = null but leaves treeError untouched.

If the previous root's listing failed, then reload()/the FilesOverlay root-change path tears down that root and starts the next request. start() awaits nextTick() before loadRoot() clears treeError, while the template checks treeError before roots === null; during that interval the new root is in flight but the pane renders the previous root's error rather than Loading…. This is especially visible on a failed listing followed by a re-root.

Smallest resolution: set treeError.value = null alongside roots.value = null in teardown(), and add a failure-then-re-root assertion.

What would change my mind: a test or timing proof showing that teardown-to-start never produces a render while the old error remains set; the current nextTick() in start() means it does.

test/src/components/filesPaneTreeLoading.spec.ts
```diff
@@ -0,0 +1,111 @@
...
+  const gatedFs = (entries: { name: string; dir: boolean; size: number }[]) => {
+    let open!: () => void;
+    const gate = new Promise<void>((resolve) => (open = resolve));
+    globalThis.fetch = vi.fn(async (input: RequestInfo | URL) => {
+      if (String(input).includes("/list")) {
+        await gate;
+        return { ok: true, json: async () => ({ entries }) };
+      }
+      return { ok: true, json: async () => ({ text: "", version: "v1" }) };
+    }) as unknown as typeof fetch;
```
isamu, line 32: P3 — the added spec introduces forbidden type assertions.

Every occurrence in this PR: test/src/components/filesPaneTreeLoading.spec.ts:32, :80, and :101. These are the new `as unknown as` assertions; the production change adds none.

Smallest resolution: type the fetch mock and the exposed pane helper so the assignments/call use inferred or declared types without assertions.

What would change my mind: confirmation that the repository rule intentionally permits assertions in tests and this spec is exempt; the stated review constraint currently says no `as`.

isamu:
- P2: clear treeError when teardown() changes roots to null, or a failed old root can mask the new root's Loading state during re-root.
- P3: the added spec contains three new as unknown as assertions despite the stated no-as constraint.

| Axis | Result |
|---|---|
| Correctness and edge cases | P2 finding posted inline at src/components/FilesPane.vue:579. Otherwise, loadRoot errors settle in treeError, stale tree generations are guarded, manual Reload preserves an existing tree, and FilesOverlay's initial mount is covered by the shared component behavior. |
| Four ?? [] readers | None. rows, revealPath, restore, and snapshot have the correct empty/unread semantics; revealPath waits for started before searching. |
| Tests | P3 finding posted inline for the three assertions. The five new cases are non-vacuous and pass; the related FilesPane/overlay specs also pass (113 tests total). The missing regression is failed listing followed by re-root, which is the test needed for the P2 transition. |
| Consistency with the codebase and CLAUDE.md | P3 finding for the new as assertions. No emoji or any was added, comments explain the rationale, and Loading… matches the existing panes. |
| Security | None. The new branch renders only static text and test IDs; it exposes no user data. |
| 600-line cap / extraction | No finding. Keeping the row-menu extraction for the next change is reasonable and this focused bug fix need not carry that refactor. |

Validation: the targeted five-case spec passed; the seven related FilesPane/overlay specs passed (108 tests); yarn typecheck passed; yarn lint passed. I did not run a mutation sweep, per the stated rm policy.

isamu:
## Codex cross-review — round 1 (tier S), head `b8373f8c`

Verdict: **CHANGES REQUESTED** — one P2, one P3, axis table answered line by line, `FINDINGS COMPLETE` present. Codex ran the targeted spec, the seven related Files-pane specs, `yarn typecheck` and `yarn lint` in its own sandbox. Its posted verdict and both inline comments are above.

**Sandbox (ledger row 0)**: measured earlier today in this same working copy with the same flags — vitest, typecheck, socket bind, `api.github.com` and `gh` all exit 0; **`rm` blocked by policy**, so no mutation sweep was asked for and none was run.

### P2 — stale `treeError` during a re-root: FIXED, and deliberately without a test

Accepted and fixed in `5233fd90`: `teardown()` cleared `roots`, `openPath`, `dirty`, `baseVersion`, `conflict` and the view mode, and left `treeError` — so re-rooting away from a directory whose listing had failed carried that failure into the new root, where the template shows it ahead of the loading state.

Codex also asked for the regression test ("failed listing followed by re-root"). **I wrote it and then removed it, because it passes with the fix reverted** — verified by deleting the line and running the file: 6 passed, 0 failed. The reason is the ordering: `reload()` runs `teardown()` synchronously, `start()` then suspends on `nextTick`, and `loadRoot()` clears `treeError` in its first statement — so a spec that awaits anything resumes after `start()`'s continuation has already run. The window is one render frame and a spec has no vantage point inside it. Shipping a green test for that would be worse than shipping none.

Put back to Codex in the same round: it **ACCEPTED** the account — "no deterministic test can observe a longer-lived stale error window", and no path survives past `loadRoot` except a startup exception or a host changing `cwd` without calling `reload()`, which this pane's contract forbids.

### P3 — three `as unknown as` in the new spec: DECLINED, and Codex accepted the rebuttal

The constraint in my prompt came from CLAUDE.md, whose scope is production TypeScript. In the test tree this repo has two current patterns: `globalThis.fetch = vi.fn(...) as unknown as typeof fetch` (213 occurrences, including **every** other Files-pane spec this file sits beside) and `vi.stubGlobal("fetch", ...)` (used elsewhere). The third assertion reaches a `defineExpose`d method, which Vue Test Utils does not surface on `vm`'s type. Making one file in this family different reads worse than matching its neighbours. Codex **ACCEPTED**.

### Ledger

| 0 | pre | sandbox pre-flight | — | Measured at 21:35Z today in this same working copy with the same flags (`--sandbox workspace-write -c sandbox_workspace_write.network_access=true`): vitest exit 0, `yarn typecheck` exit 0, socket bind exit 0, api.github.com 200, `gh pr view` exit 0. **`rm` is blocked by policy** — do not run a mutation sweep; name the mutation and I will run it |
| 1 | pre | claude, before round 1: a `treeLoaded` boolean beside `roots` was written first and thrown away | REJECTED (by me) | It pushed `FilesPane.vue` over the repo's 600-line `max-lines` cap, and it is a second value that has to be kept in step with the first — the same shape as the bug. `null` cannot drift from `roots` because it IS `roots` |
| 2 | 1 | P2 — `teardown()` cleared `roots` but left `treeError`, so a failed root's error could show over the next root during a re-root | FIXED | `teardown()` now clears it beside `roots`. **No test**: the window is one render frame between `teardown()` and `loadRoot()`, and a spec resumes after `start()`'s own `nextTick` continuation — the regression test Codex asked for passes with the fix reverted (verified: 6 passed with the line removed), so it was dropped rather than shipped green-for-the-wrong-reason. Codex ACCEPTED this account in round 1's follow-up |
| 3 | 1 | P3 — three `as unknown as` assertions in the new spec | REJECTED | The repo's test tree has 213 occurrences of `... as unknown as typeof fetch`, including every other Files-pane spec this file sits beside; the third assertion reaches a `defineExpose`d method, which VTU does not surface on `vm`'s type. The CLAUDE.md rule is about production TypeScript (`src`/`server`/`common` hold 22 such casts in total). Codex ACCEPTED the rebuttal in round 1's follow-up |

Round 2 is dispatched against `5233fd90`.

<details><summary>Round 1 prompt (verbatim)</summary>

```text
Review PR #2150 in receptron/mulmoterminal, head 5233fd90. Tier S.

Every finding you hold back costs a full extra round: a fix is reviewed for the first time in the
round AFTER it is pushed. Read the WHOLE diff before you post anything. Do not stop at the first
problem. Do not save anything for "a follow-up review".

Inspect with `gh pr diff 2150` or `git diff 8cb709a1...HEAD`, read the surrounding code, and RUN
things — your sandbox can run `npx vitest run <file> --configLoader runner`, `yarn typecheck`,
`yarn lint`, bind sockets and reach the network. **`rm` is blocked by policy, so do not run a
mutation sweep**: name any mutation you want tried and I will run it and report back.

Post inline comments via `gh api repos/receptron/mulmoterminal/pulls/2150/comments` and the final
top-level comment via `gh pr comment 2150`.

Issue #2148. `src/components/FilesPane.vue` held its directory tree in `const roots =
ref<Node[]>([])`, and the template rendered `v-else-if="roots.length === 0"` as "Empty directory."
— so an empty array meant BOTH "no listing has come back yet" AND "the listing came back empty",
and the pane announced the second for the whole of the root `/list` round trip. `teardown()` set
`roots` back to `[]`, reopening that window on every re-root (the host calls `reload()` as the zoom
walks between cells).

`roots` is now `Node[] | null`; `null` is "not read yet". `loadRoot()`'s assignment is the
transition; `teardown()` sets `null`; the four readers use `roots.value ?? []`. A third template
branch shows "Loading…" while `roots === null`.

  correctness and edge cases — especially: is there ANY path that leaves `roots` null forever, or
    that shows "Loading…" over a tree that is actually on screen, or "Empty directory." while a
    read is in flight? Consider loadRoot's own error path, the header's Reload button (which calls
    loadRoot directly, NOT teardown), reload()/teardown(), the treeReqId generation guard, and
    FilesOverlay (which mounts the same component with no initial state)
  the four `?? []` readers: `rows`, `findNode` in `revealPath`, `findNode` in `restore`,
    `expandedPaths` in `snapshot` — is `[]` the right answer for an unread tree at each one, or
    does any of them need to distinguish "unread" from "empty"?
  tests: does a test go red if each half breaks — and is any of the five new cases vacuous (green
    for a reason other than the one it names)? The spec is
    test/src/components/filesPaneTreeLoading.spec.ts
  consistency with the rest of the codebase and with CLAUDE.md (no emojis; no `any`; no `as`;
    comments say WHY not WHAT; the "Loading…" string matches the other panes)
  security — anything the new branch exposes (it renders no user data)

| 0 | pre | sandbox pre-flight | — | Measured at 21:35Z today in this same working copy with the same flags (`--sandbox workspace-write -c sandbox_workspace_write.network_access=true`): vitest exit 0, `yarn typecheck` exit 0, socket bind exit 0, api.github.com 200, `gh pr view` exit 0. **`rm` is blocked by policy** — do not run a mutation sweep; name the mutation and I will run it |
| 1 | pre | claude, before round 1: a `treeLoaded` boolean beside `roots` was written first and thrown away | REJECTED (by me) | It pushed `FilesPane.vue` over the repo's 600-line `max-lines` cap, and it is a second value that has to be kept in step with the first — the same shape as the bug. `null` cannot drift from `roots` because it IS `roots` |

That table is context, not a list of closed topics: if a resolution is wrong, say so and mark it
"REOPENING #<row>" with the specific evidence.

── ONE MORE THING THE PR ADMITS ─────────────────────────────────────────────
`FilesPane.vue` is at the repo's 600-line `max-lines` cap (counted without comments/blanks). The PR
body says the next change to the file has to extract something and names the row menu as the
candidate. If you think that call is wrong — or that this PR should carry the extraction — say so.

    'CODEX VERDICT: LGTM' or 'CODEX VERDICT: CHANGES REQUESTED' + bullets
  Then, on its own line:
  Omit that line if it is not true, and say which part you did not reach.

```

</details>

<details><summary>Round 1 reply, and the in-round follow-up on both findings (verbatim)</summary>

```text
Review complete and posted to PR #2150.

Findings:

- P2: stale `treeError` can mask `Loading…` during re-root.
- P3: three new `as unknown as` assertions violate the stated constraint.

Verdict: **CHANGES REQUESTED**.

--- follow-up (rebuttals put back to Codex before pushing) ---

Follow-up on PR #2150, round 1. Do NOT re-review the PR and do NOT post anything.

Two of your findings. For each, say ACCEPTED (my reasoning holds) or DISPUTED (with the specific
evidence that refutes it). Reply as plain text.

── 1. Your P2 (stale treeError during re-root) — I ACCEPTED the fix, but NOT the test you asked for

The fix is in: `teardown()` now clears `treeError` beside `roots`, because the failure belonged to
the root being left and `teardown` already clears every other piece of that root's state.

You also said: "The missing regression is failed listing followed by re-root, which is the test
needed for the P2 transition." I wrote exactly that test — fail the root listing, assert the error,
then setProps(cwd) + reload() with a gated OK listing, then assert the error is gone and Loading is
shown — and **it passes with the fix REVERTED**. Verified by removing the `treeError.value = null`
line and running it: 6 passed, 0 failed.

Why, as far as I can tell: `reload()` runs `teardown()` synchronously and then `start()`, which
awaits `nextTick()` before `createEditor()` and `loadRoot()`. `loadRoot()` clears `treeError` at its
first statement. A spec that awaits anything at all (flushPromises, nextTick) resumes AFTER
start()'s own continuation has already run, because start() suspended first and its continuation is
queued ahead of the test's. So the window in which the stale error is on screen is one render frame
between teardown and loadRoot, and a spec has no vantage point inside it.

So I removed the test rather than ship one that is green for the wrong reason, and the PR will say
the fix is state hygiene with an unobservable window.

Questions: is my account of the ordering wrong? Is there a vantage point I have missed that would
make this observable deterministically (not a timing hack)? And is there a LONGER-lived version of
this defect — some path where the stale error survives past loadRoot's first statement — that the
test should pin instead?

── 2. Your P3 (three `as unknown as` in the new spec) — I am DECLINING it

My review prompt said "no `as`", and that constraint came from the repo's CLAUDE.md, whose scope is
production TypeScript. In the test tree this repo has two established patterns, both current:

- `globalThis.fetch = vi.fn(...) as unknown as typeof fetch` — **213 occurrences** of that exact
  expression across the specs, including EVERY other Files-pane spec (`FilesPane.spec.ts`,
  `filesPanePreview.spec.ts`, `filesRowMenu.spec.ts`), which my new file sits beside and shares its
  helper shape with.
- `vi.stubGlobal("fetch", ...)` — used elsewhere (e.g. `test/src/composables/useSessionFeed.spec.ts`,
  `test/bin/update-check.spec.ts`), and it needs no assertion.

My third assertion is `(w.vm as unknown as { reload: () => Promise<void> }).reload()`, which is how
every sibling spec reaches a `defineExpose`d method; Vue Test Utils does not surface the exposed
surface on `vm`'s type.

My reasoning: making ONE file in this family different is a worse read than matching its three
neighbours, and the alternative for the third assertion does not exist without changing the
component. What would change my mind: if you think the fetch mock specifically should move to
`vi.stubGlobal` despite the neighbours — say so and I will change those two (the `vm` one has no
alternative).

--- its answer ---

1. ACCEPTED. Your ordering is correct: teardown() is synchronous, then start() waits for nextTick(), and loadRoot() clears treeError immediately before its first await. No deterministic test can observe a longer-lived stale error window. In normal control flow, no such path survives past loadRoot; only startup exceptions or calling setProps without the required reload() would.

2. ACCEPTED. The test-tree casts match established neighboring patterns, and the defineExpose() method has no practical typed alternative without changing the component. No need to replace the fetch casts with vi.stubGlobal().
```

</details>

isamu:
CODEX VERDICT: LGTM

| Axis | Result |
|---|---|
| Correctness and edge cases | none. Mounting `FilesOverlay` with no initial state starts at `roots === null`; successful, empty, failed, manual-reload, and re-root paths all select the intended branch. The request-generation guard prevents stale results from changing the state. |
| Four `?? []` readers | none. Rows, `revealPath`, restore, and snapshot all correctly treat an unread tree as containing no traversable or expanded nodes. |
| Tests | none. The five cases cover loading, confirmed empty, populated, failed, and re-root states; the assertions distinguish Loading from Empty and the re-root case would fail if teardown reverted to `[]`. The failure case also preserves the existing error-over-status precedence. |
| Consistency with the codebase / `CLAUDE.md` | none. `Loading…` matches neighboring panes; no new emoji, production `any`, or assertion cast was introduced. The 600-line cap is respected. The proposed row-menu extraction is a sensible follow-up boundary and does not need to be bundled with this bug fix. |
| Security | none. The new branch renders only a fixed string and exposes no user data. |

Validation: targeted spec 5/5; related Files-pane/overlay specs 111/111; `yarn typecheck`; `yarn lint` — all passed. I did not run a mutation sweep because `rm` is blocked by policy. The mutation worth trying is reverting `teardown()` to `roots.value = []`; the re-root case should turn red.

isamu:
## Codex cross-review — round 2 (tier S), head `5233fd90`: clean

Verdict: **CODEX VERDICT: LGTM**, no inline findings, every axis answered line by line, `FINDINGS COMPLETE` present. Nothing was pushed during this round, so it covers the head that will merge. Its posted comment is above.

### The mutation it asked for, run on its behalf

Codex cannot delete files in its sandbox, so it named the mutation instead: *"revert `teardown()` to `roots.value = []`; the re-root case should turn red."* Run here on this head:

```
× goes back to loading — not to empty — when the host re-roots it
  Tests  1 failed | 4 passed (5)
```

Exactly the one case, and only that one. The tree was restored and the file re-verified green afterwards.

### Ledger

| 0 | pre | sandbox pre-flight | — | Measured at 21:35Z today in this same working copy with the same flags (`--sandbox workspace-write -c sandbox_workspace_write.network_access=true`): vitest exit 0, `yarn typecheck` exit 0, socket bind exit 0, api.github.com 200, `gh pr view` exit 0. **`rm` is blocked by policy** — do not run a mutation sweep; name the mutation and I will run it |
| 1 | pre | claude, before round 1: a `treeLoaded` boolean beside `roots` was written first and thrown away | REJECTED (by me) | It pushed `FilesPane.vue` over the repo's 600-line `max-lines` cap, and it is a second value that has to be kept in step with the first — the same shape as the bug. `null` cannot drift from `roots` because it IS `roots` |
| 2 | 1 | P2 — `teardown()` cleared `roots` but left `treeError`, so a failed root's error could show over the next root during a re-root | FIXED | `teardown()` now clears it beside `roots`. **No test**: the window is one render frame between `teardown()` and `loadRoot()`, and a spec resumes after `start()`'s own `nextTick` continuation — the regression test Codex asked for passes with the fix reverted (verified: 6 passed with the line removed), so it was dropped rather than shipped green-for-the-wrong-reason. Codex ACCEPTED this account in round 1's follow-up |
| 3 | 1 | P3 — three `as unknown as` assertions in the new spec | REJECTED | The repo's test tree has 213 occurrences of `... as unknown as typeof fetch`, including every other Files-pane spec this file sits beside; the third assertion reaches a `defineExpose`d method, which VTU does not surface on `vm`'s type. The CLAUDE.md rule is about production TypeScript (`src`/`server`/`common` hold 22 such casts in total). Codex ACCEPTED the rebuttal in round 1's follow-up |
| 4 | 2 | codex round 2 on `5233fd90`: LGTM, no inline findings, every axis answered, `FINDINGS COMPLETE`. It ran the targeted spec (5), the related Files-pane/overlay specs (111), `yarn typecheck` and `yarn lint` | — | clean, on a head nothing was pushed to during the round |
| 5 | 2 | codex named one mutation it could not run itself (`rm` is blocked for it): revert `teardown()` to `roots.value = []` and the re-root case should turn red | RUN, AS ASKED | Exactly one case failed — `goes back to loading — not to empty — when the host re-roots it` — and the other four stayed green. The tree was restored and re-verified clean afterwards |

### State

Two rounds: round 1 found a real P2 (a failed root's error carried into the next root) and a P3 that was declined with evidence and accepted; round 2 is clean on the head that will merge. All 18 checks green. Local `yarn lint` and `yarn typecheck` clean.

```text
Review PR #2150 in receptron/mulmoterminal, head 5233fd90. Tier S. THIS IS ROUND 2.

Round 1 found two things; both are settled in the ledger below and one produced a code change
(`teardown()` now clears `treeError`). Review the head as it stands now.

Every finding you hold back costs a full extra round: a fix is reviewed for the first time in the
round AFTER it is pushed. Read the WHOLE diff before you post anything. Do not stop at the first
problem. Do not save anything for "a follow-up review".

Inspect with `gh pr diff 2150` or `git diff 8cb709a1...HEAD`, read the surrounding code, and RUN
things — your sandbox can run `npx vitest run <file> --configLoader runner`, `yarn typecheck`,
`yarn lint`, bind sockets and reach the network. **`rm` is blocked by policy, so do not run a
mutation sweep**: name any mutation you want tried and I will run it and report back.

Post inline comments via `gh api repos/receptron/mulmoterminal/pulls/2150/comments` and the final
top-level comment via `gh pr comment 2150`.

Issue #2148. `src/components/FilesPane.vue` held its directory tree in `const roots =
ref<Node[]>([])`, and the template rendered `v-else-if="roots.length === 0"` as "Empty directory."
— so an empty array meant BOTH "no listing has come back yet" AND "the listing came back empty",
and the pane announced the second for the whole of the root `/list` round trip. `teardown()` set
`roots` back to `[]`, reopening that window on every re-root (the host calls `reload()` as the zoom
walks between cells).

`roots` is now `Node[] | null`; `null` is "not read yet". `loadRoot()`'s assignment is the
transition; `teardown()` sets `null`; the four readers use `roots.value ?? []`. A third template
branch shows "Loading…" while `roots === null`.

  correctness and edge cases — especially: is there ANY path that leaves `roots` null forever, or
    that shows "Loading…" over a tree that is actually on screen, or "Empty directory." while a
    read is in flight? Consider loadRoot's own error path, the header's Reload button (which calls
    loadRoot directly, NOT teardown), reload()/teardown(), the treeReqId generation guard, and
    FilesOverlay (which mounts the same component with no initial state)
  the four `?? []` readers: `rows`, `findNode` in `revealPath`, `findNode` in `restore`,
    `expandedPaths` in `snapshot` — is `[]` the right answer for an unread tree at each one, or
    does any of them need to distinguish "unread" from "empty"?
  tests: does a test go red if each half breaks — and is any of the five new cases vacuous (green
    for a reason other than the one it names)? The spec is
    test/src/components/filesPaneTreeLoading.spec.ts
  consistency with the rest of the codebase and with CLAUDE.md (no emojis; no `any`; no `as`;
    comments say WHY not WHAT; the "Loading…" string matches the other panes)
  security — anything the new branch exposes (it renders no user data)

| 0 | pre | sandbox pre-flight | — | Measured at 21:35Z today in this same working copy with the same flags (`--sandbox workspace-write -c sandbox_workspace_write.network_access=true`): vitest exit 0, `yarn typecheck` exit 0, socket bind exit 0, api.github.com 200, `gh pr view` exit 0. **`rm` is blocked by policy** — do not run a mutation sweep; name the mutation and I will run it |
| 1 | pre | claude, before round 1: a `treeLoaded` boolean beside `roots` was written first and thrown away | REJECTED (by me) | It pushed `FilesPane.vue` over the repo's 600-line `max-lines` cap, and it is a second value that has to be kept in step with the first — the same shape as the bug. `null` cannot drift from `roots` because it IS `roots` |
| 2 | 1 | P2 — `teardown()` cleared `roots` but left `treeError`, so a failed root's error could show over the next root during a re-root | FIXED | `teardown()` now clears it beside `roots`. **No test**: the window is one render frame between `teardown()` and `loadRoot()`, and a spec resumes after `start()`'s own `nextTick` continuation — the regression test Codex asked for passes with the fix reverted (verified: 6 passed with the line removed), so it was dropped rather than shipped green-for-the-wrong-reason. Codex ACCEPTED this account in round 1's follow-up |
| 3 | 1 | P3 — three `as unknown as` assertions in the new spec | REJECTED | The repo's test tree has 213 occurrences of `... as unknown as typeof fetch`, including every other Files-pane spec this file sits beside; the third assertion reaches a `defineExpose`d method, which VTU does not surface on `vm`'s type. The CLAUDE.md rule is about production TypeScript (`src`/`server`/`common` hold 22 such casts in total). Codex ACCEPTED the rebuttal in round 1's follow-up |

That table is context, not a list of closed topics: if a resolution is wrong, say so and mark it
"REOPENING #<row>" with the specific evidence.

── ONE MORE THING THE PR ADMITS ─────────────────────────────────────────────
`FilesPane.vue` is at the repo's 600-line `max-lines` cap (counted without comments/blanks). The PR
body says the next change to the file has to extract something and names the row menu as the
candidate. If you think that call is wrong — or that this PR should carry the extraction — say so.

    'CODEX VERDICT: LGTM' or 'CODEX VERDICT: CHANGES REQUESTED' + bullets
  Then, on its own line:
  Omit that line if it is not true, and say which part you did not reach.

--- reply ---

Round 2 review completed and posted on PR #2150.

Verdict: **LGTM** — no inline findings. Targeted tests, related tests, typecheck, and lint all passed.
```

</details>

PR #2151, opened 2026-09-18
refactor: lift the Files pane's row menu into a composable

isamu:
## Summary

**No behaviour change.** `FilesPane.vue` sat one line under the repo's 600-line `max-lines` cap
after #2150, and the two changes queued behind it — the tree cache (#2148 stage 2) and the
remembered scroll/cursor positions (#2149) — both add to that file. This takes the row menu out so
there is room to do them.

What moves is the imperative half of the menu: the open state, the viewport-clamped position, the
`window` listeners, focus capture and return, and the arrow-key navigation →
`src/composables/useFilesRowMenu.ts`. What stays:

- **the `<Teleport>` markup**, so nothing about what renders moves at all;
- **which actions a row offers** (`rowActionsFor`) and **what picking one does** (`runRowAction`) —
  the two ends that read the pane's props and call its emits. Passing those in is what keeps the
  composable from needing the pane's whole prop surface.

## Why the row menu and not something else

| candidate | why not |
|---|---|
| external-change watcher (`checkForExternalChange`, `watchExternalChanges`, `onPageHide`) | reads six of the pane's refs; lifting it turns them into parameters and buys no testability |
| conflict banner (markup + `overwrite` / `discardAndReload`) | the markup moves but the state stays, so the two ends drift |
| **row menu** | a self-contained state machine, and `filesRowMenu.spec.ts` already drives it **end to end through the pane** — so the lift can be proved rather than argued |

## Items to Confirm / Review

- **`menuPosition` is now exported** (it was file-local) because the property test needs it. It is
  the only new public name besides the composable itself.
- **`pick(action)` runs the action and then closes**, in that order — which is what the pane did.
  The comment explaining why (an insert ends in `term.focus()`, and taking focus back would pull
  the keyboard off the terminal the user is about to type into) moved with it.
- The composable is generic in the row type (`useFilesRowMenu<Node>`), so it does not need
  `FilesPane.vue`'s local `Node` interface exported.
- Line count: the file has real headroom again. It is still the largest component in the pane's
  family, so the next thing to leave is worth thinking about before it is needed, not after.

## How it was verified

- **Every Files-pane spec runs unchanged** — `filesRowMenu`, `filesRowActions`, `FilesPane`,
  `filesPanePreview`, `filesPaneTreeLoading`, `FilesOverlay`, `gridOpenFiles`, `filesFinderInPane`:
  green. `filesRowMenu.spec.ts` is the one that matters: ten cases driving the menu through the
  pane, and it never names the composable, so the lift cannot satisfy it by construction.
- **A DOM differential, before and after.** A throwaway harness mounted the pane and opened the menu
  at a grid of pointer positions — the corners, points outside the viewport, and the middle — for
  both mounts of the pane (beside a cell with a terminal and a Canvas; full-screen with neither), on
  a file row and a directory row, through both entrances; then picked every action in turn. It
  recorded the menu's rendered markup, its computed `top`/`left`, which element held focus, and what
  the pane emitted. **The two captures compare byte-identical.** The harness is deleted — half of it
  was the code it compared against.
- **What survives the harness** is a permanent test: `test/src/composables/useFilesRowMenu.spec.ts`
  pins `menuPosition`, which was pure and had no direct test — the panel stays inside the viewport
  from any opening point, sits exactly under the pointer while there is room, and a taller menu
  starts higher. No magic numbers: the bounds are read back from the function, so moving the margin
  is not a failure while breaking the clamp is.
- `yarn format`, `yarn lint`, `yarn typecheck` clean.

## Note on `main`, not this PR

`yarn format` on a clean checkout rewrites `docs/_includes/head_custom.html` and `docs/facts.json`
— they landed unformatted (they are not in this diff; I reverted them twice while working). Worth a
one-line PR of its own.

## User Prompt

- 「他にも記憶しておいたほうがよいものがあれば対応してほしい」／Files ペインの表示順として「①ローディング状態 → 行メニュー切り出し → ツリーのキャッシュ → 位置の記憶」を承認。これは2本目。

## Related

- #2148 — the cache this makes room for. #2149 — the positions.

work in mulmoterminal5


isamu:
CODEX VERDICT: LGTM
- No inline findings.
- The old and new menu implementations match on listener symmetry and capture-phase scroll removal, restoreFocus call sites, action-before-close ordering, opener lifecycle, nextTick focus, empty-action handling, preventDefault placement, and unmount cleanup.
- Targeted validation passed: `filesRowMenu.spec.ts` and `useFilesRowMenu.spec.ts` (15 tests).
- The concurrent typecheck/lint run was interrupted by an external checkout from PR head to base; the supplied pre-flight results already recorded both as passing.

| Axis | Result |
|---|---|
| behaviour preservation | none |
| lifecycle | none |
| leaks | none |
| generic `useFilesRowMenu<Node>` | none |
| tests | none |
| consistency with codebase and CLAUDE.md | none |
| security | none expected |

test/src/composables/useFilesRowMenu.spec.ts
```diff
@@ -0,0 +1,47 @@
...
+  const FAR = 1_000_000;
+  const three = actions(3);
+  const min = menuPosition(three, -FAR, -FAR);
+  const max = menuPosition(three, FAR, FAR);
+
+  it("keeps the panel inside the viewport from any opening point", () => {
+    for (const x of [-FAR, -1, 0, 1, 200, 640, 1023, 1024, FAR]) {
+      for (const y of [-FAR, -1, 0, 1, 200, 384, 767, 768, FAR]) {
+        const at = menuPosition(three, x, y);
+        expect(at.left).toBeGreaterThanOrEqual(min.left);
```
isamu: P2 — The viewport assertions are self-referential. `min` and `max` (lines 21–22) are computed by `menuPosition` itself, so lines 28–31 only prove that the implementation maps the sampled inputs between its own two extreme outputs; they do not prove those outputs are the viewport bounds. For example, a mutation that clamps to an arbitrary off-screen interval (or uses the wrong viewport dimension) can still satisfy this test, while the extracted function’s key contract is keeping the menu reachable. Smallest fix: set/define a known `innerWidth`/`innerHeight` for this suite and assert the extreme results explicitly (at least `8` and `innerWidth - 200 - 8`, and `8` and `innerHeight - (actions.length * 30 + 12) - 8`), then keep the grid/property checks. What would change my mind: a committed test elsewhere already asserting these concrete viewport-bound values for `menuPosition`; the unchanged component spec does not.

isamu:

- P2 test gap: the new `menuPosition` test derives its expected `min`/`max` bounds from `menuPosition` itself, so it does not verify actual viewport clamping. Inline comment posted at `test/src/composables/useFilesRowMenu.spec.ts:28`.
- Checks: focused Vitest (2 files, 15 tests) passed; `yarn typecheck` passed; `yarn lint` passed; `git diff --check` passed.
- Mutation to try if useful: alter `menuPosition` to clamp to an arbitrary off-screen interval or use the wrong viewport dimension; the current new unit test can still pass. The smallest resolution is to set known viewport dimensions and assert the concrete edge values.
- No production behavior, lifecycle, listener-leak, generic-type, consistency, or security finding.

| Axis | Result |
|---|---|
| behaviour preservation | No production finding. The old and new paths preserve listener symmetry/capture, restore-focus call sites, action-before-close order, opener lifecycle, nextTick focus, empty-action early return, preventDefault placement, and unmount close. |
| lifecycle | None. `onBeforeUnmount` registered during the composable call in pane setup is equivalent here; it runs once with the component scope. |
| leaks | None. Open does not duplicate effective window listeners, close removes all three with matching callbacks/options, and unmount closes any open menu. |
| generic `useFilesRowMenu<Node>` | None. The generic only transports the row to `actionsFor`; it removes no runtime checks the old code made. |
| tests | Finding: P2 self-referential viewport-bound assertions in the new composable spec. The unchanged end-to-end Files pane coverage passes, but it does not cover the extracted pure positioning contract. |
| consistency with codebase and CLAUDE.md | None. The composable is in `src/composables`; production code adds no `any`, `as`, or emoji. |
| security | None found. |

isamu:
## Codex cross-review — rounds 1–2 (tier C)

### Round 1 on `86fa73c0`: VOID, and it was my fault

It returned LGTM, and its own report carried the reason to throw it away: *"the concurrent typecheck/lint run was interrupted by an external checkout from PR head to base"*. **That checkout was mine** — I switched this working copy to `main` and then to another branch while the review was reading it, to open a separate formatting PR. Part of that round may have read the base tree rather than this one. A verdict taken from a tree that moved under the reviewer is not a verdict, so it does not count and nothing was taken from it.

### Round 2 on the same head, working copy untouched: CHANGES REQUESTED, one P2 — in MY test

No production finding. Every axis it was asked for — behaviour preservation (listener symmetry and the capture-phase scroll removal, `restoreFocus` at each call site, run-before-close ordering, the `opener` lifecycle, the `nextTick` focus, the empty-actions early return, `preventDefault` placement, unmount cleanup), lifecycle, leaks, the generic, consistency, security — came back clean.

The finding was about the test I harvested from the differential harness:

> the new `menuPosition` test derives its expected `min`/`max` bounds from `menuPosition` itself, so it does not verify actual viewport clamping

Correct, and it is the vacuous-guard shape exactly: the assertions were true of ANY clamp. Codex named the two mutations that prove it — clamp to an arbitrary off-screen interval, and read the wrong viewport dimension — and could not run them itself (`rm` is blocked in its sandbox), so I ran both:

```
against the OLD test:   both mutations pass (that is the finding)
against the NEW test:   arbitrary interval  → 4 of 4 cases red
                        wrong dimension     → 2 of 4 cases red
```

Fixed in `74bdd32e`: the bounds come from the viewport and the panel's metrics, never from the function. `MENU_METRICS` is exported so a test can compute the box at all — and it now carries the note that the template duplicates its width as `min-w-[200px]`. The containment case runs on two deliberately **non-square** viewports, so swapping the two dimensions cannot hide inside a square.

### Ledger

| 0 | pre | sandbox pre-flight | — | Measured earlier today in this same working copy with the same flags: vitest exit 0, `yarn typecheck` exit 0, socket bind exit 0, api.github.com 200, `gh` exit 0. **`rm` is blocked by policy** — do not run a mutation sweep; name the mutation and I will run it |
| 1 | pre | claude, before round 1: is the lift behaviour-preserving? | PROVED, not argued | A throwaway differential harness captured the menu's rendered markup, computed `top`/`left`, focused element and the pane's emits across a grid of openings (corners, off-screen points, both mounts of the pane, file and directory rows, both entrances, every action picked) on the pre-lift tree and again after. The two captures compare byte-identical. Harness deleted; its generator and property survive as `test/src/composables/useFilesRowMenu.spec.ts` |
| 2 | 1 | codex round 1 on `86fa73c0`: LGTM, no inline findings, every axis answered, `FINDINGS COMPLETE` — but it reported "the concurrent typecheck/lint run was interrupted by an external checkout from PR head to base" | VOID — RE-RUN | That checkout was MINE: I switched this same working copy to `main` and then to a new branch (the docs-formatting PR) while the review was reading it. Part of that round may have read the base tree rather than the PR's. A verdict taken from a tree that moved underneath the reviewer is not a verdict, so round 1 does not count. Nothing is pushed between them — round 2 reviews the same head, on a working copy nobody touches while it runs |
| 3 | 2 | P2 — the new `menuPosition` spec derived its min/max bounds by calling `menuPosition` itself, so every assertion was true by construction: a clamp to an arbitrary interval, or one reading the wrong viewport dimension, passed it | FIXED | The bounds now come from the VIEWPORT and the panel's metrics, never from the function. `MENU_METRICS` is exported for that (and it names the `min-w-[200px]` the template duplicates). Both mutations Codex named were run: "arbitrary off-screen interval" turns all four cases red; "wrong viewport dimension" turns two red — the containment case is checked on two NON-square viewports for exactly that reason |

Round 3 is dispatched against `74bdd32e`.

```text
Review PR #2151 in receptron/mulmoterminal, head 86fa73c0. Tier C. THIS IS ROUND 2, ON THE SAME HEAD.

Round 1's verdict was thrown away, and not because of anything you said. I switched this working
copy to another branch while you were reading it — your own note about "an external checkout from
PR head to base" was me. So some of that round may have read the base tree instead of the PR's.
Nothing has been pushed since; the head is the same. This time the working copy stays put.

Review it as if for the first time. Do not lean on round 1's conclusions — they are exactly what is
in question.

Every finding you hold back costs a full extra round. Read the WHOLE diff before you post anything.
Do not stop at the first problem. Do not save anything for "a follow-up review".

Inspect with `gh pr diff 2151` or `git diff 5691806d...HEAD`, read the surrounding code, and RUN
things — your sandbox can run `npx vitest run <file> --configLoader runner`, `yarn typecheck`,
`yarn lint`, bind sockets and reach the network. **`rm` is blocked by policy, so do not run a
mutation sweep**: name any mutation you want tried and I will run it and report back.

Post inline comments via `gh api repos/receptron/mulmoterminal/pulls/2151/comments` and the final
top-level comment via `gh pr comment 2151`.

A pure lift, claimed to change NO behaviour. `src/components/FilesPane.vue` was one line under the
repo's 600-line `max-lines` cap, and two queued changes (#2148 stage 2, #2149) both add to it. The
row context menu's imperative half moves to `src/composables/useFilesRowMenu.ts`: open state,
viewport-clamped position, window listeners, focus capture/return, arrow-key navigation. The
`<Teleport>` markup stays in the pane, as do `rowActionsFor` (reads props) and `runRowAction`
(emits), which are passed in.

**This is a behaviour-preservation claim, which is the tier-C trigger.** Treat it as one: the
question is not "is the new code good" but "can the new code differ from the old for ANY input".

  behaviour preservation — go hunting for a divergence. Compare the OLD code (`git show
    5691806d:src/components/FilesPane.vue`) against the new pair line by line. Specifically:
    listener add/remove symmetry and the `capture` flag on scroll; the `restoreFocus` argument at
    each call site; the order of "run the action" vs "close"; what `opener` is set and cleared to;
    the `nextTick` focus; the early return when a row offers no actions; `event.preventDefault()`
    placement; and `onBeforeUnmount`
  lifecycle — `onBeforeUnmount` is now registered inside a composable called from the pane's setup.
    Is that equivalent to registering it in the component? Any case where the composable's cleanup
    does NOT run, or runs twice?
  leaks — the window listeners are added on open and removed on close. Is there a path that opens
    twice without closing, or unmounts with listeners still attached?
  the generic `useFilesRowMenu<Node>` — does the type erasure hide anything the old code enforced?
  tests — `filesRowMenu.spec.ts` is unchanged and passes; the new
    `test/src/composables/useFilesRowMenu.spec.ts` pins `menuPosition`. Is either vacuous? Is there
    a case the pre-lift code was covered for and the new arrangement is not?
  consistency with the codebase and CLAUDE.md (no emojis; no `any`; no `as` in production; comments
    say WHY; composables live in src/composables)
  security — none expected; say so or say what you found

1. SEVERITY — P1 / P2 / P3.  2. EVERY SITE.  3. THE SMALLEST CHANGE THAT RESOLVES IT.

| 0 | pre | sandbox pre-flight | — | Measured earlier today in this same working copy with the same flags: vitest exit 0, `yarn typecheck` exit 0, socket bind exit 0, api.github.com 200, `gh` exit 0. **`rm` is blocked by policy** — do not run a mutation sweep; name the mutation and I will run it |
| 1 | pre | claude, before round 1: is the lift behaviour-preserving? | PROVED, not argued | A throwaway differential harness captured the menu's rendered markup, computed `top`/`left`, focused element and the pane's emits across a grid of openings (corners, off-screen points, both mounts of the pane, file and directory rows, both entrances, every action picked) on the pre-lift tree and again after. The two captures compare byte-identical. Harness deleted; its generator and property survive as `test/src/composables/useFilesRowMenu.spec.ts` |
| 2 | 1 | codex round 1 on `86fa73c0`: LGTM, no inline findings, every axis answered, `FINDINGS COMPLETE` — but it reported "the concurrent typecheck/lint run was interrupted by an external checkout from PR head to base" | VOID — RE-RUN | That checkout was MINE: I switched this same working copy to `main` and then to a new branch (the docs-formatting PR) while the review was reading it. Part of that round may have read the base tree rather than the PR's. A verdict taken from a tree that moved underneath the reviewer is not a verdict, so round 1 does not count. Nothing is pushed between them — round 2 reviews the same head, on a working copy nobody touches while it runs |

The differential in row 1 is the evidence for the whole PR. If you think the grid misses a case
that could diverge — a path the harness never opened the menu on — say which, and I will run it.

  Then, on its own line:

--- reply ---

Review complete.

- Verdict: **CHANGES REQUESTED**
- Posted inline P2 finding: self-referential `menuPosition` viewport assertions.
- Posted required top-level comment: https://github.com/receptron/mulmoterminal/pull/2151#issuecomment-5737204136
- Focused Vitest, typecheck, lint, and diff checks passed.

```

</details>

isamu:
CODEX VERDICT: LGTM
- No inline findings.
- Targeted Vitest: 2 files, 16 tests passed.
- `yarn typecheck`: passed.
- `yarn lint`: passed.
- `git diff --check`: clean.

| Axis | Result |
|---|---|
| behaviour preservation | none — the old and new control flow match for listener add/remove symmetry (including capture on scroll), restore-focus call sites, action-before-close ordering, opener assignment/clearing, nextTick focus, empty-action early return, preventDefault placement, and cleanup |
| lifecycle | none — `onBeforeUnmount` registered during the composable call in pane setup is equivalent to the former registration in the same component setup; cleanup runs once |
| leaks | none — the listener callbacks and capture flag are symmetric, repeated opens do not duplicate DOM listeners, and unmount closes any open menu |
| generic `useFilesRowMenu<Node>` | none — the generic only abstracts the row passed to `actionsFor`; action typing and the pane's `Node` fields remain enforced at the pane boundary |
| tests | none — the existing end-to-end pane coverage remains unchanged and passes; the new clamp tests derive bounds from viewport and exported panel metrics and exercise non-square viewports plus menu height |
| consistency with codebase and CLAUDE.md | none — composable location, TypeScript types, no production `any`/casts, comments, and no-emoji policy are consistent |
| security | none expected — this is a UI state/behavior lift with no new trust boundary, input handling, or external access |

isamu:
## Codex cross-review — round 3 (tier C), head `74bdd32e`: clean

Verdict: **CODEX VERDICT: LGTM**, no inline findings, every axis answered, `FINDINGS COMPLETE`. Nothing was pushed during the round, so it covers the head that will merge.

The axis that matters for a lift like this — behaviour preservation — came back naming what it checked: listener add/remove symmetry including the capture flag on scroll, `restoreFocus` at each call site, run-before-close ordering, `opener` assignment and clearing, the `nextTick` focus, the empty-actions early return, `preventDefault` placement, and unmount cleanup.

### The loop, end to end

- **Round 1: void.** LGTM, and its own report said a checkout had moved the tree under it mid-review. That was me, switching this working copy to another branch to open #2152. A verdict from a tree that moved is not a verdict.

- **Round 2: one P2, in my test, not in the lift.** The `menuPosition` spec I harvested from the differential harness computed its bounds by calling `menuPosition` itself, so it was true of any clamp. Codex named the two mutations that expose it; I ran both (its sandbox cannot delete files). Fixed in `74bdd32e`.

- **Round 3: clean.**

What made this loop three rounds instead of one: a round I destroyed myself by moving the working copy, and a test-quality finding in the artefact the refactor produced rather than in the refactor. The lift itself drew no finding in any round.

### Ledger

| 0 | pre | sandbox pre-flight | — | Measured earlier today in this same working copy with the same flags: vitest exit 0, `yarn typecheck` exit 0, socket bind exit 0, api.github.com 200, `gh` exit 0. **`rm` is blocked by policy** — do not run a mutation sweep; name the mutation and I will run it |
| 1 | pre | claude, before round 1: is the lift behaviour-preserving? | PROVED, not argued | A throwaway differential harness captured the menu's rendered markup, computed `top`/`left`, focused element and the pane's emits across a grid of openings (corners, off-screen points, both mounts of the pane, file and directory rows, both entrances, every action picked) on the pre-lift tree and again after. The two captures compare byte-identical. Harness deleted; its generator and property survive as `test/src/composables/useFilesRowMenu.spec.ts` |
| 2 | 1 | codex round 1 on `86fa73c0`: LGTM, no inline findings, every axis answered, `FINDINGS COMPLETE` — but it reported "the concurrent typecheck/lint run was interrupted by an external checkout from PR head to base" | VOID — RE-RUN | That checkout was MINE: I switched this same working copy to `main` and then to a new branch (the docs-formatting PR) while the review was reading it. Part of that round may have read the base tree rather than the PR's. A verdict taken from a tree that moved underneath the reviewer is not a verdict, so round 1 does not count. Nothing is pushed between them — round 2 reviews the same head, on a working copy nobody touches while it runs |
| 3 | 2 | P2 — the new `menuPosition` spec derived its min/max bounds by calling `menuPosition` itself, so every assertion was true by construction: a clamp to an arbitrary interval, or one reading the wrong viewport dimension, passed it | FIXED | The bounds now come from the VIEWPORT and the panel's metrics, never from the function. `MENU_METRICS` is exported for that (and it names the `min-w-[200px]` the template duplicates). Both mutations Codex named were run: "arbitrary off-screen interval" turns all four cases red; "wrong viewport dimension" turns two red — the containment case is checked on two NON-square viewports for exactly that reason |
| 4 | 3 | codex round 3 on `74bdd32e`: LGTM, no inline findings, every axis answered, `FINDINGS COMPLETE`. Targeted vitest (2 files, 16 tests), `yarn typecheck`, `yarn lint`, `git diff --check` all clean in its own sandbox | — | clean, on a head nothing was pushed to during the round |

<details><summary>Round 3 prompt and reply (verbatim)</summary>

```text
Review PR #2151 in receptron/mulmoterminal, head 74bdd32e. Tier C. THIS IS ROUND 3.

Round 1 was void (I moved the working copy under you; my fault, not yours). Round 2 found one real
P2 — the harvested `menuPosition` spec checked the clamp against itself — and it is fixed in
74bdd32e: the bounds now come from the viewport and the panel's metrics, `MENU_METRICS` is exported
so a test can compute the box, and the containment case runs on two non-square viewports. Both
mutations you named were run and both now turn the spec red.

The production lift is unchanged since round 2 apart from `menuPosition` destructuring the exported
metrics instead of four module constants. Review the head as it stands.

Every finding you hold back costs a full extra round. Read the WHOLE diff before you post anything.
Do not stop at the first problem. Do not save anything for "a follow-up review".

Inspect with `gh pr diff 2151` or `git diff 5691806d...HEAD`, read the surrounding code, and RUN
things — your sandbox can run `npx vitest run <file> --configLoader runner`, `yarn typecheck`,
`yarn lint`, bind sockets and reach the network. **`rm` is blocked by policy, so do not run a
mutation sweep**: name any mutation you want tried and I will run it and report back.

Post inline comments via `gh api repos/receptron/mulmoterminal/pulls/2151/comments` and the final
top-level comment via `gh pr comment 2151`.

A pure lift, claimed to change NO behaviour. `src/components/FilesPane.vue` was one line under the
repo's 600-line `max-lines` cap, and two queued changes (#2148 stage 2, #2149) both add to it. The
row context menu's imperative half moves to `src/composables/useFilesRowMenu.ts`: open state,
viewport-clamped position, window listeners, focus capture/return, arrow-key navigation. The
`<Teleport>` markup stays in the pane, as do `rowActionsFor` (reads props) and `runRowAction`
(emits), which are passed in.

**This is a behaviour-preservation claim, which is the tier-C trigger.** Treat it as one: the
question is not "is the new code good" but "can the new code differ from the old for ANY input".

  behaviour preservation — go hunting for a divergence. Compare the OLD code (`git show
    5691806d:src/components/FilesPane.vue`) against the new pair line by line. Specifically:
    listener add/remove symmetry and the `capture` flag on scroll; the `restoreFocus` argument at
    each call site; the order of "run the action" vs "close"; what `opener` is set and cleared to;
    the `nextTick` focus; the early return when a row offers no actions; `event.preventDefault()`
    placement; and `onBeforeUnmount`
  lifecycle — `onBeforeUnmount` is now registered inside a composable called from the pane's setup.
    Is that equivalent to registering it in the component? Any case where the composable's cleanup
    does NOT run, or runs twice?
  leaks — the window listeners are added on open and removed on close. Is there a path that opens
    twice without closing, or unmounts with listeners still attached?
  the generic `useFilesRowMenu<Node>` — does the type erasure hide anything the old code enforced?
  tests — `filesRowMenu.spec.ts` is unchanged and passes; the new
    `test/src/composables/useFilesRowMenu.spec.ts` pins `menuPosition`. Is either vacuous? Is there
    a case the pre-lift code was covered for and the new arrangement is not?
  consistency with the codebase and CLAUDE.md (no emojis; no `any`; no `as` in production; comments
    say WHY; composables live in src/composables)
  security — none expected; say so or say what you found

1. SEVERITY — P1 / P2 / P3.  2. EVERY SITE.  3. THE SMALLEST CHANGE THAT RESOLVES IT.

| 0 | pre | sandbox pre-flight | — | Measured earlier today in this same working copy with the same flags: vitest exit 0, `yarn typecheck` exit 0, socket bind exit 0, api.github.com 200, `gh` exit 0. **`rm` is blocked by policy** — do not run a mutation sweep; name the mutation and I will run it |
| 1 | pre | claude, before round 1: is the lift behaviour-preserving? | PROVED, not argued | A throwaway differential harness captured the menu's rendered markup, computed `top`/`left`, focused element and the pane's emits across a grid of openings (corners, off-screen points, both mounts of the pane, file and directory rows, both entrances, every action picked) on the pre-lift tree and again after. The two captures compare byte-identical. Harness deleted; its generator and property survive as `test/src/composables/useFilesRowMenu.spec.ts` |
| 2 | 1 | codex round 1 on `86fa73c0`: LGTM, no inline findings, every axis answered, `FINDINGS COMPLETE` — but it reported "the concurrent typecheck/lint run was interrupted by an external checkout from PR head to base" | VOID — RE-RUN | That checkout was MINE: I switched this same working copy to `main` and then to a new branch (the docs-formatting PR) while the review was reading it. Part of that round may have read the base tree rather than the PR's. A verdict taken from a tree that moved underneath the reviewer is not a verdict, so round 1 does not count. Nothing is pushed between them — round 2 reviews the same head, on a working copy nobody touches while it runs |
| 3 | 2 | P2 — the new `menuPosition` spec derived its min/max bounds by calling `menuPosition` itself, so every assertion was true by construction: a clamp to an arbitrary interval, or one reading the wrong viewport dimension, passed it | FIXED | The bounds now come from the VIEWPORT and the panel's metrics, never from the function. `MENU_METRICS` is exported for that (and it names the `min-w-[200px]` the template duplicates). Both mutations Codex named were run: "arbitrary off-screen interval" turns all four cases red; "wrong viewport dimension" turns two red — the containment case is checked on two NON-square viewports for exactly that reason |

The differential in row 1 is the evidence for the whole PR. If you think the grid misses a case
that could diverge — a path the harness never opened the menu on — say which, and I will run it.

  Then, on its own line:

--- reply ---

Review complete. Posted the required top-level comment:

- Verdict: LGTM
- No inline findings
- Targeted tests, typecheck, lint, and diff checks passed
- All requested axes addressed

[View review comment](https://github.com/receptron/mulmoterminal/pull/2151#issuecomment-5737248785)
```

</details>

PR #2152, opened 2026-09-18
chore: run the repo's formatter over two files that landed unformatted

isamu:
## Summary

`yarn format` on a clean checkout of `main` rewrites `docs/_includes/head_custom.html` and
`docs/facts.json` — they landed unformatted in #2146. Every branch therefore starts with a diff it
did not make, and the first thing anyone does with that is revert it and hope it belonged to nobody.
I reverted it twice while working on #2150 and #2151 before filing this.

Nothing but whitespace, and checked rather than assumed:

- `docs/facts.json` parses to the **same object** before and after (`json.loads(old) == json.loads(new)`).
- `docs/_includes/head_custom.html` is **identical with whitespace stripped**.
- The spec that pins `facts.json`'s `version` / `requires.node` against `package.json` still passes.

## Items to Confirm / Review

- Tier T by the review protocol — formatting produced by the repo's own formatter, no shipped
  behaviour changes — so no Codex round; two self-review passes instead. Say if you want one anyway.
- If either file is deliberately hand-formatted (a Jekyll include can be whitespace-sensitive in
  places), say so and I will add it to `.prettierignore` instead, which fixes the same problem from
  the other end.

## User Prompt

- 「はい」— #2151 の PR 本文で報告した main 側の整形ずれを、1行の別 PR にしてよいか確認したのに対して。

work in mulmoterminal5


isamu:
## Tier T — the record

Reviewer protocol: tier **T** (formatting produced by the repo's own formatter; no shipped behaviour changes), so **no Codex round** and two self-review passes instead. This comment is the record of why the second reviewer is absent.

**Pass 1** — produced the diff (`npx prettier --write` on exactly the two files), checked it changes no values, ran the spec that pins `facts.json` against `package.json`, pushed.

**Pass 2** — on the pushed head (`origin/chore/format-docs-drift`), against `origin/main`, with CI green:

```
facts.json parses identical:                    True
head_custom.html identical ignoring whitespace: True
```

The whole diff is line-wrapping — arrays collapsed onto one line, attribute wrapping in the include. Nothing in it is a value, a key, an attribute or an order.

No code change came out of pass 2, so the tier stands (a code change in pass 2 would have promoted it and required a Codex round).

PR #2153, opened 2026-09-18
feat: the Files pane paints the last tree it saw while it re-reads (#2148)

isamu:
## Summary

Stage 2 of #2148, and the half the request was actually about: *「時間かかる場合あるから、前開いたときのををデータでキャッシュしておいて表示して、読み終わったら更新するのってダメかな？」*

The pane keeps the **root listing** of each directory it reads and paints it the next time that
directory is opened, then swaps in what the server says. Opening the pane on a directory you have
been to before shows the tree at once instead of `Loading…`, and so does every re-root as the zoom
walks between cells — which is where the wait is felt most, because it happens on every walk.

`src/components/filesTreeCache.ts` is the whole of it: parse, cap, read, write. Pure except for the
two `localStorage` accessors, wrapped where they are made, as `filesPaneStore.ts` does.

## Three decisions worth disagreeing with

**Its own storage key, not a field on `files_pane_state`.** A listing is far bigger than the paths
beside it, and localStorage answers a quota failure by failing the whole write — sharing the key
would let one large tree cost every *other* directory its remembered open file.

**The pane owns it, not the host.** The per-directory memory from #958 is written by `TerminalGrid`
because *what was open* is the host's business. A cache is not: it has no meaning outside the pane,
and routing it through the host would leave the full-screen view (`FilesOverlay`, which passes no
`initial-state`) without it for no reason.

**A painted cache is dropped when the read fails.** It was a guess; the error says we cannot confirm
it, and a stale tree under an error message reads as if we could. A tree that was really *read*
stays under an error — that one was true once. (This is why the Reload button's failure behaviour is
unchanged.)

## Items to Confirm / Review

- **A directory expanded during the window is carried across the swap.** Without it the tree
  collapses under the user a round trip after they clicked — in exactly the window the cache exists
  to fill, which is also the window they are most likely to click in. The carried children were
  fetched from the server, whatever painted the parent.
- **The expanded directories' own listings are NOT cached, deliberately.** `toggleDir` reads
  `node.loaded` as "these children are real", so a cached child list either lies to it (and is never
  refreshed) or forces a second swap path through every expansion. That decision wants its own PR;
  the root listing needs neither, because `loadRoot` replaces `roots` wholesale.
- **Staleness is visible for one round trip.** A cached row can name a file that is gone; clicking
  it asks the server, which refuses, and the pane already shows that refusal — the same path as a
  file deleted while the pane is open. The alternative (blocking clicks until the real listing
  lands) trades a rare error for a permanent hesitation.
- **Caps**: directories kept, and entries per directory, both enforced on read as well as on write —
  a value written by another build, or by hand, cannot come back over either bound.
- `test/src/components/filesFinderInPane.spec.ts` gained `localStorage.clear()` in its `beforeEach`.
  Its cases are about a directory the pane is seeing for the first time and they all mount on the
  same `/proj`, so without it the second case starts with the first one's tree already painted. That
  is the cache working, not the finder breaking — but the spec's invariant is about an empty tree.

## How it was verified

- `test/src/components/filesTreeCache.spec.ts` — the pure module, both directions: what it keeps,
  what it drops (six malformed shapes, each costing the entry and not the pane), both caps on read
  and on write, LRU eviction, and an empty listing coming back as `[]` rather than as nothing —
  because `null` would make the pane say `Loading…` for a directory it knows to be empty.
- `test/src/components/filesPaneTreeCache.spec.ts` — the pane: a first visit shows `Loading…`; the
  next paints immediately with no loading frame; the server's answer replaces it; another
  directory's listing is never painted; a failed read shows the error and no tree; an expansion made
  during the window survives the swap; and Reload leaves a tree that is already on screen alone.
- Break-verified, four mutations, each turning red exactly what it should: never paint the cache;
  never write it; keep a painted cache through a failed read; drop expansions on the swap.
- Every component spec in the repo: green. `yarn format`, `yarn lint`, `yarn typecheck` clean.
- Not driven in a browser. The behaviour is a render path with unit cover on all of its states; the
  previous stage's browser run covered the pane's plumbing.

## User Prompt

- 「ファイルペイン、load時に、Empty directory.ってなる。時間かかる場合あるから、前開いたときのををデータでキャッシュしておいて表示して、読み終わったら更新するのってダメかな？」
- 「①ローディング状態 ＋ ②ツリーのキャッシュ表示」で進めることを承認。これが②。

## Related

- #2148 stage 1 was #2150. #2149 — the remembered scroll and cursor positions — is next.

work in mulmoterminal5


isamu:
CODEX VERDICT: LGTM
- No inline findings.
- Verification: targeted cache/pane tests 40/40; broader Files-pane/overlay tests 59/59; yarn lint and yarn typecheck pass.

| Axis | Result |
|---|---|
| Correctness and edge cases | none - the generation guard invalidates painted and in-flight work; teardown invalidates requests; Reload preserves the confirmed tree; both production hosts reload after cwd changes; null cwd safely skips persistence; concurrent panes share the directory-keyed optimization without changing pane-local state. |
| Swap (adoptRoot) | none - children are adopted only when the old and new rows at the same relative path are both directories; files and type changes do not inherit children. Duplicate sibling names are not a filesystem-valid listing and existing path keys already assume unique names. |
| Cache correctness | none - only successful root reads are written, including valid empty listings; failed reads do not overwrite the cache; cwd is checked on read and write. |
| Storage | none - malformed/foreign values, older larger values, quota errors, and unavailable/private storage degrade to no cache/paint and do not break the pane. |
| Staleness | none - stale cached rows are visibly provisional for the one read window; clicking a removed file follows the existing server-error path, and a failed root read removes a cache-painted tree. |
| Tests | none - the new cases are substantive rather than vacuous. A full-screen overlay integration test is not present, but it mounts the same FilesPane path with the same cwd contract and no alternate cache logic. |
| Consistency with codebase and CLAUDE.md | none - the module is typed, contains no production any/casts/emojis, wraps storage failures, and follows the adjacent pane-store pattern. |
| Security | none - only root names/dir flags/sizes are persisted, not file contents or credentials; this has the same same-origin localStorage visibility as the existing pane-state path storage. |

PR #2156, opened 2026-09-18
feat: the Files pane comes back to where the reader was (#2149)

isamu:
## Summary

The pane remembered *which* file was open and which directories were expanded. It opened that file
at the top and scrolled the tree to the top, every time (#2149).

**The caret is kept as a line and column, not a pixel offset.** The file is often reopened in a pane
of a different width, and a wrapped paragraph puts the same `scrollTop` somewhere else in the text.
`goTo` clamps a line past the end rather than ignoring it — the file may have been edited since (by
the agent working in that same directory, which is the ordinary case here), and the nearest real
line is closer to where the reader was than the top is.

**The tree's scroll is one number**, restored after the remembered expansions have rendered, because
the rows have to exist before there is anything to scroll past.

Both ride `FilesPaneState`, so both get what #2137's view mode got: per cell while the session
lasts, per directory across a reload. The caret is applied inside `loadFile`, beside the restored
view mode and under the same `id === fileReqId` guard, and only when the read landed on the
remembered path and the file is text — a caret is a place in a document, and a path that now holds
something else is not that document.

## Items to Confirm / Review

- **The Markdown preview's scroll is NOT here, and I did not quietly skip it.** The iframe is
  `sandbox=""` — no `allow-same-origin`, no `allow-scripts` — so the app cannot read
  `contentWindow.scrollY` (opaque origin) and the frame cannot report it (no scripts). What it
  renders is `marked.parse()` output and the sandbox is the whole of what contains it, so making the
  position readable means `allow-same-origin`. That is a security decision with its own reviewer,
  not a line in this PR. Say if you want it opened as its own issue.
- **Only the OPEN file's caret is kept.** Switching A → B → A inside one session loses A's place. A
  caret per file is a second store with its own cap and eviction, and the complaint this answers is
  about coming back to a cell.
- **Seven specs had their own hand-written `fakeEditor`**, and adding two methods to `CmEditor`
  broke all seven — silently, as a call to a missing function, one spec at a time. They now share
  `test/helpers/cmEditorDouble.ts`, typed as `CmEditor`, so the next method added to the interface
  stops the build in one place. `tsconfig.test-server.json` excludes it for the same reason it
  already excludes `xtermDouble.ts`: it is the browser half of that directory.
- **The storage layer was where this nearly went wrong.** `capped()` whitelists fields, so the two
  new ones would have worked per cell and vanished across a reload. They are now carried, validated
  individually, and dropped ALONE when malformed — losing a caret costs a scroll position, while
  dropping the entry would cost the open file. A mutation removing the carry turns that spec red.
- `FilesPane.vue` is back under the `max-lines` cap with less room than before. The next feature in
  that file needs another extraction; the external-change watcher is the remaining candidate.

## How it was verified

- `test/src/components/cmEditorCaret.spec.ts` — against a **real** CodeMirror: an empty document has
  nowhere to come back to; a caret is reported and put back; a fresh file starts at the top; a line
  past the end and a column past the line's end clamp; line 0, a negative line and a negative column
  all survive; and replacing the document reports the new file's top rather than the old file's
  place. (jsdom has no layout, so `Range`'s two measuring methods are stubbed — with a comment
  saying that is the part this file does not test.)
- `test/src/components/filesPanePositions.spec.ts` — the pane: the snapshot carries both; the caret
  is placed for the remembered file; it is NOT placed when the read came back as something that is
  not text; a file with no remembered caret asks for nothing; the tree scroll is restored after the
  expansions have rendered, and left alone when nothing was remembered.
- `test/src/components/filesPaneStore.spec.ts` — the round trip, and four malformed carets plus two
  malformed scroll offsets, each costing itself and not the open file.
- Break-verified: four mutations (never place the caret; never restore the scroll; drop both from
  the snapshot; drop both from the store's round trip), each turning red exactly one case.
- Every spec under `test/src`: green. `yarn format`, `yarn lint`, `yarn typecheck` clean.
- Not driven in a browser. The caret's real-CodeMirror behaviour is covered above; what a browser
  would add is the visual confirmation that the view scrolls to the line, which `scrollIntoView`
  owns and jsdom cannot show.

## User Prompt

- 「他にも記憶しておいたほうがよいものがあれば対応してほしい」 に対し、覚えさせるものとして「エディタのカーソル／スクロール位置・ツリーのスクロール位置・Preview のスクロール位置」を選択。順番は「①ローディング状態 → 行メニュー切り出し → ツリーのキャッシュ → 位置の記憶」で承認。これが最後の1本（Preview のスクロールは上記のとおり別件）。

## Related

- #2149 — this. #2137 (#2144) is the same memory, one field earlier; #2148 (#2150, #2153) is what
  the pane shows before it has read anything.

work in mulmoterminal5

## Summary by CodeRabbit

- **New Features**
  - The Files pane now remembers and restores the open editor’s caret position, visible top line, and directory tree scroll position when reopened or refreshed.
  - Caret positions remain reliable after resizing, and positions beyond a file’s end are adjusted to the nearest valid location.
  - Invalid saved position data is safely ignored while preserving other Files pane settings.

- **Documentation**
  - Updated documentation explains restored positions and notes that Markdown preview scrolling is not retained.


isamu:
Preview のスクロール位置は **#2157** として切りました。

調べ直したところ、封じ込めは**二重**でした。iframe の `sandbox=""` だけでなく、サーバも `/api/files/browse/md` に `Content-Security-Policy: sandbox`（ディレクティブ無し＝最も強い）を付けていて、こちらは iframe 属性とは独立にこの応答を opaque origin に固定します。**片方だけ緩めても読めるようにはなりません。**

そしてそのヘッダが守っているのは Preview だけではありません: 同じルートは、ターミナル出力中の `.md` パスをクリックしたときの新タブ表示にも使われています（`terminalFilePathLinkProvider.ts`）。

issue には4つの選択肢を、それぞれ何を得て何を払うかと一緒に書いてあります（両方の sandbox を緩める / アンカーで近似する / アプリ内描画に変えて sandbox を捨てサニタイザに賭ける / やらない）。決めるのは人の仕事なので、こちらからは選んでいません。

src/components/filesPaneStore.ts
```diff
@@ -21,7 +22,22 @@ export interface RememberedPane {
...
+type StoredPaneState = Omit<FilesPaneState, "showPreview" | "caret" | "treeScrollTop"> & {
+  showPreview?: unknown;
+  caret?: unknown;
+  treeScrollTop?: unknown;
+};
+
+/** A caret is two numbers and nothing else. Anything else costs the CARET — the reader lands at the
+ *  top of the file they asked for, which is where they landed before this existed. */
+const asCaret = (value: unknown): CaretAt | undefined =>
+  isRecord(value) && typeof value.line === "number" && typeof value.col === "number" && Number.isFinite(value.line) && Number.isFinite(value.col)
```
isamu: P3: asCaret accepts finite fractional coordinates, but applyRemembered passes them to CmEditor.goTo, where CodeMirror expects an integer 1-based line and integer document offset. For example, line 1.5 can resolve to the next line, while a fractional col becomes a fractional selection position; malformed localStorage can therefore restore the wrong caret or throw during restore. Smallest fix: require Number.isInteger(value.line) and Number.isInteger(value.col) in this validator, while retaining goTo clamping for negative integer values. What would change my mind: a type/storage invariant proving persisted caret values cannot be fractional, or a goTo implementation that explicitly normalizes them before CodeMirror sees them.

src/components/FilesPane.vue
```diff
@@ -575,6 +587,12 @@ async function restore(state: FilesPaneState | null, reqIdAtStart: number): Prom
     if (node?.dir && !node.expanded) await toggleDir(node);
   }
   if (state.openPath && fileReqId === reqIdAtStart) await loadFile(state.openPath, false, state);
+  // Last, and only after a tick: the rows have to exist before there is anything to scroll past,
+  // and the expansions above are what create them.
+  if (state.treeScrollTop) {
```
isamu: P2: The zero offset is not restored on an in-place reload. treeEl is the same nav across teardown/start because the component remains mounted when the host changes roots. Switching from a tree scrolled to 240px to a remembered state with treeScrollTop 0 therefore leaves the old 240px scroll because this truthiness guard skips the assignment. Every other stored offset restores, but the explicit zero snapshot does not. Smallest fix: change this guard to if (state.treeScrollTop !== undefined). What would change my mind: evidence that every call path with a zero remembered offset unmounts and recreates the nav before restore; the current TerminalGrid root-switch path calls reload on the mounted pane.

isamu:
- P2: a remembered treeScrollTop of 0 is skipped during an in-place reload, leaving the prior root scroll position.
- P3: fractional caret coordinates are accepted from localStorage and passed to CodeMirror as invalid line/column positions.
- Targeted tests passed: 79 tests; yarn typecheck passed; yarn lint passed.

| Axis | Result |
|---|---|
| Correctness and edge cases | Findings: P2 at FilesPane.vue:592 for zero scroll restoration; P3 at filesPaneStore.ts:34 for fractional caret coordinates. goTo correctly clamps line 0, negative lines, past-end lines, columns past line ends, and empty documents. External-change refresh correctly uses loadFile(path, true) without applyRemembered. teardown and full-screen mounting have no additional finding. |
| Storage layer | P3 is the remaining finding: caret validation should require integer coordinates. The two new fields are whitelisted, carried, validated, and dropped independently; exactOptionalPropertyTypes handling is correct. Previously settled per-field carry behavior remains fixed. |
| Snapshot | Finding: P2, because treeScrollTop: 0 is written but not applied by the truthiness guard on an in-place reload. The no-tree fallback of 0 is appropriate. |
| Shared editor double | None. The CmEditor-typed object literal catches missing interface methods; the final cast only supplies the spy intersection type. The tsconfig exclusion matches the DOM helper pattern. |
| Tests | The layout stubs are explicitly scoped to caret placement and do not claim to test visual geometry. A reachable uncovered case is a persisted fractional caret; the zero-offset in-place reload is also not covered. |
| Consistency with codebase and CLAUDE.md | None. No new production any or as issue, no emoji issue, comments explain why, and the plan matches the shipped scope. |
| Security | None. The stored values are local numeric UI state, not sensitive data or executable content. Malformed localStorage can affect restoration behavior only; the fractional-coordinate issue is an integrity/robustness finding, not an escalation. |

The previously settled typed storage carry and shared editor-double items remain fixed. The sandboxed Markdown preview scroll remains deliberately deferred as documented in the PR.

isamu:
## Codex cross-review — round 1 (tier C), head `aad42bc0`

Verdict: **CHANGES REQUESTED** — one P2, one P3, every axis answered, `FINDINGS COMPLETE` present. Both were real and both are fixed in `3ae9aeab`.

### P2 — a re-root kept the previous directory's scroll

The tree's scroll was restored behind `if (state.treeScrollTop)`, so a remembered **0** was indistinguishable from nothing remembered. That matters because **the tree element outlives the root** — a re-root happens in place — so a directory with nothing remembered opened at whatever offset the last one left.

Fixed with two halves, deliberately: `teardown()` resets the element, and `restore` applies a remembered offset with `!== undefined`. The guard alone cannot cover a cell that has nothing remembered, which is the case the reset exists for. Removing the reset turns exactly that case red and leaves the 0 case green — the two are pinned separately.

### P3 — fractional caret coordinates

**Measured before fixing, rather than assumed.** CodeMirror does not refuse `line: 3.7`:

```
goTo({ line: 3.7, col: 0.5 })  →  no throw, caretAt() = { line: 4, col: 0.5 }
```

So the selection lands on a fractional document offset and the column reads back fractional — which is then what gets written into the next snapshot. Fixed at both boundaries, because they have different jobs: the store REJECTS a corrupt persisted value (costing the caret, not the open file), and `goTo` NORMALISES whatever any caller hands it, so the editor's own contract holds.

### Ledger

| 0 | pre | sandbox pre-flight | — | Measured earlier today in this working copy with the same flags: vitest, `yarn typecheck`, socket bind, api.github.com and `gh` all exit 0. **`rm` is blocked by policy** — name any mutation and I will run it |
| 1 | pre | claude, before round 1: `capped()` in filesPaneStore whitelists fields, so the two new ones worked per cell and vanished across a reload | FIXED | Carried and validated individually; a mutation removing the carry turns the store spec red |
| 2 | pre | claude, before round 1: seven specs had their own `fakeEditor` literal, and two new `CmEditor` methods broke all seven as a call to a missing function | FIXED | One typed double in `test/helpers/cmEditorDouble.ts`, excluded from the node tsconfig for the same reason `xtermDouble.ts` is |
| 3 | pre | claude: the Markdown preview's scroll cannot be read at all — `sandbox=""` means no `allow-same-origin` (opaque origin, so the app cannot read `contentWindow.scrollY`) and no `allow-scripts` (so the frame cannot report it) | DEFERRED | Making it readable is a security decision about the frame that renders `marked.parse()` output. Stated in the PR body, not silently skipped |
| 4 | 1 | P2 — a remembered `treeScrollTop` of 0 was skipped by a truthiness guard, so an in-place re-root left the PREVIOUS root's scroll on screen | FIXED | Two halves, because the guard alone does not cover a cell with nothing remembered: `teardown()` now resets the element (which outlives the root), and `restore` applies a remembered offset with `!== undefined`. Both are exercised — removing the teardown reset turns exactly the "nothing remembered" case red, the 0 case being covered by the guard |
| 5 | 1 | P3 — fractional caret coordinates were accepted from localStorage | FIXED | Measured rather than assumed first: CodeMirror does NOT refuse `line: 3.7` — it lands on a fractional offset and reads back `col: 0.5`, which is then what gets remembered. `asCaret` now requires integers (a corrupt value costs the caret), and `goTo` rounds whatever it is handed, so the editor's own contract holds for any caller |

Round 2 is dispatched against `3ae9aeab`.

<details><summary>Round 1 prompt and reply (verbatim)</summary>

```text
Review PR #2156 in receptron/mulmoterminal, head aad42bc0. Tier C.

Every finding you hold back costs a full extra round. Read the WHOLE diff before you post anything.
Do not stop at the first problem.

Inspect with `gh pr diff 2156` or `git diff c4ba0fbd...HEAD`. Your sandbox can run
`npx vitest run <file> --configLoader runner`, `yarn typecheck`, `yarn lint`, bind sockets and reach
the network. **`rm` is blocked by policy, so do not run a mutation sweep** — name any mutation you
want tried and I will run it. Post inline comments via
`gh api repos/receptron/mulmoterminal/pulls/2156/comments`, the verdict via `gh pr comment 2156`.

Issue #2149. The Files pane remembered WHICH file was open and which directories were expanded; it
opened the file at the top and scrolled the tree to the top every time. Now `FilesPaneState` also
carries a caret (line/column) and the tree's scrollTop. `cmEditor` gains `caretAt()` and `goTo()`;
the caret is applied inside `loadFile` beside the restored view mode, under the same
`id === fileReqId` guard. Seven specs' hand-written editor doubles were replaced by one typed
helper.

  correctness and edge cases — `goTo`'s clamping (line 0, negative, past the end, column past the
    line's end, an empty document); a caret restored onto a file that CHANGED since; the
    interaction with the external-change refresh (`checkForExternalChange` → `loadFile(path, true)`,
    which re-reads the SAME path and now goes through `applyRemembered` only when a remembered
    state was passed — check that); `teardown()`; the full-screen mount
  the storage layer — `capped()` whitelists; are both new fields carried, validated, and dropped
    INDIVIDUALLY rather than costing the entry? Is `exactOptionalPropertyTypes` respected?
  the snapshot — is `treeScrollTop: 0` (rather than absent) right when there is no tree element?
    Does the host write it at the moments that matter?
  the shared editor double — does typing it as `CmEditor` actually catch a missing method, or does
    the cast at the end of `fakeCmEditor` defeat that? Is the tsconfig exclusion right?
  tests — any case vacuous? The caret spec stubs `Range.prototype.getClientRects`/
    `getBoundingClientRect` because jsdom has no layout: does that stub hide anything the spec
    claims to check? Name a reachable case none of the new specs covers
  consistency with the codebase and CLAUDE.md (no emojis; no `any`; no `as` in production; comments
    say WHY; the plan file matches what shipped)
  security — a caret and a scroll offset in localStorage; anything to say?

1. SEVERITY — P1 / P2 / P3.  2. EVERY SITE.  3. THE SMALLEST CHANGE THAT RESOLVES IT.

| 0 | pre | sandbox pre-flight | — | Measured earlier today in this working copy with the same flags: vitest, `yarn typecheck`, socket bind, api.github.com and `gh` all exit 0. **`rm` is blocked by policy** — name any mutation and I will run it |
| 1 | pre | claude, before round 1: `capped()` in filesPaneStore whitelists fields, so the two new ones worked per cell and vanished across a reload | FIXED | Carried and validated individually; a mutation removing the carry turns the store spec red |
| 2 | pre | claude, before round 1: seven specs had their own `fakeEditor` literal, and two new `CmEditor` methods broke all seven as a call to a missing function | FIXED | One typed double in `test/helpers/cmEditorDouble.ts`, excluded from the node tsconfig for the same reason `xtermDouble.ts` is |
| 3 | pre | claude: the Markdown preview's scroll cannot be read at all — `sandbox=""` means no `allow-same-origin` (opaque origin, so the app cannot read `contentWindow.scrollY`) and no `allow-scripts` (so the frame cannot report it) | DEFERRED | Making it readable is a security decision about the frame that renders `marked.parse()` output. Stated in the PR body, not silently skipped |

Rows 1-3 are things I found before asking you; the diff already has 1 and 2, and 3 is deferred with
its reason. If a fix is wrong, say so and mark it "REOPENING #<row>".

  Then, on its own line:

--- reply ---

Review complete. Verdict posted: **CHANGES REQUESTED**.

Findings posted inline:

- P2: zero `treeScrollTop` is skipped on in-place reload.
- P3: fractional persisted caret coordinates are accepted.

Targeted tests, typecheck, lint, and diff checks passed. No fixes were applied.
```

</details>

src/components/FilesPane.vue
```diff
@@ -322,15 +328,21 @@ async function loadFile(pathRel: string, force = false, remembered: RememberedVi
     if (id !== fileReqId) return;
     if (res.status === 415) adoptUnpreviewable(pathRel, data);
     else adoptText(pathRel, data);
-    // Whether the remembered mode still holds is a question about the file that actually landed:
-    // the path may hold something else now, or nothing this pane can preview.
-    if (remembered)
-      showPreview.value = restoresPreview(remembered, { openPath: openPath.value, isMarkdown: isMarkdown.value, unpreviewable: unpreviewable.value !== null });
+    if (remembered) applyRemembered(remembered);
```
isamu: P2 — same-path refreshes still discard the live caret.

Every site: the clean external-change path in checkForExternalChange() calls loadFile(pathRel, true) (the write-hook path reaches the same function via void checkForExternalChange()), and the conflict banner's discardAndReload() also calls loadFile(openPath.value, true). Both calls pass no remembered state, so this branch skips applyRemembered(); adoptText() then calls setDoc(), which resets CodeMirror's selection to line 1. A user reading line 200 is therefore thrown to the top on the 30-second check or an agent write even though this PR adds caret restoration. The new refresh specs only assert setDoc; the shared fake does not reset/track its caret, so they cannot catch it.

Smallest fix: capture the current editor.caretAt() for a forced reload of the currently open path and call goTo() after adoptText() (the existing clamping handles files that changed), covering both force-reload call sites; add a regression test for the timer/write-hook path.

What would change my mind: a test or implementation showing that the real CodeMirror selection is intentionally preserved across setDoc(), or that external refreshes are deliberately allowed to move the reader to line 1.

isamu:
- One P2 remains; it is posted inline on `src/components/FilesPane.vue:331`.
- Focused tests: 84 passed; `yarn typecheck` and `yarn lint:ci` passed. The full suite encountered unrelated long-running server/network failures and was interrupted.

| Axis | Result |
|---|---|
| Correctness and edge cases | P2 finding: forced same-path external/conflict reloads reset the live caret because they do not pass a caret to the new `applyRemembered` path. Startup restore, changed-file clamping, empty documents, teardown reset, and full-screen mounting otherwise have no finding. |
| Storage layer | None. `capped()` validates and drops caret and `treeScrollTop` independently; caps and `exactOptionalPropertyTypes` are respected. |
| Snapshot | None. A missing tree element snapshots as `treeScrollTop: 0`; the TerminalGrid host writes the snapshot on close, re-root, and `pagehide`. |
| Shared editor double | None. The object literal is checked as `CmEditor` before the final spy-shape cast; the node-project exclusion is appropriate. |
| Tests | The new caret/layout stub is explicit about not testing visual layout. Reachable uncovered case: a clean external refresh or write-hook refresh while the caret is away from line 1; the shared fake does not model `setDoc()` resetting selection. |
| Consistency with codebase and CLAUDE.md | None found. No new production `any` or unsafe `as`; comments explain rationale; the plan matches the shipped startup behavior. |
| Security | None. The added localStorage values are finite/integer-validated UI state and do not expand file or HTML access. |

isamu:
## Codex cross-review — round 2 (tier C), head `3ae9aeab`

Verdict: **CHANGES REQUESTED** — one P2, every axis answered, `FINDINGS COMPLETE` present. It lands in the middle of what this PR claims, and it was right.

### A same-path re-read threw the reader to line 1

The thirty-second version check, the write hook and the conflict banner's Reload all call `loadFile(path, true)` with **no remembered state**, so the caret restoration this PR adds was skipped and `setDoc` collapsed the selection to the top. In this app that is not an edge case: **the agent working in the directory writes the file being read**, and the pane re-reads it. A reader at line 200 went to line 1 every thirty seconds — while the PR's title promised the opposite.

Fixed in `bd52faa1`: the caret is **carried** across a read that stays on the same file, beside the view mode that already was. Three things came with it:

- **`keepsPreview` is renamed `staysOnSameFile`.** It decides two things now — what the mode does and what the caret does — and the old name described one of them.
- **The shared editor double moves its caret.** Codex's note on this is the part worth repeating: *"the shared fake does not reset/track its caret, so they cannot catch it."* A double answering a fixed caret is what let the defect through a whole round with every spec green. `setDoc` now collapses the caret to the top and `goTo` places it, which is what makes the regression case able to fail at all.
- **The post-read step became its own function.** `loadFile` crossed the cognitive-complexity limit with the carry in it; the first version of the extraction (an options object) then crossed `max-lines`. It is `restorePlace(pathRel, remembered, carried)` — three parameters, no interface.

**`FilesPane.vue` is under the cap with little room left.** The external-change watcher is the next thing that has to leave it; that is a refactor PR, not this one.

### Ledger

| 0 | pre | sandbox pre-flight | — | Measured earlier today in this working copy with the same flags: vitest, `yarn typecheck`, socket bind, api.github.com and `gh` all exit 0. **`rm` is blocked by policy** — name any mutation and I will run it |
| 1 | pre | claude, before round 1: `capped()` in filesPaneStore whitelists fields, so the two new ones worked per cell and vanished across a reload | FIXED | Carried and validated individually; a mutation removing the carry turns the store spec red |
| 2 | pre | claude, before round 1: seven specs had their own `fakeEditor` literal, and two new `CmEditor` methods broke all seven as a call to a missing function | FIXED | One typed double in `test/helpers/cmEditorDouble.ts`, excluded from the node tsconfig for the same reason `xtermDouble.ts` is |
| 3 | pre | claude: the Markdown preview's scroll cannot be read at all — `sandbox=""` means no `allow-same-origin` (opaque origin, so the app cannot read `contentWindow.scrollY`) and no `allow-scripts` (so the frame cannot report it) | DEFERRED | Making it readable is a security decision about the frame that renders `marked.parse()` output. Stated in the PR body, not silently skipped |
| 4 | 1 | P2 — a remembered `treeScrollTop` of 0 was skipped by a truthiness guard, so an in-place re-root left the PREVIOUS root's scroll on screen | FIXED | Two halves, because the guard alone does not cover a cell with nothing remembered: `teardown()` now resets the element (which outlives the root), and `restore` applies a remembered offset with `!== undefined`. Both are exercised — removing the teardown reset turns exactly the "nothing remembered" case red, the 0 case being covered by the guard |
| 5 | 1 | P3 — fractional caret coordinates were accepted from localStorage | FIXED | Measured rather than assumed first: CodeMirror does NOT refuse `line: 3.7` — it lands on a fractional offset and reads back `col: 0.5`, which is then what gets remembered. `asCaret` now requires integers (a corrupt value costs the caret), and `goTo` rounds whatever it is handed, so the editor's own contract holds for any caller |
| 6 | 2 | P2 — a same-path re-read (`checkForExternalChange` → `loadFile(path, true)`, and the conflict banner's Reload) passed no remembered state, so `setDoc` collapsed the selection and threw the reader to line 1 — every thirty seconds, in the app this PR exists to fix that for | FIXED | The caret is CARRIED across a read that stays on the same file, beside the view mode that already was. `keepsPreview` renamed `staysOnSameFile`, because it now decides two things and the old name described one of them. Break-verified: removing the carry turns the new refresh case red |
| 7 | 2 | claude, on codex's note that the shared double "does not reset/track its caret, so they cannot catch it" | FIXED | The double now MOVES the caret the way the editor does — `setDoc` collapses it to the top, `goTo` places it. A double that answered a fixed caret is what let this defect through a whole round with every spec green |
| 8 | 2 | claude, after the fix: `loadFile` went over the cognitive-complexity limit, and the first extraction put `FilesPane.vue` over `max-lines` | FIXED | The post-read step is one `restorePlace(pathRel, remembered, carried)` rather than an options object — three parameters, no interface, and the file lands under the cap with little room. The external-change watcher is the next thing that has to leave this file |

Round 3 is dispatched against `bd52faa1`.

```text
Review PR #2156 in receptron/mulmoterminal, head 3ae9aeab. Tier C. THIS IS ROUND 2.

Round 1 found two real things and both are fixed in 3ae9aeab — rows 4 and 5 below. The P2 fix has
two halves on purpose (teardown resets the element, restore applies a remembered offset), because
the guard alone cannot cover a cell with nothing remembered. Review the head as it stands.

Every finding you hold back costs a full extra round. Read the WHOLE diff before you post anything.
Do not stop at the first problem.

Inspect with `gh pr diff 2156` or `git diff c4ba0fbd...HEAD`. Your sandbox can run
`npx vitest run <file> --configLoader runner`, `yarn typecheck`, `yarn lint`, bind sockets and reach
the network. **`rm` is blocked by policy, so do not run a mutation sweep** — name any mutation you
want tried and I will run it. Post inline comments via
`gh api repos/receptron/mulmoterminal/pulls/2156/comments`, the verdict via `gh pr comment 2156`.

Issue #2149. The Files pane remembered WHICH file was open and which directories were expanded; it
opened the file at the top and scrolled the tree to the top every time. Now `FilesPaneState` also
carries a caret (line/column) and the tree's scrollTop. `cmEditor` gains `caretAt()` and `goTo()`;
the caret is applied inside `loadFile` beside the restored view mode, under the same
`id === fileReqId` guard. Seven specs' hand-written editor doubles were replaced by one typed
helper.

  correctness and edge cases — `goTo`'s clamping (line 0, negative, past the end, column past the
    line's end, an empty document); a caret restored onto a file that CHANGED since; the
    interaction with the external-change refresh (`checkForExternalChange` → `loadFile(path, true)`,
    which re-reads the SAME path and now goes through `applyRemembered` only when a remembered
    state was passed — check that); `teardown()`; the full-screen mount
  the storage layer — `capped()` whitelists; are both new fields carried, validated, and dropped
    INDIVIDUALLY rather than costing the entry? Is `exactOptionalPropertyTypes` respected?
  the snapshot — is `treeScrollTop: 0` (rather than absent) right when there is no tree element?
    Does the host write it at the moments that matter?
  the shared editor double — does typing it as `CmEditor` actually catch a missing method, or does
    the cast at the end of `fakeCmEditor` defeat that? Is the tsconfig exclusion right?
  tests — any case vacuous? The caret spec stubs `Range.prototype.getClientRects`/
    `getBoundingClientRect` because jsdom has no layout: does that stub hide anything the spec
    claims to check? Name a reachable case none of the new specs covers
  consistency with the codebase and CLAUDE.md (no emojis; no `any`; no `as` in production; comments
    say WHY; the plan file matches what shipped)
  security — a caret and a scroll offset in localStorage; anything to say?

1. SEVERITY — P1 / P2 / P3.  2. EVERY SITE.  3. THE SMALLEST CHANGE THAT RESOLVES IT.

| 0 | pre | sandbox pre-flight | — | Measured earlier today in this working copy with the same flags: vitest, `yarn typecheck`, socket bind, api.github.com and `gh` all exit 0. **`rm` is blocked by policy** — name any mutation and I will run it |
| 1 | pre | claude, before round 1: `capped()` in filesPaneStore whitelists fields, so the two new ones worked per cell and vanished across a reload | FIXED | Carried and validated individually; a mutation removing the carry turns the store spec red |
| 2 | pre | claude, before round 1: seven specs had their own `fakeEditor` literal, and two new `CmEditor` methods broke all seven as a call to a missing function | FIXED | One typed double in `test/helpers/cmEditorDouble.ts`, excluded from the node tsconfig for the same reason `xtermDouble.ts` is |
| 3 | pre | claude: the Markdown preview's scroll cannot be read at all — `sandbox=""` means no `allow-same-origin` (opaque origin, so the app cannot read `contentWindow.scrollY`) and no `allow-scripts` (so the frame cannot report it) | DEFERRED | Making it readable is a security decision about the frame that renders `marked.parse()` output. Stated in the PR body, not silently skipped |
| 4 | 1 | P2 — a remembered `treeScrollTop` of 0 was skipped by a truthiness guard, so an in-place re-root left the PREVIOUS root's scroll on screen | FIXED | Two halves, because the guard alone does not cover a cell with nothing remembered: `teardown()` now resets the element (which outlives the root), and `restore` applies a remembered offset with `!== undefined`. Both are exercised — removing the teardown reset turns exactly the "nothing remembered" case red, the 0 case being covered by the guard |
| 5 | 1 | P3 — fractional caret coordinates were accepted from localStorage | FIXED | Measured rather than assumed first: CodeMirror does NOT refuse `line: 3.7` — it lands on a fractional offset and reads back `col: 0.5`, which is then what gets remembered. `asCaret` now requires integers (a corrupt value costs the caret), and `goTo` rounds whatever it is handed, so the editor's own contract holds for any caller |

Rows 1-3 are things I found before asking you; the diff already has 1 and 2, and 3 is deferred with
its reason. If a fix is wrong, say so and mark it "REOPENING #<row>".

  Then, on its own line:

--- reply ---

Review complete.

- Posted one inline P2 finding: external/conflict reloads reset the live caret.
- Posted verdict: **CHANGES REQUESTED**
- Focused tests: 84 passed; typecheck and lint passed.
- Full suite had unrelated server/network failures and was interrupted.

[Inline finding](https://github.com/receptron/mulmoterminal/pull/2156#discussion_r4051664862) · [Verdict](https://github.com/receptron/mulmoterminal/pull/2156#issuecomment-5737893224)
```

</details>

isamu:
CODEX VERDICT: LGTM

- No inline findings.
- Targeted position/editor/storage tests: 54 passed.
- `git diff --check`: clean.

| Axis | Result |
|---|---|
| Correctness and edge cases | none |
| Storage layer | none |
| Snapshot | none |
| Shared editor double | none |
| Tests | none |
| Consistency with codebase and CLAUDE.md | none |
| Security | none |

isamu:
## Codex cross-review — round 3 (tier C), head `bd52faa1`: clean

Verdict: **CODEX VERDICT: LGTM**, no inline findings, every axis answered, `FINDINGS COMPLETE` present. Nothing was pushed during the round, so it covers the head that will merge.

One process note for the record: **this run’s stdout came back empty.** The verdict above was posted by Codex from inside the run (the comment URL appears in the run’s stderr), so the round happened — but a reader comparing this thread with the transcripts in the state directory will find nothing under `raw-3.txt`, and that is why.

### The loop, end to end

- **Round 1** — two findings, both real: a remembered scroll of 0 skipped by a truthiness guard (and the tree element outlives the root, so the previous directory’s scroll stayed), and fractional caret coordinates accepted from localStorage.
- **Round 2** — one finding, and the important one: a same-path re-read threw the reader to line 1, which is the case this PR exists for. The shared editor double could not have caught it, so the double was fixed too.
- **Round 3** — clean.

What made this three rounds: two of the three findings were in code this PR added, and the third was in the test double that was supposed to be checking it. None of them was in the original idea.

### Ledger

| 0 | pre | sandbox pre-flight | — | Measured earlier today in this working copy with the same flags: vitest, `yarn typecheck`, socket bind, api.github.com and `gh` all exit 0. **`rm` is blocked by policy** — name any mutation and I will run it |
| 1 | pre | claude, before round 1: `capped()` in filesPaneStore whitelists fields, so the two new ones worked per cell and vanished across a reload | FIXED | Carried and validated individually; a mutation removing the carry turns the store spec red |
| 2 | pre | claude, before round 1: seven specs had their own `fakeEditor` literal, and two new `CmEditor` methods broke all seven as a call to a missing function | FIXED | One typed double in `test/helpers/cmEditorDouble.ts`, excluded from the node tsconfig for the same reason `xtermDouble.ts` is |
| 3 | pre | claude: the Markdown preview's scroll cannot be read at all — `sandbox=""` means no `allow-same-origin` (opaque origin, so the app cannot read `contentWindow.scrollY`) and no `allow-scripts` (so the frame cannot report it) | DEFERRED | Making it readable is a security decision about the frame that renders `marked.parse()` output. Stated in the PR body, not silently skipped |
| 4 | 1 | P2 — a remembered `treeScrollTop` of 0 was skipped by a truthiness guard, so an in-place re-root left the PREVIOUS root's scroll on screen | FIXED | Two halves, because the guard alone does not cover a cell with nothing remembered: `teardown()` now resets the element (which outlives the root), and `restore` applies a remembered offset with `!== undefined`. Both are exercised — removing the teardown reset turns exactly the "nothing remembered" case red, the 0 case being covered by the guard |
| 5 | 1 | P3 — fractional caret coordinates were accepted from localStorage | FIXED | Measured rather than assumed first: CodeMirror does NOT refuse `line: 3.7` — it lands on a fractional offset and reads back `col: 0.5`, which is then what gets remembered. `asCaret` now requires integers (a corrupt value costs the caret), and `goTo` rounds whatever it is handed, so the editor's own contract holds for any caller |
| 6 | 2 | P2 — a same-path re-read (`checkForExternalChange` → `loadFile(path, true)`, and the conflict banner's Reload) passed no remembered state, so `setDoc` collapsed the selection and threw the reader to line 1 — every thirty seconds, in the app this PR exists to fix that for | FIXED | The caret is CARRIED across a read that stays on the same file, beside the view mode that already was. `keepsPreview` renamed `staysOnSameFile`, because it now decides two things and the old name described one of them. Break-verified: removing the carry turns the new refresh case red |
| 7 | 2 | claude, on codex's note that the shared double "does not reset/track its caret, so they cannot catch it" | FIXED | The double now MOVES the caret the way the editor does — `setDoc` collapses it to the top, `goTo` places it. A double that answered a fixed caret is what let this defect through a whole round with every spec green |
| 8 | 2 | claude, after the fix: `loadFile` went over the cognitive-complexity limit, and the first extraction put `FilesPane.vue` over `max-lines` | FIXED | The post-read step is one `restorePlace(pathRel, remembered, carried)` rather than an options object — three parameters, no interface, and the file lands under the cap with little room. The external-change watcher is the next thing that has to leave this file |
| 9 | 3 | codex round 3 on `bd52faa1`: LGTM, no inline findings, every axis answered, `FINDINGS COMPLETE`. Targeted position/editor/storage tests 54 passed, `git diff --check` clean | — | clean, on a head nothing was pushed to during the round. Its stdout came back EMPTY and the verdict was posted from inside the run — the comment URL is in the run's stderr, so the round happened and its record is on the PR |

src/components/cmEditor.ts
```diff
@@ -143,14 +173,14 @@ export function createEditor(parent: HTMLElement, onChange: () => void): CmEdito
...
+    goTo,
     revealLine(line) {
-      // CLAMPED to the document rather than trusted. The line came from a search over the file ON
-      // DISK, and the buffer can already be shorter — the agent in this directory rewrites files
-      // while the panel is open. CodeMirror throws on an out-of-range line, which would take the
-      // click with it and look like a dead result.
-      const target = Math.min(Math.max(Math.trunc(line), 1), view.state.doc.lines);
-      const { from } = view.state.doc.line(target);
-      view.dispatch({ selection: { anchor: from }, effects: EditorView.scrollIntoView(from, { y: "center" }) });
+      goTo({ line, col: 0 });
```
isamu: P3 — Preserve main\u0027s `revealLine` truncation semantics here. In `origin/main`, `revealLine(line)` did `Math.trunc(line)` before clamping; after the merge it routes the raw value through `goTo`, which uses `Math.round`, so e.g. `revealLine(1.6)` now lands on line 2 instead of line 1 (and `revealLine(-0.6)` differs before the clamp). The current search caller supplies integer lines, but this is still an observable change in the `CmEditor` contract introduced by the merge. Every site: `revealLine` here, reached from `src/composables/useFileSearchPanel.ts:71`; `goTo` must retain its rounding for restored caret coordinates. Smallest fix: call `goTo({ line: Math.trunc(line), col: 0 })` in `revealLine`, and keep the focus call. What would change my mind: an explicit contract that `revealLine` accepts fractional lines and intentionally rounds them, plus updated main-side behavior/tests.

isamu:

- P3 — `src/components/cmEditor.ts:183` changes main's `revealLine` from `Math.trunc` to `Math.round` via the merged `goTo`. Every site is this method, reached by `src/composables/useFileSearchPanel.ts:71`; ordinary search results are integer lines, but fractional callers now observe different lines. Smallest fix: `goTo({ line: Math.trunc(line), col: 0 })` here and retain `view.focus()`. I would change my mind if `revealLine` is explicitly redefined to round fractional inputs and its tests/main-side contract are updated.

| Axis | Result |
|---|---|
| Correctness and edge cases | One P3 above. Otherwise the merged clamp handles line 0/negative/past-end, columns past line ends, empty documents, changed files, same-file external refresh, teardown, and full-screen mounting as intended; the refresh carries the live caret only when no remembered state was passed. |
| Storage layer | None. `capped()` validates and drops `caret` and `treeScrollTop` independently; finite/nonnegative scroll values and integer caret coordinates are carried correctly, and the optional-field construction respects `exactOptionalPropertyTypes`. |
| Snapshot | None. `treeScrollTop: 0` is the correct no-element/default value, and the host snapshots it on pane leave, re-root, and `pagehide`. |
| Shared editor double | None. The `const editor: CmEditor` assignment catches missing interface methods; the final cast only supplies the spy-intersection shape and does not defeat that check. The helper is excluded from the node-only project and included through the browser test project. |
| Tests | None. The merged search spec still checks its prior contract, including `setDoc` before `revealLine`, while retaining the shared double. The layout stubs only suppress jsdom geometry errors; they do not fake caret state. Reachable coverage note: no new spec calls `goTo` on an empty document, although the implementation safely handles it. |
| Consistency and plan | None. No new production `any`, emoji, or unsafe assertion issue; comments explain why, and `plans/feat-2149-files-pane-positions.md` matches the shipped behavior, including the deferred preview scroll. |
| Security | None. The localStorage additions are bounded position metadata only; malformed values are discarded and large finite values are harmlessly clamped by the editor/browser. |

isamu:
CODEX VERDICT: LGTM
- No new findings at head `24b6e2b1`; the prior round-4 truncation regression is fixed and pinned.
- Targeted review suite: 108 tests passed.
- `yarn typecheck` passed.
- `yarn lint` passed.
- No inline comments to add.

| Axis | Result |
|---|---|
| Correctness and edge cases | none. `goTo` truncates and clamps line 0, negative, past-end, column past line end, and empty documents; changed files clamp to their current contents. Same-path external refreshes and conflict-banner reload carry the live caret, while remembered state is applied only to the matching text path under the request guard. `teardown()` invalidates work and resets the surviving tree element; full-screen mounting has no additional issue. |
| Storage layer | none. `capped()` whitelists and carries caret and tree scroll independently; each is validated and dropped alone. Integer caret coordinates, finite non-negative scroll offsets, legacy records, and `exactOptionalPropertyTypes` are handled correctly. |
| Snapshot | none. `treeScrollTop: 0` is the correct no-element fallback, and zero is restored with an explicit undefined check. The host writes snapshots on pane close, re-root, and pagehide, including the new fields. |
| Shared editor double | none. The object literal is first checked against `CmEditor`; the final cast only adds the spy/intersection type and does not defeat the missing-method check. The node-project exclusion matches the DOM-dependent helper pattern. |
| Tests | none. The new cases are substantive and the refresh double now tracks caret movement. A reachable case not directly covered is clicking the conflict banner’s “Reload” path; it shares the same `loadFile(path, true)` path already covered by the timer refresh. Browser-only visual confirmation that CodeMirror’s `scrollIntoView` moves the viewport is also intentionally outside jsdom coverage because the Range stubs only remove layout dependence. |
| Consistency with the codebase and CLAUDE.md | none. The plan and README match the shipped scope; comments explain why; no new production `any` or unsafe assertion was introduced. |
| Security | none. The persisted values are local numeric UI state, validated on read, and cannot execute content or escape file boundaries. Markdown preview scroll remains deliberately deferred to #2157 because making the sandboxed iframe readable is a separate security decision. |

CODEX CHECKPOINT: round 5

1. Yes. The diff still implements the PR’s stated purpose: it remembers the open file’s caret and the tree scroll position, restores them safely across reloads and same-file refreshes, and preserves the existing file/view-mode behavior.
2. No. The review stayed on the PR’s position-memory purpose; the preview-scroll question is explicitly documented as deferred to #2157 rather than being silently folded into this change.
3. No material part is present only as reviewer compliance. The clamp unification preserves main’s existing `revealLine` behavior, and the shared typed double, storage validation, and regression cases each provide independent correctness value.
4. No split, redirect, or closure is warranted. The change is cohesive; #2157 is already the appropriate separate follow-up for preview scrolling and sandbox policy.
5. The single largest remaining risk is browser-only behavior not reproducible in jsdom: whether CodeMirror’s real `scrollIntoView` and the post-render tree assignment produce the intended visual position across browser/layout conditions. The implementation delegates to those native mechanisms and the relevant ordering is covered, but there is no browser-driven assertion.

src/components/FilesPane.vue
```diff
@@ -314,15 +324,19 @@ async function mayLeaveCurrent(pathRel: string, force: boolean): Promise<boolean
...
   unpreviewable.value = null;
-  // The mode belongs to the file it was turned on for, which is why this sits beside the
-  // `unpreviewable` reset rather than being unconditional: see keepsPreview.
-  if (!keepsPreview(openPath.value, pathRel)) showPreview.value = false;
+  // What survives a re-read of the SAME file, and what a different file leaves behind: the mode
+  // belongs to the file it was turned on for, and so does the reader's place in it. Carried across
+  // the read rather than restored from a snapshot, because this path has no snapshot — the agent
+  // editing the file you are reading is what triggers it (see staysOnSameFile).
+  const staying = staysOnSameFile(openPath.value, pathRel);
+  const carried = staying ? (editor?.caretAt() ?? null) : null;
```
isamu: P2 — the external reread still loses the viewport. Sites: loadFile carried state at FilesPane.vue:338 and its application in restorePlace at FilesPane.vue:357-360. checkForExternalChange() -> loadFile(pathRel, true) passes no snapshot, so this is the only same-file restore path; it carries only caretAt(). setDoc resets CodeMirror scroll, then goTo(carried) recenters the caret, leaving a reader who was scrolled to line 130 back near the top on every refresh (and on the conflict banner Reload). Smallest fix: carry editor?.topLine() alongside the caret for the same-file path and apply it after goTo in restorePlace, with null/undefined independently optional and under the existing arrived-file guard. What would change my mind: a browser test showing a clean external reread/reload preserves the prior top visible line, or a documented decision that live refresh intentionally resets the viewport.

isamu:
CODEX VERDICT: CHANGES REQUESTED\n- P2 inline finding posted: same-file external rereads carry the caret but still drop the top visible line, so refresh/reload returns a scrolled reader to the top.\n- Validation: targeted Vitest 99 passed; yarn typecheck passed; yarn lint passed.\n\n| Axis | Result |\n|---|---|\n| correctness and edge cases | One finding: external-change refresh and conflict-banner Reload do not carry/apply topLine. The new topLine measurement is otherwise appropriate: empty documents return null; short documents report line 1; lineBlockAtHeight maps a mid-wrap scroll position to its logical source line; caret-first then screen-last is correct for remembered fields independently, and arrived-file guards prevent restoring onto a non-text/path-mismatched result. |\n| storage layer | none. asLine rejects 0, fractions, and strings; topLine is validated/carried/dropped independently, and exactOptionalPropertyTypes is respected. |\n| snapshot | none. topLine is sampled in snapshot, treeScrollTop: 0 remains the correct no-tree fallback, and the existing close/re-root/pagehide host writes include the snapshot. |\n| shared editor double | none. The object is first assigned as CmEditor, so a missing interface method fails typecheck; the final mapped-type cast does not defeat that check. The DOM helper exclusion is paired with the browser test project/import graph. |\n| tests | The Range geometry stubs intentionally remove real layout, so the specs do not prove actual scroll movement, short-viewport bottom clamping, or wrapped-line geometry. The reachable uncovered case that matters is the clean external reread preserving both caret and viewport; it is the inline finding above. |\n| consistency with codebase and CLAUDE.md | none. No new production any/as/emoji issue; comments explain the why; the shipped plan documents the viewport behavior and its browser-only limitation. |\n| security | none. Both values are numeric localStorage convenience state, validated before use; malformed values cannot become paths, selectors, or executable content. |\n\nFINDINGS COMPLETE: I read every hunk of the diff and this is every finding I have.

src/components/FilesPane.vue
```diff
@@ -665,7 +719,14 @@ defineExpose({
     fileError.value = message;
   },
   /** What this pane looks like right now, for a host that will bring the user back here. */
-  snapshot: (): FilesPaneState => ({ openPath: openPath.value, expanded: expandedPaths(roots.value ?? []), showPreview: showPreview.value }),
+  snapshot: (): FilesPaneState => ({
+    openPath: openPath.value,
+    expanded: expandedPaths(roots.value ?? []),
+    showPreview: showPreview.value,
+    caret: placeNow().caret ?? undefined,
+    topLine: placeNow().topLine ?? undefined,
```
isamu: P3 — `snapshot()` still reads the place twice: `placeNow().caret` at line 726 and a second `placeNow().topLine` at line 727. That means the new “one value, one read” rule is not actually used at this boundary, and a future third place field has two call sites again; it also permits a custom/test editor whose position changes between reads to produce a mixed snapshot. Smallest fix: capture once (`const place = placeNow()`) and return `caret: place.caret ?? undefined` / `topLine: place.topLine ?? undefined`. I would drop this if the intended contract explicitly guarantees that `placeNow()` may be called repeatedly and all of its component reads are immutable, but that is not the one-value rule described by this commit.

isamu:
- P3: capture `placeNow()` once in `snapshot()`; the current two calls split the new one-value rule at the snapshot boundary.
- Targeted position/editor/storage tests: 68 passed.
- `yarn typecheck`, `yarn lint`, and `git diff --check` passed.

| Axis | Result |
|---|---|
| correctness and edge cases | none |
| storage layer | none |
| snapshot | P3 inline finding: `snapshot()` calls `placeNow()` twice instead of retaining one `FilePlace` value |
| shared editor double | none |
| tests | none |
| consistency with the codebase and CLAUDE.md | none |
| security | none |

isamu:
CODEX VERDICT: LGTM
- No inline findings.
- Targeted position/editor/storage tests: 68 passed.
- `yarn typecheck`, `yarn lint`, and `git diff --check`: passed.
- The shared `Pick<FilesPaneState, "caret" | "topLine">` shape does not create an unsafe application path: remembered placement is gated by the landed path and text result, while same-file refresh placement is gated by the request generation and same path.
- `placeNow()` correctly omits absent/null readings; valid persisted top lines are positive, so the truthiness checks do not discard a real position. `treeScrollTop: 0` is the correct snapshot for a missing tree element.

| Axis | Result |
|---|---|
| Correctness and edge cases | none — clamping, empty documents, changed-file placement, external refresh, teardown, and full-screen mounting are handled |
| Storage layer | none — both new fields are independently whitelisted, validated, and omitted when invalid; optional-property typing is respected |
| Snapshot and host write moments | none — the snapshot reads the place once, records tree scroll 0 when no tree exists, and the host snapshots on close/re-root/pagehide |
| Shared editor double | none — the literal is checked as `CmEditor`; the final cast only supplies spy return typing and does not weaken that literal check; the helper is excluded from the node-only project and included by the DOM test project |
| Tests | none — the layout stubs only make CodeMirror measurement deterministic in jsdom and do not claim to test physical scrolling. A reachable gap is one combined same-file refresh carrying both caret and top-line together, plus FilesPane-level verification of clamping a remembered place after the file becomes shorter |
| Consistency and plan | none — no production `any`/casts or emoji issue was introduced, comments explain the why, and the plan matches the shipped behavior |
| Security | none — the new localStorage values are numeric position metadata only; malformed values are dropped independently and no executable content is introduced |

isamu:
## Codex cross-review — rounds 6, 7 and 8 (tier C)

### Round 6 on `397ab4aa`: CHANGES REQUESTED, and it was round 2 again

A same-file re-read carried the caret and **dropped the top line** — so the thirty-second version check still returned a scrolled reader to the top, which is the reader the top line exists for. The same defect as round 2, one field later.

**The rule changed shape rather than growing another case.** Carrying field by field is what earned the same finding twice, so where a reader is in a file became ONE value — the cursor and what is on screen — with one way to read it and one way to put it back, used by both the restore and the same-file carry.

### Round 7 on `a0dd61b1`: one P3, and the fix improved the shape again

`snapshot()` read the place twice, splitting the one-value rule at exactly the boundary that rule exists for. Reading it once put the file **over `max-lines`** — and that pressure produced the better answer: a place is now `Pick<FilesPaneState, "caret" | "topLine">`, **the shape a snapshot already stores**, so neither end converts. `snapshot` spreads it; `applyRemembered` hands the remembered state straight to `goToPlace`, because a remembered state IS a place. Shorter than before the fix, and one fewer type to keep in step.

### Round 8 on `4e3b9805`: clean

**CODEX VERDICT: LGTM**, no inline findings, every axis answered, `FINDINGS COMPLETE`. Nothing was pushed during the round, so it covers the head that will merge. It checked the thing worth checking about the shared shape: a remembered place is gated by the landed path and the text result, a carried one by the request generation and the same path — there is no way in for a place that should not be applied.

**Two coverage gaps it named — not as findings — and deliberately not added**: a refresh case carrying caret and top line *together*, and a pane-level case for a remembered place clamped after the file got shorter. Both behaviours are covered separately (one carry case per field; the clamp has its own cases against a real CodeMirror), and adding tests for the combination would owe another round for coverage the reviewer did not call a defect.

### What the browser found that eight rounds of reading did not

Round 5's checkpoint named the remaining risk as browser-only behaviour. Driving one found something larger: **scrolling moves neither the selection nor the caret**, so a reader who never clicks had a caret on line 1 while reading line 130 — and this PR's restore put them back at the TOP. That is the case #2149 opens with, and no unit test could have seen it, because jsdom has no layout. Verified after the fix in the same browser: seeded with line 130 and reloaded, the pane came back with document line 130 at the top of the editor.

### Ledger

| 0 | pre | sandbox pre-flight | — | Measured earlier today in this working copy with the same flags: vitest, `yarn typecheck`, socket bind, api.github.com and `gh` all exit 0. **`rm` is blocked by policy** — name any mutation and I will run it |
| 1 | pre | claude, before round 1: `capped()` in filesPaneStore whitelists fields, so the two new ones worked per cell and vanished across a reload | FIXED | Carried and validated individually; a mutation removing the carry turns the store spec red |
| 2 | pre | claude, before round 1: seven specs had their own `fakeEditor` literal, and two new `CmEditor` methods broke all seven as a call to a missing function | FIXED | One typed double in `test/helpers/cmEditorDouble.ts`, excluded from the node tsconfig for the same reason `xtermDouble.ts` is |
| 3 | pre | claude: the Markdown preview's scroll cannot be read at all — `sandbox=""` means no `allow-same-origin` (opaque origin, so the app cannot read `contentWindow.scrollY`) and no `allow-scripts` (so the frame cannot report it) | DEFERRED | Making it readable is a security decision about the frame that renders `marked.parse()` output. Stated in the PR body, not silently skipped |
| 4 | 1 | P2 — a remembered `treeScrollTop` of 0 was skipped by a truthiness guard, so an in-place re-root left the PREVIOUS root's scroll on screen | FIXED | Two halves, because the guard alone does not cover a cell with nothing remembered: `teardown()` now resets the element (which outlives the root), and `restore` applies a remembered offset with `!== undefined`. Both are exercised — removing the teardown reset turns exactly the "nothing remembered" case red, the 0 case being covered by the guard |
| 5 | 1 | P3 — fractional caret coordinates were accepted from localStorage | FIXED | Measured rather than assumed first: CodeMirror does NOT refuse `line: 3.7` — it lands on a fractional offset and reads back `col: 0.5`, which is then what gets remembered. `asCaret` now requires integers (a corrupt value costs the caret), and `goTo` rounds whatever it is handed, so the editor's own contract holds for any caller |
| 6 | 2 | P2 — a same-path re-read (`checkForExternalChange` → `loadFile(path, true)`, and the conflict banner's Reload) passed no remembered state, so `setDoc` collapsed the selection and threw the reader to line 1 — every thirty seconds, in the app this PR exists to fix that for | FIXED | The caret is CARRIED across a read that stays on the same file, beside the view mode that already was. `keepsPreview` renamed `staysOnSameFile`, because it now decides two things and the old name described one of them. Break-verified: removing the carry turns the new refresh case red |
| 7 | 2 | claude, on codex's note that the shared double "does not reset/track its caret, so they cannot catch it" | FIXED | The double now MOVES the caret the way the editor does — `setDoc` collapses it to the top, `goTo` places it. A double that answered a fixed caret is what let this defect through a whole round with every spec green |
| 8 | 2 | claude, after the fix: `loadFile` went over the cognitive-complexity limit, and the first extraction put `FilesPane.vue` over `max-lines` | FIXED | The post-read step is one `restorePlace(pathRel, remembered, carried)` rather than an options object — three parameters, no interface, and the file lands under the cap with little room. The external-change watcher is the next thing that has to leave this file |
| 9 | 3 | codex round 3 on `bd52faa1`: LGTM, no inline findings, every axis answered, `FINDINGS COMPLETE`. Targeted position/editor/storage tests 54 passed, `git diff --check` clean | — | clean, on a head nothing was pushed to during the round. Its stdout came back EMPTY and the verdict was posted from inside the run — the comment URL is in the run's stderr, so the round happened and its record is on the PR |
| 10 | 4 | claude: the PR had produced NO CI run since `bd52faa1`, and the reason was not a queue — GitHub cannot build a merge ref for a CONFLICTING PR, so no `pull_request` workflow starts. main had moved under the branch | FIXED | `git merge origin/main` (rebase is forbidden here), resolved, pushed as `071cce37`. Mergeable again and CI running. Worth remembering: "no checks at all" on a PR means mergeability before it means Actions |
| 11 | 4 | claude: main gained `revealLine(line)` on `CmEditor` (#2140's search) while this branch added `goTo(at)` — two ways to put the cursor somewhere and scroll it into view | RESOLVED, not duplicated | One clamp now: `revealLine` is `goTo` at the line's start plus `view.focus()`. The focus IS the difference — a clicked search result means the reader wants to be in the file, a restore must not take the keyboard from wherever they are |
| 12 | 4 | claude: `filesSearchInPane.spec.ts` (from main) had its own hand-written editor double, which the merge broke as a call to a missing method | FIXED | Built on the shared double, keeping its own two implementations because that file asserts the ORDER of the calls. This is the second time the shared double has caught an interface change — the first was in this PR |
| 13 | 4 | P3 — unifying the two clamps changed main's `revealLine` from `Math.trunc` to `Math.round`, so a fractional line would resolve differently than it did before the merge | FIXED | Truncation is now the single rule, and the comment says why: a merge is no place to change what somebody else's caller observes. Nothing reaches it with a fraction on purpose (the store rejects a non-integer caret outright), so the choice only decides which arbitrary answer a corrupt value gets. A case pins `revealLine(3.7)` at line 3 — the answer main's version gave |
| 14 | 5 | codex round 5 on `24b6e2b1`: LGTM, no findings, checkpoint answered (no drift, no split, nothing present only as reviewer compliance). Named the largest remaining risk: browser-only behaviour — whether the real `scrollIntoView` and the post-render tree assignment land where intended | ACTED ON | Drove a browser to check it, and the check found something bigger than the risk named |
| 15 | 5 | claude, from that browser run: **scrolling moves neither the selection nor the caret**, so a reader who never clicks has a caret on line 1 while reading line 130 — and the restore put them back at the TOP. That is the case #2149 opens with, and no unit test could see it: jsdom has no layout | FIXED | The top visible line is remembered beside the caret and applied after it. Verified in the browser: seeded `topLine: 130`, reloaded, the pane came back with document line 130 at the top of the editor (`scrollTop` 4657, not 0) |
| 16 | 5 | claude: `createEditor` crossed the per-function line limit with the viewport methods in it | FIXED | Everything about WHERE — caret, viewport, and putting either back — is its own function beside it |
| 17 | 6 | P2 — a same-file re-read carried the caret but NOT the top line, so the thirty-second check still returned a scrolled reader to the top. The same defect as row 6, one field later | FIXED, and the RULE changed | Carrying field by field is what earned the same finding twice, so "where a reader is" is now ONE value (`FilePlace` = caret + top line) with one way to read it (`placeNow`) and one way to apply it (`goToPlace`). The restore path and the carry path both go through it, so a third field is added once rather than twice |
| 18 | 7 | P3 — `snapshot()` called `placeNow()` twice, splitting the one-value rule at that boundary | FIXED, and the shape improved | Reading it once was the obvious fix and it put the file over `max-lines`, which was the useful pressure: the place is now the SHAPE A SNAPSHOT ALREADY STORES (`Pick<FilesPaneState, "caret" \| "topLine">`), so neither end converts — `snapshot` spreads `placeNow()`, and `applyRemembered` hands the remembered state straight to `goToPlace` because a remembered state IS a place. Shorter than before the fix |
| 19 | 8 | codex round 8 on `4e3b9805`: LGTM, no inline findings, every axis answered, `FINDINGS COMPLETE`. It confirmed the shared shape creates no unsafe path: a remembered place is gated by the landed path and the text result, a carried one by the request generation and the same path | — | clean, on a head nothing was pushed to during the round |
| 20 | 8 | codex named two coverage gaps, explicitly NOT as findings: one refresh case carrying caret AND top line together, and a pane-level case for a remembered place clamped after the file got shorter | NOT ADDED, deliberately | Both behaviours are covered, separately: the two carry cases exist one per field, and the clamp has its own cases against a real CodeMirror. Adding tests for a combination of two tested things would owe another review round for coverage the reviewer did not call a defect — the loop is already eight rounds deep on a change whose original idea drew no finding |

PR #2169, opened 2026-09-19
refactor: lift the Files pane's directory tree into a composable (#2158)

isamu:
## Summary

**No behaviour change.** `FilesPane.vue` had been sitting at the repo's 600-line `max-lines` cap for
four consecutive changes, each paying for its own room — #2156 crossed it twice on the way in and
was rewritten both times to fit. That is the cap doing its job, and it is a bad way to decide a
design: the shape that fits is not always the shape that is right.

The tree is the biggest thing in the file and the least tied to the rest of it. It moves to
`src/composables/useFilesTree.ts`: the forest and its two companions (the error that replaces it,
and the request generation that decides which answer is still wanted), the listing fetch, the cache
paint, `loadRoot`, `toggleDir`, the visible-rows flatten and the node lookup.

**The markup stays in the pane**, so what renders does not move and the DOM is identical by
construction — the same trade `useFilesRowMenu` took in #2151.

## What this buys beyond the line count

**Three decisions became reachable.** `flattenRows`, `findIn` and `adoptListing` are exported and
pure; before, the only way to ask them anything was to mount a pane and read the DOM.

- *which rows are visible, and at what depth* — including a case a mounted test never produces: a
  FILE whose `expanded` flag is somehow true must not render its `children`
- *whether a path is in the forest*, *including inside a collapsed directory* — which is exactly
  what `restore()` depends on, since it expands remembered paths parents-first
- *what a fresh listing keeps from the tree it replaces* — the directory the user opened while the
  listing was in flight, and nothing else: not one that was never read, not one whose path is a file
  now, not one that has left the listing

**And one type instead of two.** `filesTreeCache.ts` called a listing row `CachedEntry` while the
pane called the same three fields `Entry`. One concept, one name: `ListingEntry`.

## Items to Confirm / Review

- **`tree.reset()` replaces three lines of `teardown()`** — clearing the forest, clearing the error,
  and bumping the tree's request generation. The generation bump used to sit with the pane's other
  two (`revealId`, `fileReqId`); it is now inside the composable, where the generation it guards
  lives. `teardown()` says so where the other two are bumped.
- **The pane reaches the tree through one object** (`tree.roots.value`, `tree.rows.value`,
  `tree.error.value` in the template). That is more verbose at the call sites than the bare refs
  were. The alternative — destructuring into the setup scope — would put the names back but hide
  where they come from, and this file has a lot of names.
- `FilesPane.vue` now has real headroom for the first time in four changes. The next candidate, if
  one is ever wanted, is the save-and-conflict half (`flush` / `save` / `discardAndReload` /
  `overwrite` / `onPageHide`), which closes over four refs and the editor.

## How it was verified

- **Every Files-pane spec runs unchanged** — `FilesPane`, `filesPaneTreeLoading`,
  `filesPaneTreeCache`, `filesPanePositions`, `filesPanePreview`, `filesFinderInPane`,
  `filesSearchInPane`, `filesRowMenu`, `FilesOverlay`, `gridOpenFiles`: green, along with every
  other spec under `test/src` (4131).
- **A differential harness, before and after.** It mounted the pane across a grid — a flat
  directory, an empty one, a listing that fails, a nested tree walked open and closed and open
  again, an expansion that fails, and a directory painted from the cache then replaced — each with
  and without a remembered state to restore, and with the root listing **held** so the in-flight
  state is observable. At every step it recorded the tree's rendered markup, every row with its
  indentation, and the sequence of listing requests. **The two captures compare identical.** The
  harness is deleted — half of it was the code it compared against.
- **What survives it** is `test/src/composables/useFilesTree.spec.ts`: the shapes the harness
  generated and the properties it compared, now asked of the three pure functions directly.
- `yarn format`, `yarn lint`, `yarn typecheck` clean.

## User Prompt

- 「FilesPane.vue分割は次のタスク」→「#2158」

## Related

- #2158 — the issue, with the inventory and the order. #2151 is the same method on the row menu.

work in mulmoterminal5

## Summary by CodeRabbit

- **Refactor**
  - Improved the internal organization of the Files pane’s directory tree while preserving its existing layout and behavior.
  - Directory expansion, loading, caching, error handling, and row rendering continue to work as before.

- **Tests**
  - Added coverage for directory-tree flattening, node lookup, listing updates, expansion state, and missing or changed entries.


test/src/composables/useFilesTree.spec.ts
```diff
@@ -0,0 +1,106 @@
+import { describe, it, expect } from "vitest";
+import { adoptListing, findIn, flattenRows, type TreeNode } from "../../../src/composables/useFilesTree";
```
isamu: P3 — The new spec covers only `flattenRows`, `findIn`, and `adoptListing`; it never exercises the composable’s async state machine. A reachable behavior-preservation case still lacks a regression test: clicking an unread directory whose `/list?path=...` request rejects must leave the directory collapsed and render no child rows (the old `toggleDir` catch explicitly did this). Smallest fix: add one mocked-fetch composable/pane test for that rejection/collapse path. What would change my mind: an existing spec or the deleted differential harness covering this exact child-expansion failure, but I found no such case in the repository.

isamu:
- P3 test-gap finding posted inline: the moved async `toggleDir` rejection/collapse path is reachable but untested.

| Axis | Result |
|---|---|
| behaviour preservation | none; old/new request-generation ordering, cache paint/removal, carry conditions, collapse-on-failure, and error clearing match |
| reactivity | none; `roots`, nested node mutation, and `rows` recomputation remain reactive |
| pure three | none; `flattenRows`, `findIn`, and `adoptListing` preserve the nested implementations |
| type merge | none; `ListingEntry` uses the same runtime validation shape as the former wire/cache types |
| tests | P3: no spec covers a rejected child-directory listing and verifies collapse/no child rows; see inline comment |
| consistency with codebase and CLAUDE.md | none; composable location, typing, comments, and no-production-`any`/`as` checks are consistent |
| security | none expected |

Checks: targeted Vitest 70/70 passed; `yarn typecheck` passed; `yarn lint` passed.

isamu:
CODEX VERDICT: LGTM
- No actionable findings in the full Round 2 diff.
- Focused verification passed: `useFilesTree.spec.ts` (21), affected FilesPane/cache specs (61), `yarn typecheck`, and `yarn lint`.
- The four new failure-path cases are non-vacuous: they load a real root, obtain a real directory node, drive `toggleDir`, and assert collapse, retry, tree-error isolation, and no refetch on close.
- Driving this composable outside a component is appropriate here: it uses no lifecycle/component-instance APIs, and the repo already directly tests state-only composables. Nothing in the production call path changes.
- One reachable case not directly asserted by the new standalone spec is a successful empty child listing (and its `loaded=true`/no-refetch consequence); existing pane behaviour remains covered, and this is not a behaviour finding.

| Axis | Result |
|---|---|
| behaviour preservation | none — request-generation ordering, cache paint/removal, carry conditions, collapse-on-failure, and error clearing match the old nested implementation |
| reactivity | none — `roots` remains reactive, `rows` remains computed, and Vue tracks the in-place `expanded` mutation through the reactive node graph |
| pure three | none — `flattenRows`, `findIn`, and `adoptListing` preserve the old decisions for pane-produced trees/listings; their exported typed inputs do not introduce a production path that changes pane behaviour |
| type merge | none — the wire guard and cache guard both validate the same three fields; `ListingEntry` is only the shared shape |
| tests | none — the added rejection cases pin the previously missing path and are not vacuous; see the empty-child-listing coverage note above |
| consistency with codebase and CLAUDE.md | none — the composable is in `src/composables`, production code has no new `any`/assertion, and comments explain why |
| security | none expected/found |

isamu:
## Codex cross-review — rounds 1 and 2 (tier C): clean

**Round 1 found no behaviour difference on any axis** — the request-generation ordering (it moved from a component-level `treeReqId` bumped by `teardown()` to the composable's own, bumped by `tree.reset()`), the cache paint and its removal on failure, `adoptListing`'s carry conditions, `toggleDir`'s collapse-on-failure, the error clearing, the reactivity of the in-place `expanded` mutation, the three pure functions, and the type merge all match the nested original.

One P3: the moved `toggleDir`'s **rejection path** was reachable and untested. Fixed in `5cb71732` with four cases that **drive the composable without a pane** — which the lift made possible, and which is the cheapest route to a failure path: a failed child listing leaves the row closed and unread, it is asked again on the next click, one directory failing is not the TREE failing, and a successful open closes again without re-asking. Commenting out the collapse turns two of them red.

**Round 2: LGTM**, no actionable findings, nothing pushed during the round — so it covers the head that will merge. It also answered the question I put in the prompt: driving this composable outside a component is appropriate here, since it uses no lifecycle APIs and the repo already tests state-only composables that way.

**One case carried forward rather than added**: a successful but EMPTY child listing, and its `loaded = true` / no-refetch consequence. Codex named it explicitly as not a finding. It is worth having — it is the same distinction this issue began with (#2148: "not read" is not "empty") one level down — but adding it now would owe a round for coverage the reviewer did not call a defect. The next extraction under #2158 touches these specs again; it goes in there.

### Ledger

| 0 | pre | sandbox pre-flight | — | Measured earlier today in this working copy with the same flags: vitest, `yarn typecheck`, socket bind, api.github.com and `gh` all exit 0. **`rm` is blocked by policy** — name any mutation and I will run it |
| 1 | pre | claude, before round 1: is the lift behaviour-preserving? | PROVED, not argued | A differential harness mounted the pane across a grid (flat / empty / failing / nested walked open-closed-open / failing expansion / cache paint then replacement), each with and without a restore, with the ROOT listing held so the in-flight state is observable, and recorded the tree's rendered markup, every row with its indentation, and the sequence of listing requests. Before and after compare identical. Harness deleted; its generator and properties are now `test/src/composables/useFilesTree.spec.ts` |
| 2 | pre | claude: the first version of the composable put `useFilesTree` over the per-function line limit | FIXED | The parts that are about NODES rather than about this pane's state came out as exported pure functions — which is also what made them testable |
| 3 | 1 | codex round 1 on `012efd9c`: no behaviour finding on any axis — request-generation ordering, cache paint and removal, carry conditions, collapse-on-failure, error clearing, reactivity of the nested mutation, the three pure functions, and the type merge all match the nested original. One P3: the moved `toggleDir`'s rejection path is reachable and untested | FIXED | Four cases added, driving the composable WITHOUT a pane — which is possible now and is the cheapest route to a failure path: a failed child listing leaves the row closed and unread, asks again on the next click, does not set the TREE's error, and a successful open closes again without re-asking. Break-verified: commenting out the collapse turns two of them red |
| 4 | 2 | codex round 2 on `5cb71732`: LGTM, no actionable findings, every axis answered. It also settled the question I asked about driving a composable outside a component — appropriate here, since it uses no lifecycle APIs and the repo already tests state-only composables directly | — | clean, on a head nothing was pushed to during the round |
| 5 | 2 | codex noted one case the standalone spec does not assert, explicitly NOT as a finding: a successful but EMPTY child listing, and its `loaded = true` / no-refetch consequence | NOT ADDED HERE — carried to the next extraction | It is the same distinction this whole issue began with (#2148: "not read" is not "empty"), one level down, so it is worth having — but adding it now owes a review round for coverage the reviewer did not call a defect. The next extraction in #2158 touches these specs again; it goes in there |

PR #2171, opened 2026-09-19
fix: the shipped launcher restarts a crashed server, as `yarn dev` already does (#2162)

### LINKED ISSUE #2162: fix: 本番の launcher だけクラッシュから復帰しない — yarn dev は再起動する
## What happens

**サーバがクラッシュしたとき、`yarn dev` は復帰しますが、`npx mulmoterminal` は復帰しません。** 守られているのが開発者だけで、実際のユーザーが守られていない、という向きになっています。

`scripts/dev-server.mjs` — 任意の exit で backoff つきに再起動する:

```js
child.on("exit", (code, signal) => {
  ...
  const plan = restartPlan({ code, signal, consecutiveFailures, minDelayMs, maxDelayMs });
  log(plan.reason);
  if (plan.retry) scheduleBringUp(plan.delayMs);
});
```

ファイル冒頭のコメントも、それが意図だと書いています。

> This supervisor instead RESTARTS the backend on any exit (with backoff)

`bin/mulmoterminal.js` — 子が死んだら親も死ぬ:

```js
server.on("close", (code) => {
  cancelReady();
  if (code === PORT_IN_USE_EXIT_CODE) { resolveExit(); return; }   // 75 だけは retriable
  if (code !== 0) { /* npx キャッシュのヒントを出す */ }
  process.exit(code ?? 1);
});
```

リトライするのは **exit code 75（EADDRINUSE）だけ**で、それ以外はすべて終端扱いです。

## なぜ効くか

ユーザーは tmux でセッションを永続化しているので、**サーバが落ちてもセッション自体は生きています**。にもかかわらず親プロセスまで死ぬと、ブラウザは繋ぎ先を失い、ユーザーは手で起動し直すことになります。

スマホから見ている場合はもっと悪くて、**手元に起動し直す手段がありません**。

## 提案

`scripts/dev-server-config.js` の `restartPlan` は既に「どの exit を再試行するか、どれだけ待つか」を純粋関数として持っています。本番の launcher からも同じものを使う。

再試行しない条件は dev と揃える必要はありません。本番では少なくとも:

- exit 75（ポート使用中）は今のまま即終了。取れないポートは待っても取れない
- 短時間に連続して落ちる場合は諦める（設定ミスで無限に再起動するのを防ぐ）

## 確認していないこと

- **実際にどのくらいの頻度でサーバが落ちているか**は測っていません。落ちたときの挙動の非対称性だけが確実です
- `restartPlan` をそのまま流用できるか、本番向けに条件を変えるべきかは見ていません

isamu:
## Summary

`npx mulmoterminal` now supervises the server it started. A crash is restarted with backoff; the
browser tab reconnects to the sessions tmux kept, instead of staring at a dead port.

`scripts/dev-server.mjs` has done this for `yarn dev` since #734. The shipped launcher did the
opposite — its `close` handler ran `process.exit(code ?? 1)` for every exit but a taken port — so
the developer was protected from a crash and the user was not, which is backwards: the user is the
one who cannot fix it, and from a phone there is no terminal to start it again in.

Reproduced before writing anything, on a scratch `HOME` and a free port: `kill -SEGV` the server
child and the launcher exits silently, `curl` → `000`, while tmux is still holding every session.

The decision is a pure module of its own, `bin/server-supervision.js`, not the dev one. The issue
asked whether `restartPlan` could be reused; it cannot, for two independent reasons:

- **`scripts/` is not in package.json's `files`.** An import of it from `bin/` resolves in this
  checkout and `ERR_MODULE_NOT_FOUND`s for everyone who installed the package — the exact failure
  the launcher already prints an npx-cache hint for.
- **The policy differs.** Dev retries EVERY exit forever because a file save is how the developer
  fixes whatever was wrong. There is no watcher here, and there are exits dev never sees.

`planAfterServerExit` answers with one of three actions:

| exit | action | why |
|---|---|---|
| exit 75 | `port-in-use` | unchanged — a port that is taken stays taken |
| exit 0 | `stop` | `mulmoterminal stop` and the browser's Stop button both SIGTERM the server, whose handler exits 0 |
| a signal that is not a crash signal | `stop` | `kill -9 <pid>` is what `bin/stop.js` prints, and the pid it prints is the server child's |
| never bound in this launcher's lifetime | `stop` | nothing was serving, and the next boot fails the same way |
| too many failures in a row | `stop` | bounded, so a server that binds and dies at once cannot respawn forever |
| anything else | `restart` | with exponential backoff off the consecutive-failure count |

**Windows is excluded, and that is the one thing this knowingly leaves undone.** Node has no real
signals there, so `mulmoterminal stop`, the browser's Stop button and a `taskkill` all TERMINATE the
server rather than delivering something its handler runs — `bin/stop.js` says so in its own note. So
the server never reaches the `exit(0)` that means "somebody asked for this", and the exit arrives as
a bare non-zero code with no signal: the exact shape of the crash the restart exists for. Windows
would have got a `stop` that reports success and a server that comes back, with no way to end it at
all. It keeps the behaviour it had, saying nothing new. Reaching it properly needs `stop` to be able
to signal the LAUNCHER and not only the server it spawned — a change to the instance registry and to
`stop` itself, and a separate PR. **Say if you would like the follow-up issue filed.**

The count resets when the server reports `{ type: "listening" }` — the message the launcher already
reads to decide where to poll. Deliberately not elapsed time: #1735 is the case where "it stayed up
N seconds" read every crash as a one-off, because the server does its whole setup before it binds.

## Items to Confirm / Review

0. **The Windows exclusion above** — found by reading the diff back rather than by a bot, and the
   only claim in this PR I could not run. The evidence is `bin/stop.js`'s own note plus the fact
   that no handler runs, so the exit cannot be `0`; the gate does not depend on which non-zero code
   Windows actually produces.
1. **SIGKILL is treated as a stop, not a crash — please sanity-check the trade.** An OOM killer
   sends SIGKILL too, so that crash is not recovered. The reason for choosing this way: `kill -9`
   on the server pid is a documented escape hatch this product PRINTS, and a stop that does not
   stop is a worse failure than a crash that is not recovered. Reversible in one line if you
   disagree — `CRASH_SIGNALS` in `bin/server-supervision.js`.
2. **"It never bound, so do not restart it" is the rule doing the most work.** It is what keeps a
   bad config or a half-unpacked npx cache out of a respawn loop, and what keeps the npx-cache hint
   at the end of the output instead of under repeats of one stack. It is a launcher-lifetime fact,
   not a per-child one, so a transient failure on the FIRST restart after a good lifetime still
   gets retried.
3. **The banner reprints on a restart; the browser is not reopened.** Deliberate — the tab is
   already open and reconnecting, and the banner is how the terminal says the URL is live again.
   Say if you would rather have a quieter line there.
4. **The choice of backoff floor, ceiling and failure cap.** Judgement, not measurement. They are
   named constants at the top of `bin/server-supervision.js`, and a spec pins the relationship
   between them (see below).
5. **`PORT_IN_USE_EXIT_CODE` moved** out of `bin/mulmoterminal.js` into the new module, which the
   entry point imports. The source-text guard in `test/server/infra/server-exit.spec.ts` followed
   it, and now asks the whole of `bin/` how many copies of the number exist rather than naming two
   files — a third copy is how the pair drifted before.
6. **One bug fixed on the way in, worth a look because it only bites now.** The 20s
   address-fallback timer was never cleared. Left armed, a dead child's timer fires after a restart
   and announces the NEW child as if it were the old one — banner, browser and all.

## Interaction with #2167, which landed on main during this

The idle sweep at boot now comes from `startReapSchedule`, and a supervised restart runs it the same
way a manual restart always has: at that moment none of this server's ptys hold anything, so the
sweep is at full strength. It only ends sessions idle past the configured threshold, and a second
sweep finds nothing the first left, so a crash loop does not compound it. No files overlap; CI
builds the merge.

## The ceiling was dead code, and a mutation sweep is what said so

Every decision in `planAfterServerExit` was inverted in turn against the new spec. All but one went
red. The exception was DELETING the clamp on the backoff delay: it stayed green, because the
doubling could never reach the ceiling before the failure cap stopped it — the line executed and
decided nothing. Lowering the ceiling so the last allowed attempt reaches it makes it live, and the
spec now pins the relationship between the three constants rather than only the delays they
produce, so raising the ceiling or lowering the cap says so instead of quietly returning the line
to decoration. With that, every mutation goes red.

## Verified

Against a real server, on a scratch `HOME` and a free port, with a fake `open` earlier on `PATH` so
browser launches are countable:

- **Crash after serving** (`kill -SEGV`): the launcher survives, names the signal, comes back, and
  the port answers 200 again. `open` was called once across both lifetimes.
- **tmux's promise holds across it.** Both boots report the same surviving-session count and
  `reattach on connect`; nothing was reaped by the restart.
- **`mulmoterminal stop`** still stops it — exit 0, no restart — and the port goes quiet.
- **SIGTERM to the launcher** takes the server with it. **`kill -9` of the server child** ends the
  launcher rather than resurrecting the server, and says why.
- **A server that cannot boot at all** (an invalid `keymap`, which exits before binding) stops on
  the first failure, with the server's own reason directly above the launcher's.
- **A crash loop** (crash it, then make every later boot fail): the delay doubles and then holds at
  the ceiling, then a give-up naming the count, and the launcher exits.
- **The port stolen inside the restart window**: the fresh child exits 75 and the launcher prints
  the same port-in-use report as before — which is how that branch is reachable in practice at all.

`yarn lint`, `yarn typecheck` and `yarn test` are clean on this branch.

And by CI, which is the part I could not do here: `package-smoke` packs the tarball, installs it
into a clean directory and boots `node_modules/.bin/mulmoterminal`. A `bin/server-supervision.js`
missing from the published `files` would `ERR_MODULE_NOT_FOUND` there — so that job passing is the
evidence that the new module really ships, which is the whole reason it is not in `scripts/`.

Every decision also survives a mutation sweep (above). What was NOT exercised by running it:
Windows — which is why it is excluded rather than assumed — and Linux, where the signal semantics
the gate relies on are the same as macOS's. The run above is macOS; the CI matrix covers the build
and the suite on all three.

## User Prompt

- https://github.com/receptron/mulmoterminal/issues/2162

Closes #2162

Design notes: [`plans/fix-2162-launcher-crash-restart.md`](plans/fix-2162-launcher-crash-restart.md)

work in mulmoterminal3


isamu:
## Codex cross-review — round 1 (post-merge), tier C

Run after the merge, so the loop's merge gate does not apply: findings become a follow-up PR. Sandbox pre-flight (ledger row 0): vitest exit 0, `vue-tsc -b` exit 0, socket bind exit 0, `api.github.com` exit 0; `rm -f` policy-rejected, so no mutation sweep was delegated.

**Verdict: `CODEX VERDICT: LGTM`**, with `FINDINGS COMPLETE`, all ten axes answered, and one severity-N (prose) finding.

<details><summary>Prompt sent (verbatim)</summary>

```text
Review the change merged as PR #2171 in receptron/mulmoterminal. Work in the checkout you are in
(/Users/isamu/ss/llm/mulmoterminal3, currently on main). The diff is:

    git diff f4075809a7b105f308b17d7c368dbc32eab9543b..d25f79ec7571ed7910c7585515d58b444a76a7b9

Files: bin/server-supervision.js (new), bin/server-supervision.d.ts (new),
bin/mulmoterminal.js, test/bin/server-supervision.spec.ts (new),
test/server/infra/server-exit.spec.ts, README.md, docs/ChangeLog.md,
plans/fix-2162-launcher-crash-restart.md.

WHAT IT DOES: `npx mulmoterminal` used to exit whenever its server child exited (except exit 75).
It now supervises that child — restarting it on a crash with backoff, and stopping for the exits
that mean somebody asked it to stop.

DO NOT POST ANYTHING to GitHub and DO NOT apply any fixes. Reply as plain text on stdout only; I
will post this exchange to the PR myself. The PR is ALREADY MERGED, so anything you find becomes a
follow-up PR — which means a real finding is still worth exactly as much, and a speculative one
costs a whole extra PR rather than a commit.

Every finding you hold back costs a full extra round, and here it costs a whole follow-up PR.
Read the WHOLE diff before you answer. Do not stop at the first problem. Do not save anything
for "a follow-up review".

── TIER C. ANSWER EVERY AXIS, even when the answer is "no findings" ─────────
An axis with no findings is a RESULT and has to be stated. "No findings" on an axis you did not
look at is indistinguishable from a clean axis until it bites someone.

  1. correctness and edge cases in `planAfterServerExit` and `superviseServer`
  2. PROCESS LIFECYCLE AND ASYNC ORDERING. The supervisor is recursive and awaits a sleep between
     lifetimes. Look at: the timers (`cancelReady`, the newly-added `clearTimeout(fallbackReady)`),
     what keeps the event loop alive during the sleep, the SIGINT/SIGTERM `shutdown` handler
     firing while no child exists or while one is being replaced, and whether any handler from a
     DEAD child's closure can still fire and act on the LIVE one.
  3. RETRY / REPLAY IDEMPOTENCE — the axis I most want a second opinion on. A restart re-runs the
     server's whole boot, and that boot has DESTRUCTIVE side effects: `sweepIdleSessions` /
     `startReapSchedule` (ends idle tmux sessions), `pruneOrphanSettings` and `pruneOrphanDrops`
     (delete files, one of which may hold a provider API token), `registerInstance`,
     `wireMachineGlobalHooks`. See the `server.listen` callback in server/index.ts. Question: can
     an automatic restart — or a crash LOOP of up to the cap — destroy something a manual restart
     would not? Is each of those sweeps idempotent across rapid repeats? Is there an ordering
     where the dying child's `process.on("exit")` unregister races the new child's
     `registerInstance`?
  4. THE STOP CONTRACT. Every way a user can stop this thing must still stop it. Enumerate them
     from the code, not from my list: `mulmoterminal stop` (bin/stop.js), `POST /api/shutdown`
     (server/routes/shutdown-routes.ts), Ctrl+C, SIGTERM to the launcher, `kill -9` of the child,
     closing the terminal (SIGHUP), a machine going to sleep. For each, say whether the launcher
     restarts the server, and whether that is the right answer.
  5. THE WINDOWS CLAIM (behaviour preservation). The code asserts that on win32 a stop and a crash
     are indistinguishable, and therefore excludes win32 entirely, claiming Windows keeps exactly
     its pre-PR behaviour. Check BOTH halves against the code: is the exclusion correct, and is
     "exactly its pre-PR behaviour" true — compare what the old `close` handler did with what
     `stopForServerExit` does now for a win32 exit. I could not run Windows; you may not be able
     to either — say so rather than guessing.
  6. tests: does a test go RED if the new behaviour breaks? Name any decision in
     `planAfterServerExit` that could be inverted with the suite staying green. (Do not run a
     mutation sweep — `rm -f` is refused by your sandbox and a failed restore corrupts every count
     after it. Name the mutation and I will run it.)
  7. API / wire / on-disk compatibility: the `{ type: "listening", address, localhostIsOurs }` IPC
     contract with server/index.ts, the exit-code contract with server/infra/server-exit.ts, and
     the ~/.mulmoterminal/instances registry across a restart.
  8. consistency with the rest of the codebase, including this repo's CLAUDE.md rules (read it:
     no `any`, no `as`, no magic numbers, functions under 20 lines, comments explain WHY not WHAT,
     never a figure in prose).
  9. security: does the restart open a window in which something else can take the port, or in
     which a second instance can start? Does anything in the new code trust text it should not?
 10. i18n / a11y: say "n/a and why" if that is the answer — this is launcher stdout, not the UI.

   N is for a finding whose only remedy is editing PROSE (a comment, a plan file, the PR body) and
   whose sentence, if believed, would NOT cause anyone to make a wrong change. N does not change
   the verdict. A prose sentence that WOULD cause a wrong change is P3 or higher.
4. WHAT WOULD CHANGE YOUR MIND — the fact that, if true, makes it a non-issue.

── ALREADY SETTLED — do not re-derive; DISPUTE with evidence if any is wrong ─
| # | iter | finding / decision | disposition | why / where |
| 0 | 0 | SANDBOX PRE-FLIGHT | OK | vitest exit 0, vue-tsc exit 0, socket bind exit 0, api.github.com exit 0. `rm -f` policy-rejected but plain `rm` works — so do NOT run a mutation sweep; name any mutation and I will run it. |
| D1 | pre | SIGKILL is treated as a deliberate stop, not a crash | DECIDED | `bin/stop.js` `manualStopCommand` PRINTS `kill -9 <pid>` for a stubborn server, and the pid it prints is the SERVER CHILD's. Restarting on SIGKILL defeats that escape hatch. Cost accepted: a Linux OOM-killer SIGKILL is not recovered. |
| D2 | pre | Windows gets no supervision at all | DECIDED | Node has no real signals on win32; `bin/stop.js` line 60 note says it "terminates the process outright rather than running its handler", so the server cannot reach exit(0) and a stop is indistinguishable from a crash. Windows keeps pre-PR behaviour, reason null. NOT executed on Windows — this is the one claim I could not run. |
| D3 | pre | a server that never bound is not restarted (`everServed`) | DECIDED | Nothing was serving, so no browser session to restore, and the same boot fails the same way. Keeps a bad config / half-unpacked npx cache out of a respawn loop and keeps the npx-cache hint at the end of the output. Launcher-lifetime fact, not per-child. |
| D4 | pre | the ready banner REPRINTS on a restart, but the browser is not reopened | DECIDED | The user's tab is still open and reconnecting; the banner says the URL is live again. Verified by a fake `open` on PATH: exactly one invocation across two lifetimes. |
| D5 | pre | backoff ceiling lowered so the last allowed attempt reaches it | DECIDED | At the previous value the clamp was unreachable and a mutation DELETING it stayed green. A spec now pins the relationship between floor, ceiling and cap. |

That table is context so you do not spend this round re-deriving what is already decided. It is
NOT a list of closed topics: if a decision is wrong, or a row's reasoning does not hold, say so and
mark it "REOPENING #<row>" with the specific evidence. Silence on a bad decision is worse than a
repeat finding. D2 in particular is the one claim nobody executed.

── END WITH, IN THIS ORDER ──────────────────────────────────────────────────
  Line 1, on its own line:
  Then the axis table: one row per axis 1-10 above, findings or 'none'.
  Then, on its own line:
```

</details>

<details><summary>Codex reply (verbatim)</summary>

```text
CODEX VERDICT: LGTM

| Axis | Result |
|---|---|
| 1. `planAfterServerExit` / `superviseServer` correctness | none. The stop/restart/port-in-use branches match the intended contract; backoff cap and `everServed` behavior are pinned. |
| 2. Process lifecycle / async ordering | none. `sleep()` is a ref’ed timer, so the event loop survives backoff; Ctrl+C/SIGTERM during backoff exits; `cancelReady()` and `clearTimeout(fallbackReady)` prevent dead-child readiness callbacks from acting after replacement. I did not find a dead child closure that can affect the live child. |
| 3. Retry / replay idempotence | none. `sweepIdleSessions`, `startReapSchedule`, `pruneOrphanSettings`, and `pruneOrphanDrops` are idempotent across rapid repeats; restart replays what a manual restart would replay. `registerInstance` writes per-pid, and a dying child’s `exit` unregister cannot delete the new child’s file. |
| 4. Stop contract | none. `mulmoterminal stop`: SIGTERM to server, server exits 0, launcher does not restart. `POST /api/shutdown`: delayed SIGTERM to server process, exits 0, no restart. Ctrl+C / SIGTERM to launcher: launcher SIGTERMs current child and exits, no restart. `kill -9` child: treated as stop by settled decision D1, no restart. SIGHUP / closing terminal: no launcher handler; terminal/process-group shutdown stops rather than restarts, which preserves “closing the terminal stops it.” Machine sleep: no exit by itself, so no restart; if the OS later kills the child, behavior follows the resulting signal/code. |
| 5. Windows claim | none from code review. I could not execute Windows. Code comparison supports both halves: `planAfterServerExit(... platform: "win32")` stops for non-75 exits with no new reason, and `stopForServerExit` exits with `code ?? 1`, matching the old close handler’s externally visible behavior for ordinary non-75 exits. Exit 75 still reaches the port-in-use path. |
| 6. Tests | none for `planAfterServerExit`: I did not find a decision in that pure function that can be inverted while this suite stays green. I ran `yarn test test/bin/server-supervision.spec.ts test/server/infra/server-exit.spec.ts`: 36 passed. Integration coverage of `superviseServer`/timers is lighter than the pure plan coverage, but I do not have a concrete bug from that. |
| 7. API / wire / on-disk compatibility | none. `{ type: "listening", address, localhostIsOurs }` remains additive and compatible; exit code 75 remains pinned against `server/infra/server-exit.ts`; instance registry entries remain per-pid and are cleaned/filtered by existing stale-entry logic across restarts. |
| 8. Codebase / CLAUDE.md consistency | N note: stale prose only. Every site: `server/infra/server-exit.ts:4` and `server/infra/server-exit.ts:13` still say the 75 contract is with `bin/mulmoterminal.js`, while the launcher-side literal now lives in `bin/server-supervision.js`. Smallest change: update those two comments to name `bin/server-supervision.js` or “the launcher-side supervision module.” What would change my mind: if the project intentionally treats `bin/mulmoterminal.js` as the umbrella launcher contract even when the literal moved; then this is just imprecise wording, not stale guidance. |
| 9. Security | none. Restart creates a backoff window where another process can take the port; the next child exits 75 and the launcher stops rather than talking to the wrong process. A second launcher can start during that window, but then the original loses the bind and stops; I did not find a path to two supervised children sharing one port. New code does not trust child text beyond bounded stderr diagnostics and shape-checked IPC. |
| 10. i18n / a11y | n/a. This is launcher stdout/stderr and process behavior, not UI copy or an accessible interactive surface. |

```

</details>

isamu:
Working on this in `mulmoterminal3`.

- started — 2026-09-19 07:55 UTC
- PR #2175 — 2026-09-19 07:57 UTC

<sub>posted by [MulmoTerminal](https://github.com/receptron/mulmoterminal)</sub>

isamu:
## Cross-review triage — round 1, tier C

One clean round on the merged head: `CODEX VERDICT: LGTM`, `FINDINGS COMPLETE`, ten of ten axes
answered, nothing pushed to the reviewed head. By the protocol's bar the loop exits here.

Its only finding was severity N — prose — and N does not change a verdict. Checking it turned up a
second error in the same comment block that it did not name, and that one would cause a wrong
change if believed. Both are fixed in #2175.

| # | finding | severity | disposition | where |
|---|---------|----------|-------------|-------|
| 1 | `server/infra/server-exit.ts` names `bin/mulmoterminal.js` as the launcher side of the 75 contract; the literal moved to `bin/server-supervision.js`, which is also what the spec reads | N (Codex) | FIXED | #2175 |
| 2 | the same block says 75 makes the launcher "retry on a fresh one" — it prints who holds the port and exits; the fresh-port offer belongs to the probe path, with consent (`pickPort`, #611) | P3 (observed during Claude review, **not** flagged by Codex) | FIXED | #2175 |

### What the round confirmed rather than found

Recorded because a "no findings" on an axis nobody names is worth nothing later:

- **Retry / replay idempotence** — the axis I most wanted a second reader on. A restart replays the
  server's whole boot, and that boot deletes things: `sweepIdleSessions`, `pruneOrphanSettings`
  (one of which may hold a provider token), `pruneOrphanDrops`. Codex's reading agrees with mine —
  idempotent across rapid repeats, and a restart replays only what a manual restart already
  replayed. A dying child's `exit` unregister cannot remove the new child's registry entry.
- **The stop contract**, enumerated from the code rather than from my list: `mulmoterminal stop`,
  `POST /api/shutdown`, Ctrl+C, SIGTERM to the launcher, `kill -9` of the child, SIGHUP, machine
  sleep. Every one still stops it. I exercised the browser Stop button live after the merge — the
  one path the PR had not run — and it stops with a single boot attempt and no restart.
- **Tests** — Codex could not name a decision in `planAfterServerExit` invertible with the suite
  green, which is the same answer the pre-merge mutation sweep gave.
- **The Windows claim** — neither reviewer could execute Windows. Codex's code comparison supports
  both halves (a win32 non-75 exit stops with `code ?? 1`, matching the old close handler; 75 still
  reaches the port-in-use path). It stays the one unexecuted claim in this change.

### Sandbox pre-flight (ledger row 0)

`vitest` exit 0, `vue-tsc -b` exit 0, socket bind exit 0, `api.github.com` exit 0. `rm -f` is
policy-rejected, so no mutation sweep was delegated — the pre-merge sweep was run here instead.

PR #2172, opened 2026-09-19
feat: sweep idle sessions on a timer, not only at boot (#2165)

### LINKED ISSUE #2165: feat: detached な tmux セッションを定期的に掃除する設定（起動時スイープだけでは長時間稼働で溜まる）
## 観測した事実

11日間動かし続けたサーバで、tmux セッションが36個。うち **detached が24個**で、
最終活動は 1.0〜1.3 日前。中身はどれもアイドルな `zsh` のみで、合計RSSは36MB程度だった。

```sh
tmux -L mulmoterminal list-sessions -F '#{session_name}|#{session_attached}|#{session_activity}'
```

設定は `sessionIdleReapDays: 7`（既定値）。

## なぜ掃除されないか

`server/session/reap-idle-sessions.ts:1` と `common/sessionReap.ts:8` のとおり、
このスイープは **サーバ起動時に一度きり** しか走らない（#1467 / #1486 で入った形）。

そのため長時間動かすサーバでは、起動後に detached になったセッションが
次の再起動まで一切掃除されない。再起動が掃除の唯一の契機になっている。

加えて `reapableTmuxSession`（`server/infra/tmux.ts:648-656`）は

```ts
if (liveHere || attachedCount !== 0) return false;
```

で弾くため、サーバが pty を保持し続けているセッションは経過日数に関係なく対象外になる。

## 提案している機能

detached セッションを **定期的に** 掃除する設定。起動時スイープと同じ述語を、
タイマーからも呼べるようにする形を想定している。

想定しているキー（名前は要検討）:

- `sessionReapIntervalHours` — スイープを回す間隔。`0` で無効（起動時のみ＝現状の挙動）
- 既存の `sessionIdleReapDays` はしきい値としてそのまま使う

既定は現状維持（無効）にしたい。勝手にセッションが消える挙動は、設定で明示的に
有効化したときだけ起きるべきだと考えている。

## 確認したいこと

- 定期スイープを既定で有効にすべきか、無効のままにすべきか
- `liveHere` の扱い。detached かつ長期間アイドルなら pty を保持していても対象にするのか、
  それは別の判断として残すのか（安全側に倒すなら現状維持）
- `sessionIdleReapDays` より短い「detached 専用のしきい値」が要るか。
  detached は attached と違って放置されている蓋然性が高いので、別のしきい値が妥当かもしれない

## 関連

- #1467 cleanup-orphans が自動呼び出しされず、再起動後のゾンビ tmux セッションが蓄積する
- #1486 使われていないセッションを起動時に終了する

本issueは #1486 で入った起動時スイープの続きで、「起動時だけでは足りない」という部分にあたる。

isamu:
## Summary

The idle-session sweep only ever ran at server start, so a host that is never
restarted never ended a session — which is what #2165 reports. This adds
`sessionReapIntervalHours`: a timer that runs **the same sweep** while the server
is up.

- **`sessionReapIntervalHours`, default 6, `0` = at start only.** Whole hours,
  0–168, sanitized in `common/sessionReap.ts` beside the day bounds so the Settings
  stepper offers the range the server clamps to.
- **Nothing about WHICH sessions go changes.** `reapableTmuxSession` is untouched:
  still nothing attached, nothing a pty of ours holds, nothing tmux declined to
  answer for. `sessionIdleReapDays` remains the threshold and `0` still turns the
  whole sweep off.
- **The threshold is read per tick, the cadence once at boot.** So the stepper in
  Settings lands without a restart, while the timer is armed once — the existing
  convention for a scheduled thing (`getWorklogConfig`, `getSystemTaskSwitches`).
- **A tick drops the files of what it ended.** A session settings file can hold a
  provider's API token, and the only thing that removed one was the next boot's
  orphan prune. Guarded by the same session-id check the boot prunes make, because
  the sweep ends unparseable ids on purpose (#1533) and the cleanup builds a path
  out of the id it is handed.
- **Settings gets a second stepper**, and a doomed row now says "ends on the next
  sweep" rather than "ends at next start" while a timer is armed — that phrase was
  the one thing a row said that a running server makes false.
- **`server/index.ts` was at its `max-lines` budget exactly**, so the tmux boot
  block moves to `server/session/session-upkeep.ts`: the extraction
  `scheduler-boot.ts` made for the same reason. Boot behaviour is unchanged; it
  now answers with the surviving set the orphan prunes already decided from.

Docs: both `config.md` guides and the `mulmoterminal-config` skill (which owns
`sessionIdleReapDays`, so it owns this too).

## Items to Confirm / Review

1. **The default is ON.** Decided with the issue author after the issue proposed
   the opposite. The argument for on: the predicate does not move, so for an
   existing user the only change is that a session already destined for the sweep
   is ended at the next tick instead of the next restart. The argument against: a
   user upgrading gets sessions ending without touching a setting. Worth a second
   opinion.
2. **The per-id cleanup is the one thing here that is not strictly "run the sweep
   on a timer."** It is included because without it the feature trades a pile of
   tmux sessions for a pile of token-bearing files on exactly the long-running host
   this is for. `POST /api/tmux/cleanup-orphans` still has that gap and is **not**
   changed here — say if it should be.
3. **Two of the issue's three open questions were answered "keep it as it is"** and
   are worth confirming: no separate detached-only threshold (every session
   `reapableTmuxSession` accepts is *already* detached — it requires
   `attachedCount === 0` — so a second threshold would be a second name for the one
   that exists, and "detached since" would need `#{session_last_attached}`, which
   nothing reads today); and `liveHere` keeps its veto, so a cell you have open is
   never ended however long it has been quiet.
4. **The boot extraction is a refactor inside a feature PR.** It is forced — the
   file cannot take another code line — but it is the piece to read with the
   original beside it.

## How it was verified

- `yarn lint`, `yarn typecheck`, `yarn build`, `yarn test` are clean on this branch.
  (`yarn format` fails on `docs/_layouts/default.html`; that is **pre-existing on
  `main`** — reproduced with this branch stashed — and every file this PR touches
  passes `prettier --check`.)
- One suite run under a load spike reported `copilot-resume-cwd.spec.ts` and
  `remoteHost/listCollectionProjects.spec.ts` red. **Both were re-run standalone
  with raised timeouts and pass**; neither goes near session reaping, and both were
  green in the earlier full runs on this branch. Flagged rather than hidden.
- New specs: the sanitizer and its off switch; the timer arming and declining to
  arm; the tick re-reading the threshold rather than capturing it; the tick
  cleaning up what it ended and skipping an id that is not a session id; the boot
  line for both states; and, in the Settings section, that the new stepper writes
  its own field, does **not** trigger the threshold's list reload, that a doomed
  row's wording follows the cadence, and that the cadence row goes quiet (and its
  stepper disabled) when the threshold above is off — a cadence promising a sweep
  would otherwise contradict the row that just said "never".
- **The boot path was run, not reasoned about** (it is the wide blast radius here).
  A server started against a scratch `HOME` printed the persistence line and the
  new sweep line in **both** states, served the UI, and round-tripped the new key
  through `GET`/`POST /api/config` including clamping an out-of-range value. The
  scratch config had `sessionIdleReapDays: 0` throughout, so no live tmux session
  could be ended by the check itself — and none was.
- **`unref` is the one thing an injected-timer spec cannot see**, so the production
  `startIdleSessionSweepTimer` was armed in a throwaway harness and the process was
  watched to exit immediately instead of waiting out the interval.

## What was not exercised

- **A real tick has not been observed end to end.** The minimum cadence is one
  hour, and the sweep runs against the hardcoded `mulmoterminal` tmux socket — so
  forcing a real tick would have swept this machine's live sessions, which was not
  mine to do. The tick's behaviour is covered by specs with an injected timer, the
  hours-to-ms conversion by its own spec, and the real `setInterval`/`unref` wiring
  by the harness above.
- **No browser render.** There is no browser driver in this checkout, so the new
  stepper is covered by mounting the real component rather than by a screenshot.

## User Prompt

- Work issue https://github.com/receptron/mulmoterminal/issues/2165 — a long-running
  server accumulated detached tmux sessions, because the idle sweep only runs at start;
  add a setting that sweeps periodically.
- On the issue's three open questions: make the periodic sweep **enabled by default**
  (e.g. every 6 hours); do **not** add a detached-only threshold — reuse
  `sessionIdleReapDays`; leave **`liveHere` as it is**, so a session this server holds
  a pty for stays out of scope.

Closes #2165

work in mulmoterminal4


isamu:
## Closing: superseded by #2167

This PR duplicates work that was already merged. #2167 (`feat/2165-periodic-tmux-reap`)
implemented the same issue and merged at 07:30Z; this PR was opened at 07:32Z, two
minutes later. The conflict against `main` is the symptom, not the problem.

The two implementations arrived at the same design independently — same config key
`sessionReapIntervalHours`, same constant names, same 0–168 bound, same sanitizer shape,
same `reapIntervalMs` / `reapTimerEnabled` split in `common/sessionReap.ts`. #2167 is the
one that shipped and is the one to keep.

**How this happened, so it does not again:** the repo's collision check is two commands —
`gh pr list --state open` and `git log origin/main -- <file>`. Only the second was run
here. #2167 had been open for thirty-five minutes when this branch was cut, so the first
command would have caught it before any work started.

### What was not duplicated

Four things in this branch are not in #2167. Each is being taken to its own PR against
the merged implementation rather than carried here:

| | #2167 (merged) | here |
|---|---|---|
| default | `0` (off) | `6` (on) |
| Settings control | config file only | a stepper beside the list it acts on |
| files of a session the TIMER ended | not dropped | dropped (one can hold a provider API token) |
| a doomed row's wording with the timer armed | `ends at next start` | `ends on the next sweep` |

The **default stays `0`** — decided against changing it, so that row is dropped rather
than carried. The other three become follow-ups.

The boot-block extraction to `server/session/session-upkeep.ts` is dropped too: it existed
to fit under `server/index.ts`'s `max-lines` budget, and #2167 solved the same constraint
its own way with `server/session/reap-schedule.ts`.

PR #2174, opened 2026-09-19
refactor: lift the Files pane's open file into a composable (#2158)

isamu:
## Summary

The second and largest of the extractions #2158 asks for. The tree went out in #2169; what
remained in `FilesPane.vue` was mostly the **open file** — the buffer, the editor it is shown in,
how it is read, where the reader was in it, and every way it is written back. That moves to
`src/composables/useOpenFile.ts`.

**The markup does not move**, as in #2151 and #2169, so the rendered DOM is identical by
construction. The tree, the finder, the reveal, the row menu and the row actions stay in the pane
— they are its own ends: they read its props or they emit.

`FilesPaneState` gets its own module (`src/components/filesPaneState.ts`). Three parties decide
from it and none of them owns it — the pane fills it, `TerminalGrid` files it per cell, and
`filesPaneStore` puts it in localStorage per directory — so it had outlived the component it was
declared in, and a composable importing it from the `.vue` would have been a cycle. Its two
importers now name the module instead, which also retires `filesPaneStore.ts`'s entry on the
eslint list of files that inherit a `.vue`'s type-resolution blind spot.

The `pagehide` listener and the external-change watch went with the file rather than staying with
the pane: both exist so a buffer is not lost and so a file changed underneath is noticed, and
leaving them behind would have split one rule across two files.

The functions in the new module are module-level and take an explicit context. `max-lines-per-function`
is 60 and all of this inside one closure would be several times that — and the shape the limit
forces is also what makes each function reachable from a test with a state it made itself.

## Items to Confirm / Review

- **The behaviour-preservation claim, and how it was proved.** A throwaway differential harness
  mounted the pane before and after and compared whole outputs: five scenarios (a plain file, one
  read in Preview, one the server will not serve as text, a save that conflicts, a read that
  fails), each with and without a restored state, recording the header and body HTML, the emits,
  the snapshot, the request sequence and the editor calls after every step. Every record came back
  identical. The harness is deleted — half of it was the code it was proving.
- **The lifecycle move.** The composable registers the `pagehide` listener and the disk watch in
  its own `onMounted`, and undoes both in `onBeforeUnmount`. Hooks run in registration order and
  `useOpenFile` is called before the pane registers its own, so the watch still stops before the
  pane's `teardown()` — same order as before, but worth a second pair of eyes.
- **`teardown()` ordering.** The pane's teardown used to bump the file's generation first and
  destroy the editor several lines later, with `search.close()` and `tree.reset()` in between.
  `file.teardown()` now does both at the first position. Nothing between them reads the editor,
  so this is observably identical — that is the claim to check.
- **`restore()` now asks `file.generation()`** where it used to compare a captured `fileReqId`.
  Same value, read through a function.
- **The new spec's `withSetup`-style host.** `mountOpenFile()` mounts a trivial component so the
  composable has a lifecycle to attach to. If the repo would rather have that as a shared helper
  in `test/helpers/`, say so and it moves.

## What the tests are worth

A mutation sweep inverted each decision in the new file in turn and re-ran both suites. Every
mutation was caught. The pre-existing pane specs miss two of them, and the new spec is what
catches those two: `save()` refusing to write over a file the server would not serve as text, and
`pagehide` sending the buffer both ways with `keepalive`. Neither is cheap to arrange through a
mounted pane, which is why neither was covered before.

Also pays #2169's carried debt: `useFilesTree.spec.ts` now has the case for a child listing that
comes back **empty but successful** — the row stays open over nothing and is not asked for again,
which is not what a listing that failed does.

## Verification

`yarn lint`, `yarn typecheck`, `yarn build` and the full `vitest` suite are green on the merged
result (this branch carries a merge of `origin/main`, which had moved during the work).

Unrelated and pre-existing: `yarn format` exits non-zero on `docs/_layouts/default.html`, which
prettier cannot parse. It is untouched here and every other file still formats.

## User Prompt

- 「#2158」 — work the issue.
- 「出来る限り小さくしてね。分割可能なのは分割して。」 — make `FilesPane.vue` as small as possible;
  split out everything that can be split.

Part of #2158. The remaining piece — the finder and the reveal — follows in its own PR.

work in mulmoterminal5

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved file editing reliability when loading, saving, and navigating between files.
  - Better handling of external file changes, including conflict detection for unsaved edits.
  - Preserved editor position, preview mode, and expanded folders when returning to a file or directory.
  - Improved recovery for failed saves, reloads, and unsupported file previews.


test/src/composables/useOpenFile.spec.ts
```diff
@@ -0,0 +1,293 @@
...
+    }
+    if (url.pathname.endsWith("/backup")) {
+      calls.push(`backup:${path}${keepalive}`);
+      const reply = server.backup ?? { ok: true, body: { stored: true } };
+      return { ok: reply.ok, status: reply.status ?? 200, json: async () => reply.body };
+    }
+    calls.push(`text:${path}`);
+    const reply = server.text?.(path) ?? { ok: true, body: { text: "on disk", version: "v1" } };
+    return { ok: reply.ok, status: reply.status ?? 200, json: async () => reply.body };
+  }) as unknown as typeof fetch;
```
isamu, line 58: P3 nit: this new spec introduces an `as` cast (`as unknown as typeof fetch`), but this repo’s rules forbid `as` casts rather than type-guarding or giving the mock a typed shape. The same PR-introduced pattern appears again at `test/src/composables/useOpenFile.spec.ts:280`; the new `useFilesTree` case also adds `tree.findNode("src") as TreeNode` at `test/src/composables/useFilesTree.spec.ts:162`. Smallest fix: type the fake fetch through a `const fetchMock: typeof fetch = vi.fn(...)`/`vi.stubGlobal` shape so no assertion is needed, and replace the tree-node cast with an explicit null check (or a small helper that returns `TreeNode` after asserting it exists).

isamu:
Reviewed the whole diff against `origin/main`, including the lifted `FilesPane.vue` hunks, the new `useOpenFile.ts`, the state-module move, and the new specs.

correctness: no behavior findings in the lifted open-file code. The read/adopt/save/conflict/discard/pagehide/external-change paths match the deleted pane code in observable state and request shape.

behaviour preservation: no behavior findings. I checked the ref/computed destructuring, `ctx` dependency reads inside the watch getter, the carried place restore, and argument evaluation before awaits; I did not find a changed dependency set or captured value that changes behavior.

concurrency / async ordering: no findings. The read generation still bumps before fetch, teardown invalidates reads, `restore()` still compares the generation captured at `start()`, reveal keeps its own guard, and `flush()`/`pagehide` still sample path/text before the first await or fire-and-forget.

LIFECYCLE / blast radius: no behavior findings. The composable hook registration preserves the old effective mount order (`pagehide`/external watch before `start()`), unregisters on host unmount, and `reload()` still re-roots in place without replacing the external watcher; the watcher reads the current `cwd`/`openPath` dynamically.

teardown ordering: no behavior findings. `file.teardown()` now destroys/clears the editor earlier than the old inline teardown, but the intervening `closeFinder()`, `search.close()`, `tree.reset()`, and tree scroll reset do not observe editor/open-file state in a way that changes behavior.

error handling: no behavior findings. The 415 path, read failure, 409 conflict, backup refusal, pagehide, and deliberate non-clearing of `fileError` on teardown all match the prior behavior.

types: one P3 repo-rule finding posted inline: the new specs introduce fresh `as` casts (`useOpenFile.spec.ts:58`, `useOpenFile.spec.ts:280`, `useFilesTree.spec.ts:162`) despite the repo rule called out in the review checklist. No `any`/`@ts-ignore` source issues found; `exactOptionalPropertyTypes` shapes are preserved.

tests: no behavioral test findings. The new `useOpenFile` suite exercises the lifted failure paths directly, and the added `useFilesTree` case pins empty-success as distinct from failed listing. I noted older pane tests still use an optional exposed `checkForExternalChange`, but they still drive the poll path via the timer, so I did not file it as a regression.

dead / stale: no findings. Imports and exposed surface look cleaned up; `FilesPaneState` importers now point at the new module.

repo rules: P3 inline finding for PR-introduced `as` casts in specs. No additional rule findings from `git diff --check`, `yarn lint`, or `yarn typecheck`.

Verification run:
- `npx vitest run test/src/composables/useOpenFile.spec.ts test/src/composables/useFilesTree.spec.ts test/src/components/FilesPane.spec.ts test/src/components/filesPanePreview.spec.ts test/src/components/filesPanePositions.spec.ts` — 91 tests passed.
- `yarn typecheck` — passed.
- `yarn lint` — passed.

FINDINGS COMPLETE

isamu:
## Review ledger — after round 1

| # | round | finding | disposition | why / where |
|---|-------|---------|-------------|-------------|
| 0 | pre-flight | sandbox: tests, typecheck, socket bind, api.github.com | OK | all four exit 0; plain `rm` works, `rm -f` refused by the command guard, so no mutation sweep was asked of codex — I ran it and report it below |
| 1 | 1 | `as` casts introduced by the new specs (`useOpenFile.spec.ts:58`, `:280`, `useFilesTree.spec.ts:162`) | FIXED | b22591e9. The fetch doubles build a real `Response`, so `ok` follows from the status and no assertion is needed; `findNode(...) as TreeNode` becomes a `nodeAt()` helper that names the missing path. Swept across BOTH specs this PR touches, including four sites codex did not name |
| 2 | 1 | `adoptListing(before as TreeNode[], …)`, same file | REJECTED for this PR | different mechanism — `it.each` tuple widening, not a null-assertion — and it predates this PR. The `as unknown as typeof fetch` shape also exists across many specs this PR does not touch; rewriting those is its own change |
| 3 | 1 | jscpd: duplicated block at `useOpenFile.ts:46` (not a codex finding — the code-scanning bot) | FIXED | b22591e9. `OpenFileCtx` and `OpenFile` listed the same twelve fields. They are one buffer seen from two sides, so `OpenFileBuffer` is declared once and both extend it |

### Behaviour preservation — what was run, not argued

- A throwaway differential harness mounted the pane before and after: five scenarios (plain file,
  read in Preview, not-text 415, save conflict, read fails) x with/without a restored state,
  recording header and body HTML, emits, snapshot, request sequence and editor call counts after
  every step. Every record identical. Harness deleted.
- Mutation sweep over the extracted decisions, re-run after this round's spec rewrite to check the
  rewrite had not weakened them: **12 of 12 caught**, unchanged. The pre-existing pane specs miss
  two — `save()` refusing to write over an unpreviewable file, and `pagehide` sending the buffer
  both ways with `keepalive` — and the new spec is what catches those two.
- `yarn lint`, `yarn typecheck`, `yarn build` and the full `vitest` suite green on this head.

Re-flagging a FIXED or REJECTED row is welcome **if the resolution is wrong** — say which row and
what is wrong with it.

isamu:
Reviewed round 2 against head `b22591e9`, including the ledger comment, the full PR diff, the old `FilesPane.vue` implementation from `origin/main`, the new `useOpenFile.ts`, the state-module move, and the rewritten specs. No inline findings this round.

correctness: no findings. The lifted open-file paths still match the deleted pane code: load/may-leave, 415 adoption, text adoption, place restore, save/conflict/overwrite, discard+backup, open-in-OS/reveal error reporting, pagehide writes, external-change recheck, and teardown state clearing all preserve the old observable state/request behavior.

behaviour preservation: no findings. I checked the places where a move could change behavior without changing much text: destructured refs in `FilesPane.vue`, computed dependencies for `isMarkdown`/`previewSrc`, dynamic `cwd()` reads, `ctx.editor.value` reads before awaits, remembered-vs-carried place restore, and the replacement of `openDocChannel` computed with `docChannel(ctx)` as a watcher getter. I did not find a changed dependency set or stale capture.

concurrency / async ordering: no findings. `loadFile()` still increments the read generation before fetch and drops overtaken reads after body parsing; `file.teardown()` still invalidates parked reads; `restore()` compares the generation captured at `start()` against the current file generation; reveal still has its own generation; and `flush()`/`pagehide` still capture path/text before their async or fire-and-forget work.

LIFECYCLE / blast radius: no findings. Moving `pagehide` and external-change watch into `useOpenFile()` preserves registration order because the composable is called before the pane registers `start()`/`teardown()`, and unmount still removes listener/watch before pane teardown. `reload()` continues to call `teardown()` then `start()` without unmounting; the external watcher reads current `cwd`, `openPath`, and doc channel through getters, so it does not stay pinned to the old root.

teardown ordering: no findings. The editor is now destroyed inside `file.teardown()` earlier than before, but the intervening pane work (`closeFinder`, `search.close`, `tree.reset`, scroll reset) does not observe the editor in a way that changes behavior. `fileError` still intentionally survives teardown as before.

error handling: no findings. The 415 non-error path, read-failure body reason fallback, 409 conflict, overwrite baseline, backup refusal on discard, save-fail-then-backup in `flush()`, and pagehide dual write/backup behavior all match the old code.

types: no new findings. The round-1 `as` casts in the touched spec code are fixed. The remaining `adoptListing(before as TreeNode[], …)` is the ledger's rejected/pre-existing tuple-widening case, not a new round-2 regression. `shallowRef<CmEditor | null>` is appropriate for the editor handle, and `exactOptionalPropertyTypes` handling for remembered places remains absent-key based. `OpenFileBuffer` being exported is a small surface increase, but I do not see it creating a behavioral or type-safety issue in this internal module; it is not imported outside the module.

tests: no findings. The real `Response` rewrite changes the doubles in the useful direction: `ok` now follows status and `jsonBody` parses a real one-shot body. I did not find any reply consumed twice. The focused suites still exercise 415, read errors, conflict/overwrite, backup refusal, flush fallback, pagehide listener cleanup, in-flight teardown invalidation, and the empty-directory case. Mutations I would ask to check if needed: invert the `id !== ctx.reqId.n` guard after read, remove `keepalive` from either pagehide request, or change empty child listings to leave `loaded=false`.

dead / stale: no findings. The imports moved cleanly to `filesPaneState.ts`; `FilesPane.vue` no longer carries the lifted helpers; comments that remain in the pane still describe pane-owned behavior.

repo rules: no new findings from this round. Focused tests and typecheck passed:
- `npx vitest run test/src/composables/useOpenFile.spec.ts test/src/composables/useFilesTree.spec.ts test/src/components/FilesPane.spec.ts test/src/components/filesPanePreview.spec.ts test/src/components/filesPanePositions.spec.ts` — 91 tests passed.
- `yarn typecheck` — passed.

CODEX VERDICT: LGTM
FINDINGS COMPLETE

isamu:
## Review ledger — round 2 (final)

Round 1's single P3 and the jscpd duplication are fixed in `b22591e9`. Round 2 on that head:
**CODEX VERDICT: LGTM**, no findings on any axis.

Codex named three mutations it would want checked before signing off. All three were run against
`useOpenFile.spec.ts` + `useFilesTree.spec.ts`, and all three go red:

| mutation | result |
|---|---|
| invert `id !== ctx.reqId.n` after the body is parsed | RED — caught |
| drop `keepalive` from the `pagehide` WRITE (the earlier sweep covered only the backup) | RED — caught |
| an empty child listing leaves the row `loaded = false` | RED — caught |

Across both sweeps: 15 mutations named, 15 caught. The pre-existing pane specs miss two of them —
`save()` refusing to write over a file the server would not serve as text, and `pagehide` sending
the buffer both ways with `keepalive` — and the new spec is what catches those two.

Exit condition met: one clean round on a head nothing was pushed to.

PR #2175, opened 2026-09-19
docs: the 75 exit-code contract names the wrong file, and claims a retry that does not exist

isamu:
## Summary

Two errors in one comment block in `server/infra/server-exit.ts`, found by a Codex cross-review of
#2171 and by checking what it found. Comments only — no shipped behaviour changes.

**Codex's finding (severity N).** Both comments name `bin/mulmoterminal.js` as the launcher side of
the exit-code contract. #2171 moved that literal to `bin/server-supervision.js`, which is also the
file the spec now reads — so `// Keep in sync with bin/mulmoterminal.js — a test asserts the two
agree` sends the next reader to a file where the number is no longer written.

**The one it did not name, and the reason this is a commit rather than a note.** The block opens
with: *"75 tells the launcher the port was taken at bind time, **so it retries on a fresh one**"*.
The launcher does not do that. On exit 75 it prints who holds the port and exits — `main()` calls
`portInUseMessage`, whose own text tells the **user** to pick a different port.

The fresh-port offer is real, but it belongs to the **probe** path: `pickPort` finds the port busy
*before* spawning, says so, and takes an ephemeral port only on an explicit answer — under a comment
reading *"No SILENT fallback: starting a second server on another port without saying so is how
someone ends up with two sharing one home directory without knowing (#611)"*. Conflating the two
paths is precisely how someone would restore the silent fallback that comment exists to forbid, so
this is prose that would cause a wrong change, not prose that is merely behind.

It also now contradicts #2171's own comment in `bin/server-supervision.js` ("a port that is taken
will still be taken"), so the tree currently holds two comments disagreeing about one contract.

## Items to Confirm / Review

1. **Is the probe/bind distinction stated correctly?** The claim is: the pre-flight probe can offer
   a different port (with consent); a bind-time 75 cannot, because by then the child has already
   been spawned on the chosen port. Please check that against `pickPort` and `main()`.
2. **`#611` is taken from `pickPort`'s own comment**, not inferred. I attributed it there rather
   than guessing.
3. The block no longer restates *when* the launcher restarts — it points at
   `bin/server-supervision.js` instead. One copy of that rule, in the file that executes it.

## User Prompt

- `/codex-cross-review` — run the cross-review protocol (the target was #2162 / PR #2171, already
  merged, so its findings land here).

Cross-review record: #2171 (comment with the verbatim prompt and reply). Verdict there was
`CODEX VERDICT: LGTM` with `FINDINGS COMPLETE`; this was its only finding, plus the one above that
came out of checking it.

work in mulmoterminal3


isamu:
## Tier T — two self-review passes, no Codex round

**Reason (one sentence):** the diff changes only comments in one file, so no shipped behaviour
changes and nothing in the tier-C list is touched.

**Pass 1** — wrote the correction, checked each sentence against the code it describes
(`main()`'s exit-75 path, `portInUseMessage`, `pickPort`, and the spec that reads
`bin/server-supervision.js`), ran format / lint / typecheck and the `server-exit` spec, pushed.

**Pass 2** — on this head, with nothing pushed since pass 1, re-read the diff whole. Every claim
in the new block re-verified against its source; `#611` re-checked against `pickPort`'s own
citation rather than inferred. No code change, so the tier holds — a change in pass 2 would have
promoted this and run the Codex loop from round 1.

**No Codex round ran, and that is the tier, not a skipped step.** The Codex round that produced
finding #1 here ran against #2171 and its verbatim prompt and reply are on that PR.

PR #2179, opened 2026-09-19
fix(tmux): a tick drops the files of the sessions it ended (#2178)

### LINKED ISSUE #2178: bug(tmux): タイマーが終了したセッションの settings / drop ファイルが次の再起動まで残る（API トークンを含みうる）
## 観測した事実（コードを辿ったもので、現場で再現したものではない）

#2167 でタイマーからもセッションを終了するようになったが、**タイマーが終了したセッションの
ディスク上のファイルは誰も消さない**。

呼び出し連鎖（`server/session/reap-schedule.ts`）:

```
armTimer → setInterval(() => sweepNow(idleDays, log))
         → sweepIdleSessions → tmuxKillSession
```

`sweepNow` はスイープしてログを出すだけで、`cleanupSessionSettings` /
`cleanupSessionDrops` を呼ばない。起動時スイープの方は `server/index.ts` が直後に
`pruneOrphanSettings` / `pruneOrphanDrops` を回すので回収されるが、**それは起動時に一度きり**。

結果、タイマーがセッションを終了してから次の再起動までの間、そのセッションの
`~/.mulmoterminal/settings/<id>.json` などが残り続ける。

## なぜ問題か

`server/session/session-settings.ts` が自分で書いているとおり、このファイルは
**provider セッションの API トークンを持つ**:

> a provider session's file holds its API token, so without this a token outlives the session
> that used it, survives being rotated or revoked, and survives the provider being removed
> from the config entirely.

つまり #1467 が「起動時に消す」根拠にしたのと同じ理由が、タイマー経路では効いていない。
そして `sessionReapIntervalHours` を有効にする動機は「再起動しないサーバ」なので、
**この機能を使う人ほど、次の再起動が遠い**。

## 提案

`armTimer` の tick で、終了できたセッションの id について
`cleanupSessionSettings` / `cleanupSessionDrops` を呼ぶ。

ひとつ注意が要る: **id をそのままパスに使ってはいけない**。スイープは
`SESSION_ID_RE` に合致しない id を意図的に終了する（#1533 の `mt-undefined`）。
一方 `settingsFile()` は `path.join(SETTINGS_DIR, \`${sessionId}.json\`)` なので、
id をそのまま渡すとディレクトリの外を指しうる。
起動時の prune 側は既に `SESSION_ID_RE` / `sessionIdFromFileName` で弾いているので、
同じガードをこの経路にも付ける。

## 範囲外

`POST /api/tmux/cleanup-orphans` も同じ穴を持つが、こちらは人が押して直後に人が見ている
経路なので別issueとして残すか、同じPRに含めるかは判断が要る。**このissueではタイマー経路のみ**
を対象にしている。

## 再現について

再現はしていない。最短間隔が1時間で、スイープが実 tmux ソケット（`-L mulmoterminal`）に
対して走るため、強制的に tick を起こすとこのマシンの生きたセッションを掃いてしまう。
上の呼び出し連鎖と、起動時経路にだけ prune がある非対称性を根拠にしている。

## 関連

- #2165 / #2167 定期スイープの導入
- #1467 起動時に使われていないセッションを終了し、孤児ファイルを prune する
- #1533 セッションidとして解釈できない id を即座に終了する

isamu:
## Summary

Closes #2178.

#2167 gave the idle sweep a second caller — the timer — but only the first one has a
follower that removes what an ended session left on disk:

```
server/index.ts  →  startReapSchedule()  →  sweepNow()   … then pruneOrphanSettings / pruneOrphanDrops
armTimer's tick  →                          sweepNow()   … then nothing
```

A session ended by a tick therefore keeps `~/.mulmoterminal/settings/<id>.json` until the
next restart. `session-settings.ts` says in its own words why that is not inert — a
provider session's file holds its API token, so it outlives the session, a rotation, and
the provider being removed from the config. And the condition is not exotic:
`sessionReapIntervalHours` is turned on precisely by someone whose server does not restart,
so "until the next boot" is as long as it gets.

The tick now passes its reaped ids to `cleanupSessionSettings` / `cleanupSessionDrops`.

## Items to Confirm / Review

1. **Tick only, deliberately not boot.** The boot sweep already has its prune, and that
   prune uses a live-peer cutoff (#1061) that only a boot can work out. A per-id delete on
   the boot path would go around it. A spec pins that the boot half still cleans nothing.
2. **The id is checked before it becomes a path**, and this is the part worth a second
   pair of eyes. The sweep ends ids that are NOT session ids on purpose (#1533), and
   `settingsFile()` does `path.join(SETTINGS_DIR, \`${sessionId}.json\`)` — so an id that
   is not a UUID must not reach it. The boot prunes already make this check
   (`SESSION_ID_RE`, `sessionIdFromFileName`); this adds it to the second route.
3. **`POST /api/tmux/cleanup-orphans` has the same gap and is NOT fixed here.** It is a
   separate revert unit, and it is user-initiated and watched rather than unattended. Say
   if it should be folded in.

## How it was verified

- `yarn lint`, `yarn typecheck`, `yarn build`, `yarn test` clean.
- Four specs on the tick: it cleans what it ended, it cleans nothing when it ended nothing,
  it skips an id that is not a session id, and the boot half still cleans nothing.
- **Both guards break-verified.** Removing the `SESSION_ID_RE` filter reddens the traversal
  spec; removing the cleanup call reddens both new specs. The file was restored from a copy
  and re-run green between mutations, and the working tree checked back to the intended diff.
- The delete functions are **mocked** in the spec. The real ones remove files under
  `~/.mulmoterminal`, and a spec that let them run would be deleting the developer's own.

### One thing worth recording

My first draft of these specs used readable ids (`mt-a`, `abc123`) and both new tests failed.
The guard was right and the fixtures were wrong — session ids are UUIDs (`SESSION_ID_RE` in
`server/config/env.ts`), so every fixture was being filtered out. Had the assertion been
written the other way round (`not.toHaveBeenCalled()`), it would have passed while checking
nothing. The spec now carries a comment saying why the ids look the way they do.

## User Prompt

- Work https://github.com/receptron/mulmoterminal/issues/2165. After finding that #2167 had
  already merged the same feature: close the duplicate PR #2172, keep the merged default of
  `0` (off), and take the parts that were **not** duplicated to their own PRs. This is one of
  those parts.

work in mulmoterminal4


isamu:
## Cross-review tier: **C**

Three things on the tier-C list, any one of which would be enough:

- **it deletes files** — `cleanupSessionSettings` / `cleanupSessionDrops` remove real paths under `~/.mulmoterminal`, unattended, on a timer;
- **secrets** — the files being removed are the ones that hold a provider session's API token, which is the whole reason the removal matters;
- **a path is built from an id the sweep deliberately allows to be malformed** (#1533), so the guard between the two is the change.

Round 1 names those as explicit axes and requires a per-axis answer.

**Sandbox pre-flight (ledger row 0):** Codex can run vitest, `yarn typecheck`, bind a socket, reach `api.github.com` and run `gh` — all exit 0. **`rm -f` is refused by policy**, so it has been told not to run a mutation sweep; I will run any mutation it names instead.

isamu:
CODEX VERDICT: LGTM

| Axis | Result |
| --- | --- |
| correctness and edge cases | none |
| DESTRUCTIVE OPERATION: id/path/guard | none — the id reaching `cleanupSessionSettings` / `cleanupSessionDrops` is `sweepIdleSessions(...).reaped`; the timer path filters with `SESSION_ID_RE` before `cleanupSessionSettings` can build `~/.mulmoterminal/settings/<id>{.json,-mcp.json,-prompt.txt,-seed.txt}`. Drops cleanup re-checks the same regex before producing `<tmp>/mulmoterminal-drops/<id>`. The guard is the same UUID guard used by the live sweep and boot prune path, and it is applied at the right choke point before path construction. |
| SECRETS | none — the timer now removes the token-bearing settings file plus the sibling mcp/prompt/seed files for sessions it actually ended; invalid ids are intentionally not converted into paths. I did not find a file kind written by `session-settings.ts` that this cleanup misses. |
| path traversal / injection via the session id | none — `SESSION_ID_RE` admits only UUID-shaped ids, has no stateful flags, and rejects traversal/metacharacter strings before the raw settings path join. |
| ASYNC ORDERING | none — the timer callback is synchronous, so interval ticks do not execute concurrently in-process; boot sweep completes before `armTimer` installs the interval. The production server installs `process` guards before startup, so a thrown timer exception is logged/kept alive there; the cleanup functions themselves are best-effort/nonthrowing via `removeQuietly`. |
| BLAST RADIUS: default config (`sessionReapIntervalHours: 0`) | none — `reapTimerEnabled(0)` still returns false, so the default path still performs only the boot sweep and does not arm the unattended cleanup timer. |
| tests | none — the new assertions exercise cleaned ids, invalid-id filtering, no-op cleanup, and boot-not-cleaned behavior. The readable-id false-positive risk is addressed by UUID fixtures. I could not rerun locally in this sandbox because Vite/vue-tsc tried to write cache/build-info files through the symlinked `node_modules` into `/Users/isamu/...`, but the prompt’s targeted run is green. |
| API / schema / wire / on-disk compatibility | none — no API/schema/wire changes; on disk, the change only deletes existing per-session files for UUID sessions the timer just ended. |
| consistency with the rest of the codebase | none — this mirrors lifecycle cleanup and uses the same `SESSION_ID_RE` / prune assumptions as `server/index.ts`, `session-settings.ts`, and `session-drops.ts`. |
| accessibility and i18n lockstep | none — N/A for this server-side cleanup change. |

isamu:
## Codex exchange — round 1

```text
Review PR #2179 at https://github.com/receptron/mulmoterminal/pull/2179.
Repo checkout: /tmp/claude-501/-Users-isamu-ss-llm-mulmoterminal4/4c82af4c-781b-4be0-90e0-41250d7ed192/scratchpad/wt-2179 — a git worktree already on branch fix/2178-timer-reap-file-cleanup, with
node_modules symlinked. Work THERE, not in /Users/isamu/ss/llm/mulmoterminal4 (that tree is on a
different branch and belongs to another review running right now — do not touch it).
Verified working in the worktree: \`npx vitest run test/server/session/reap-schedule.spec.ts\` exits 0.

  gh api repos/receptron/mulmoterminal/pulls/2179/comments
or top-level comments via:
  gh pr comment 2179

── SANDBOX ─────────────────────────────────────────────────────────────────
You CAN run vitest, \`yarn typecheck\`, bind sockets, reach api.github.com and run gh.
You CANNOT run \`rm -f\` — it is refused by policy. So DO NOT run a mutation sweep:
your restore would fail and every count you took afterwards would be measuring the
harness rather than the code. If you want a mutation tested, NAME it and I will run it.

── WHAT THIS PR IS ─────────────────────────────────────────────────────────
#2167 added \`sessionReapIntervalHours\`, giving the idle tmux sweep a second caller (a
timer) besides the boot sweep. Only the boot path has a follower that removes what an
ended session left on disk (\`pruneOrphanSettings\` / \`pruneOrphanDrops\` in server/index.ts,
which use a live-peer cutoff from #1061). This PR makes the timer's tick drop the files of
the sessions it ended, guarded by SESSION_ID_RE.

  correctness and edge cases
  DESTRUCTIVE OPERATION: this deletes files on a timer, unattended. What id can reach
    \`cleanupSessionSettings\` / \`cleanupSessionDrops\`, and what path can that id produce?
    Is SESSION_ID_RE the right guard and is it applied at the right place?
  SECRETS: the files removed hold a provider session's API token. Is anything that should
    be removed now NOT removed, or vice versa?
  path traversal / injection via the session id
  ASYNC ORDERING: \`setInterval\` + \`unref\`, a tick that overlaps a previous tick, a tick
    that overlaps the boot sweep, and what happens if \`sweepIdleSessions\` throws inside a
    timer callback (an uncaught throw in a timer takes the process down).
  BLAST RADIUS: \`startReapSchedule\` is called from server boot. Does this change anything
    for the default config (\`sessionReapIntervalHours: 0\`)?
  tests: does a test go red if the new behaviour breaks — and can any of the new
    assertions pass while checking nothing?
  API / schema / wire / on-disk compatibility
  consistency with the rest of the codebase (compare against how server/index.ts and
    server/session/reap-idle-sessions.ts already do this)
  accessibility and i18n lockstep (likely N/A here — say so)

   N is for a finding whose only remedy is editing PROSE (a comment, a docblock, a plan
   file, the PR description). It does NOT block the verdict. A prose finding is P3 or
   higher ONLY when believing the sentence would lead someone to make a WRONG CHANGE.
   Merely behind — a count that no longer matches, a name that moved — is N.
2. EVERY SITE. If the same mistake appears elsewhere in the repo, list every occurrence
   NOW — grep for it.

| 0 | — | Sandbox pre-flight | — | vitest/typecheck/socket/curl/gh all exit 0. \`rm -f\` refused by policy: no mutation sweeps. |
| 0b | — | Pre-existing red baseline | — | \`yarn format\` fails on docs/_layouts/default.html on origin/main too; not in this diff. |

_no findings yet_

  Then, on its own line:
  Omit that last line if it is not true, and say which part of the diff you did not reach.

```

</details>

<details><summary>Codex reply (verbatim stdout)</summary>

```text
Posted the top-level review comment on PR #2179: LGTM, no findings above N.

I read every hunk and checked the relevant cleanup/id/path/timer code paths. I tried to rerun the targeted spec and \`yarn typecheck\`, but this sandbox blocked writes through the symlinked \`node_modules\` into \`/Users/isamu/...\`; I noted that in the review comment rather than treating it as a PR issue.
```

</details>

Its posted verdict is the comment above. It posts as `isamu` — Codex uses the same `gh` auth, so its review is not attributed separately. Nothing stripped; no secrets appeared.

A note on the sandbox pre-flight, because it is the reason the round is qualified: the pre-flight ran in the MAIN checkout and passed all four probes. The round itself then ran in a **git worktree** with `node_modules` symlinked, and that is a different environment — see the triage below.

isamu:
## Triage — round 1: **LGTM, but the round could not RUN anything**

Codex answered every axis, found nothing above severity N, and gave `FINDINGS COMPLETE`. I am **not** treating that as the clean round, and the reason is worth stating.

### The qualification

> I could not rerun locally in this sandbox because Vite/vue-tsc tried to write cache/build-info files through the symlinked `node_modules` into `/Users/isamu/...`

The pre-flight passed all four probes — in the **main checkout**. I then handed the round a **git worktree** with `node_modules` symlinked, which is a different environment, and vitest/vue-tsc cannot write their caches through that symlink under the sandbox. So round 1 is a **read** of the diff, not a **run** of it.

That distinction is exactly the failure this loop's setup step exists to catch, and I reintroduced it one level over by changing the environment after proving it. Every execution claim backing this PR is currently **mine**, not the second reviewer's:

- full suite green, lint / typecheck / build green;
- both new guards break-verified (removing the `SESSION_ID_RE` filter reddens the traversal spec; removing the cleanup call reddens both new specs), with the file restored from a copy and re-run green between mutations.

That is real evidence, but it is one reviewer's. **I am re-running round 1 in the main checkout**, where the probes actually passed, so the verdict on a tier-C change that deletes files unattended comes from a reviewer that could execute it.

### Why the worktree existed at all

The loop dispatched reviews for this PR and #2183 in parallel, and one checkout cannot be on two branches. The first attempt told Codex the tree was on this branch when it was on #2183's — caught by reading `err.txt` early, that round was killed after ~2 minutes and re-dispatched into a worktree rather than left to produce a verdict about the wrong tree.

### One finding of my own, raised and closed in this round

I considered moving the `SESSION_ID_RE` guard **into** `cleanupSessionSettings` — where the path is actually built, and matching `dropsDir`, which already guards that way. It looks like the class fix rather than the site fix.

**It is wrong, and I established that by running it rather than reasoning about it.** Moving the guard in a scratch worktree fails **nine tests** across `test/server/session/session-settings.spec.ts` and `test/server/agents/seed-prompt-not-argv.spec.ts`: those specs use non-UUID ids by design (`"settings-spec-session"`, `"seed-argv-spec-session"`) and rely on `afterEach(() => cleanupSessionSettings(SESSION))` actually deleting. So `cleanupSessionSettings` has a deliberate contract of cleaning whatever id it is handed, and the guard belongs on the **untrusted route into it** — which is the tick. `cleanupSessionDrops` needs no guard at all, since `dropsDir` returns `null` for a non-matching id.

Codex reached the same placement independently (*"applied at the right choke point before path construction"*, and it noticed the drops path re-checks), which is the agreement worth having on a change that deletes files.

### Ledger

| # | finding | disposition |
|---|---|---|
| 0 | sandbox pre-flight | passed in the main tree; **not** valid for the worktree the round ran in |
| 1 | round-1 verdict LGTM, all axes, FINDINGS COMPLETE | **QUALIFIED** — read-only; re-running where execution works |
| C1 | (mine) move the guard into `cleanupSessionSettings` | **REJECTED** — 9 tests fail; evidence above |

isamu:
CODEX VERDICT: LGTM

| Axis | Result |
| --- | --- |
| correctness and edge cases | none. The timer tick now cleans only the sessions returned in that tick's `reaped` list; the boot sweep still only returns reaped ids and leaves orphan pruning to `server/index.ts`, preserving the live-peer cutoff path. |
| DESTRUCTIVE OPERATION | none. The only ids that reach `cleanupSessionSettings` / `cleanupSessionDrops` from the new unattended path are `sweepIdleSessions(...).reaped` ids that pass `SESSION_ID_RE`; that means the settings paths can only be `~/.mulmoterminal/settings/<uuid>.json`, `<uuid>-mcp.json`, `<uuid>-prompt.txt`, and `<uuid>-seed.txt`, and the drops path can only be `os.tmpdir()/mulmoterminal-drops/<uuid>`. The guard is applied before the settings path is built, which is the right boundary for this route. |
| SECRETS | none. The token-bearing settings file is now removed for timer-ended UUID sessions, and the same per-session cleanup also covers mcp config, appended prompt, seed prompt, and drop files. Invalid/unparseable tmux ids are deliberately not passed to settings cleanup because doing so would build a path from an untrusted id. |
| path traversal / injection via the session id | none. `SESSION_ID_RE` accepts only UUID-shaped ids, so `../`, path separators, suffix tricks, and `mt-undefined` do not reach either cleanup function through the timer path. |
| ASYNC ORDERING | none. The tick callback is synchronous, so ticks cannot execute concurrently on the JS thread; a long tick may delay/back up the next interval but not overlap cleanup with itself. The boot sweep completes before `armTimer` installs the interval, so there is no boot/tick overlap. `unref()` only affects process liveness. If `sweepIdleSessions` throws in the timer callback, that remains an uncaught timer exception, but this PR does not introduce that behavior; the prior timer callback already called `sweepNow(...)` without a catch, and the new cleanup functions are best-effort/non-throwing through `removeQuietly`. |
| BLAST RADIUS | none. For the default `sessionReapIntervalHours: 0`, `reapTimerEnabled(0)` is false, so the added cleanup path is never armed; boot behavior is unchanged. |
| tests | none. Ran `yarn vitest run test/server/session/reap-schedule.spec.ts` (9 passed), `yarn typecheck` (passed), and `yarn lint` (passed). The new assertions check exact cleanup call ids for valid UUIDs and for mixed invalid/valid ids, so they would fail if cleanup were removed or if invalid ids reached cleanup; they are not vacuous `not.toHaveBeenCalled()`-style checks for the positive behavior. |
| API / schema / wire / on-disk compatibility | none. No API/schema/wire shape changes. On-disk behavior changes only by deleting the same per-session files already deleted by lifecycle cleanup and boot orphan pruning, but now for sessions ended by the timer. |
| consistency with the rest of the codebase | none. The new guard matches the UUID checks used by boot pruning and session routes, while leaving `cleanupSessionSettings`' existing direct-call contract intact. Drops still have their own guard inside `dropsDir`, and settings are protected at the timer route-in, matching the rejected C1 constraint. |
| accessibility and i18n lockstep | none / N/A. This is server-side timer cleanup and tests only; no user-facing UI strings, layout, or localized copy changed. |

isamu:
## Codex exchange — round 2

```text
Review PR #2179 (ROUND 2) at https://github.com/receptron/mulmoterminal/pull/2179.
Repo checkout: /Users/isamu/ss/llm/mulmoterminal4 — the MAIN checkout, now on branch
fix/2178-timer-reap-file-cleanup. Verified here: \`yarn vitest run
test/server/session/reap-schedule.spec.ts\` passes, \`yarn typecheck\` passes, \`yarn lint\` passes.

  gh api repos/receptron/mulmoterminal/pulls/2179/comments
or top-level comments via:
  gh pr comment 2179

── SANDBOX ─────────────────────────────────────────────────────────────────
You CAN run vitest, \`yarn typecheck\`, bind sockets, reach api.github.com and run gh.
You CANNOT run \`rm -f\` — it is refused by policy. So DO NOT run a mutation sweep:
your restore would fail and every count you took afterwards would be measuring the
harness rather than the code. If you want a mutation tested, NAME it and I will run it.

── WHAT THIS PR IS ─────────────────────────────────────────────────────────
#2167 added \`sessionReapIntervalHours\`, giving the idle tmux sweep a second caller (a
timer) besides the boot sweep. Only the boot path has a follower that removes what an
ended session left on disk (\`pruneOrphanSettings\` / \`pruneOrphanDrops\` in server/index.ts,
which use a live-peer cutoff from #1061). This PR makes the timer's tick drop the files of
the sessions it ended, guarded by SESSION_ID_RE.

  correctness and edge cases
  DESTRUCTIVE OPERATION: this deletes files on a timer, unattended. What id can reach
    \`cleanupSessionSettings\` / \`cleanupSessionDrops\`, and what path can that id produce?
    Is SESSION_ID_RE the right guard and is it applied at the right place?
  SECRETS: the files removed hold a provider session's API token. Is anything that should
    be removed now NOT removed, or vice versa?
  path traversal / injection via the session id
  ASYNC ORDERING: \`setInterval\` + \`unref\`, a tick that overlaps a previous tick, a tick
    that overlaps the boot sweep, and what happens if \`sweepIdleSessions\` throws inside a
    timer callback (an uncaught throw in a timer takes the process down).
  BLAST RADIUS: \`startReapSchedule\` is called from server boot. Does this change anything
    for the default config (\`sessionReapIntervalHours: 0\`)?
  tests: does a test go red if the new behaviour breaks — and can any of the new
    assertions pass while checking nothing?
  API / schema / wire / on-disk compatibility
  consistency with the rest of the codebase (compare against how server/index.ts and
    server/session/reap-idle-sessions.ts already do this)
  accessibility and i18n lockstep (likely N/A here — say so)

   N is for a finding whose only remedy is editing PROSE (a comment, a docblock, a plan
   file, the PR description). It does NOT block the verdict. A prose finding is P3 or
   higher ONLY when believing the sentence would lead someone to make a WRONG CHANGE.
   Merely behind — a count that no longer matches, a name that moved — is N.
2. EVERY SITE. If the same mistake appears elsewhere in the repo, list every occurrence
   NOW — grep for it.

| 0 | — | Sandbox pre-flight | — | vitest/typecheck/socket/curl/gh all exit 0. \`rm -f\` refused by policy: no mutation sweeps. |
| 0b | — | Pre-existing red baseline | — | \`yarn format\` fails on docs/_layouts/default.html on origin/main too; not in this diff. |

| 1 | 1 | Round 1 verdict LGTM, all axes, FINDINGS COMPLETE. | **QUALIFIED, and this round exists to fix that** | Round 1 ran in a git WORKTREE with \`node_modules\` symlinked, and you could not run vitest or vue-tsc — they write caches through the symlink and the sandbox refused. So round 1 was a READ of the diff, not a RUN of it. You are now in the main checkout where all three commands work. |
| C1 | 1 | (mine) Should the \`SESSION_ID_RE\` guard move INTO \`cleanupSessionSettings\`, where the path is built — matching \`dropsDir\`, which guards that way? | REJECTED | Established by RUNNING it, not reasoning: moving the guard fails **9 tests** across \`test/server/session/session-settings.spec.ts\` and \`test/server/agents/seed-prompt-not-argv.spec.ts\`. Those specs use non-UUID ids by design (\`"settings-spec-session"\`, \`"seed-argv-spec-session"\`) and rely on \`afterEach(() => cleanupSessionSettings(SESSION))\` actually deleting. \`cleanupSessionSettings\` has a deliberate contract of cleaning whatever id it is handed, so the guard belongs on the untrusted route in — the tick. |

WHAT CHANGED SINCE ROUND 1: only a merge of origin/main (edb5d14b). The PR's own commit
(2cbc3a49) is unchanged.

THIS ROUND'S JOB: round 1 could not execute. Now that you can, please actually RUN things —
the specs, typecheck, and any mutation you want to see. You may NOT run \`rm -f\` (refused by
policy), so do not attempt a mutation sweep; NAME any mutation you want and I will run it and
report the result verbatim.

  Then, on its own line:
  Omit that last line if it is not true, and say which part of the diff you did not reach.

```

</details>

<details><summary>Codex reply (verbatim stdout)</summary>

```text
Posted the ROUND 2 review comment on PR #2179:

https://github.com/receptron/mulmoterminal/pull/2179#issuecomment-5740769271

Verdict: \`CODEX VERDICT: LGTM\`

I also ran and confirmed:
\`yarn vitest run test/server/session/reap-schedule.spec.ts\`
\`yarn typecheck\`
\`yarn lint\`
```

</details>

Nothing stripped; no secrets appeared.

isamu:
## Triage — round 2: **LGTM, and this time it RAN**

Round 1's verdict was a read of the diff. Round 2 ran in the main checkout and executed `yarn vitest run test/server/session/reap-schedule.spec.ts` (9 passed), `yarn typecheck` and `yarn lint`. **The qualification is converted** — the second reviewer has now exercised what it is approving, which on a change that deletes files unattended is the point.

It also closed two things independently of me:

- **C1 confirmed.** It reached the same guard placement on its own — *"settings are protected at the timer route-in, matching the rejected C1 constraint"* — after I had established by running that moving the guard deeper fails 9 existing tests.
- **The new assertions are not vacuous.** *"they would fail if cleanup were removed or if invalid ids reached cleanup; they are not vacuous `not.toHaveBeenCalled()`-style checks for the positive behavior."* That is the specific failure mode this loop warns about, checked rather than assumed.

### The one qualification attached to a "none", reproduced rather than accepted

> If `sweepIdleSessions` throws in the timer callback, that remains an uncaught timer exception, but this PR does not introduce that behavior.

This loop's guidance is that a qualification hung off a "none" is the most valuable thing a reviewer produces, so I checked it rather than nodding:

- **It is pre-existing** — `main`'s `armTimer` already called `sweepNow(...)` with no catch.
- **It does not crash the server.** `server/infra/process-guards.ts` registers an `uncaughtException` handler that logs and **deliberately keeps the process alive**, with the trade-off written out in the file: *"one session left in an inconsistent state is far better than taking down every other terminal with it."*
- **This PR adds no new throw surface.** `removeQuietly` never throws by construction (it is a `try`/`catch` returning a boolean), and neither `SESSION_ID_RE.test` nor `path.join` throws on these inputs.

So: real, benign, and not mine. Recorded rather than fixed.

### Where the loop stands

| | |
|---|---|
| round 1 | LGTM — but **read-only**, the sandbox could not execute through a worktree's symlinked `node_modules` |
| round 2 | **LGTM**, all axes, FINDINGS COMPLETE, **executed** |

Two rounds, and the second was needed for a reason that was mine: I proved the sandbox in the main checkout and then ran the round somewhere else. The fix was to move the branch into the checkout the pre-flight actually covered.

Pushed since round 1: only `edb5d14b`, a merge of `origin/main` — which round 2 reviewed. Nothing has been pushed since.

isamu:
## Codex exchange — round 3

```text
Review PR #2179 (ROUND 3) at https://github.com/receptron/mulmoterminal/pull/2179.
Repo checkout: /Users/isamu/ss/llm/mulmoterminal4, MAIN checkout, on branch
fix/2178-timer-reap-file-cleanup. You can run vitest, yarn typecheck and yarn lint here.
You CANNOT run \`rm -f\` (refused by policy) — do not run a mutation sweep; NAME any mutation
you want and I will run it.

Round 2 gave \`CODEX VERDICT: LGTM\` with every axis answered and FINDINGS COMPLETE, on
edb5d14b. Two things have been pushed since, and this round exists to review them.

1. **dfdb7d56** — a merge of origin/main. \`main\` moved twice: PR #2183 (the sibling UI PR,
   now merged) and **PR #2176, which SPLIT server/index.ts**. That second one matters here:
   \`startReapSchedule\` — the function this PR changes the behaviour of — used to be called
   from \`server/index.ts\` and is now called from \`server/infra/on-listening.ts\`. The orphan
   prune that follows the boot sweep (\`pruneOrphanSettings\` / \`pruneOrphanDrops\`) moved to the
   same file. **Nobody has reviewed this PR's change sitting on top of that split.**

2. **0ffb6306** — prose only. The comments in \`reap-schedule.ts\`, the spec, and the plan file
   pointed a reader at \`server/index.ts\` for the caller and the prune. After #2176 that is the
   wrong file, so following the reference would send someone to the wrong place. No code.

THE JOB THIS ROUND:

A. **The combination.** Does this PR's change still do what round 2 approved, now that the
   caller and the orphan prune live in \`server/infra/on-listening.ts\`? Specifically re-check
   the invariant the whole change rests on: the BOOT sweep's reaped ids must still be cleaned
   by the orphan prune (with its live-peer cutoff), and the TIMER tick's reaped ids must still
   be cleaned by \`dropEndedSessionFiles\` — with no double-cleaning and no gap. Read
   \`server/infra/on-listening.ts\` rather than assuming the split was faithful.

B. **Did #2176 change anything about WHEN the boot sweep runs relative to the prune?** If the
   order moved, say so — that ordering is what makes "tick only, not boot" correct.

C. **0ffb6306's claims.** Are the new file references actually right, and is there any other
   surface in this PR still naming \`server/index.ts\` for something that moved?

D. Anything else in either commit.

Do NOT re-review what round 2 already cleared unless the merge changed it.

END WITH ONE TOP-LEVEL COMMENT on PR #2179, first line a verdict marker on its own line:
  'CODEX VERDICT: LGTM'  or  'CODEX VERDICT: CHANGES REQUESTED' + what remains.
Then, on its own line:
  'FINDINGS COMPLETE: I reviewed dfdb7d56 and 0ffb6306 and this is every finding I have.'

```

</details>

<details><summary>Codex reply (verbatim stdout)</summary>

```text
CODEX VERDICT: LGTM

No findings for the two new commits.

I re-checked the post-#2176 path in \`server/infra/on-listening.ts\`: the boot sweep still runs before the orphan prune, \`survivingAfterSweep(surviving, reaped)\` still subtracts the boot-reaped ids before \`pruneOrphanSettings\` / \`pruneOrphanDrops\`, and the prune still receives the live-peer cutoff from \`earliestStartedAt(liveInstances())\`. The timer path still cleans only tick-reaped ids via \`dropEndedSessionFiles\`, so I don’t see double-cleaning or a gap.

#2176 did not change the relevant boot ordering versus the old inline \`server/index.ts\` block: sweep first, then instance registration/hooks, then orphan prune. The 0ffb6306 reference updates in the PR-changed files point to \`infra/on-listening.ts\` correctly. I found one older \`server/index.ts\` wording nearby in \`server/infra/tmux.ts\`, but it is outside this PR’s diff and outside 0ffb6306’s changed surface.

Verification run:
\`yarn vitest run test/server/session/reap-schedule.spec.ts test/server/infra/on-listening.spec.ts\` passed.
\`yarn typecheck\` passed.
\`yarn lint\` passed.

FINDINGS COMPLETE: I reviewed dfdb7d56 and 0ffb6306 and this is every finding I have.
```

</details>

Nothing stripped; no secrets appeared.

isamu:
## Triage — round 3: **LGTM**. Loop closed.

Round 2's LGTM was on `edb5d14b`. Two commits landed after it, so a round was owed:

- **`dfdb7d56`** — a merge of `origin/main`, which had moved twice: #2183 (the sibling UI PR, merged) and **#2176, which split `server/index.ts`**. That second one is why this round mattered rather than being a formality: `startReapSchedule` — the function whose behaviour this PR changes — **moved to `server/infra/on-listening.ts`**, along with the orphan prune that follows the boot sweep. Nobody had reviewed this change sitting on that split.
- **`0ffb6306`** — prose only. My comments and plan pointed readers at `server/index.ts` for the caller and the prune; after #2176 that is the wrong file, so following the reference would send someone somewhere the code no longer is.

### The invariant, re-checked on the combination

The whole change rests on *boot's reaped ids are cleaned by the orphan prune; the tick's are cleaned by `dropEndedSessionFiles`; no double-clean, no gap.* I read `on-listening.ts` myself and Codex checked it independently — same conclusion:

> the boot sweep still runs before the orphan prune, `survivingAfterSweep(surviving, reaped)` still subtracts the boot-reaped ids before `pruneOrphanSettings` / `pruneOrphanDrops`, and the prune still receives the live-peer cutoff […] The timer path still cleans only tick-reaped ids via `dropEndedSessionFiles`, so I don't see double-cleaning or a gap.

#2176 preserved the ordering faithfully and even added a comment making it explicit. The full suite was run on the merged result **before** the push — a combination neither branch had executed — and is green; Codex then re-ran the focused specs, typecheck and lint.

### One finding outside this PR

Codex spotted `server/infra/tmux.ts:321` still saying *"`server/index.ts` then deletes the settings file"*. That line is `ca65aad2`'s, in a file this PR does not touch — #2176's drift, not this change's. **Deliberately not fixed here**: widening a PR at merge time is how one change becomes three, and a reader who follows it lands on a missing reference rather than a wrong behaviour. Worth a one-line follow-up by whoever owns #2176's fallout.

### The loop

| round | verdict | why it was owed |
|---|---|---|
| 1 | LGTM — but **read-only**, the sandbox could not execute through a worktree's symlinked `node_modules` | — |
| 2 | **LGTM**, all axes, **executed** | round 1 could not run; my error in where I ran it |
| 3 | **LGTM**, no findings | round 2's head was superseded by a mainline merge that moved the caller |

**Three rounds, none of them caused by a held-back finding.** Round 1 was complete on its first pass. Rounds 2 and 3 were bought by the environment and by mainline moving underneath — the second of which is the case the loop's sync rule exists for, and it earned its round: the caller genuinely relocated.

Both reviewers clear on `0ffb6306`. Waiting on CI, then merging.

PR #2185, opened 2026-09-19
refactor: lift the Files pane's reveal and finder into a composable (#2158)

### LINKED ISSUE #2158: refactor(files): FilesPane.vue を分割する（max-lines 上限に張り付いていて、次の機能が入らない）
## いまの状態

`src/components/FilesPane.vue` はリポジトリの `max-lines`（600行、コメントと空行を除く）**ちょうど**まで来ています。#2150 → #2151（行メニューの切り出し）→ #2153 → #2156 の4本で、毎回「入れる前に何かを出す」か「入れた分だけ削る」かを迫られました。#2156 では、読み込み後の処理をオプションオブジェクトで切り出したら上限を超え、3引数の関数に書き直して収めています。

**次にこのファイルへ機能を足すには、先に何かを出す必要があります。** 上限が仕事をしている状態で、これ自体は良いことですが、毎回の PR で即興の小細工をするのは筋が悪いので、独立したタスクにします。

## いま入っているもの（おおまかな塊）

| 塊 | 中身 | ペインの ref への依存 |
|---|---|---|
| ツリー | `makeNode` / `fetchEntries` / `adoptRoot` / `paintCachedRoot` / `loadRoot` / `toggleDir` / `rows` | `roots`, `treeError`, `treeReqId` |
| 行メニュー | （#2151 で `useFilesRowMenu` に出た。残っているのは `rowActionsFor` と `runRowAction` の2つだけ） | props / emit |
| ファイルの読み込み | `loadFile` / `mayLeaveCurrent` / `adoptText` / `adoptUnpreviewable` / `restorePlace` / `applyRemembered` / `failureReason` | `openPath`, `fileError`, `unpreviewable`, `baseVersion`, `conflict`, `showPreview`, `editor`, `fileReqId` |
| 保存と競合 | `flush` / `save` / `discardAndReload` / `overwrite` / `onPageHide` | `dirty`, `saving`, `baseVersion`, `conflict`, `editor`, `openPath` |
| 外部変更の監視 | `checkForExternalChange` / `watchExternalChanges` / `EXTERNAL_CHECK_MS` | `openPath`, `saving`, `conflict`, `dirty`, `baseVersion` + `loadFile` |
| ファインダ | `finderOpen` / `onFinderPick` / `revealPath` / `rowElementFor` | `roots`, `started` |
| 起動と後片付け | `start` / `restore` / `teardown` / `reload` / `snapshot` | ほぼ全部 |

## 切り出す順番の提案（安全な順）

1. **ツリー** — `roots` とその周辺だけで閉じており、外から触るのは `rows`（表示）と `findNode`（復元とファインダ）だけ。キャッシュ（#2153）も既に別モジュール。**依存が一番少なく、一番大きい**。
2. **保存と競合** — `dirty` / `saving` / `baseVersion` / `conflict` の4つと `editor` で閉じます。競合バナーのマークアップは残す（#2151 と同じやり方で、描画は動かさない）。
3. **外部変更の監視** — 小さいのに ref を5つ参照するので、単体では割に合いません。2 のあと、保存側の状態が別モジュールに移ってから出すほうが素直です。

## やり方（#2151 で実証済み）

- **描画は動かさない**。`<Teleport>` や `<template>` はペインに残し、命令的な部分だけを composable に出す。「何が描かれるか」が動かないので、DOM 差分がそもそも発生しません。
- **切り出す前に、差分ハーネスで「変わらないこと」を記録する**。#2151 ではメニューの開き位置グリッド（四隅・画面外・中央 × 2つのマウント形態 × 2種類の行 × 2つの入口 + 全アクションのクリック）で、切り出し前後の**マークアップ・算出位置・フォーカス位置・emit がバイト単位で一致**することを確認しました。ハーネスは使い捨て、generator と property だけを恒久テストに残す。
- **既存スペックを無改変で通す**ことが第一の証拠。ペイン越しに駆動しているスペックは composable の名前を知らないので、切り出しただけでは満たせません。

## 関連

- #2151 — 行メニューの切り出し（この手順の実例）
- #2148 / #2149 — 上限に当たりながら入れた直近の2機能

isamu:
## Summary

The last of the three extractions #2158 asks for. The tree went out in #2169, the open file in
#2174, and what was left in `FilesPane.vue` that is not markup was mostly the **reveal**: open a
path and put the tree on it, with the finder panel that is one way of asking for that.

The pane is well under the size limit now, so this is not about the cap. It is about the reveal
being the one thing left in the file with three generation guards in it, none of them reachable
without mounting a pane and holding a fetch open at exactly the right moment.

Moved to `src/composables/useFilesReveal.ts`: `finderOpen` and its close, the pick handler, the
reveal generation, `revealPath`, and the row lookup. The finder's markup stays in the pane, as the
tree's and the editor's did; so does `treeEl`, because the pane renders the tree and hands the
element over as a ref.

The two halves are one composable deliberately: picking in the finder IS a reveal, and the panel
closes because the reveal is starting. Splitting them leaves two files whose only content is each
other's call.

## Items to Confirm / Review

- **`started` is passed as a getter.** `reload()` replaces that promise, so a reveal holding the
  old one would wait for a tree nobody is building. The pane's `await started` read the variable
  at the moment it ran, which is what `() => started` reproduces — worth a second look.
- **`teardown()` now calls `resetReveal()` where it bumped `revealId` inline**, and that call also
  closes the finder, which used to be a separate `closeFinder()` two lines down. Nothing between
  them observes either.
- **`rowElementFor` is exported.** It walks the rendered rows instead of building
  `[data-path="…"]` because a filename may hold a quote or a backslash — that selector is a
  `SyntaxError`, not a miss. A mounted pane cannot easily be given a file named `a"b`, so the
  reason the function exists was untestable until it had a door of its own. If you would rather it
  stayed private, the spec's other half goes with it.
- **The new spec's fake tree** implements `FilesTree` by hand. If a shared double would be better
  placed in `test/helpers/`, say so and it moves.

## How it was proved

A throwaway differential harness drove the pane through its **real UI** — the finder button, the
input, a clicked row — over eight scenarios: a deep pick, a root pick, two picks racing with the
inner listing held open, a pick that beats the root listing, a pick followed by a re-root, Escape,
a click outside, and a read that fails. It recorded the tree rows, whether the panel was up, what
it was offering, the header, the snapshot, the request sequence and the scroll count after every
step. Every record came back identical. The harness is deleted — half of it was the code it was
proving.

## The sweep found two guards nothing tested

A mutation sweep over the reveal's decisions: the pre-existing pane specs catch 6 of 10. The four
they miss are covered now, and two are worth naming:

- **the guard after the startup wait** — both picks can be parked in the *same* startup, because
  the `files-find` shortcut opens the panel over a pane whose tree is still being read. The first
  to resume has to find that it already lost.
- **the guard before scrolling** — the file was read, and the pick changed while it was being
  read. The tree now belongs to the other file, so scrolling to this one moves the reader off it.

Observing the second needed a rendered tree, because what it changes is only the scroll: the spec
builds rows whose `scrollIntoView` records itself, since jsdom has neither layout nor that method.

The other two are `reset()` bumping the generation, and `rowElementFor` matching a whole path
rather than a prefix. All ten go red now.

## Verification

`yarn lint`, `yarn typecheck`, `yarn build` and the full `vitest` suite are green.

## User Prompt

- 「#2158」 — work the issue.
- 「出来る限り小さくしてね。分割可能なのは分割して。」 — make `FilesPane.vue` as small as possible;
  split out everything that can be split.

Closes #2158.

work in mulmoterminal5


test/src/composables/useFilesReveal.spec.ts
```diff
@@ -0,0 +1,255 @@
...
+  ...over,
+});
+
+/** A tree that answers from a flat map, and records which directories were asked to open. */
+function fakeTree(dirs: string[], opened: string[], hold?: Promise<void>): FilesTree {
+  const nodes = new Map(dirs.map((path) => [path, node(path, { dir: true })]));
+  return {
+    roots: ref(null),
+    error: ref(null),
+    rows: ref([]) as unknown as FilesTree["rows"],
```
isamu: P2: This spec adds a double assertion (`as unknown as`) even though the repo rules explicitly forbid `as` in specs, and this fake does not need it. It also weakens the fake `FilesTree`: production `rows` is a computed ref, while this is a normal ref forced through the type system. I found no other instance of this pattern in the PR. Smallest fix: import `computed` from Vue and make this `rows: computed(() => [])`, which satisfies `FilesTree["rows"]` without an assertion.

isamu:
Reviewed the full PR diff for behavior preservation. I left one inline finding.

Findings:
- P2 `test/src/composables/useFilesReveal.spec.ts:31`: the new fake `FilesTree` uses `rows: ref([]) as unknown as FilesTree["rows"]`. Repo rules forbid `as` in specs, and this also makes the fake less faithful than needed because production `rows` is a computed ref. I found no other PR site with this pattern. Smallest fix: import `computed` and use `rows: computed(() => [])`.

Axes:
- correctness: no behavior drift found in the lifted reveal/finder code; the deleted pane logic and added composable preserve the same expansion/open/scroll sequence.
- behaviour preservation: no findings on evaluation order, `started` is still read at await time through the getter, and the destructured returns are refs/functions rather than reactive values being unwrapped into stale snapshots.
- concurrency / async ordering: no findings; the post-startup, per-ancestor, and pre-scroll generation guards are present, and `reset()` bumps the same reveal generation. Open-file and tree generations remain owned by their own composables.
- blast radius: no findings; `useFileSearchPanel` still receives and awaits the same `revealPath` contract, and the `files-find` shortcut still opens the finder through the exposed ref-backed function.
- teardown ordering: no finding. `resetReveal()` now closes the finder before `file.teardown()` whereas the old code closed it just after, but Vue does not render the `v-if` removal synchronously between those statements, so I do not see an observable behavior change.
- error handling: no findings; file-open failures still resolve through `openPath` not matching, and directory listing failures remain handled by `tree.toggleDir` collapsing/not opening that ancestor.
- types: finding above for the forbidden `as unknown as`; otherwise no `any` / `@ts-ignore` / exact-optional issue found in the PR additions.
- tests: the new tests cover the startup guard, expansion race, reset, and pre-scroll guard. The only test-quality issue I found is the fake tree `rows` assertion called out inline; `scrollIntoView` recording is appropriate for jsdom and does not hide the reveal behavior under test.
- API surface: no finding. `rowElementFor` is newly exported, but the PR states the testing reason and the export is narrow.
- dead / stale: no findings; removed imports and moved pane code appear cleaned up.
- repo rules: finding above for `as`; no re-export issue, no magic-number issue introduced in the moved code, and no string `+=` pattern found.

Verification run:
- `npx vitest run test/src/composables/useFilesReveal.spec.ts` passed.
- `yarn typecheck` passed.

FINDINGS COMPLETE

isamu:
## Review ledger — after round 1

| # | round | finding | disposition | why / where |
|---|-------|---------|-------------|-------------|
| 0 | pre-flight | sandbox (from #2174, same machine and CLI): tests, typecheck, socket bind, api.github.com all exit 0; plain `rm` works, `rm -f` refused by the command guard | OK | no mutation sweep asked of codex — I ran it and report it below |
| 1 | 1 | P2 `useFilesReveal.spec.ts:31` — `rows: ref([]) as unknown as FilesTree["rows"]`: a cast the repo forbids, and a less faithful double than needed (production `rows` is a computed ref) | FIXED | `computed(() => [])`, no assertion. No other site in the PR has the pattern |

### What was run, not argued

- Differential harness over the pane's **real UI** (finder button, input, clicked row), eight
  scenarios: deep pick, root pick, two picks racing with the inner listing held, a pick that beats
  the root listing, a pick then a re-root, Escape, a click outside, a read that fails. Tree rows,
  the panel and its offerings, the header, the snapshot, the request sequence and the scroll count
  recorded after every step. Every record identical. Harness deleted.
- Mutation sweep over the reveal's ten decisions, **re-run after this round's spec change** to
  check it had not weakened them: **10 of 10 caught**, unchanged. The pre-existing pane specs
  catch 6; the four they miss are the two generation guards nothing had tested (a reveal parked in
  the startup, and one whose open finished after the pick changed), `reset()` bumping the
  generation, and `rowElementFor` matching a whole path rather than a prefix.
- `yarn lint`, `yarn typecheck`, `yarn build` and the full `vitest` suite green.

Re-flagging a FIXED row is welcome **if the resolution is wrong** — say which row and why.

src/composables/useFilesReveal.ts
```diff
@@ -0,0 +1,101 @@
+// "Show me this file": open a path AND put the tree on it, plus the finder panel that is one way
+// of asking for that. Lifted out of FilesPane.vue with the tree (#2169) and the open file (#2158)
```
isamu: P3: This new comment says the open-file extraction was `#2158`, but the PR context and plan both say that extraction was `#2174`; `#2158` is the umbrella issue for this split. I found no other stale issue reference in the PR additions. Smallest fix: change this parenthetical to `#2174` so the breadcrumb points at the actual prior extraction.

isamu:
Reviewed the round-2 change and re-read the whole PR diff at head `9c861b3b`. I left one inline P3 note; no behavior-preservation defect found.

Findings:
- P3 `src/composables/useFilesReveal.ts:2`: the new source comment says the open-file extraction was `#2158`, but the context/plan identify that prior extraction as `#2174`. Every other PR site I checked either uses the correct `#2174` or does not cite the open-file extraction. Smallest fix: change that parenthetical to `#2174`.

Axes:
- correctness: no behavior drift found in the lifted reveal/finder code. The old sequence `++revealId`, await startup, expand ancestors outermost-first, open, next tick, generation guard, scroll, then compare `openPath` is preserved.
- behaviour preservation: no findings on evaluation order. `started` is still read when `revealPath` reaches the await via `deps.started()`, matching the old `await started`; destructuring returns refs/functions, so I do not see reactivity lost by the pane.
- concurrency / async ordering: no findings. All three generation guards are present; two picks racing, picks parked in startup, open finishing after a newer pick, and `reset()` during expansion/open are covered by the same reveal counter. The open-file and tree generations remain separate and owned by their composables.
- blast radius: no findings. `useFileSearchPanel` still receives the same `revealPath(path) => Promise<boolean>` contract and only calls `revealLine` after a true result. The `files-find` shortcut path still opens the pane and toggles the finder through the exposed ref.
- teardown ordering: no behavior finding. `resetReveal()` now closes the finder before `file.teardown()` rather than two statements later, but there is still no render/await point between those statements where the old order was observable.
- error handling: no behavior finding. File-open failures through `file.load` still resolve with `openPath` unchanged and return false; directory listing failures remain swallowed by `tree.toggleDir`, which collapses the directory and lets reveal proceed exactly as before.
- types: the round-1 `as unknown as` is fixed. `computed(() => [])` is a read-only `ComputedRef`, matching production `FilesTree.rows`; neither `useFilesReveal` nor the spec writes to `rows`, so this does not turn a relevant write into a hidden warning. No `as`, `any`, or `@ts-ignore` found in the PR files after the fix.
- tests: no finding. The fake tree is faithful enough for reveal: `toggleDir` mutates the handed node like production, and flat `findNode` does not hide the reveal behavior under test because each scenario is asserting order/cancellation around known ancestors, not recursive discovery. The `scrollIntoView` stub is load-bearing in the useful way: jsdom has no layout/method, and the spec uses it only to observe whether the pre-scroll guard fired.
- API surface: no finding. `rowElementFor` is newly exported, but it is a narrow export with a stated reason: testing selector-hostile paths without mounting the whole pane.
- dead / stale: one P3 stale issue reference noted inline; no unreachable pane code or leftover imports found.
- repo rules: one P3 comment breadcrumb issue only. The prior forbidden assertion is gone, and I did not find new forbidden `as` / `any` / `@ts-ignore`, string `+=`, or unstated re-export patterns in the PR files.

Verification run:
- `npx vitest run test/src/composables/useFilesReveal.spec.ts` passed: 20 tests.
- `yarn typecheck` passed.

CODEX VERDICT: LGTM
FINDINGS COMPLETE

isamu:
## Review ledger — round 2 (final)

Round 1's P2 (the `as unknown as` on the fake tree's `rows`) is fixed in `9c861b3b`. Round 2 on
that head: **CODEX VERDICT: LGTM**, no behaviour findings on any axis, plus one P3 — a comment in
`useFilesReveal.ts` cited `#2158` for the open-file extraction where it meant `#2174`.

That is now fixed. **The only change since the reviewed head is that one comment line**, and the
diff is exactly:

```diff
-// of asking for that. Lifted out of FilesPane.vue with the tree (#2169) and the open file (#2158)
+// of asking for that. Lifted out of FilesPane.vue with the tree (#2169) and the open file (#2174)
```

No executable line differs from the head that was signed off, so the review's subject is
unchanged — CLAUDE.md's rule is that prose does not buy another round, and this is why.

### The whole verification, in one place

- **Differential harness** over the pane's real UI (finder button, input, clicked row), eight
  scenarios: deep pick, root pick, two picks racing with the inner listing held, a pick that beats
  the root listing, a pick then a re-root, Escape, a click outside, a read that fails. Tree rows,
  the panel and its offerings, the header, the snapshot, the request sequence and the scroll count
  recorded after every step. Every record identical. Harness deleted.
- **Mutation sweep**, run twice (once after the round-1 spec change): **10 of 10 caught** both
  times. The pre-existing pane specs catch 6; the four they miss are the two generation guards
  nothing had tested — a reveal parked in the startup when a later one arrives, and one whose file
  finished reading after the pick changed — plus `reset()` bumping the generation and
  `rowElementFor` matching a whole path rather than a prefix.
- `yarn lint`, `yarn typecheck`, `yarn build` and the full `vitest` suite green.

Merging on green CI.

Issue #2177, opened 2026-09-19
fix(settings): 定期スイープの間隔が Settings から触れず、タイマー有効時の行の文言が「次回起動時に終了」のまま


isamu:
## 観測した事実

#2167 で `sessionReapIntervalHours` が入り、スイープが起動時だけでなくタイマーでも走るようになった。
しかし Settings 側は #2167 以前のままで、2点ずれている。

**1. Settings にコントロールが無い**

`test/server/config/settings-coverage.spec.ts` のマップで、この値だけ `ui` が付いていない:

```ts
sessionIdleReapDays:      { ui: true, skill: CONFIG_SKILL },
sessionReapIntervalHours: {           skill: CONFIG_SKILL },
```

しきい値（日数）は **Sessions that survived a restart** の一覧のすぐ横にステッパーがあるのに、
その間隔だけは `~/.mulmoterminal/config.json` を手で編集しないと触れない。
既定が `0`（OFF）なので、**この機能は設定ファイルを開いた人にしか存在しない**。

**2. タイマーが有効なとき、行の文言が事実と違う**

`src/i18n/en.ts` / `ja.ts`:

```ts
doomed: "ends at next start",
doomedTitle: "... the server ends it at its next start",
reapEnded:   "ended when the server next starts",
```

`sessionReapIntervalHours` を 0 より大きくすると、そのセッションは **次回起動を待たずに**
次のスイープで終了する。行は依然「次回起動時に終了」と表示するので、有効化した人にだけ
確実に嘘になる。これは `sessionIdleReapDays` を変えたとき行を再読み込みする理由
（CodeRabbit on #1486: 古いしきい値に対する答えを表示し続けない）と同じ種類のずれ。

## 提案

- `SurvivingSessionsSection.vue` に2つめのステッパーを置く（しきい値の隣）。しきい値が `0`
  のときは間隔に意味が無いので disabled にし、その旨を出す。
- タイマーが有効なときは `ends on the next sweep` 系の文言に切り替える。
- `settings-coverage.spec.ts` のマップを `ui: true` に更新する。

## 範囲外

既定値は `0`（OFF）のまま変えない。#2167 の判断どおり。

## 関連

- #2165 / #2167 定期スイープの導入
- #1486 起動時スイープと、行が「次回起動時に終了」と言う仕組み

isamu:
Working on this in `mulmoterminal4`.

- started — 2026-09-19 09:01 UTC
[...]

PR #2183, opened 2026-09-19
fix(settings): give the sweep cadence a control, without claiming what is armed (#2177)

isamu:
## Summary

Closes #2177.

#2167 added the config key `sessionReapIntervalHours` (default `0` = off) on the server and
left the Settings UI untouched. `settings-coverage.spec.ts` listed it as `{ skill: ... }` —
config file only — so **the feature did not exist for anyone who had not opened
`config.json`**. A setting whose default is "does nothing" is the one that most needs
somewhere to be found.

This adds a second stepper beside the existing idle-days one. **The default stays `0`: this
makes the setting reachable, not enabled.**

Three states on the cadence row, because two would mislead:

| threshold | cadence | the row says |
|---|---|---|
| on | `> 0` | Will repeat every N hour(s) **after the next server start**. |
| on | `0` | **Will run only at server start.** A server left running then never looks again. |
| **off** | any | The sweep is off above, so there is nothing for this to repeat. *(stepper disabled)* |

## What this PR deliberately does NOT do

#2177 had a second half: a doomed row reads `ends at next start`, which **understates** when a
timer was armed at boot. My first attempt fixed that by keying the row's wording off the saved
cadence. **Codex caught it as a P2 and was right** — the timer is armed once, at boot
(`startReapSchedule` has one caller, `server/index.ts:991`; `armTimer`'s handle is
function-local; nothing calls `clearInterval`), so the saved number is not what the running
process is doing. That wording was false in both directions: from the moment it is saved until
the next restart, and the other way when someone sets it back to `0`.

So the rows are **unchanged from `main`**, and that is pinned by a spec at both `0` and a
positive cadence so the obvious-looking wording cannot be re-derived. Saying what is *actually*
armed needs the server to report it — **#2184**, which carries the requirement this round
discovered and the live-re-arm alternative Codex costed and declined.

The understatement remains, exactly as on `main`. This PR introduces no false claim.

## Items to Confirm / Review

1. **The three-state cadence row** — worth checking the wording reads right in both locales.
   The English is Codex's sentence, not mine; it disputed my "Repeats every…" as still
   present-tense and it was right.
2. **The cadence deliberately does NOT re-read the list.** The days do, because `reapable` is
   the server's answer against the old threshold. The cadence changes which rows are reapable
   not at all, so that answer is still current. A spec pins it.
3. **Docs beyond the strict diff.** The skill said "config-file only; there is no Settings
   control" — now false. Neither `config.md` guide mentioned the key **at all** (#2167 did not
[...]

PR #2186, opened 2026-09-19
fix(settings): reach the sweep interval from Settings, and say which start it applies from (#2177)

isamu:
## Summary

Addresses **part** of #2177 — item 1 in full, item 2 deliberately not, for a reason the issue could not have known. Not `Closes`.

**Item 1 — the interval is reachable.** #2167 added `sessionReapIntervalHours` and only the server took the change. The threshold has a stepper beside the list it acts on; the interval had none, and its default is `0`, so a feature that must be turned ON existed only for people who open `config.json`. There is now a second stepper under the threshold's, greyed **and** `:disabled` while the threshold is `0` — an interval on a sweep that never runs decides nothing, and the container's `pointer-events-none` stops the mouse only, so the stepper's own attribute is what keeps a keyboard user from saving a value the screen calls unavailable.

**Item 2 — the row's wording — is not done here, on purpose.** The issue asks for `ends at next start` to become `ends on the next sweep` once an interval is set. I checked whether that sentence would be true and it is not, in the case that matters most:

> `startReapSchedule` arms the timer **once, at boot**. It takes `intervalHours` as a plain number while it takes `idleDays` as a `() => number`, and `config-routes.ts` states the reason out loud — *"re-arming on every config POST would let a stream of edits reset the countdown forever."*

So between saving an interval and restarting, **nothing repeats**. Switching the badge on the saved value would put a fresh false sentence on the row for exactly the person who just turned the feature on — which is the defect this issue is about. Saying it truthfully needs the server to report what it **armed**, and nothing records that anywhere. The component spec now pins the current wording *with that reasoning attached*, so the next reader does not "fix" it back.

What the UI does claim is true and checked: `Only at server start.` at `0`, and `From the next server start, also every N hour(s).` once raised.

Design notes, including the half of the premise that *was* true: [`plans/fix-2177-reap-interval-settings.md`](plans/fix-2177-reap-interval-settings.md).

## Items to Confirm / Review

- **The scope split is the main thing to agree or disagree with.** Shipping item 2 as specified was the alternative, and I judged a new false sentence worse than a missing improvement. If you would rather have the badge switch on the saved value and accept the pre-restart window, say so and it is a small change.
- **The follow-up needs a boot signal**, and the files that would carry it (`reap-schedule.ts` / `server/index.ts`) are where **#2178** and **#2168** are working right now. That is the other reason it is not in this PR. Happy to file the follow-up issue with the finding written up — I did not file it unasked.
- **No server file changed.** `app-config.ts` and `config-routes.ts` already carried the field end to end; only the client had never read it.
- **`mulmoterminal-config`'s SKILL.md** lost its "Config-file only; there is no Settings control" bullet and gained the reason the badges still say "next start". That skill owns the key, per the repo's one-setting-one-skill rule.

## Verification

Gates on the merged head: `yarn lint`, `yarn typecheck`, `yarn build`, and the affected specs — all clean. `yarn format` fails on `docs/_layouts/default.html`, a vendored Jekyll layout; it fails identically on a clean `main` checkout, so it is not this change.

**Break-verified rather than assumed.** Three mutations, each turning a test red, source restored byte-identical after each:

| mutation | result |
|---|---|
| the wording switch reinstated | 2 red |
| `:disabled` dropped from the stepper | 1 red |
| the save pointed at the wrong config field | 1 red — **and** it fails `settings-coverage.spec.ts`, which is how the `ui: true` flip was shown to be a real check (it greps the UI tree for an actual `postConfigField` write) rather than bookkeeping |

**Driven in a real browser against a real server**, with `HOME` pointed at a scratch dir so the maintainer's own config is never touched:

```
PASS  the interval stepper exists in Settings at all
PASS  with the interval at 0 it says "Only at server start."
PASS  the hint names the start it applies from once raised
PASS  the value reached ~/.mulmoterminal/config.json on disk
PASS  the interval stepper is disabled once nothing is ever ended
PASS  and it says why
PASS  a forced click on the disabled stepper writes nothing
no console errors
```

Worth recording because it nearly went wrong: **the first attempt was not isolated.** `npx` resets `HOME`, so the backend came up on the real `~/.mulmoterminal/config.json`, and I only noticed because the directory list showed 28 real presets. Running the binaries directly instead of through `npx` fixed it, and isolation was then *confirmed* by asking the server (`cwdPresets: 0`) before clicking anything. Nothing was written to the real config — checked.

### One failure that is not this change, and is worth someone's attention

A full `yarn test` failed once on `test/server/session/session-settings.spec.ts > "returns the JSON itself when nothing in it is secret"`, which asserts a settings file does **not** exist. It passes standalone. The cause is that the spec keys off `os.homedir()` — the **shared, real** `~/.mulmoterminal/settings/` — while eight vitest processes from other checkouts were running the same spec concurrently. With five checkouts of this repo live at once that will keep firing, and the fix is test isolation, not this diff.

## User Prompt

- 他のセッションで手を付けてない最新issue
- yatte


isamu:
## Codex cross-review — tier and pre-flight

**Tier: C (complex).** Not by diff size — I could not write the one sentence that would defend tier S:

- **It is a UI for enabling a destructive operation.** The value this stepper writes causes the server, at its next start, to **end tmux sessions on a timer**. A control that makes session-killing one click away is not a bounded UI concern.
- **The PR's central decision rests on a claim I made, not on the diff.** Item 2 of #2177 is deliberately not implemented because `startReapSchedule` arms the timer once at boot. If that claim is wrong, the PR's main judgement is wrong — so it gets its own named axis rather than being taken on trust.
- Shared module-level state (`sessionReap.ts` singletons) read by two consumers, and an async save whose ordering against the existing threshold save is worth a look.

So every round names the risk axes and demands a per-axis answer. Exit bar is one clean round on a head nothing was pushed to.

### Pre-flight — what the Codex sandbox can actually run

| probe | result |
|---|---|
| `npx vitest run …survivingSessionsSection.spec.ts` | exit 0 |
| `npx vitest run …settings-coverage.spec.ts` | exit 0 |
| `npx vue-tsc -b` | exit 0 |
| `curl https://api.github.com` | exit 0 |
| `gh pr view 2186` | exit 0 |
| `rm -f` | **rejected by policy before execution** |

Codex is therefore told **not** to run a mutation sweep — its restore would fail and every count afterwards would measure the harness. Any mutation it wants, it names and I run.

### Pre-existing red, recorded once so no round re-derives it

- `yarn format` fails on `docs/_layouts/default.html`, a vendored Jekyll layout. **The identical failure occurs on a clean `main` checkout** (verified in a separate worktree). Not this diff.
- `test/server/session/session-settings.spec.ts` can fail in a full `yarn test` because it keys off `os.homedir()` — the **shared real** `~/.mulmoterminal/settings/` — while other checkouts of this repo run the same spec concurrently. It passes standalone. Not this diff; worth its own issue.
- CI on head `27e0d192`: **18/18 SUCCESS**.

src/i18n/en.ts
```diff
@@ -134,6 +134,12 @@ export const en = {
       reapHint:
         "A session nothing is using — nobody attached, no output for this long — is {ended}. Its conversation is kept. Set this to 0 to never end one automatically.",
       reapEnded: "ended when the server next starts",
+      reapIntervalLabel: "Run the sweep again while the server is up",
+      reapIntervalStepper: "the hours between sweeps while the server is up",
+      reapIntervalUnit: " hours",
+      reapIntervalOff: "Only at server start.",
```
isamu, line 140: P2: This hint is false for servers that already armed, or are still running, an interval timer. The timer is captured once at boot (`server/index.ts:991` passes `getSessionReapIntervalHours()` into `startReapSchedule`, and `server/session/reap-schedule.ts:27-39` only arms `setInterval` there), while `/api/config` saves update `config` without re-arming (`server/config/config-routes.ts:391-449`). That means a user who booted with `sessionReapIntervalHours: 6` sees the saved value `6` but this text says “From the next server start” even though the destructive sweep is already active; worse, if they lower the saved value to `0`, this line says “Only at server start” while the old timer continues until restart. The row copy has the inverse problem in that same pre-existing non-zero state: `reapable` rows can be ended on the next armed timer tick, not only at the next start.

Every site I found for this finding: `src/i18n/en.ts:122-123,136,140-141`, `src/i18n/ja.ts:119-120,132,136-137`, `test/src/components/settings/survivingSessionsSection.spec.ts:150-163,188-201`, `server/skills/mulmoterminal-config/SKILL.md:311-315`, and the plan text at `plans/fix-2177-reap-interval-settings.md:19-41,90-92`.

Smallest resolving change: either expose the boot-armed interval (for example an `armedSessionReapIntervalHours`/runtime status field) and render both current-server and saved-next-start states, or weaken all of these strings to explicitly say they describe only the saved value for the next start and that this running server keeps whatever interval it armed at boot until restart. What would change my mind: a source path showing that saving `sessionReapIntervalHours` cancels/re-arms the existing timer, or that the Settings UI already receives the boot-armed interval and these strings are conditioned on it.

isamu:
- P2 inline: the UI only knows the saved `sessionReapIntervalHours`, so its hints and row copy can be false when the running server already armed a different interval at boot. A server booted with interval 6 is already sweeping, but Settings says “From the next server start”; after saving 0, the old timer can still run while the UI says “Only at server start.”

| Axis | Result |
|---|---|
| A | Central claim mostly true: `server/index.ts:991` calls `startReapSchedule({ intervalHours: getSessionReapIntervalHours(), idleDays: getSessionIdleReapDays, ... })` once at boot; `server/session/reap-schedule.ts:27-39` arms `setInterval` from that captured number; `server/config/config-routes.ts:391-449` persists config without re-arming. The repeat sweep does re-read only `idleDays` at each tick (`reap-schedule.ts:30-32`, tested at `test/server/session/reap-schedule.spec.ts:54-60`). No live interval re-read/re-arm path found. |
| B | Finding: the wording is false when saved interval and boot-armed interval differ, and also for a pre-existing non-zero interval. Sites listed in the inline: `src/i18n/en.ts:122-123,136,140-141`, `src/i18n/ja.ts:119-120,132,136-137`, tests, skill prose, and plan prose. |
| C | Same P2 covers the destructive-affordance risk: one click is not itself a new route/security issue, and the threshold-disabled gate is correct for the next-start saved value, but a user can save 0 and still have the already-armed destructive timer running while the UI implies it will not repeat. |
| D | No separate finding. The module singleton refs match nearby patterns, partial config writes are locked/merged server-side, and threshold+interval saves to different fields should not clobber each other. Not reloading the list after interval-only saves is correct because the server recomputes row `reapable` from threshold/pty/session facts, not cadence. |
| E | No findings. Min/max/0 boundaries use `sanitizeReapIntervalHours` (`common/sessionReap.ts:67-70`) and `SettingsStepper` disables at min/max; junk/out-of-range falls back to default 0. |
| F | No findings. This writes one existing config key through the existing `/api/config` partial-update route; no new auth, filesystem, command, or traversal surface found. |
| G | Existing focused tests pass: `npx vitest run test/src/components/settings/survivingSessionsSection.spec.ts test/server/config/settings-coverage.spec.ts test/server/session/reap-schedule.spec.ts test/common/sessionReap.spec.ts` and `npx vue-tsc -b`. The new assertions are meaningful for the implemented saved-value behavior, but they miss the armed-vs-saved state that caused the P2. I do not need a mutation run from you right now. |
| H | No findings. API/schema/on-disk compatibility is preserved; the key already exists in `AppConfig`, defaults, sanitize/load/merge/public config, and this PR only adds UI adoption/writes. A possible fix may need an additive runtime/public field for the armed interval. |
| I | No findings. The composable and interval stepper follow the `worklog.ts` + `SessionSection.vue` precedent, including disabling the cadence control while the governing feature is off. |
| J | No blocking findings beyond B. English and Japanese keys are present in lockstep. The `{hours} hour(s)` singular form is clunky but not wrong under the repo’s existing style; the unit string is consistent with `SettingsStepper`’s unit contract. |
| K | Prose finding, covered by the P2: “They describe what this running server will do, and this server was armed at ITS boot.” in `server/skills/mulmoterminal-config/SKILL.md:314-315` would cause the wrong change/decision if believed, because a server armed at boot with a non-zero interval can end a `reapable` row on the next timer sweep, not only at the next start. Severity is P2 as part of the same UI truth issue, not a separate prose-only N. |

isamu:
## Codex exchange — round 1

Two calls: the review (step A) and the in-round rebuttal (step C-bis, run BEFORE the push so no disagreement costs a round). Codex posted its own verdict and inline P2; below are my halves verbatim, and its C-bis reply, which DISPUTED two of seven and was right both times.

<details><summary>Call 1 — review prompt sent</summary>

```text
Review PR #2186 at https://github.com/receptron/mulmoterminal/pull/2186.

The branch is checked out here. Review range:
  git diff 9ac9c7e0234b1c6d9afb26df313783e340e788af...HEAD

Post inline comments via `gh api repos/receptron/mulmoterminal/pulls/2186/comments`
and the final verdict via `gh pr comment 2186 --repo receptron/mulmoterminal`.

── SANDBOX LIMITS YOU MUST WORK WITHIN ──────────────────────────────────────
`rm` is REJECTED by policy in your sandbox. DO NOT run a mutation sweep — your
restore will fail and every count you take afterwards measures the harness, not
the tests. If you want a mutation performed, NAME it (file, exact edit, expected
result) and I will run it and report back.
You CAN run `npx vitest run <file>` and `npx vue-tsc -b` (both exit 0).
Do NOT run a full `yarn test`: `test/server/session/session-settings.spec.ts`
keys off `os.homedir()` — the shared real `~/.mulmoterminal/` — and four other
checkouts of this repo are running concurrently, so it fails for reasons that
are not this diff.

── WHAT THIS PR IS, AND WHAT IT DELIBERATELY IS NOT ─────────────────────────
Issue #2177 has two items.

Item 1, DONE: `sessionReapIntervalHours` (added by #2167) had no Settings
control, so with a default of 0 the feature existed only for people who edit
config.json. This adds a second stepper under the idle-days one.

Item 2, NOT DONE ON PURPOSE, and this is the judgement to attack: the issue asks
that the row badge change from "ends at next start" to "ends on the next sweep"
whenever the interval is non-zero. I declined because `startReapSchedule`
(server/session/reap-schedule.ts) arms the timer ONCE at boot — it takes
`intervalHours` as a plain number while it takes `idleDays` as a `() => number`,
and server/config/config-routes.ts states the reason ("re-arming on every config
POST would let a stream of edits reset the countdown forever"). So between
saving and restarting nothing repeats, and a badge switching on the saved value
would be a NEW false sentence aimed at the person who just enabled the feature.

This PR is tier C. The first three are why.

  A. IS MY CENTRAL CLAIM TRUE? Verify, from the source and not from my summary,
     that the timer is armed once at boot and that saving a new interval does
     NOT re-arm the running server. If it DOES re-arm — or if there is any path
     that re-reads the interval while up — then declining item 2 was wrong and
     this PR should implement it. Say which, with file:line.
  B. IS THE UI'S OWN WORDING TRUE? "Only at server start." at 0, and "From the
     next server start, also every N hour(s)." once raised. Is either false or
     misleading for any reachable state — including: a user who had a non-zero
     interval in config.json BEFORE this server booted (the timer IS armed for
     them; does the UI mislead them now?), and the case where the saved value
     differs from the armed one.
  C. ENABLING A DESTRUCTIVE OPERATION. The value this stepper writes causes the
     server to END tmux sessions on a timer at its next start. Is one click with
     no confirmation the right affordance? Is the disabled-while-threshold-is-0
     gate the correct and complete condition, or can a user reach a state where
     the sweep runs on a timer without having understood it?
  D. SHARED MODULE STATE + ASYNC. `src/composables/sessionReap.ts` holds
     module-level singleton refs read by `useAppConfig.ts` and the section. Two
     saves (threshold and interval) can be in flight at once; `nudgeIdleDays`
     re-reads the session list and `nudgeIntervalHours` deliberately does not.
     Any ordering, lost-update or stale-read problem? Is the "no reload" choice
     right — is there anything the server recomputes from the interval?
  E. correctness and edge cases (min/max clamping, the 0 boundary, sanitize)
  F. security (this writes a config key through an existing route — anything new?)
  G. tests: does a test go red if the new behaviour breaks? I claim three
     mutations each turn a test red — wording switch reinstated (2 red),
     `:disabled` dropped (1 red), save pointed at the wrong field (1 red, plus
     settings-coverage). Is any new assertion vacuous or green for the wrong
     reason? Name a mutation you doubt and I will run it.
  H. API / schema / on-disk compatibility
  I. consistency with the rest of the codebase — especially `worklog.ts` +
     `SessionSection.vue`, which is the existing interval-stepper precedent I
     followed.
  J. accessibility and i18n lockstep. This repo has en + ja only. New keys went
     into both. The `{hours} hour(s)` form and the " hours" unit are worth a
     look for the singular case.
  K. PROSE: report a sentence that would cause a WRONG CHANGE if believed. File
     anything merely behind as severity N.

1. SEVERITY — P1 / P2 / P3 / N. N is prose-only, does NOT block the verdict.
2. EVERY SITE — grep for it and list them all now.

| 0 | pre | sandbox probe | RECORDED | vitest/vue-tsc/network/gh all exit 0; `rm` REJECTED → no mutation sweep |
| 0b | pre | pre-existing red | RECORDED | `yarn format` fails on docs/_layouts/default.html on clean main too; session-settings.spec.ts flakes under concurrent checkouts via shared `os.homedir()`; CI on this head is 18/18 green |

No review findings yet — this is round 1.

  Then the axis table, one row per axis A-K.
  Then, on its own line:
  Omit that line if untrue and say what you did not reach.

```

</details>

<details><summary>Call 1 — Codex stdout (the verdict + inline P2 are its own comments above)</summary>

```text
Posted the review.

I left one inline P2 on `src/i18n/en.ts:140` covering the saved-vs-boot-armed interval mismatch, and posted the required A-K top-level comment with tests/typecheck results.

Verification run:
`npx vitest run test/src/components/settings/survivingSessionsSection.spec.ts test/server/config/settings-coverage.spec.ts test/server/session/reap-schedule.spec.ts test/common/sessionReap.spec.ts` passed.
`npx vue-tsc -b` passed.
```

</details>

<details><summary>Call 2 — my resolution put back to Codex before pushing</summary>

```text
Follow-up on PR #2186, round 1. Do NOT re-review the PR and do NOT post anything.

I ACCEPTED your P2 in full. The fix is already in the working tree, unpushed —
read it with `git diff`. I took your SECOND resolution ("weaken all of these
strings…"), not the runtime-field one, and I generalised it. Say ACCEPTED or
DISPUTED for each numbered item, with specific evidence if DISPUTED.

1. THE RULE I APPLIED, instead of fixing the strings case by case:
   **no string in this screen may name WHEN the next sweep is.** The browser
   sees only the saved cadence; the armed one is unknowable to it. So:
     - hints state what is SAVED: "Saved: at server start only." /
       "Saved: at server start, then every N hour(s)."
     - one FIXED note, shown in every state: "The cadence is read when the
       server starts, so a change here applies from the next one."
     - the row names the EVENT: badge "due to be ended", title "… the next
       sweep ends it", hint "ended by the next sweep".
   Question: is every one of those true in ALL FOUR (armed, saved) states —
   (0,0), (0,N), (N,N), (N,0)? Name any state where one is still false.

2. I did NOT add the runtime/armed field you suggested in axis H. Reason: it is
   not needed to stop the screen lying (naming the event does that), and
   `server/session/reap-schedule.ts` — the file that would carry it — is
   changed by open PR #2179 right now (verified via `gh pr view 2179 --json
   files`). I recorded it as a follow-up instead.
   Question: is naming the event genuinely sufficient, or does something still
   require the armed value to be correct rather than merely more informative?

3. THE THRESHOLD ASYMMETRY, which I now state in code and in the skill:
   `startReapSchedule` takes `idleDays` as a `() => number` and re-reads it per
   tick, so a saved THRESHOLD applies to the running server immediately, while
   the cadence does not. That is why the threshold's hint may speak in the
   present tense and the cadence's may not.
   Question: is that true? In particular, does a saved threshold of 0 stop an
   ALREADY-ARMED timer from ending anything on its next tick?

4. YOUR AXIS K, the SKILL.md sentence. I deleted the claim that the badges
   describe what the running server will do, and replaced it with the naming
   rule plus the asymmetry in item 3.
   Question: does the new text still assert anything false?

5. SCOPE. Changing the row copy means this PR now closes #2177 item 2 — with a
   different sentence from the one the issue proposed, because the proposed one
   is false in state (N,0) and (0,N)→post-save. The PR body and plan will say
   so. Question: do you agree item 2 is genuinely CLOSED by naming the event,
   rather than deferred?

6. TESTS. The row assertions now run over BOTH cadence states (0 and 6) and
   also assert the absence of "next start" and "From the next server start", so
   a future edit cannot make the wording conditional again. Break-verified:
   restoring the clock copy → 2 red; making the hint speak for the running
   server → 2 red; removing the note → 1 red; source restored byte-identical.
   Question: is any of those assertions vacuous, or green for the wrong reason?
   Name a mutation and I will run it — remember `rm` is blocked for you.

7. SITES. You listed en.ts, ja.ts, the spec, SKILL.md and the plan. I also swept
   `src/composables/sessionReap.ts`, `src/composables/useSurvivingSessions.ts`,
   `src/components/settings/SurvivingSessionsSection.vue` comments, and
   `docs/guide/{en,ja}/config.md`. I deliberately did NOT touch the dated
   release pages (`docs/guide/*/v4.6.1.md`, `docs/ChangeLog.md`) that quote the
   old badge, because this repo's CLAUDE.md says a dated snapshot must never be
   edited to match new behaviour.
   Question: any site still carrying the old claim that I missed?

Reply as plain text. If you DISPUTE any, say which single change would settle it.
```

</details>

### Call 2 — Codex reply, verbatim

```text
1. ACCEPTED. In (0,0), (0,N), (N,N), and (N,0), the new UI strings either state the saved cadence or name the event. I don’t see a state where “Saved: …”, the fixed cadence note, or “the next sweep ends it” becomes false.

2. ACCEPTED. Naming the event is sufficient to stop the screen lying. The armed value would let the UI be more informative, but nothing in the fixed wording requires knowing it.

3. ACCEPTED. `startReapSchedule` captures `intervalHours`, but calls `idleDays()` inside `sweepNow` on every tick. If the saved threshold becomes `0`, the already-armed timer can still tick, but `sweepIdleSessions(..., 0)` returns without ending anything.

4. DISPUTED. The new `sessionReapIntervalHours` text is fine, but `server/skills/mulmoterminal-config/SKILL.md` still says under `sessionIdleReapDays`: “each row there says whether the next start will take it.” That is now the old false claim. Single settling change: replace that sentence with the event wording, e.g. “each row there marks sessions due to be ended by the next sweep.”

5. ACCEPTED. Conceptually, #2177 item 2 is closed by naming the event instead of the clock. That fixes the user-facing lie without needing the armed cadence.

6. ACCEPTED. The assertions are not vacuous: the row is forced present via `w.get('[data-testid="surviving-doomed"]')`, the badge/title/hint are positive assertions, and the negative assertions guard the old clock language across both saved cadence states.

7. DISPUTED. I found stale old-claim sites outside your listed sweep:

- `common/survivingSessions.ts`: “The server will end it at its next start”
- `docs/guide/en/config.md` all-keys row for `sessionIdleReapDays`: “before the server ends it at its next start”
- `server/config/app-config.ts`: “before the server ends it at its next start”
- `server/skills/mulmoterminal-config/SKILL.md`: “row there says whether the next start will take it”
- `plans/fix-2177-reap-interval-settings.md`: verification still says the browser checked “names the next start once raised”

Single settling change: sweep those living comments/docs/plan lines to the same “next sweep” / “due to be ended” event wording.
```

isamu:
## Round 1 triage — every finding and its disposition

Codex's P2 was **the same hole I had found independently** while its round was running, and its C-bis reply then disputed two of my seven resolutions and was right on both. Ledger:

| # | finding | disposition | why |
|---|---|---|---|
| 1 | **P2** — the UI only knows the SAVED cadence, so its hints and row copy are false when the running server armed a different one at boot | **FIXED** | see below — fixed by inverting the rule, not patching cases |
| 2 | **axis K** — SKILL.md claimed the badges "describe what this running server will do" | **FIXED** | deleted and replaced with the naming rule + the asymmetry |
| 3 | **axis H** — expose the boot-armed cadence as a runtime field | **REJECTED, and Codex accepted the rejection** | *"Naming the event is sufficient to stop the screen lying… nothing in the fixed wording requires knowing it."* And `server/session/reap-schedule.ts`, the file that would carry it, is changed by open PR #2179 right now. Recorded as a follow-up |
| 4 | my claim that the threshold applies immediately while the cadence does not | **CONFIRMED** by both of us in source | `sweepNow` calls `idleDays()` per tick; `reapIdleSessions` returns early at 0, so a saved 0 neuters an already-armed timer at once |
| 5 | **C-bis DISPUTE** — five more sites carried the old claim | **FIXED** | see the sweep note below |
| 6 | **C-bis DISPUTE** — is item 2 closed or deferred? | **CLOSED** | Codex agreed naming the event genuinely closes #2177 item 2 |
| 7 | dated release pages still quote the old badge | **deliberately not fixed** | repo rule: a dated snapshot is never edited to match new behaviour |

### What actually changed, and why it is a rule rather than three edits

The first draft switched the hint on the saved cadence and left the row alone. Both were wrong, in mirror images, and the table is the whole argument:

| armed | saved | old row: "ends at next start" | old hint: "from the next server start, also every N" |
|---|---|---|---|
| 0 | 0 | true | true |
| 0 | N | true | true |
| N | N | **false** — the sweep comes sooner | **false** — it is already running |
| N | 0 | **false** — the old timer runs on | **false** — it is not "start only" |

The issue's proposal fixes row 3 and breaks rows 1–2. My first draft fixed nothing and added the mirror error. So rather than a fourth attempt at the right sentence: **no string in this screen names WHEN the next sweep is.** The hints state what is *saved*, one fixed note states when a cadence change is read, and the row names the **event** — "the next sweep ends it" — which is true whatever was armed. Codex confirmed that state by state.

That also means this PR now **closes #2177 item 2**, with a different sentence from the one the issue proposed.

### The sweep miss, recorded because it is the reusable lesson

My claims sweep searched for the sentence `ends at next start`. Five sites say **"ends it at its next start"** and were invisible to it — `common/survivingSessions.ts` (the wire type's own docblock), `server/config/app-config.ts`, the skill's threshold section, the all-keys row in `docs/guide/{en,ja}/config.md`, and the plan's own verification line. Codex found all five. Re-swept on the durable token `next start` instead of the sentence.

### Verification of this round's batch

Break-verified, source restored byte-identical after each: the clock-naming row copy restored → **2 red**; the hint made to speak for the running server again → **2 red**; the always-on note removed → **1 red**. The row assertions run over **both** cadence states and also assert the absence of the old strings, so a future edit cannot quietly make the wording conditional again — which is the shape the first draft had.

`yarn lint`, `yarn typecheck`, `yarn build` and the five affected specs (58 tests) all clean.

isamu:
## Recommendation: CLOSE this PR as superseded — and one real defect survives it

While round 1 was running, **PR #2183 merged at 09:40:59Z** — a different session working the same issue #2177. That is one minute after this review loop started, and after my claim comment on the issue. The loop turned into a comparison, and the comparison has a clear answer.

### What #2183 already did, making this PR duplicate

- The cadence stepper under the threshold, wired to `sessionReapIntervalHours`, disabled while the threshold is `0`. Functionally what this PR does.
- It reached **the same armed-vs-saved analysis** independently — its component comment states it, and it filed **#2184** for reporting the armed value.

So item 1 is shipped and the analysis is not unique to this branch. Rebasing would ask a reviewer to disentangle a duplicate implementation plus conflicts in the component, the spec and both locale files.

### What did NOT ship, and is still wrong on `main` today

Verified against `origin/main`, by me and then independently by Codex quoting the file:

| string on main | state (armed=N, saved=N) | state (armed=N, saved=0) |
|---|---|---|
| `doomed` / `doomedTitle` / `reapEnded` — "ends at next start" | **false** — the armed timer ends it sooner | **false** — the old timer is still running |
| `sweepHint` — "Will repeat every N hour(s) after the next server start." | **false** — it is already repeating | not rendered; the "start only" copy shows and is misleading for the same reason |

This branch fixes both **without needing the armed value**, by inverting the rule: no string names a clock, the row names the EVENT ("the next sweep ends it"), the hints state what is SAVED, and one fixed note says when a cadence change is read. Codex confirmed that is true in all four states and that it is sufficient — *"if the UI names the event instead… the armed value is not needed for correctness."*

### What replaces this PR

1. A small PR against current `main` carrying **only** the wording rule and the five-site claims sweep (`common/survivingSessions.ts`, `server/config/app-config.ts`, the skill's threshold section, `docs/guide/{en,ja}/config.md`, and the guard tests that keep clock-naming from creeping back). No stepper, no duplicate i18n.
2. A comment on **#2184** recording that the armed value is no longer needed for *correctness* once the event is named — it stays worth having for *informativeness* ("this server is armed every N hours"), which is a smaller and different claim than the issue currently makes.

### Codex's judgement of this recommendation, verbatim

<details><summary>prompt sent</summary>

```text
Follow-up on PR #2186. Do NOT post anything. Judge a CLOSE recommendation.

While round 1 was running, PR #2183 — a DIFFERENT session working the same
issue #2177 — merged into main at 09:40:59Z. Read it:
  git log origin/main --oneline --grep=2177
  git show origin/main:src/i18n/en.ts   (the settings.surviving block)
  git show origin/main:src/components/settings/SurvivingSessionsSection.vue

What I believe, and want you to check rather than accept:

1. #2183 implements the SAME item 1 my PR does — a cadence stepper under the
   threshold, disabled while the threshold is 0. My stepper is therefore
   duplicate work and my branch conflicts with main in the spec and the i18n.

2. #2183 reached the SAME analysis we did about the armed-vs-saved cadence —
   its component comment says so and it filed issue #2184 for reporting the
   armed value. So that is not a gap my PR uniquely found.

3. BUT #2183 kept the clock-naming copy, and I believe two strings on merged
   main are still false:
   - `doomed` / `doomedTitle` / `reapEnded` — "ends at next start" — false in
     state (armed=N, saved=N) and (armed=N, saved=0), the same states we
     tabulated.
   - `sweepHint` = "Will repeat every {hours} hour(s) after the next server
     start." — false in state (armed=N, saved=N): that server is ALREADY
     repeating, and this says it will only after the next start.
   My branch's surviving value is the rule that fixes both WITHOUT the armed
   value — name the event, never a clock — which you confirmed is sufficient.

My recommendation: CLOSE #2186 as superseded, and open a small PR against the
new main carrying ONLY the wording rule and the five-site claims sweep, plus a
comment on #2184 saying the armed value is no longer needed for correctness
(only for informativeness).

Answer these:
  A. Is point 1 right — is my PR's stepper genuinely duplicate, such that
     rebasing would be a worse outcome than a fresh small PR?
  B. Are the two strings in point 3 actually false on merged main, in the
     states I name? Quote them from origin/main and say yes or no per state.
  C. Is CLOSE the right call, or should I instead rebase #2186 onto main and
     keep it as the vehicle? Say which and why.
  D. Does anything else in my branch have value that would be LOST by closing
     it — anything #2183 did not do?
  E. Is my read of #2184 right, that naming the event makes the armed value
     unnecessary for CORRECTNESS while still useful for informativeness?

Plain text. Disagree freely; I would rather be told to rebase than close.
```

</details>

```text
A. Yes. #2183 already added the cadence stepper under the threshold, wired to `sessionReapIntervalHours`, and disabled while `sessionIdleReapDays === 0`. Your stepper is duplicate now. Rebasing #2186 would be mostly disentangling duplicate UI/spec/i18n from a branch that also predates later main changes, so a fresh small PR is cleaner.

B. Yes, your false-string read is right.

From `origin/main`:

`doomed`: “ends at next start”  
`doomedTitle`: “Nothing is using it and it has been silent for {days} day(s) — the server ends it at its next start”  
`reapEnded`: “ended when the server next starts”

State `(armed=N, saved=N)`: false. The running server already has a repeating timer, so a reapable session may be ended by the next timer sweep, not the next start.

State `(armed=N, saved=0)`: false. The saved value says off, but the already-armed timer is still running, so “next start” is still not the honest event.

Also from `origin/main`:

`sweepHint`: “Will repeat every {hours} hour(s) after the next server start.”

State `(armed=N, saved=N)`: false. It is already repeating in this running server; it is not only something that will begin after the next start.

For `(armed=N, saved=0)`, `sweepHint` is not rendered. Main renders the “start only” copy instead, and that copy is also misleading while the old armed timer keeps running.

C. CLOSE is the right call. #2186 has been superseded as the vehicle for item 1, and rebasing it would make reviewers sort through duplicated implementation plus conflicts in the component, tests, and i18n. The better move is a fresh PR against current `main` carrying only the wording rule and the claims sweep.

D. I do not see unique functional value that would be lost by closing. The value to preserve is the copy rule, supporting comments/docs, and tests that prevent clock-naming from creeping back. Do not carry the stale branch wholesale; its diff against current `main` also collides with later merged work.

E. Yes. #2184 is right that the armed value is needed if the UI wants to say the actual clock. But if the UI names the event instead, “the next sweep”, then the armed value is not needed for correctness. It remains useful for informativeness: “this server is currently armed every N hours” or “saved differs from armed until restart.”
```

### The process failure worth recording

I claimed #2177 on the issue before starting, named the files, and checked the other checkouts and the open PRs first — the branch `fix/2177-settings-sweep-interval` did not exist at that moment. Two sessions still built the same thing. The claim comment is only as good as the other session reading it before it starts, and neither of us did badly here; the coordination did.

**Not closing this myself** — that is the user's call.

isamu:
Closing as **superseded by #2183**, per the recommendation above and the user's go-ahead.

What survives it is #2189, which carries the one thing #2183 left behind: the row and hint copy that names a clock no state can make true. The reasoning, the four-state table and Codex's judgement of the close are all in this thread and are reproduced in #2189's body and plan file, so nothing here needs to be read to follow that PR.

The branch `fix/2177-reap-interval-settings` is left in place rather than deleted — say the word if you want it gone.

PR #2188, opened 2026-09-19
chore: prettier skips Jekyll's Liquid templates (#2187)

### LINKED ISSUE #2187: chore: yarn format が docs/_layouts/default.html でこけ、prettier が Jekyll の Liquid を HTML として整形している
## 症状

クリーンな main で `yarn format` が exit 2 で終わります。

```
[error] docs/_layouts/default.html: SyntaxError: Opening tag ":svg:svg" not terminated. (34:98)
```

`format:check` も同じです。他のファイルの整形自体は通るので、実害は「毎回 exit 2 が返る」ことと、
その裏にある下の2点です。

## 根本原因

`docs/` の Jekyll テンプレートは **Liquid であって HTML ではなく**、prettier に Liquid パーサは
ありません。対象が2ファイルあり、prettier はそれぞれ別の壊れ方をしています。

**1. パースできない方 — `docs/_layouts/default.html`**

`{% include %}` の属性文字列の中に SVG が入っていて、prettier の HTML パーサがそこで落ちます。
ただしこのファイルは、**落ちていること自体が幸運**です。先頭にこう書いてあります:

> just-the-docs 0.12.0 の `_layouts/default.html` を、1行だけ変えて持っている。…
> テーマを上げたときは、**この1行以外を本家に合わせ直すこと**。

upstream と diff が取れることがこのファイルの存在意義で、整形されたらそれが失われます。

**2. パースできた方 — `docs/_includes/head_custom.html`**

こちらは `56fe298e` で実際に prettier を通り、Liquid タグが行をまたいで割れました:

```diff
-{%- comment -%}
-just-the-docs が head の末尾に差し込む場所。…
-{%- endcomment -%}
-{%- if page.url == "/" -%}
+{%- comment -%} just-the-docs が head の末尾に差し込む場所。…
+…ためでもある。トップページにだけ出す。 {%-
+endcomment -%} {%- if page.url == "/" -%}
```

Liquid はタグ内の改行を許すので**レンダリングは通ります**。当時の検証も「空白を除けば同一」
まででした。動くが読めない、が見逃された理由です。

## やること

`.prettierignore` に `docs/_layouts/` と `docs/_includes/` を理由付きで追加し、あわせて
`head_custom.html` を prettier に通される前の形に戻す。ignore が同じコミットに入るので、
`56fe298e` が解消した「どのブランチも身に覚えのない diff から始まる」問題は再発しません。

`default.html` を prettier が読める形に書き換える案は取りません。ファイル自身のコメントが
禁じている方向です。

## 確認方法

`yarn format` / `npx prettier --check .` が 0 で終わること。加えて、Jekyll をローカルで
ビルドして生成物を突き合わせ、出力が変わっていないことを確認する。

isamu:
## Summary

`yarn format` exited 2 on a clean `main`. The parse error names
`docs/_layouts/default.html`, but the cause is that `docs/` holds **Liquid templates, not HTML**,
and prettier has no Liquid parser. Two files are affected and prettier breaks each one
differently.

**The file that fails is the one that was safe.** `docs/_layouts/default.html` is just-the-docs
0.12.0's own `default.html` held with a single line changed, and its own comment asks the next
reader to re-align it with upstream on a theme upgrade. Being diffable against upstream is the
whole point of it — if prettier could parse it, it would reformat it and end that. The parse
error was protecting it by accident.

**The file that parsed is the one that broke.** `docs/_includes/head_custom.html` went through
prettier in `56fe298e` and came back with `{%- endcomment -%}` split across a line break and the
Japanese above it run together into one paragraph:

```diff
-{%- comment -%}
-just-the-docs が head の末尾に差し込む場所。…
-{%- endcomment -%}
-{%- if page.url == "/" -%}
+{%- comment -%} just-the-docs が head の末尾に差し込む場所。…
+…ためでもある。トップページにだけ出す。 {%-
+endcomment -%} {%- if page.url == "/" -%}
```

Liquid tolerates a newline inside a tag, so it still **renders** — which is why nobody saw it.
That commit's own verification was "identical with whitespace stripped", and that was true; it
just does not mean readable.

So: `.prettierignore` gains the two Jekyll template directories with the reason attached, and
`head_custom.html` is restored. The ignore lands in the same commit, so the problem `56fe298e`
fixed — every branch starting with a diff it did not make — does not come back. That commit's
intent is better served by the ignore than by the reformat.

## Items to Confirm / Review

- **The option not taken.** Rewriting `default.html` so prettier can parse it (hoisting the SVG
  into a `{% capture %}`) was rejected: it is the direction the file's own comment forbids, since
  prettier would then reformat the whole file and "this one line aside, it matches upstream" would
  stop being true.
- **The ignore is two directories, not a glob.** `docs/_*/` would have been shorter and wrong —
  `_sass` is SCSS, which prettier formats correctly, and `docs/_config.yml` stays a target for the
  same reason.
- **`head_custom.html` is restored from `56fe298e^`**, and nothing has touched the file since that
  commit, so this is a clean return rather than a hand-unformat.

## Verification — the built site, not the whitespace

The claim "the output does not change" is proved by building it, not by reading the diff:

- `bundle exec jekyll build` before and after, `diff -r` over the two trees: **all 327 generated
  files byte-identical**.
- The homepage still carries **both** `ld+json` blocks (`WebSite` and `SoftwareApplication`) and
  both parse as JSON — that block is what `head_custom.html` exists to emit, and a Liquid comment
  mistake would silently drop it.
- No Liquid comment text leaks into the rendered page, and a guide page still has no
  `SoftwareApplication` block (it is homepage-only, which is what the template's `{%- if
  page.url == "/" -%}` is for).
- `yarn format` and `npx prettier --check .` both exit 0.

No TypeScript changed, so `lint` / `typecheck` / `build` are left to CI.

## User Prompt

- 「docs/_layouts/default.html の prettier エラーなおせる？format の対象にすれば良い？」 —
  the answer turned out to be the other way round: the file is already a format target, and that
  is the problem.
- Scope chosen with the user: add the ignore **and** restore `head_custom.html`, rather than the
  ignore alone or making `default.html` parseable.

Closes #2187

work in mulmoterminal5

## Summary by CodeRabbit

* **Documentation**
  * Added a planning document describing formatting support for Jekyll Liquid templates and associated verification steps.
  * Reformatted structured-data markup in documentation templates without changing its content or behavior.
* **Chores**
  * Updated formatting exclusions for documentation template files.


isamu:
Review complete. I found no inline findings.

Axes checked:

- Correctness of the ignore: `docs/_layouts/` and `docs/_includes/` match the intended directories under gitignore-style `.prettierignore` syntax; the trailing slash limits the pattern to directories. I also searched for Liquid markers across the repo. The remaining real docs Liquid outside those directories is in Markdown, which is already ignored by `*.md`, plus `docs/llms.txt`, for which Prettier reports no inferred parser and does not format it during `prettier --check .`. `_config.yml` and `_sass` remain covered and Prettier can parse them as YAML/SCSS.
- Restore: `docs/_includes/head_custom.html` is byte-identical to `56fe298e^:docs/_includes/head_custom.html`. The file history since its introduction is `17adfe6b` -> formatter commit `56fe298e` -> this restore, so I did not find a legitimate intervening change being thrown away.
- Behaviour preservation: I built archived trees for `2f36b1b3^` and `2f36b1b3` and `diff -r` produced no differences across 327 generated files. The homepage still has two valid `ld+json` blocks (`WebSite`, `SoftwareApplication`), the Liquid comment text does not leak, and `guide/en/index.html` has no `SoftwareApplication` block. Liquid whitespace can be significant around whitespace-control tags, captures, and preformatted content, so the byte-identical build is the right verification here; it is a content-specific proof, not a general “Liquid formatting is harmless” rule.
- Scope: ignoring the whole includes/layouts template directories is a reasonable default for this repo because those files are Jekyll/Liquid template surfaces and future files there would have the same parser mismatch. It is intentionally wider than the two current files, but not too wide for the stated maintenance goal. I agree with leaving `docs/_sass` and `docs/_config.yml` in scope because Prettier has real parsers for SCSS and YAML.
- Rejected alternative: making `default.html` parseable by rewriting the Liquid/SVG shape would work mechanically, but it fights the file’s stated upstream-diffability contract. A `prettier-ignore`-style local escape would also be another local delta in the vendored layout. Ignoring the template path is the cleaner preservation move.
- Future traps: new files under these two directories are now silently unformatted by Prettier. Given Prettier’s lack of Liquid support, that is the better default than silently HTML-formatting Liquid or failing on parser edge cases. I did not find tooling that assumes every future file under these paths is Prettier-covered.
- CI: I did not find a workflow running `format:check`; CI runs lint/typecheck/build/test with docs/plans filtering, and `pages.yml` builds the docs with Jekyll on `main` pushes. This change affects local formatter coverage, not the Pages build path. The archive build verifies the generated Pages content is unchanged.
- Repo rules: issue #2187 exists, the plan file exists, the commit uses the `chore:` prefix, and the `.prettierignore` comment explains why rather than merely what. Its length is earned because the rule changes future formatter coverage and preserves an upstream-diffability invariant.
- Docs: I found `CLAUDE.md` mentioning that `.prettierignore` excludes `*.md`; that remains true and does not claim to be an exhaustive coverage list. I did not find README/CLAUDE/docs text that becomes materially stale from adding the Jekyll template directories.

Verification run:

- `gh pr diff 2188`
- `git diff --no-index -- docs/_includes/head_custom.html <(git show 56fe298e^:docs/_includes/head_custom.html)` -> no diff
- `npx prettier --file-info` for `docs/_includes/head_custom.html`, `docs/_layouts/default.html`, `docs/_config.yml`, `docs/_sass/custom/custom.scss`, and `docs/llms.txt`
- `cd docs && bundle exec jekyll build -d /tmp/site-pr2188 --quiet` -> exit 0, with existing Sass deprecation warnings
- archived before/after Jekyll builds for `2f36b1b3^` and `2f36b1b3`, then `diff -r` -> no differences, 327 generated files
- rendered structured-data check -> two homepage JSON-LD blocks parse; no Liquid comment leak; guide page has no `SoftwareApplication`
- `yarn format:check` -> exit 0
- `npx prettier --check .` -> exit 0
- `yarn format` -> exit 0; reported files unchanged

Note: my first parallel attempt at `prettier --check .` raced against a Jekyll build creating/removing `docs/.jekyll-cache` and failed with transient ENOENTs from ignored generated cache files. After cleaning that generated cache and running the commands sequentially, the exact formatter commands passed.

CODEX VERDICT: LGTM
FINDINGS COMPLETE

PR #2189, opened 2026-09-19
fix(settings): name the sweep, not a clock — no time this screen can name is true (#2177)

isamu:
## Summary

Follow-up to #2183, which closed #2177's item 1. This fixes the copy that item 1 left behind.

#2183 gave the cadence a Settings control and **deliberately** left the row copy alone, reasoning that the saved cadence is not what the running server armed so the row must not be keyed off it. That reasoning is right, and it stops one step short: the copy it left in place is false in exactly the states it was protecting against, and the cadence hint added alongside it is false in one of them too.

`startReapSchedule` arms the timer **once, at boot** — it takes `intervalHours` as a plain number while it takes `idleDays` as a `() => number`, and `config-routes.ts` says why (*"re-arming on every config POST would let a stream of edits reset the countdown forever"*). So the SAVED cadence and the ARMED cadence are different numbers, and the browser only ever receives the saved one:

| armed | saved | `doomed` — "ends at next start" | `sweepHint` — "…after the next server start" |
|---|---|---|---|
| 0 | 0 | true | true |
| 0 | N | true | true |
| N | N | **false** — the armed timer ends it sooner | **false** — it is already repeating |
| N | 0 | **false** — the old timer runs on | not rendered; the "start only" copy shows, equally wrong |

Keying the row off the saved value — what #2177 originally proposed — fixes row 3 and breaks rows 1 and 2. **No sentence naming a time is true in all four**, so the fix is not a better sentence:

> **Nothing in this section names when the next sweep is.**

- the row names the **event** — `due to be ended`, "the next sweep ends it", "ended by the next sweep"
- the cadence hints state what is **saved** — "Saved: repeats every N hour(s)." / "Saved: at server start only."
- one line, shown in **every** state including the disabled one, says when a change is read — "The cadence is read when the server starts, so a change here applies from the next one."

### This changes what #2184 is for

#2184 exists because the row cannot know the armed cadence. Naming the event means it no longer needs to: the armed value stays worth exposing because *"this server is armed every N hours"* is more informative, but it is no longer what stands between the screen and the truth. Codex confirmed that explicitly. I have commented on #2184 rather than editing it.

### The asymmetry the section now states

[...]

PR #2191, opened 2026-09-19
feat(settings): the hint states the cadence the server ARMED, not the saved one (#2184)

### LINKED ISSUE #2184: feat(settings): 行が「実際に仕掛かっているスイープ間隔」を知らないので、タイマー有効時に終了時期を遅く見積もる
## 背景

#2183 のレビューで見つかった。`sessionReapIntervalHours`（#2167）は**起動時に一度だけ**
タイマーを仕掛ける:

```
server/index.ts:991  →  startReapSchedule({ intervalHours: getSessionReapIntervalHours(), ... })
                     →  armTimer  →  const timer = setInterval(...)   ← 関数ローカル
```

`startReapSchedule` の呼び出し元はこの1箇所だけで、`clearInterval` はリポジトリ全体に存在しない。
つまり**仕掛けられた間隔は boot 時点で固定**され、保存された設定値とは独立に drift する。

## 何が困るか

Settings の「Sessions that survived a restart」の行は、対象になるセッションに
**ends at next start** と出す。タイマーが有効な状態で起動したサーバでは、これは
**実際より遅く見積もっている** — そのセッションは次の起動を待たずにスイープで終了する。

#2183 で一度これを直そうとして、**保存値**から行の文言を切り替える実装にした。それは
両方向に嘘になるので取り下げた:

- 既定 `0` から `+` を押した直後 → UI は「次のスイープで終了」と言うが、**このプロセスには
  タイマーが無い**。実際には次の起動まで終了しない。
- boot 時に有効だった状態から `0` に下げた直後 → UI は「起動時のみ」と言うが、**古い
  タイマーは回り続けている**。

## 必要なもの

行が正しいことを言うには、**サーバが「実際に仕掛けた間隔」を報告する**必要がある。保存値では
不十分。案:

- `startReapSchedule` が仕掛けた値を記録し、読み出せるようにする
- それを UI に渡す。`/api/tmux/sessions` はこの節が既に叩いているので、新しいルートを増やさず
  レスポンスに足せる（`{ sessions, armedReapIntervalHours }`）
- 行の文言は **armed** の値で決める。保存値が armed と違うときは「次回起動時に反映」と出す

## 代替案（検討済み・不採用）

**設定変更時にタイマーを張り直す**（live re-arm）。そうすれば保存値 == 実際の値になり、UI は
何も新しく知る必要が無い。Codex の見解と合わせて不採用:

- #2167 が「次回起動時に反映」を**意図的に**契約として文書化したばかりで、それを覆す
- モジュールスコープのタイマー所有、clear/re-arm の意味論、config POST の配線、`0` への
  変更の扱い、実行中スイープとの競合、それぞれのテストが要る

どちらを採るかは実装時の判断。**live re-arm を選ぶなら #2167 の契約変更として扱い、
skill と両方のガイドの「次回起動時に反映」を書き換えること。**

## 現状（#2183 で入れた分）

行の文言は**保存値に依存しない**ことを spec で固定してある
（`test/src/components/settings/survivingSessionsSection.spec.ts`、cadence 0 と 6 の両方で
`ends at next start` を表明）。嘘は残っていないが、タイマー有効時に**遅く見積もる**ままで、
それがこの issue。

## 関連

- #2165 / #2167 定期スイープの導入
- #2177 / #2183 Settings にステッパーを出した回
- #1486 行が「次回起動時に終了」と言う仕組みと、しきい値変更後の再読み込み

isamu:
## Summary

Closes #2184.

**Read the history first — this PR is now much smaller than it started.**

`sessionReapIntervalHours` (#2167) repeats the idle sweep on a timer, armed **once at boot** and
deliberately not re-armed on a config POST. So the saved number and the running one are different
things until a restart, and the browser was only ever told the saved one. Every clock-naming
sentence was therefore false in half the reachable states.

**#2189 fixed that while this PR was in review**, by removing the claim: the row names the *event*
("due to be ended", "the next sweep ends it"), the hints said `Saved: …`, and a standing note says
the cadence is read at server start. All true whatever is armed. That is the better base, and this
PR now **sits on top of it rather than replacing it**.

What is left here is the one thing #2189 could not do, and said so in its own comment:

> #2184 would let this say what IS armed, which is strictly more informative

So: **the server now reports the cadence it armed**, and the hint stops hedging.

| | |
|---|---|
| the row | **unchanged from #2189** — still names the event |
| hint, saved == armed | *Repeating every N hour(s).* — a fact the server reported |
| hint, saved ≠ armed | *Saved — starts repeating at the next server start.* + names what is **still** sweeping until then |
| hint, server silent | falls back to the saved-value wording rather than inventing a present-tense claim |

`armTimer` returns the cadence it started, `startReapSchedule` records it,
`armedReapIntervalHours()` reads it back, and it rides on the `/api/tmux/sessions` response the
section already fetches.

## Items to Confirm / Review

1. **The scope narrowing is the main thing to agree with.** I originally changed the row's wording
   too; the merge with #2189 showed that was redundant — its `doomedTitle` already says *"the next
   sweep ends it"*. Those keys (`doomedSoon`, `doomedSoonTitle`, `reapEndedSweep`) are **dropped**.
   If you would rather this PR had been closed entirely, that was my recommendation before the
   merge and it is still a defensible call — the case is in the comment above.
2. **A superseded schedule is now cancelled.** Codex's round-1 P2: `startReapSchedule(6)` then
   `(0)` left the six-hour interval ticking while the getter said OFF. **Reproduced** by advancing
   the clock and watching a sweep fire. `armTimer` clears any previous interval on **every** call,
   including one that arms nothing — cancelling only when the new schedule arms something misses
   that case exactly. This is independent of all the UI wording and is the part worth reading
   hardest.
3. **Not the live re-arming #2167 declined.** That was re-arming on every config POST. This
   cancels only when a caller explicitly starts a new schedule — once, at boot. Codex confirmed
   in-round.

## How it was verified

- `yarn format`, `yarn lint`, `yarn typecheck`, `yarn build` clean on the merged result; the five
  affected spec files pass (61 tests).
- **Break-verified**, each with the mutation's application asserted by count first: reverting the
  hint to the saved value reddens the pending-state spec; removing the timer cancellation reddens
  two; moving the cancel below the enabled check reddens the OFF case specifically.
- **Driven against a real server** before the merge: scratch `HOME`, `sessionIdleReapDays: 0` so
  the timer arms but can never reap, isolation confirmed by asking the server before trusting any
  result, tmux session count unchanged. Booting with 6 then saving 2, then 0, kept `armed = 6`
  throughout — the divergence this PR exists to expose — and booting with 0 then saving 6 kept
  `armed = 0`.
- #2189's invariants are **pinned, not assumed**: its specs for the row naming the event and for
  the standing note being shown in every state are kept as they were.

## What was not exercised

**No browser render** — no driver in this checkout, so the states are covered by mounting the real
component. Five locales changed; the Korean and Chinese strings are worth a native eye.

**The full local suite deadlocked** once during this work (an hour at 0% CPU, killed); CI is the
ground truth for it.

## User Prompt

- `#2184やって`, then `merge` — implement the deferred half of #2177, and land it rather than
  closing it when mainline superseded part of it.

work in mulmoterminal4


isamu:
## Cross-review tier: **C**

Three things on the tier-C list:

- **a wire format change** — `/api/tmux/sessions` gains `armedReapIntervalHours`, a new field on an existing response;
- **module-level mutable state that decides a user-facing claim** — `armedIntervalHours` in `reap-schedule.ts` is what the row's promise about *when a session is deleted* now rests on;
- **the claim is about a destructive operation** — the row tells someone when their session goes. Overstating it is the defect this PR exists to avoid creating.

Round 1 names those as explicit axes and requires a per-axis answer.

**Sandbox pre-flight (ledger row 0):** vitest, `yarn typecheck`, socket bind, `curl api.github.com` (200) and `gh` all exit 0 — and this time the round runs in the **main checkout**, the directory the pre-flight actually covered. Last loop I proved the sandbox here and then ran the round in a worktree whose symlinked `node_modules` blocked vitest, which made an LGTM a read rather than a run. `rm -f` is still refused by the command guard, so Codex is told not to run a mutation sweep — it names them and I run them.

**Pre-existing red baseline: none.** `yarn format` passes for the first time this session (#2188 fixed the Jekyll/Liquid failure), so the full gate set is runnable.

isamu:
## Codex exchange — round 1

Both calls of the round. Prompts are in `/tmp/codex-cross-review-2191/`.

<details><summary>Round 1 review — prompt sent</summary>

```text
Review PR #2191 at https://github.com/receptron/mulmoterminal/pull/2191.
Repo checkout: /Users/isamu/ss/llm/mulmoterminal4 — the MAIN checkout, on branch
feat/2184-armed-reap-interval. Verified working here: \`yarn vitest run <file>\`, \`yarn typecheck\`,
\`yarn lint\`, \`yarn build\`, \`yarn format\`.

Every finding you hold back costs a full extra round: a fix is reviewed for the first time in
the round AFTER it is pushed, so a finding raised late is not a later finding, it is a repeat
of this entire exchange. Read the WHOLE diff before you post anything. Do not stop at the
first problem. Do not save anything for 'a follow-up review'.

── SANDBOX ─────────────────────────────────────────────────────────────────
You CAN run tests, typecheck, lint, build, bind sockets, reach api.github.com and run gh.
You CANNOT run \`rm -f\` — the command guard refuses it. DO NOT run a mutation sweep: a restore
that fails leaves you measuring the harness rather than the code. NAME any mutation you want
and I will run it and report the result verbatim.

── WHAT THIS PR IS ─────────────────────────────────────────────────────────
Background, because the history is the point:

\`sessionReapIntervalHours\` (#2167) repeats the idle-session sweep on a timer. The timer is
armed ONCE, at boot, by \`startReapSchedule\`, and deliberately NOT re-armed on a config POST —
\`config-routes.ts\` says why: a stream of edits would reset the countdown forever.

#2183 added the Settings stepper for that key and REFUSED to let the doomed row's wording
follow it, because the saved number is not what the running process is doing. That was correct
but left the row UNDERSTATING: on a server that booted with a repeat armed, a session goes at
the next sweep while the row still points at a restart.

This PR closes that. \`armTimer\` now returns the cadence it started (hours, or OFF when it
started none), \`startReapSchedule\` records it in a module-level \`armedIntervalHours\`,
\`armedReapIntervalHours()\` reads it back, and it rides on the \`/api/tmux/sessions\` response.
The row reads "ends on the next sweep" when a repeat is genuinely scheduled and "ends at next
start" otherwise; the cadence stepper gains a fourth state for the saved-but-not-armed window.

  correctness and edge cases
  WIRE / API COMPATIBILITY: \`/api/tmux/sessions\` gains a field. What reads that response?
    Is an older or third-party client harmed? Is the field's absence handled, and is the
    fallback the SAFE direction?
  MODULE-LEVEL MUTABLE STATE: \`armedIntervalHours\` in reap-schedule.ts. Who writes it, when,
    and can it ever disagree with the timer that is actually running? Consider: repeated
    \`startReapSchedule\` calls, a call that arms nothing after one that armed something, tmux
    absent at boot so the schedule never starts, and test pollution across spec files.
  THE CLAIM ABOUT A DESTRUCTIVE OPERATION: the row tells a user when their session will be
    ended. Enumerate the states (armed vs saved vs threshold-off) and say for EACH whether the
    sentence shown is true. Overstating — promising a sweep that is not scheduled — is the
    defect this PR exists to avoid creating, so look hardest there.
  BOOT / LIFECYCLE ORDERING: routes are mounted before the server listens, and the schedule
    does not arm until it does. Is reading the getter per request right, and is there a window
    where a client sees a value that is about to change?
  tests: does a test go red if the new behaviour breaks — and can any new assertion pass while
    checking nothing?
  security (XSS via i18n interpolation, anything user-controlled reaching a template)
  accessibility and i18n lockstep: this repo now has FIVE locales — en, ja, ko, zh-CN, zh-TW.
    ko/zh arrived in #2180 mid-work. Check every new key exists in all five and that the two
    keys whose MEANING changed (sweepHint, sweepOffTitle) were updated in all five, not just en/ja.
  documentation: report a sentence that would cause a WRONG CHANGE if believed. File anything
    merely behind as N.

An axis with no findings is a RESULT and has to be stated. 'No findings' on an axis you did
not look at is the one answer that makes this loop longer.

1. SEVERITY — P1 blocker / P2 should fix / P3 nit / N note. N is prose-only and does NOT block
   the verdict; a prose finding is P3+ only when believing the sentence would cause a WRONG
   CHANGE.
4. WHAT WOULD CHANGE YOUR MIND — the fact that, if true, makes it a non-issue.

| 0 | — | Sandbox pre-flight | — | tests/typecheck/lint/build/socket/curl/gh all exit 0 in this checkout. \`rm -f\` refused: no mutation sweeps. |
| 0b | — | Pre-existing red baseline | — | None; \`yarn format\` passes since #2188. |

_no findings yet — this is round 1._

Two things already decided, so do not spend the round re-arguing them:
  - **Not live re-arming.** Making the config POST clear and re-arm the timer would make saved
    and armed agree with no new field. Declined: #2167 documents why re-arming on every POST is
    wrong (a stream of edits resets the countdown forever), and you yourself costed and
    declined it during #2183's review. Reversing that contract is a separate decision.
  - **The default stays 0 (off).** Out of scope.
  Findings about whether those were the right calls are out of scope; findings about the code
  MISREPRESENTING them are in scope.

    'CODEX VERDICT: CHANGES REQUESTED' followed by a bulleted summary
  Then, on its own line:

```

</details>

<details><summary>Round 1 review — Codex reply (verbatim stdout)</summary>

```text

- P2: \`startReapSchedule(6)\` followed by \`startReapSchedule(0)\` reports OFF while the first interval is still alive, so the new module-level truth can disagree with the timer that is actually running.

| Axis | Result |
|---|---|
| correctness and edge cases | P2 finding below: repeated \`startReapSchedule\` calls can leave an old interval running while reporting OFF. |
| WIRE / API COMPATIBILITY | none. \`/api/tmux/sessions\` gains an additive field; the only in-repo reader is \`useSurvivingSessions\`, and absence is sanitized to OFF, the safe fallback. Older/third-party clients that read \`sessions\` ignore the extra field. |
| MODULE-LEVEL MUTABLE STATE | P2 finding below. Initial tmux-absent boot stays OFF correctly; per-request reads avoid stale route capture; spec-file pollution is limited by fake timer teardown, but the repeated-call case is real inside the module contract. |
| THE CLAIM ABOUT A DESTRUCTIVE OPERATION | P2 finding below for repeated starts. Otherwise: threshold off shows “never” / disables cadence; armed > 0 with threshold on says next sweep, true; armed 0 saved > 0 says next start / pending, true; saved 0 armed > 0 says next sweep and pending-running, true. |
| BOOT / LIFECYCLE ORDERING | none. Getter per request is right because routes mount before \`onListening\`; before listen there is no normal client window. |
| tests | P2 finding below. The new repeated-call assertion can pass while checking only the scalar, not whether a previous interval still ticks. Targeted specs, \`yarn typecheck\`, \`yarn lint\`, \`yarn build\`, and \`yarn format\` passed. |
| security | none. New i18n interpolation receives sanitized numeric values / config numbers and Vue template/i18n escaping applies. |
| consistency with the rest of the codebase | none beyond the timer-handle issue; dependency injection and route mounting follow existing patterns. |
| accessibility and i18n lockstep | none. New keys exist in en, ja, ko, zh-CN, zh-TW; changed \`sweepHint\` and \`sweepOffTitle\` meanings were updated in all five. |
| documentation | Same P2 appears in prose: the plan says recorded value and started timer cannot drift, which is false under repeated starts and would lead to preserving the bug. No separate doc-only blocker. |

**P2 — repeated schedule calls can report OFF while an old sweep still runs**

Every site:
- \`server/session/reap-schedule.ts:59-66\` starts an interval but does not store or clear its handle.
- \`server/session/reap-schedule.ts:87-93\` overwrites \`armedIntervalHours\` on every call, including a later OFF call, without cancelling any earlier interval.
- \`test/server/session/reap-schedule.spec.ts:124-128\` asserts the getter becomes \`0\` after \`startReapSchedule(6); startReapSchedule(0);\` but never advances timers afterward, so it passes while the old 6h timer would still fire.
- \`plans/feat-2184-armed-reap-interval.md:40-45\` and \`plans/feat-2184-armed-reap-interval.md:57-58\` document that the recorded value cannot drift / that this spec pins the behavior, which is the false part.

Why it matters:
After \`startReapSchedule(schedule(6))\`, \`armTimer\` creates a live \`setInterval\`. A later \`startReapSchedule(schedule(0))\` sets \`armedIntervalHours\` to OFF, but the previous interval is still scheduled and will keep sweeping. The UI would then fall back to “ends at next start” even though a next sweep is actually coming. That is exactly the kind of destructive-operation understatement this PR is trying to make precise.

Smallest change:
Keep the timer handle in module state and clear it before replacing the schedule, or make \`startReapSchedule\` explicitly single-use and remove the repeated-call test/claim. If “latest call wins” is intended, add a spec that advances the fake clock after the second OFF call and proves no tick fires.

What would change my mind:
A demonstrated invariant that \`startReapSchedule\` cannot be called more than once in a module lifetime in production or tests, plus removing/rewriting the new repeated-call claim so it no longer promises behavior the function does not implement.

```

</details>

**Codex did not manage to post its verdict comment this round** — `gh api issues/2191/comments` returns zero bodies starting `CODEX VERDICT`, and no PR review either. The verdict above is relayed verbatim from its stdout, which is the only copy. Saying so rather than letting the round look like it never happened.

<details><summary>Step C-bis — settling the remedy before pushing (prompt sent)</summary>

```text
Follow-up on PR #2191, round 1. Do NOT re-review the PR and do NOT post anything.
Repo: /Users/isamu/ss/llm/mulmoterminal4, branch feat/2184-armed-reap-interval.

I ACCEPT your P2 in full, and I reproduced it before accepting rather than reading and agreeing:
a throwaway spec called startReapSchedule(6) then startReapSchedule(0), asserted the getter was
0 (which passed), then advanced the fake clock by 6h — and a sweep fired. Exactly your claim.
The scalar was right and the timer was still alive.

I am fixing it with the FIRST of your two named options — keep the handle and clear it — not the
"make it single-use" one. \`armTimer\` now does, on EVERY call including one that arms nothing:

    if (armedTimer !== null) { clearInterval(armedTimer); armedTimer = null; }
    if (!reapTimerEnabled(intervalHours)) return REAP_INTERVAL_HOURS_OFF;
    ... setInterval ... ; armedTimer = timer; return intervalHours;

and the spec now advances the clock after the OFF call and counts sweeps, plus a second case
where 6h is replaced by 2h and only three ticks fire in six hours. Both break-verified: removing
the cancellation reddens two specs, and moving the cancel BELOW the enabled check (so it only
cancels when the new schedule arms something) reddens the OFF case specifically.

I want your judgement on THREE things, because each could cost a round if I have it wrong.

1. ACCEPTED or DISPUTED: **does cancelling here re-open the live re-arming that #2167
   deliberately declined?** That decision was "do not re-arm on every config POST, because a
   stream of edits resets the countdown forever". My reading is that this is a different thing:
   it cancels only when a caller explicitly starts a NEW schedule, which production does exactly
   once, at boot — no config path reaches it. If you think this is the declined behaviour
   arriving by another door, say so, because I would rather hear it now than after merge.

2. ACCEPTED or DISPUTED: **is "latest call wins" the right contract, or should
   \`startReapSchedule\` refuse a second call outright?** You offered both. I chose latest-wins
   because the module then behaves the way both my spec and my plan already claimed, and because
   a guard that throws on a second call turns a future in-process restart into a crash. But
   single-use is arguably the honest contract for something with one caller.

3. Is there a THIRD place this same class hides — a value reported to a client that can disagree
   with a live timer or handle — that I should sweep in this same batch rather than meet again in
   a later round? I checked \`startUpdateStatusRefresh\` in server/config/update-status.ts: it also
   keeps no handle, but it reports nothing and is never superseded, so I judged it out of class.
   Tell me if that judgement is wrong, or if you know another.

Reply as plain text. If you DISPUTE any of them, say which single change would settle it.
```

</details>

<details><summary>Step C-bis — Codex reply (verbatim stdout)</summary>

```text
1. ACCEPTED. This does not re-open #2167. The declined behavior was config POST re-arming; this path is still boot-only in production. Clearing inside \`startReapSchedule\` just makes an explicit second schedule internally coherent.

2. ACCEPTED. “Latest call wins” is the better contract here. It matches the exported getter, keeps tests/future in-process restart paths sane, and avoids turning an accidental second call into a process crash. Single-use would be defensible only if the function were private to boot wiring.

3. I don’t see a third same-class case to sweep in this batch. \`startUpdateStatusRefresh\` is out of class by my lights: it has no public “armed cadence” value and no config/client promise that can drift from a live timer. It could duplicate work if called twice, but that is not the same bug.

No disputes from me.
```

</details>

Nothing stripped; no secrets appeared in either exchange.

isamu:
## Triage — round 1: **CHANGES REQUESTED**, one P2, and it was mine

Codex found a real defect, and it is the best kind: **this PR re-created the exact bug it exists to remove, in a new place.**

> `startReapSchedule(6)` followed by `startReapSchedule(0)` reports OFF while the first interval is still alive.

Not read-and-agreed. A throwaway spec armed 6, armed 0, asserted the getter was `0` — **which passed** — then advanced the fake clock six hours:

```
sweeps fired after the OFF call: 1
```

The scalar was right and the timer was still ticking. So the list would say **ends at next start** about a session the very next sweep takes — understating when someone's session disappears, which is the one direction this PR was written to fix.

**And my own spec was what let it through.** `stops reporting a cadence once a later schedule arms none` asserted the number and never advanced the clock. It passed over a live defect while reading like coverage.

### The fix

`armTimer` clears any superseded interval **on every call, including one that arms nothing** — that last part is the whole subtlety. Cancelling only when the new schedule arms something misses the OFF case, which is precisely the reported scenario.

The specs now assert the **timer**, not the number: none fire after an OFF call, and 6h replaced by 2h leaves exactly three ticks in six hours. The scalar assertion stays — it was not wrong, only insufficient.

**Break-verified**, each with the mutation's application asserted by count first:

| mutation | result |
|---|---|
| cancellation removed (the state before this fix) | 2 red |
| cancel moved BELOW the enabled check — only cancels when the new schedule arms something | 1 red, the OFF case specifically |

### Three things settled in this round rather than the next

I put all three to Codex before pushing (C-bis, exchange above). All **ACCEPTED**, no disputes:

1. **Does cancelling re-open the live re-arming #2167 declined?** No — *"the declined behavior was config POST re-arming; this path is still boot-only in production."* Worth asking, because the prompt had put that decision out of scope and I was about to touch its neighbourhood.
2. **Latest-call-wins vs single-use?** Latest-wins — *"avoids turning an accidental second call into a process crash."*
3. **Is there a third site of this class?** I checked `startUpdateStatusRefresh` (also keeps no handle) and judged it out of class; Codex agreed independently — *"no public 'armed cadence' value and no config/client promise that can drift from a live timer."* So the class is swept, not just the site.

### Claims sweep

The plan said the recorded value and the timer *"cannot drift: there is no path that starts one without reporting it"* — false in the other direction, and it would have taught the next reader to keep the bug. Rewritten, with the reproduction that found it. `grep` across every file this PR touches now returns no such claim.

### Ledger

| # | finding | disposition |
|---|---|---|
| 1 | P2 — superseded interval keeps ticking while the getter says OFF | **FIXED**, reproduced first |
| 2 | my spec asserted the scalar and never advanced the clock | **FIXED** |
| 3 | the plan's "cannot drift" claim | **FIXED** |
| 4–6 | re-arm scope / contract choice / class sweep | **settled in-round**, Codex accepted all three |

Round 2 is owed because this round pushed.

isamu:
## Recommendation: **close this PR as superseded by #2189**

Not my call to make, so this is the case and the evidence. Both reviewers agree.

### What happened

While this was in review, **#2189 merged** — a *third* independent PR against #2177, titled *"name the sweep, not a clock — no time this screen can name is true"*. It solves the same problem by **removing the claim** rather than making it accurate:

| | before | after #2189 |
|---|---|---|
| `doomed` | ends at next start | **due to be ended** |
| `doomedTitle` | …at its next start | …**the next sweep ends it** |
| `sweepHint` | Will repeat every N hour(s) after the next server start. | **Saved:** repeats every N hour(s). |
| `sweepOffTitle` | Will run only at server start. | **Saved:** at server start only. |
| *(new)* `sweepNote` | — | The cadence is read when the server starts, so a change here applies from the next one. |
| `reapEnded` | ended when the server next starts | **ended by the next sweep** |

Every sentence is now true in every reachable state. All five locales, rendered, complete. **No server change, no wire field, no module state.**

### The decisive line is in #2189's own shipped comment

> `sweepNote` says when a cadence change is read. All true whatever was armed. **#2184 would let this say what IS armed, which is strictly more informative — it is no longer needed to stop the screen being wrong.**

That is exactly right, and it reclassifies this PR. #2191 was a **correctness** fix; after #2189 it is an **informativeness** upgrade. What it still buys: distinguishing *"this server is actually sweeping every N hours"* from *"N is merely saved for the next start"*.

What it costs for that: a server-runtime concept, a new field on `/api/tmux/sessions`, client parsing, a fourth UI state, module-level mutable state that **already produced this loop's only P2** — and now a **ten-file semantic reconciliation** (all five locale dicts, the component, its spec, both `config.md` guides, the skill), because the strings it edits were just deliberately reworded by someone else.

`mergeable` is `CONFLICTING`, and the conflicts are semantic, not textual: two different philosophies about the same sentences.

### Codex's judgement, asked adversarially

I asked it to argue *against* closing. It agreed:

> **AGREE with closing #2191 as superseded.** […] I do not think that extra precision justifies the new runtime concept, API field, client parsing, extra UI state, module mutable state, and semantic reconciliation now that #2189 has removed the false claim. #2189's answer is less precise, but **true in all reachable states. That is the better trade.**

And on whether #2189 is the worse answer: *"'Due to be ended' is vaguer, yes, but deliberately so… That leaves the user with a truthful model."*

### What to salvage — one thing, and it is real

**The superseded-timer cancellation in `armTimer`** (`b7cd3f0d`). `startReapSchedule(6)` followed by `startReapSchedule(0)` left the six-hour interval ticking while the getter reported OFF. I reproduced it by advancing the clock and watching a sweep fire. **#2189 does not touch this** — it is a server-side defect independent of any UI wording, and it survives the supersession.

Codex agrees it is the only thing worth extracting, and frames it fairly: on `main` there is one production caller, so it is *robustness/idempotence*, not a live user-facing bug. Small follow-up PR, not this one.

Everything else here — the wire field, the armed getter, the pending-cadence UI, the docs around that model — belongs to the superseded approach.

### Loop record

| round | verdict |
|---|---|
| 1 | CHANGES REQUESTED — one P2 (reproduced, fixed in `b7cd3f0d`), three questions settled in-round |
| — | **halted**: mainline superseded the change; semantic conflict in ten files |

Round 2 was owed and is not worth running: reviewing a head that should not merge is the one thing this loop's own rules say outranks a pending verdict.

**Over to the maintainer.** I do not close PRs on my own judgement.

isamu:
Reviewing this from another session (I am the one who merged #2189). I agree with the close recommendation and I am **not** running a second cross-review loop on it — yours is complete, it was checked adversarially, and `mulmoterminal4` has this branch checked out with uncommitted work, so a push from me would collide.

One correction, because it would send someone to do work that is not needed.

## The "one thing to salvage" does not survive the supersession

> **The superseded-timer cancellation in `armTimer`** (`b7cd3f0d`). […] **#2189 does not touch this** — it is a server-side defect independent of any UI wording, and it survives the supersession.

It is not reachable on `main`. `armTimer` does create a `setInterval` it never clears, but a *superseded* timer requires `startReapSchedule` to be called twice in one process, and on `main` it is called **once**:

```
server/index.ts:417   server.listen(Number(PORT), BIND_HOST, () => onListening({ … }))
server/infra/on-listening.ts:48   startReapSchedule({ intervalHours: getSessionReapIntervalHours(), … })
```

That is the only production call site (`grep -rn startReapSchedule server/` — the other hits are this file and a comment), and Node's `listen` callback is a one-shot `listening` listener, so it fires once per process. There is no port-walk retry re-calling `server.listen`, and the dev server's reload path respawns the process rather than re-listening. A second arming cannot happen, so there is no timer to supersede.

**The defect is real inside this PR**, because this PR is what adds a re-arm path — your own commit message says so (*"cancel a superseded sweep, not just its number"*). It is a bug the feature introduced and then fixed, not one it uncovered. Close this and there is nothing left unfixed on `main`.

What *is* worth one line, and is a different claim: `armTimer` holds no handle to its interval, so **the next caller to arm twice creates the bug rather than hits it**. That is latent-hazard territory, not a live defect — worth a comment there if anyone touches it, not a salvage PR.

## Otherwise

Your table of what #2189 changed is accurate, and the reclassification is the right read — I wrote that "strictly more informative" line and meant it as you took it. If the informativeness is wanted later, the cheap version is the one your own PR body implies: send the armed cadence and use it **only** where it distinguishes something the event wording cannot, rather than re-introducing clock copy across five locales.

isamu:
Closing on the user's instruction, taking the recommendation above.

**Why**, for anyone arriving later: #2189 solved the same problem by removing the false claim instead of making it accurate, so this PR stopped being a correctness fix and became an informativeness upgrade. Its cost had also grown — a server-runtime concept, a new field on `/api/tmux/sessions`, client parsing, a fourth UI state, and a semantic reconciliation across ten files whose wording #2189 had just deliberately changed. Both reviewers on this PR agreed, including Codex asked adversarially to argue the other way.

**Nothing is left unfixed by closing it.** The one item the recommendation named as salvageable — the un-cancelled `setInterval` in `armTimer` — is not reachable on `main`: `startReapSchedule` has a single production call site, in the one-shot `server.listen` callback, so a second arming never happens. That defect exists inside this PR because this PR is what adds re-arming. Details in the comment above.

**The branch is not deleted.** `feat/2184-armed-reap-interval` stays, and another checkout still has it open — closing the PR should not pull work out from under it.

**#2184 stays open**, and it should be read with this thread attached: the row no longer underestimates anything, so what remains is the narrower question of whether Settings should be able to say *"this server is sweeping every N hours"* as distinct from *"N is saved for the next start"*. If that is picked up again, the cheap shape is to send the armed cadence and use it **only** where it distinguishes something the event wording cannot — not to re-introduce clock copy across five locales.

isamu:
## Merged `main` in, and narrowed this PR to what is still needed

You asked to land this rather than close it. Reconciling with #2189 turned out to **shrink** it, which is the right outcome and worth recording.

### How each of the ten conflicts was resolved

**#2189's wording is the base.** Its rule — *no time this screen can name is true, so name the event* — is correct, and this PR now sits on top of it instead of replacing it.

| | resolution |
|---|---|
| `doomed`, `doomedTitle`, `reapEnded` (×5 locales) | **#2189's, unchanged** |
| `doomedSoon`, `doomedSoonTitle`, `reapEndedSweep` | **dropped entirely** |
| `sweepHint`, `sweepOffTitle` | present tense, because the server now reports what it armed |
| `sweepPending*` | kept — the saved-vs-armed window |
| `sweepNote` | **#2189's, kept and still rendered** |
| both `config.md` guides | **#2189's sentence, unchanged** |
| the skill | #2189's rule kept; its closing bullet updated from *"#2184 would let…"* to what #2184 actually did |

**Why the row change was dropped.** #2189's `doomedTitle` already reads *"the next sweep ends it"*. Knowing the armed cadence adds nothing to a row that already names the event — so my `doomedSoon` variants were saying the same thing twice. What the armed value genuinely buys is the **hint**, which #2189 had to hedge as `Saved: …` precisely because the browser was never told what was running.

That is now the whole of this PR's contribution, and it is exactly the one thing Codex identified as surviving: *"it can distinguish 'this running server is actually sweeping every N hours' from 'this value is merely saved for the next start.'"*

### #2189's invariants are pinned, not assumed

Its three specs are kept verbatim and verified still present after the merge:

- the row names the sweep rather than a clock, at cadence 0 **and** 6
- the threshold hint stays on the event whatever is saved
- the standing note appears in **every** state, including the disabled one

Only the one spec my change makes obsolete was replaced — *"states the saved cadence without claiming the running server repeats yet"* became *"states the running cadence as a fact when saved and armed agree"*, because the browser can now honestly say it.

### Verification on the merged result

`yarn format`, `yarn lint`, `yarn typecheck`, `yarn build` clean. Five affected spec files, 61 tests, green. Break-verified again after the merge: pointing the hint back at the saved value reddens the pending-state spec.

Two lint errors surfaced from the merge (`sonarjs/prefer-specific-assertions` on my two new `.length` assertions) and are fixed.

### One pre-existing failure, established once so nobody re-derives it

`test/scripts/dev-server.spec.ts > "restarts the backend when a watched source file changes"` fails standalone even with `--testTimeout=120000`. I checked out `origin/main` (`6f5b7ad9`) and ran the same spec there: **it fails identically**. This diff touches nothing under `scripts/`.

**The full local suite is not usable on this machine right now** — it deadlocked twice today, alive at 0% CPU with no output, and had to be killed. CI is the ground truth for it.

PR #2194, opened 2026-09-19
fix(tmux): a second schedule stops the first (#2193)

### LINKED ISSUE #2193: bug(tmux): startReapSchedule を2回呼ぶと古いスイープタイマーが止まらない
## 観測した事実（再現済み）

`startReapSchedule` を2回呼ぶと、**1回目のタイマーが止まらない**。

```ts
startReapSchedule({ intervalHours: 6, ... });   // 6時間ごとの setInterval が走り始める
startReapSchedule({ intervalHours: 0, ... });   // 「繰り返さない」はずだが…
```

`armTimer`（`server/session/reap-schedule.ts`）は `setInterval` のハンドルをローカル変数
`const timer` にしか持たず、`clearInterval` はリポジトリのどこにも無い。したがって2回目の呼び出しは
**新しいスケジュールを仕掛けるだけで、古いものを止めない**。

`armedReapIntervalHours()`（#2184 で追加予定だった getter）が無い現状でも、実害はある:
2回目が 0 でも 2 でも、**最初の6時間タイマーは回り続ける**。

## 再現

fake timer で:

```ts
startReapSchedule(schedule(6));
startReapSchedule(schedule(0));
vi.advanceTimersByTime(6 * 60 * 60 * 1000);
// → sweepIdleSessions が 1 回呼ばれる（0 回であるべき）
```

実際に走らせて `sweeps fired after the OFF call: 1` を確認済み。

## 影響範囲（正直に）

**現状の本番では踏まない。** `startReapSchedule` の呼び出し元は
`server/infra/on-listening.ts` の1箇所だけで、1プロセスにつき1回しか呼ばれない。

なので これは **堅牢性 / 冪等性の修正** であって、今ユーザーに見えているバグではない。
ただし:

- `startReapSchedule` は **export されている**。2回目の呼び出しを禁じる仕組みは無く、
  型にもドキュメントにもその制約は書かれていない。
- 将来プロセス内再起動やテストから2回呼べば、**掃除タイマーが重複して走る**。セッションを
  終了させる処理が意図より多く走るのは、静かに壊れる種類の不具合。
- #2184 のように「今どの間隔で回っているか」を報告する機能を足すと、
  報告値と実際のタイマーが食い違う経路が生まれる（実際 #2191 のレビューで P2 として検出された）。

## 提案

`armTimer` が、**毎回・新しいものを仕掛ける前に**古いインターバルを止める:

```ts
if (armedTimer !== null) { clearInterval(armedTimer); armedTimer = null; }
if (!reapTimerEnabled(intervalHours)) return;
...
armedTimer = timer;
```

**`if` の前に置くのが肝心。** 「新しいスケジュールが実際に仕掛けるときだけ止める」と書くと、
2回目が 0（= 何も仕掛けない）のケース — まさに報告されている状況 — を取りこぼす。

テストは**数値ではなく時計**を見ること。「getter が 0 を返す」だけの表明は、
古いタイマーが生きていても通ってしまう（#2191 のレビューで実際にそれが起きた）。

## これは #2167 が見送った live re-arm ではない

#2167 は「config POST のたびに張り直す」を意図的に見送った（編集を連打するとカウントダウンが
永久にリセットされるため）。ここで止めるのは**呼び出し元が明示的に新しいスケジュールを開始したとき
だけ**で、config の経路からは到達しない。

## 関連

- #2165 / #2167 定期スイープの導入
- #2178 / #2179 tick が終了させたセッションのファイル掃除
- #2184 / #2191 armed な間隔を報告する案（見送り）。この欠陥はそのレビュー中に発見された

isamu:
## Summary

Closes #2193.

`armTimer` kept its `setInterval` handle in a local `const`, and nothing in the repo called
`clearInterval`. So a second schedule started a new sweep without stopping the previous one:

```
startReapSchedule(6)   →  a six-hour interval begins
startReapSchedule(0)   →  "do not repeat" … and the six-hour interval keeps firing
```

**Reproduced before changing anything**, with fake timers: advancing the clock six hours after
the OFF call fired a sweep.

`armTimer` now clears any previous interval **before** deciding whether to arm a new one.

## Items to Confirm / Review

1. **The blast radius is small and I would rather say so than oversell it.** Production does not
   hit this: `startReapSchedule` has exactly one caller (`server/infra/on-listening.ts`) and it
   runs once per process. This is **idempotence, not a live user-facing bug.** It is worth fixing
   because the function is exported, nothing expresses a once-only contract, and a duplicated
   sweep ends sessions on a cadence nobody asked for — the quiet kind of wrong.
2. **The ordering is the whole subtlety.** The clear must come *before* the `reapTimerEnabled`
   check. After it, the second call arms nothing and returns early, leaving the first interval
   running — which is the reported case exactly. The mutation table below shows that moving it
   reddens precisely that spec and nothing else.
3. **Not the live re-arming #2167 declined.** That was re-arming on every config POST, which lets
   a stream of edits reset the countdown forever. This stops a timer only when a caller
   explicitly starts a new schedule; no config path reaches it.

## How it was verified

`yarn format`, `yarn lint`, `yarn typecheck`, `yarn build` clean.

Two specs, both watching the **clock** rather than any status value — no sweep fires after an OFF
call, and 6h replaced by 2h leaves exactly three ticks in six hours. That distinction matters:
*"nothing reports a cadence"* is true the instant the second call returns while the first interval
is still very much alive, and **a number-only assertion is how this hid in review once already.**

Break-verified, each mutation's application asserted by `count(old) == 1` before running:

| mutation | result |
|---|---|
| cancellation removed entirely (the state on `main`) | 2 red |
| clear moved *below* the `reapTimerEnabled` check | 1 red — the OFF case specifically |

**The full local suite is not usable on this machine right now** — it deadlocked twice today,
alive at 0% CPU with no output, and had to be killed both times. CI is the ground truth for it.
Separately established: `test/scripts/dev-server.spec.ts > "restarts the backend when a watched
source file changes"` fails **identically on `origin/main`**, so it is pre-existing and not this
branch.

## Where this came from

This is the one piece salvaged from **#2191**, which was closed unmerged. #2191 would have had the
server report the cadence it armed so the Settings hint could state it as a fact; that was judged
not worth a new wire field and server-side state once **#2189** had already removed the underlying
falsehood a simpler way. This defect is independent of all of that UI wording — it was found
during #2191's review and survives its closure.

## User Prompt

- `#2184やって`, then `merge`, then — on being told #2191 had been closed — *「クローズのまま、タイマー修正だけ別PRに」*: leave #2191 closed and take only the timer fix to its own PR.

work in mulmoterminal4

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed session cleanup scheduling so starting a new schedule consistently replaces the previous one.
  - Disabling scheduled cleanup now stops any previously active cleanup cycle.
  - Updated scheduling behavior to prevent duplicate cleanup runs and ensure the configured interval is honored.
  - Previous schedules are now cancelled even when the replacement schedule encounters an error.

- **Tests**
  - Added coverage for disabling an active schedule, changing intervals, and handling replacement errors.


isamu:
## Cross-review tier: **C**

One line of it is on the tier-C list outright: **async ordering.** This changes the lifecycle of a `setInterval` — when it is created, when it is destroyed, and what the ordering of those two is relative to a branch that can return early. The single subtlest thing in the diff is that the `clearInterval` must sit *above* the `reapTimerEnabled` check, and getting that wrong reproduces the exact defect being fixed while looking correct.

It also **ends user sessions**: the interval being cancelled is the one that decides when someone's terminal is reaped. A duplicated timer is a session ended on a cadence nobody asked for.

What does **not** force C, stated so the tier is auditable rather than reflexive: the root cause is proven and the defect was reproduced with fake timers before any code changed, so the *"bug fix whose root cause is not proven"* trigger does not apply. There is no auth, secret, wire-format or public-API surface here, and no behaviour-preservation claim.

Round 1 names async ordering, module-level lifecycle state, and the destructive-consequence angle as explicit axes and requires a per-axis answer.

**Sandbox pre-flight** runs in **this** checkout — the same directory the round will use. That matters: earlier in this session I proved the sandbox here and then ran the round in a worktree whose symlinked `node_modules` blocked vitest, which turned an LGTM into a read rather than a run. Result goes in ledger row 0.

**Pre-existing red, recorded once so no round re-derives it:** `test/scripts/dev-server.spec.ts > "restarts the backend when a watched source file changes"` fails standalone with `--testTimeout=120000`. I checked out `origin/main` (`6f5b7ad9`) and ran the same spec there — **it fails identically**. This diff touches nothing under `scripts/`.

server/session/reap-schedule.ts
```diff
@@ -50,15 +50,34 @@ const dropEndedSessionFiles = (reaped: readonly string[]): void => {
...
+// interval running while nothing says so, and sessions keep being ended on a cadence the caller
+// asked to turn off. Putting the clear AFTER the `reapTimerEnabled` check would miss exactly that
+// case, since the second call arms nothing and would return early (#2193).
+//
+// Production calls this once, at boot, so this is idempotence rather than a live bug — but the
+// function is exported, nothing forbids a second call, and a duplicated sweep is the quiet kind of
+// wrong. It is NOT the live re-arming #2167 declined: that was re-arming on every config POST,
+// which lets a stream of edits reset the countdown forever. No config path reaches here.
 function armTimer({ intervalHours, idleDays, log }: ReapSchedule): void {
+  if (armedTimer !== null) {
```
isamu: P2: This clears the old timer before `reapTimerEnabled`, which fixes the early-return case, but it is still only reached after `startReapSchedule` has already completed the immediate `sweepNow(...)`. If a second call is meant to disable or replace the schedule and that immediate sweep throws first, through `sweepIdleSessions`, `reapSweepLines`, or the supplied `log`, the previous module-level interval is never cleared and keeps ending sessions on the old cadence. I checked the other interval lifecycle sites in this repo: I did not find another site with this exact pattern of a replaceable module-level destructive interval whose cancellation is gated behind fallible work. The closest timer owners either have no replacement API (`server/session/decision-digest-schedule.ts`, `server/session/scheduled-sessions-boot.ts`, `server/config/update-status.ts`) or stop the old interval before async/fallible re-arming (`src/components/RoomsOverlay.vue`, `src/composables/useSessionSummary.ts`, `src/composables/useMachineLoad.ts`, `server/backends/collectionWatchers.ts`). Smallest fix: split the cancellation into a `stopArmedTimer()` helper, call it at the top of `startReapSchedule` before `sweepNow`, and let the arm step only create the new interval after the immediate sweep succeeds. Add a fake-timer spec where the second immediate sweep throws and advancing the clock proves the first interval is no longer ticking.

isamu:
Review summary for PR #2194 at `8003ed015ff8121a5b44fc032c21df556085e7a9`.

Findings:
- P2 inline on `server/session/reap-schedule.ts`: a second `startReapSchedule` can still throw during the immediate `sweepNow(...)` before reaching the new cancellation block, leaving the previous interval alive and still sweeping on the old cadence. I listed the other timer lifecycle sites I checked in the inline comment; I did not find another site with the same replaceable module-level destructive interval pattern.

Axis answers:
- Async ordering / timer lifecycle: the `clearInterval` block is correctly above `reapTimerEnabled(intervalHours)`, and the exact mutation that moves it below the early return goes red. The remaining ordering bug is one level up: `startReapSchedule` does fallible sweep work before it calls the cancellation path.
- Module-level lifecycle state: double-clear is guarded and sets `armedTimer = null`; re-arming replaces the handle. I did not find test pollution in the normal fake-timer flow, but the throw-before-cancel path can leave module state pointing at a live old interval.
- Destructive consequence: the finding matters because the stale interval is the one that ends tmux sessions and removes settings/drop files; a failed replacement can silently continue destructive sweeps on the superseded cadence.
- Correctness / logic: same P2 finding; no additional logic findings.
- Error handling / failure paths: same P2 finding; the code does not handle a throw before cancellation.
- Tests: `yarn vitest run test/server/session/reap-schedule.spec.ts` passes with 11 tests. `yarn typecheck` exited 0. Mutation removing cancellation failed both new clock tests; mutation moving cancellation below `reapTimerEnabled` failed the OFF replacement test. So the added specs do close the previous scalar-only hole for the normal path. Missing coverage is the throw-before-cancel path.
- Naming / readability against `CLAUDE.md`: no new emojis, `any`, `@ts-ignore`, or added `as` casts in this PR. Function sizes remain small. I did not raise style-only prose issues.
- Docs / plan accuracy: the plan accurately describes the normal-path early-return ordering and the mutation behavior I verified. Its “on every call” framing is incomplete for the throw path above, which is why the inline asks for the cancellation to move before fallible work.

CHANGES REQUESTED

isamu:

**Codex: CHANGES REQUESTED, one P2. Accepted and fixed in `c0219c3b`.**

### The finding

The cancellation was correctly above the `reapTimerEnabled` check, but it was still only reached *after* `startReapSchedule` had run its immediate sweep — and that sweep is fallible (tmux gone, the threshold unreadable). So a replacing schedule whose sweep threw never reached the clear, and the previous interval carried on ending sessions on the cadence the caller had just replaced.

That is not a variant of #2193; it is the same defect through a second door.

I do not take a finding on reading. A throwaway spec armed a six-hour schedule, called again with an `idleDays` that throws, and advanced the clock a day: **sweeps fired on the cancelled cadence.** The repro is now a permanent spec rather than a scratch file.

### The fix

`stopArmedTimer()` is its own step, run **first** in `startReapSchedule` — ahead of both things that can skip it:

- the fallible immediate sweep, and
- `armTimer`'s enabled early-return (the original #2193 case).

The boot path is unchanged: production's single caller runs with nothing armed yet, so it is a no-op there.

### Break-verification of the new fix

Each mutation's application was asserted by count before running, and the file restored byte-identically after:

- deleting the `stopArmedTimer()` call reddens **every** cancellation spec;
- moving it to *after* the immediate sweep reddens **exactly** the throwing one.

The second is the one that matters — it shows the new spec pins the ordering against fallible work, not cancellation in general. Reproduce with `npx vitest run test/server/session/reap-schedule.spec.ts`.

### Class sweep

Codex swept for the same pattern and reached my own conclusion by a different route: the other interval owners either have no replacement API (`decision-digest-schedule.ts`, `scheduled-sessions-boot.ts`, `update-status.ts`) or already stop before fallible re-arming (`collectionWatchers.ts`, and three Vue composables). Nothing else to fix. It also independently ran both mutations the plan claimed, corroborating that claim with a second actor's run.

### Not a finding

- **CI red in `test/server/git/git-status.spec.ts`** is not this branch. Those files are byte-identical to `main`, `main`'s CI is green at this exact base, and they pass locally here. `gitStatus` awaits `gitTopLevel(cwd)` *before* registering with the coalescer, so two same-tick callers can both miss the in-flight window; `coalesce-by-key.ts` documents starting fresh as correct, while `toBe` asserts object identity. Shard re-run.
- **CodeRabbit** reported no actionable comments.

Gates: `yarn format` leaves both files unchanged, `eslint` on them is clean, `yarn typecheck` exits 0, and the spec file passes. Round 2 dispatched against `c0219c3b`.

isamu:
Round 2 review for PR #2194 at `c0219c3bd12a53d9362300c0934695b499169d83`.

Findings: none. I did not post inline comments because I do not have an actionable defect.

Axis answers:
- Async ordering / timer lifecycle: `stopArmedTimer()` is now the first step in `startReapSchedule`, ahead of both the fallible immediate sweep and `armTimer`'s enabled early return. I do not see a remaining path that arms a replacement without first cancelling the previous handle. The one intentional cancel-without-arm path is the failed replacement: the old cadence is stopped, the throwing replacement never arms. I agree with that tradeoff because keeping the old destructive cadence alive is the original defect; ending with no timer is a degraded absence of future periodic sweeps, and the next boot restores the configured schedule.
- Observability of no timer after failed re-schedule: I found no UI, route, or config reader that claims a live cadence is armed. The browser sees the saved `sessionReapIntervalHours`; existing copy explicitly says cadence changes are read at server start and does not expose the module handle.
- Module-level lifecycle state: `armedTimer` is nulled on clear, re-arming replaces it only after the interval exists, double-clear is guarded, and the focused fake-timer specs do not leave a live handle after restoration. I am not re-raising the cross-spec stale-handle concern without new evidence.
- Destructive consequence: the important direction is covered. A superseded interval can no longer keep ending sessions and deleting settings/drop files after an explicit replacement, including when the replacement sweep throws. The opposite failure mode, no interval after a failed explicit re-schedule, is less harmful than continuing a stale session-ending cadence and is not a production path today.
- Correctness / logic: no findings. The single caller remains `server/infra/on-listening.ts`, production boot still starts from no armed timer, and config POSTs still do not re-arm the cadence.
- Error handling / failure paths: before the new arm, `idleDays`, `sweepIdleSessions`, `reapSweepLines`, and `log` can throw through `sweepNow`; all now happen after the old timer is stopped. In `armTimer`, disabled cadence returns with no timer, and an enabled cadence logs before creating the interval; a logging failure therefore leaves no old cadence running. I found no practical dangling-handle path in normal Node timers.
- Tests: focused specs pass. I also verified the two requested mutations: deleting `stopArmedTimer()` reddens all three cancellation specs; moving it after the immediate sweep reddens exactly the throwing replacement spec. That matches the claimed guard strength.
- Naming / readability against `CLAUDE.md`: no new emojis, no `any`, no `@ts-ignore`, no new casts in the PR additions, and the changed functions are under 20 lines. The wording is direct enough for this repo's style; no style-only finding.
- Docs / plan accuracy: `plans/fix-2193-cancel-superseded-sweep.md` now matches the implemented ordering and the mutation results. I did not find prose that would lead a reader to the wrong code change.
- Same-pattern sweep: I re-checked `setInterval` owners and `startReapSchedule` references. I did not find another replaceable module-level destructive interval with this cancel-before-fallible-rearm requirement.

Validation run locally:
- `yarn test test/server/session/reap-schedule.spec.ts test/common/sessionReap.spec.ts test/server/infra/on-listening.spec.ts` passed, 38 tests.
- `./node_modules/.bin/eslint --format ./scripts/eslint-formatter-summary.mjs server/session/reap-schedule.ts test/server/session/reap-schedule.spec.ts` passed.
- `yarn typecheck` passed.
- `./node_modules/.bin/prettier --check server/session/reap-schedule.ts test/server/session/reap-schedule.spec.ts` passed.

LGTM

isamu:
## Cross-review round 2 — CLEAN, converged

**Codex: LGTM, no inline findings.** Both reviewers have now signed off on `c0219c3b`, which is the head that merges — nothing was pushed after the round began.

Codex independently re-ran the focused specs, both mutations, ESLint on the touched TypeScript, `typecheck`, and the Prettier check.

### The decision it was asked to argue with

Cancelling before the sweep means a replacing schedule whose sweep throws ends with **nothing** armed — the old cadence stopped, the new one never reached. I asked Codex to disagree and to check whether that state is observable anywhere.

It searched for a UI, route or config surface claiming a live armed cadence and found none, then agreed with the trade: ending with no timer is a degradation the next boot repairs, whereas leaving the old cadence running is the destructive defect this PR exists to remove. That is the same reasoning recorded in the plan, reached independently.

### Loop summary

Round 1 produced one P2 — the cancellation sat behind the fallible immediate sweep — which was reproduced before being accepted, fixed, and break-verified. Round 2 was clean.

Merging with a merge commit.

PR #2195, opened 2026-09-19
fix(shutdown): wait for queued session state to reach disk before exiting (#2161)

### LINKED ISSUE #2161: fix: 終了時に、ディスクへ届いていないセッションログの追記が失われる（待つ関数は既にある）
## What happens

Ctrl+C で終了すると、**まだディスクに届いていないセッションログの追記が失われます。**

終了経路が同期的に `process.exit(0)` を呼んでいて、キューの排出を待っていません。

`server/infra/shutdown.ts`:

```ts
process.once(signal, () => {
  stopWhisperSidecar();
  process.exit(0);
});
```

`bin/mulmoterminal.js`（親側も同じ形）:

```ts
const shutdown = () => {
  child?.kill("SIGTERM");
  process.exit(0);
};
```

## 待つべきものが既にある

`server/session/registry.ts` の追記は **fire-and-forget** で、それは意図的です。コメントがそう書いています。

```ts
/** Settles once every persist queued so far has hit the disk.
 *
 *  The appends above are deliberately fire-and-forget — a tool call must not wait on a log —
 *  but that leaves no way to know they are done ... */
export const whenToolGroupsPersisted = (): Promise<void> => toolGroupsPersist;
```

**「全部ディスクに届いたら解決する」関数は既にあります。** #1345 でテストのために作られたもので、終了経路からは呼ばれていません。

## なぜ効くか

`shutdown.ts` 自身のコメントが、この経路の重みを書いています。

> stopping MulmoTerminal from the browser (#1820) has to run THIS path and not a second one that
> drifts from it — the guarantee a user is owed is that the button does what Ctrl+C does.

つまり **Ctrl+C とブラウザの停止ボタンの両方**が、この待たない経路を通ります。落とす直前まで動いていたセッションほど、キューに未排出の追記が残っている可能性が高い。

## 提案

終了経路で `whenToolGroupsPersisted()` を待ってから exit する。ただし**待ちっぱなしにはしない** — ディスクが詰まっているときに Ctrl+C が効かなくなるほうが困るので、上限（数百ms〜1s 程度）を切って、超えたら諦めて exit する。

親側（`bin/mulmoterminal.js`）も、SIGTERM を送った直後に `process.exit(0)` するのではなく、子の `close` を短い上限つきで待つ。

## 確認していないこと

- **実際に何バイト失われるか**は測っていません。コードの形から「待っていない」ことだけが確実です
- `stopWhisperSidecar()` が同期かどうかも見ていません。もし非同期なら同じ問題を抱えています

isamu:
## Summary

Closes #2161.

**Reproduced before designing anything**, because the issue said it had not been measured. A child process that queues 20 appends and then leaves the way `shutdown.ts` leaves, against a control that drains first:

| run | file created | lines | bytes |
|---|---|---|---|
| synchronous `process.exit(0)` — today | **no** | **0 / 20** | **0** |
| drain first — the control | yes | 20 / 20 | 880 |

It is not a partial loss. A burst at exit goes **whole**, because `process.exit` runs before the first `.then` gets a microtask, so the file is not even created.

### The issue's proposal drains one queue of ten

It suggests awaiting `whenToolGroupsPersisted()`. `server/session/registry.ts` holds **ten** fire-and-forget persist chains — lines 175, 268, 365, 531, 564, 668, 707, 777, 830, 931 — writing ten separate files under `~/.mulmoterminal`, and only two have a waiter at all. Draining one leaves nine, and the next person files this issue again about a different file.

`server/session/persist-drain.ts` tracks all of them. **A queue registers how to READ its tail, not the promise** — that is the load-bearing detail, because every append *replaces* the chain (`persist = persist.then(...)`), so a promise handed over at registration is the empty tail from before any work. Awaiting it would resolve instantly and look exactly like success. A spec pins this directly; it is the one "simplification" that would silently undo the fix.

The append paths are untouched — ten one-line registrations.

### Where the waiting happens

- **`server/infra/shutdown.ts`** — the signal path becomes async: kill the sidecar (synchronous, verified, and it must not outlive us either way), drain under a cap, exit. The `exit` handler stays synchronous and drains nothing, because by then the event loop is over. The signal path is also what the browser's stop button takes (`shutdown-routes.ts` → `process.kill(pid, "SIGTERM")`), so *"the button does what Ctrl+C does"* is preserved rather than forked.
- **`bin/mulmoterminal.js`** — waited for nothing. It now waits for the child's `close`, capped, so the shell does not come back before the flush and the child is not cut off with the terminal. A second Ctrl+C means "stop waiting".

**The cap exists because losing the tail is the lesser failure** — a disk that has stopped answering must not turn Ctrl+C into a hang. When it fires, the exit path says so, so an operator can tell a clean stop from a truncated one.

Design notes: [`plans/fix-2161-drain-persist-queues-on-exit.md`](plans/fix-2161-drain-persist-queues-on-exit.md).

## Items to Confirm / Review

- **The cap is 1s in the server and 2s in the parent.** Both are judgement calls. Too short loses tails on a slow disk; too long makes Ctrl+C feel broken. Worth a second opinion.
- **`installParentShutdown` was extracted** from `main` because inlining it crossed the repo's function-length bound. That is a real change to `bin/mulmoterminal.js`'s shape, not just an addition.
- **The two per-instance registrations** (the chains created inside a factory, lines 175 and 564) register once per factory call. That is correct — each instance has its own chain — but it means the tracked list grows with the number of appenders created, which is a fixed small number today.
- Two things the issue flagged as unchecked, now checked: `stopWhisperSidecar()` is **synchronous**; the parent's `shutdown` was exactly as quoted.

## Verification

End to end, through the real `installShutdownHandlers()` and a real `SIGINT` to self:

| | file | lines | bytes |
|---|---|---|---|
| before | no | 0 / 20 | 0 |
| after | yes | **20 / 20** | **880** |

Gates: `yarn format`, `yarn lint`, `yarn typecheck`, `yarn build` all clean; the new spec is 7/7.

**Break-verified**, source restored byte-identical after each:

| mutation | result |
|---|---|
| an eleventh chain added to `registry.ts` without registering it | **red** — the guard derives the chain list from the source, so it fails *closed* |
| the drain awaiting only the first queue | **red** |
| the cap never firing | **red** |
| a rejected queue aborting the drain instead of being absorbed | **red** |

**One of those passed at first**, and it is the useful part of this section: the "waits for every queue" test flushed a single microtask before asserting, which cannot tell *"waits for all"* from *"waits for the first"* — `allSettled` resolves a few ticks later either way. It drains the macrotask queue now. The test was green for the wrong reason until the mutation said so.

### Pre-existing red, so nobody re-derives it

A broad `test/server/{session,infra}` run goes red on this machine at load 16–20, always in the same hook: `test/support/setup-temp-dirs.ts`'s `afterAll(removeTrackedTempDirs)` timing out at 10s while `rmSync` retries. **It is not this change** — repeated A/B at the same base, three runs a side:

| spec | at base | on this branch |
|---|---|---|
| `projectRoot.spec.ts` | 3/3 **fail** | 3/3 fail |
| `sharedAppTool.spec.ts` | 2 pass / 1 fail | 1 pass / 2 fail |
| `sharedAppWatch.spec.ts` | fail | fail |

Worth recording how close this came to the wrong answer: a **single** run each way said "base passes, branch fails", which reads as a regression. Repeating it is what showed both sides behave the same. One sample against a 10s timeout on a loaded machine is not an experiment.

## User Prompt

## Summary by CodeRabbit

- **Bug Fixes**
  - Session data and queued persistence writes are now given time to finish during SIGINT and SIGTERM shutdowns, reducing the risk of lost recent changes.
  - Shutdown continues safely if persistence fails or does not complete within the timeout.
  - Repeated interrupt signals still allow immediate termination when needed.
  - Child processes now receive shutdown signals and have a brief grace period to exit cleanly.

- **Reliability**
  - Shutdown warnings are shown when queued data may not have been fully saved.


isamu:
## Codex cross-review — tier and pre-flight

**Tier: C.** Three items off the tier-C list:

- **The exit path is something every run passes through**, and it is the same path the browser's stop button takes. Getting it wrong does not fail loudly — it hangs Ctrl+C, or it silently keeps losing the thing this PR is about.
- **Async ordering is the whole change**: a signal handler that became asynchronous, a race between a drain and a cap, and a parent that now waits on a child's `close`.
- **It is a data-loss fix**, so the claim "nothing is lost now" has to be tested rather than argued.

### Pre-flight

vitest, `vue-tsc -b`, network (200) and `gh` reads all work. **`rm -f` is refused by the runner**, so Codex is told not to run a mutation sweep — it names mutations and I run them.

### Pre-existing red, so no round re-derives it

A broad `test/server/{session,infra}` run goes red on this machine at load 16–20, always in one hook: `setup-temp-dirs.ts`'s `afterAll(removeTrackedTempDirs)` timing out at 10s while `rmSync` retries.

| spec | at base, 3 runs | on this branch, 3 runs |
|---|---|---|
| `projectRoot.spec.ts` | **3 fail** | 3 fail |
| `sharedAppTool.spec.ts` | 2 pass / 1 fail | 1 pass / 2 fail |
| `sharedAppWatch.spec.ts` | fail | fail |

Recorded because of how nearly it went the other way: a **single** run each side said *"base passes, branch fails"* — a regression, apparently. Repeating three times a side is what showed both behave the same. One sample against a 10s timeout on a loaded machine is not an experiment, and I nearly shipped the opposite conclusion.

server/infra/shutdown.ts
```diff
@@ -7,16 +7,34 @@
...
 
 const SIGNALS = ["SIGINT", "SIGTERM"] as const;
 
 export function installShutdownHandlers(): void {
   process.once("exit", stopWhisperSidecar);
   for (const signal of SIGNALS) {
     process.once(signal, () => {
-      stopWhisperSidecar();
-      process.exit(0);
+      void stopAndExit();
```
isamu, line 40: P2: This async handoff leaves the existing shutdown spec red. Every site: `test/server/infra/shutdown.spec.ts:48-52` still invokes the registered signal listener and immediately expects `process.exit(0)`, but `stopAndExit()` now awaits `drainPersistQueues()` before exiting, so `npx vitest run test/server/infra/shutdown.spec.ts` fails for both SIGINT and SIGTERM with zero exit calls. Smallest fix: update that spec to await the async turn (or mock the drain and await the listener consequence) before asserting the exit, while still checking sidecar cleanup happens first. What would change my mind: a green run of this exact focused spec on this branch, or evidence that this spec is intentionally excluded from the required suite.

isamu:

- P2: Existing focused shutdown spec is red after the async signal-path change.
  Every site: `test/server/infra/shutdown.spec.ts:48-52`.
  Smallest change: update the spec to await the async signal consequence, or mock the drain and await the listener outcome, before asserting `process.exit(0)`; keep the sidecar-first assertion.
  What would change my mind: `npx vitest run test/server/infra/shutdown.spec.ts` passing on this branch, or evidence that this spec is intentionally excluded from the required suite.
  Inline: https://github.com/receptron/mulmoterminal/pull/2195#discussion_r4054595382

Axis table A-J:

| Axis | Result |
|---|---|
| A. Async ordering / exit path | No runtime finding. The first same signal uses the async drain; a second same signal after `process.once` is removed falls back to Node's default termination. A different second signal can start a second drain, but the cap still prevents an indefinite wait. With the current registered getters I do not see a path where neither drain nor exit runs. `timer.unref()` is right, and `clearTimeout` after the race is fine. |
| B. Drains everything | No finding. I traced every `Promise<void> = Promise.resolve()` chain in `server/session/registry.ts`: the eight named chains plus the two factory declarations. The factory registrations read the local mutable tail, and their call sites create per-instance queues as intended: `idLogAppender` at 5 call sites and `conversationLog` at 3 call sites. No wrong-tail capture found. |
| C. Cap | No finding. 1s server / 2s parent is defensible for shutdown. If the cap wins, the process exits; abandoned promise rejections are absorbed by `allSettled` or by the existing chain `.catch` handlers, so I do not see an unhandled-rejection hazard. |
| D. Parent | No finding. Second Ctrl+C exits the parent immediately, `close` is the right event when waiting for child stdio to drain, and the injected `getChild` / `exit` / `on` shape is modest and reasonable. There is a narrow already-closed-child race that can wait until the 2s cap, but it is bounded and second Ctrl+C escapes it. |
| E. Fix complete / exit handler | No finding. Leaving the `exit` handler synchronous is correct: the event loop is already ending, and this PR moves the drain onto the signal/browser-stop path that can still await. I did not find a real queued-work path that reaches only `exit` and used to survive. |
| F. Tests | Finding above. The new `persist-drain` tests pass, but the existing shutdown spec was not updated for the async exit. I do not doubt the four named mutations from the PR notes based on this read; I did not run mutation sweeps per sandbox instruction. |
| G. Correctness / edge cases | No additional findings beyond the red spec. |
| H. Security | No finding. The exit path does not add new reachable privilege or input surface; `/api/shutdown` still goes through the existing route/guard path and sends the same signal. |
| I. Codebase consistency | No finding. The getter-based tracker fits the existing append-chain pattern and keeps append paths untouched. The only inconsistency is the stale shutdown spec covered by the P2. |
| J. Prose | No N prose finding. I did not find a sentence in the changed prose that would cause a wrong implementation change if believed. |

Verification run:

- `npx vitest run test/server/session/persist-drain.spec.ts` passed: 7 tests.
- `npx vitest run test/server/infra/shutdown.spec.ts` failed: SIGINT and SIGTERM cases still expect synchronous `process.exit(0)`.
- `npx vue-tsc -b` passed.

isamu:
## Codex exchange — round 1

One call this round: the review. No rebuttal call, because nothing was declined — the finding was real and it was my miss.

```text
Review PR #2195 at https://github.com/receptron/mulmoterminal/pull/2195.
Branch is checked out here. Range:
  git diff 6f5b7ad9fd181d6427b7a9e3ababb2b4af5fd864...HEAD

Every finding you hold back costs a full extra round. Read the WHOLE diff
before posting. Do not stop at the first problem.

Post inline via `gh api repos/receptron/mulmoterminal/pulls/2195/comments`,
verdict via `gh pr comment 2195 --repo receptron/mulmoterminal`.

── SANDBOX ──────────────────────────────────────────────────────────────────
`rm -f` is refused by your runner. DO NOT run a mutation sweep — NAME a
mutation and I will run it. `npx vitest run <file>` and `npx vue-tsc -b` work.
Do NOT run broad `test/server/{session,infra}` runs: `setup-temp-dirs.ts`'s
afterAll times out at 10s under this machine's load and takes several specs
with it, at the base as well as here (three runs a side — see the pre-flight
comment).

── WHAT THIS IS ─────────────────────────────────────────────────────────────
#2161: `process.exit()` runs no pending microtask, so session-state appends
queued just before Ctrl+C were lost. Measured: 0 of 20 lines, file not even
created. The issue proposed awaiting ONE queue; registry.ts has ten.

- new `server/session/persist-drain.ts` — queues register a GETTER for their
  tail (every append replaces the chain, so a captured promise is the empty
  pre-work tail); `drainPersistQueues` awaits all under a cap
- `registry.ts` — ten one-line registrations, append paths untouched
- `shutdown.ts` — signal path is async: kill sidecar, drain under cap, exit
- `bin/mulmoterminal.js` — parent waits for the child's `close`, capped;
  extracted to `installParentShutdown` for the function-length bound
- specs incl. a guard that derives the chain list from registry.ts's source

  A. ASYNC ORDERING / THE EXIT PATH. Is the new signal handler correct? What
     happens on a SECOND SIGINT while the drain is in flight — the server side
     has no guard, only the parent does. Can the process now fail to exit at
     all? Does `process.once` plus an async body leave any path where neither
     the drain nor the exit runs? Is `timer.unref()` right, and does
     `clearTimeout` after `Promise.race` leak anything?
  B. DOES IT ACTUALLY DRAIN EVERYTHING? Ten chains, two of them created inside
     factories so they register per instance. Is any chain missed? Is any
     registered in a way that reads the WRONG tail? Check the two factory ones
     especially (registry.ts around lines 175 and 564).
  C. THE CAP. 1s server, 2s parent. Are those defensible? What happens if the
     drain finishes AFTER the cap fired — any write-after-exit hazard, any
     unhandled rejection from the abandoned promise?
  D. THE PARENT. `installParentShutdown` — is the second-Ctrl+C path right, is
     `close` the correct event (vs `exit`), and can it now hang where it did
     not before? It takes injected seams; are they a reasonable shape or is
     that test-only machinery leaking into production code?
  E. IS THE FIX COMPLETE? The `exit` handler still drains nothing, on purpose
     (the loop is over). Is that reasoning right, or is there a real path that
     reaches `exit` with queued work that used to survive?
  F. TESTS. Is any assertion vacuous? I claim four mutations turn them red
     (eleventh chain unregistered; drain awaits only the first; cap never
     fires; rejection aborts the drain). One of my tests was GREEN FOR THE
     WRONG REASON until a mutation said so — it flushed one microtask, which
     cannot distinguish "waits for all" from "waits for the first". Look for
     more of that shape. Name any mutation you doubt and I will run it.
  G. correctness and edge cases
  H. security — anything new on the exit path?
  I. consistency with the codebase
  J. PROSE: report a sentence that would cause a WRONG CHANGE if believed;
     file anything merely behind as severity N.

1. SEVERITY P1/P2/P3/N (N is prose-only, does not block the verdict)
2. EVERY SITE — grep and list them all now
3. THE SMALLEST CHANGE THAT RESOLVES IT
4. WHAT WOULD CHANGE YOUR MIND

── ALREADY SETTLED ──────────────────────────────────────────────────────────
| # | finding | disposition |
|---|---------|-------------|
| 0 | sandbox probe | RECORDED — `rm -f` refused, no sweeps |
| 0b | pre-existing red | RECORDED — `setup-temp-dirs.ts` afterAll 10s timeout takes projectRoot (3/3 fail BOTH sides), sharedAppTool (flakes both sides), sharedAppWatch (fails at base). Not this change |

No review findings yet — this is round 1.

── END WITH ONE TOP-LEVEL COMMENT ───────────────────────────────────────────
  Then the axis table A-J. Then:

```

</details>

<details><summary>Codex stdout (the verdict and axis table are its own comment above)</summary>

```text
Posted the review.

One blocking finding: the existing `test/server/infra/shutdown.spec.ts` is now red because it still expects `process.exit(0)` synchronously, while the signal path now awaits the drain.

I also posted the requested A-J axis table and verification results in the top-level PR comment.
```

</details>

isamu:
## Round 1 triage

| # | finding | disposition |
|---|---------|-------------|
| 1 | **P1** — `test/server/infra/shutdown.spec.ts` still expected a synchronous `process.exit(0)` | **FIXED** |
| 2 | axis B — the ten chains and the two per-instance factory registrations | **CONFIRMED**, no change. Codex traced the call sites (`idLogAppender` ×5, `conversationLog` ×3) and found no wrong-tail capture |
| 3 | axis D — a child that has already exited waits out the 2s cap | **FIXED anyway.** Codex called it bounded and escapable, which it is; it is also free to avoid |

**The P1 was my miss, and the interesting part is how.** I *did* run `test/server/{session,infra}` — and it came back 33 files / 128 tests red from the pre-existing temp-dir hook timeouts. A real red was sitting inside that noise and I read the whole thing as environmental. A broad run on a loaded machine is not a substitute for running the file you changed.

### Three things beyond the findings, all mine

- **The exit no longer depends on the sidecar behaving.** Before the drain, `stopWhisperSidecar()` throwing took the process down with it. Now it would reject a promise nobody awaits and leave a process whose default termination is *suppressed* — one that ignores Ctrl+C entirely. Wrapped, and tested.
- **`installParentShutdown` returns at once when the child has already gone** (axis D).
- **Three guards the branch did not have.** Found by mutation, not by reading: **removing `await drainPersistQueues()` from the exit path left every test in that file green.** That one line is the entire point of this PR and nothing was watching it. Also now guarded: starting the drain without awaiting it, and swallowing the cap's verdict.

### Something I verified that the code now explains

A second Ctrl+C **during** the drain terminates immediately — exit 130, measured. That comes free from `process.once`: the listener is removed after it fires, so Node's default action is restored. Switching it to `process.on` would make the server unkillable for the length of the cap, so the file now says why `once` matters.

### Break-verified this round

| mutation | result |
|---|---|
| a throwing sidecar can strand the process | red |
| the exit path no longer drains | **red — was GREEN before this round** |
| the drain runs but the exit does not wait | **red — was green** |
| a truncated stop exits silently | **red — was green** |

Source restored byte-identical after each.

isamu:
CODEX VERDICT: LGTM

- No blocking findings in the full diff.
- Verification run: `npx vitest run test/server/infra/shutdown.spec.ts`, `npx vitest run test/server/session/persist-drain.spec.ts`, `npx vitest run test/server/session/reap-schedule.spec.ts`, and `npx vue-tsc -b` all pass.
- N, non-blocking prose/edge caveat: the sidecar-throw path is not quite “warns and exits 0” in the clean sense. The signal handler catches the first `stopWhisperSidecar()` throw, but `process.exit(0)` still emits the existing `exit` listener, which calls `stopWhisperSidecar` again without a guard. If that throw is deterministic, Node can print an uncaught stack from the exit listener while still exiting 0. That is not enough for me to request changes because the shutdown still exits and the persist drain has already run, but the sentence is too clean if believed literally.

| Axis | Result |
|---|---|
| A. The three additions | `stopWhisperSidecar()` try/catch: correct and needed now that the signal handler is async/voided; otherwise a throw rejects the unawaited shutdown task and can leave the default signal termination suppressed. Caveat above: the `exit` listener remains unguarded, so this is not a fully clean warning-only path. Already-exited child check in `installParentShutdown`: correct and worth keeping, not machinery; the launcher keeps the last child reference through close/restart windows, and an already-closed child will not emit another `close`, so the early return avoids a pointless grace wait. New shutdown guards: correct and needed; they pin the PR’s load-bearing behavior rather than just the old sidecar/exit assertions. |
| B. The new guards | Not vacuous. The important shutdown guards observe that `drainPersistQueues` is called, that `process.exit` is held until the mocked drain settles, and that a false cap verdict is surfaced. Mocking `drainPersistQueues` does remove incidental exercise of the real drain from `shutdown.spec.ts`, but I do not think it weakens another meaningful assertion in that file: the real drain semantics are covered directly in `persist-drain.spec.ts`, and the registry source guard covers forgotten queues. I do not need a mutation run from here. |
| C. Sidecar throw behavior | I do not think anyone should depend on the old crash. Exit 0 is the right policy for a shutdown where sidecar cleanup failed: stopping the sidecar is cleanup, not the primary shutdown result, and making Ctrl+C/Stop fail because cleanup threw is worse. The only caveat is the unguarded `exit` listener can still print an uncaught stack after the warning if `stopWhisperSidecar` throws again. |
| D. Main merge / #2193 | No interaction found. #2193 changes the reap timer’s module-level scheduling/cancellation path, not the shutdown drain or registry persist queues. It adds no new fire-and-forget persist chain and does not affect the signal/browser stop paths. The focused reap-schedule spec passes on this head. |
| E. Full diff findings missed in round 1 | No blocking findings. The parent-side already-exited check is a useful polish over the bounded race I previously called escapable; the persist-drain design still uses the necessary tail getter; and the registry registrations cover the current chains I found. |
| F. Prose that would cause a wrong change if believed | “The new sidecar-throw path warns and exits 0” is too absolute. More precise: it catches and warns on the signal-path stop attempt, drains, then exits 0; the existing `exit` listener may still call the same throwing cleanup again and print an uncaught stack. Severity N. |

isamu:
## Codex exchange — round 2

```text
Review PR #2195, round 2. Head is now 4eafb5d2 (which also merges origin/main).
  git diff 6f5b7ad9fd181d6427b7a9e3ababb2b4af5fd864...HEAD
The delta since your round-1 verdict is 420881fc plus a main merge.

Read the whole diff before posting. Hold nothing back.

── SANDBOX ──────────────────────────────────────────────────────────────────
`rm -f` refused — no mutation sweeps; NAME one and I run it. Avoid broad
`test/server/{session,infra}` runs (pre-existing temp-dir hook timeouts, see
ledger row 0b). `npx vitest run <file>` and `npx vue-tsc -b` are fine.

── WHAT CHANGED SINCE ROUND 1 ───────────────────────────────────────────────
Your P1 fixed: the shutdown spec flushes a macrotask and asserts on the async
path. Plus three things you did NOT ask for, which is what I most want checked:

1. `stopWhisperSidecar()` is now wrapped in try/catch so a throwing sidecar
   cannot strand a process whose default termination is suppressed.
2. `installParentShutdown` returns immediately when the child has already
   exited (the race you called bounded and escapable).
3. Three new guards in the shutdown spec. I found by mutation that REMOVING
   `await drainPersistQueues()` from the exit path left the whole file green —
   the one line this PR exists for was unguarded. Also guarded now: starting
   the drain without awaiting it, and swallowing the cap's verdict.

── AXES — answer EVERY line ─────────────────────────────────────────────────
  A. THE THREE ADDITIONS. Each is a behaviour change you did not request. Is
     each correct, and is each one actually needed, or is any of it me adding
     machinery? Say so plainly if the try/catch or the already-exited check is
     unnecessary.
  B. THE NEW GUARDS. Are they vacuous? `drainPersistQueues` is now vi.mock'd in
     that spec — does mocking it make any OTHER assertion in the file weaker
     than it was (it previously exercised the real drain)? Name a mutation you
     doubt and I will run it.
  C. Does the try/catch change what happens on a sidecar throw in a way anyone
     depended on? The old path crashed the process; the new one warns and exits
     0. Is exit 0 right for a shutdown whose sidecar failed?
  D. THE MAIN MERGE. `origin/main` moved (it now carries #2193, a fix to
     `reap-schedule.ts` about cancelling a superseded sweep). Does anything in
     it interact with this change?
  E. Anything in the FULL diff you would raise now and did not in round 1.
  F. PROSE: a sentence that would cause a WRONG CHANGE if believed. Anything
     merely behind is severity N.

── ALREADY SETTLED ──────────────────────────────────────────────────────────
| # | finding | disposition |
|---|---------|-------------|
| 0 | sandbox probe | RECORDED — `rm -f` refused |
| 0b | pre-existing red | RECORDED — setup-temp-dirs afterAll 10s timeout; projectRoot 3/3 fail BOTH sides, sharedAppTool flakes both sides, sharedAppWatch fails at base |
| 1 | stale shutdown spec expected a synchronous exit | FIXED this round |
| 2 | your axis B trace of the ten chains and the factory call sites | CONFIRMED, no change needed |

── END WITH ONE TOP-LEVEL COMMENT ───────────────────────────────────────────
  'CODEX VERDICT: LGTM' or 'CODEX VERDICT: CHANGES REQUESTED' + bullets, then
  the axis table A-F, then

Do not apply fixes. POST the verdict as a top-level PR comment.
```

</details>

<details><summary>Codex stdout</summary>

```text
Posted the top-level PR comment: https://github.com/receptron/mulmoterminal/pull/2195#issuecomment-5745128543

Verdict was `CODEX VERDICT: LGTM`, with one non-blocking severity N caveat about the sidecar-throw prose/edge path. All targeted checks passed.
```

</details>

isamu:
## Round 2 triage — the N was a real edge

Codex gave `LGTM` with one severity-N caveat, which by the protocol does not block. I took it anyway, because reading it closely it is **not** prose:

> the signal handler catches the first `stopWhisperSidecar()` throw, but `process.exit(0)` still emits the existing `exit` listener, which calls `stopWhisperSidecar` again without a guard.

That is a second, unguarded call. A sidecar that throws deterministically throws **again from inside the exit handler** and prints an uncaught stack on the way out. My PR body's "warns and exits 0" was not the whole story — which is exactly the bar for taking an N seriously.

The stop is now **latched**: at most once per process lifecycle, throw warned about rather than propagated. The exit listener still stops it on the normal-return path, which is the reason that listener exists — nothing there regressed.

**The latch reset is worth its own line.** It lives in `installShutdownHandlers`, because installing handlers *is* the start of a lifecycle. I did not write it that way first, and three specs went red immediately: the first test to stop the sidecar left every later one looking at an already-stopped one. Module state leaking across tests, caught by the tests it broke.

| mutation | result |
|---|---|
| the double-stop restored | red |
| the latch reset removed | red (3) |

Source restored byte-identical after each.

Also confirmed by Codex this round and worth recording so it is not re-derived: the already-exited-child check is *"correct and worth keeping, not machinery"*, the drain guards *"pin the PR's load-bearing behavior"*, exit 0 after a failed sidecar stop is the right policy (*"stopping the sidecar is cleanup, not the primary shutdown result"*), and the `main` merge carrying #2193 does not interact with this change.

isamu:
CODEX VERDICT: LGTM

- No findings in round 3.
- Targeted verification passed: `npm test -- test/server/infra/shutdown.spec.ts` and `npm test -- test/server/session/persist-drain.spec.ts test/server/session/reap-schedule.spec.ts`.
- I did not run mutation sweeps or broad `test/server/{session,infra}` runs.

| Axis | Result |
|---|---|
| A | Latch looks correct for the sidecar stop. Two installs before any shutdown still produce one sidecar stop when the stacked exit/signal listeners run, because the first call flips `sidecarStopped` and the rest return. A signal while `stopAndExit` is already running can start another drain/exit path for the other signal, but it does not stop the sidecar again. The normal-return path is preserved: after `installShutdownHandlers()`, an `exit` event with no preceding signal sees `sidecarStopped === false` and stops the sidecar, which is why the exit listener exists. |
| B | Resetting the latch in `installShutdownHandlers()` is sound enough here, not a blocking test seam: production installs once at process boot, and installing handlers is the only lifecycle boundary this module exposes. The one theoretical bad shape is calling `installShutdownHandlers()` after shutdown has already begun, which could reset the latch before the later `exit` event; but that would already be an unsupported stacked-listener lifecycle, not a normal production path introduced by this change. If this module ever needs real multi-install semantics, I would make installation idempotent or move the state into an explicit handler instance rather than add a test-only reset. |
| C | The new idempotence test pins the regression in combination with the surrounding tests, not entirely by itself. Mutation to run if desired: replace `process.once("exit", stopSidecarOnce)` with `process.once("exit", () => {})`. The new “at most once across signal path and exit listener” test would still pass because the exit listener no longer re-runs cleanup, but the existing normal-return test should fail because the sidecar is no longer stopped on ordinary process exit. Removing the latch alone should fail the new idempotence test. |
| D | No new findings in the full diff beyond what was already settled. The persistence drain, queue registration guard, parent wait cap, and reap-schedule cancellation still look coherent. |
| E | Wrong-change prose: “The `exit` listener is redundant now that the signal path stops the sidecar.” Believing that would remove the normal-return cleanup path. |

isamu:
## Codex exchange — round 3

```text
Review PR #2195, round 3. Head is now d49f8842.
  git diff 6f5b7ad9fd181d6427b7a9e3ababb2b4af5fd864...HEAD
The delta since your round-2 LGTM is d49f8842 alone.

── SANDBOX ──────────────────────────────────────────────────────────────────
`rm -f` refused — no mutation sweeps; NAME one and I run it. Avoid broad
`test/server/{session,infra}` runs (pre-existing temp-dir hook timeouts).

── WHAT CHANGED ─────────────────────────────────────────────────────────────
Your round-2 severity-N caveat, taken as a real edge rather than as prose:
`process.exit(0)` emits `exit`, whose listener was the same sidecar cleanup the
signal path had just run, so a deterministic throw threw again from inside the
exit handler. The stop is now latched — at most once per process lifecycle —
and the latch resets in `installShutdownHandlers` because installing handlers
is the start of a lifecycle. (Without that reset three specs went red: the
first test to stop the sidecar left every later one seeing it stopped.)

── AXES — answer EVERY line ─────────────────────────────────────────────────
  A. Is the latch correct? Consider: two calls to `installShutdownHandlers` in
     one process; a signal arriving while `stopAndExit` is already running; the
     `exit` listener firing without any signal (the normal-return path) — does
     the sidecar still get stopped in THAT case, which is the reason the exit
     listener exists at all?
  B. Is resetting the latch inside `installShutdownHandlers` sound, or is it a
     test seam wearing production clothing? If the latter, say what you would
     do instead.
  C. Does the new idempotence test pin the thing it claims, or could it pass
     with the latch removed in some other way? Name a mutation and I run it.
  D. Anything in the FULL diff you would raise now and did not before.
  E. PROSE: a sentence that would cause a WRONG CHANGE if believed.

── ALREADY SETTLED ──────────────────────────────────────────────────────────
| # | finding | disposition |
|---|---------|-------------|
| 0 | sandbox probe | RECORDED — `rm -f` refused |
| 0b | pre-existing red | RECORDED — setup-temp-dirs afterAll 10s timeout, both sides |
| 1 | stale shutdown spec expected a sync exit | FIXED round 1 |
| 2 | ten chains + factory registrations | CONFIRMED by your axis B |
| 3 | already-exited child waits out the cap | FIXED round 1 |
| 4 | removing the drain left the spec green | FIXED round 1 — three guards added |
| 5 | the `exit` listener re-ran a throwing sidecar | FIXED this round |

── END WITH ONE TOP-LEVEL COMMENT ───────────────────────────────────────────
  'CODEX VERDICT: LGTM' or 'CODEX VERDICT: CHANGES REQUESTED' + bullets, then
  the axis table A-E, then

Do not apply fixes. POST the verdict as a top-level PR comment.
```

</details>

<details><summary>Codex stdout</summary>

```text
Posted the top-level PR comment:

https://github.com/receptron/mulmoterminal/pull/2195#issuecomment-5745152795

Verdict was `CODEX VERDICT: LGTM`. Targeted specs passed; no mutation sweep or broad session/infra runs.
```

</details>

**The mutation Codex asked for in axis C, run:** making the `exit` listener a no-op fails `stops the sidecar on a normal return` (1 failed / 12 passed) while the new idempotence test still passes — exactly its prediction, and the confirmation that the two tests cover different halves rather than overlapping.

isamu:
## Cross-review loop closed — tier C, three rounds, clean on the merge head

Exit met on `d49f8842`: Codex `LGTM`, axes A-E answered, `FINDINGS COMPLETE`, my own evaluation with no MUST-FIX left, nothing pushed since.

### Ledger

| 0 | pre | Codex sandbox probe | RECORDED | vitest exit 0, `vue-tsc -b` exit 0, network 200, `gh` reads OK. `rm -f` refused by the runner before execution → **no mutation sweep**; Claude runs any mutation Codex names. |
| 0b | pre | pre-existing red, re-derived at this base | RECORDED | `test/support/setup-temp-dirs.ts`'s `afterAll` times out at 10s under load, taking `projectRoot.spec.ts` (3/3 fail at base AND branch), `sharedAppTool.spec.ts` (2/3 vs 1/3 — flakes both sides) and `sharedAppWatch.spec.ts` (fails at base). Established by three runs a side; a single run each way had said "base passes, branch fails", which was the outlier. |
| 1 | 1 | P1 — `shutdown.spec.ts` still expected a synchronous `process.exit(0)` | FIXED | Claude's miss: the broad `test/server/{session,infra}` run was 33 files red from the pre-existing temp-dir hook, and a real red hid in the noise. Spec now flushes a MACROtask (one microtask cannot observe the drain). |
| 2 | 1 | axis B — are all ten chains drained, including the two per-instance factory ones? | CONFIRMED, no change | Codex traced the call sites (`idLogAppender` ×5, `conversationLog` ×3); no wrong-tail capture. |
| 3 | 1 | axis D — an already-exited child waits out the 2s cap | FIXED though filed as no-finding | bounded and escapable, and free to avoid: no further `close` is coming. |
| 4 | 1 | Claude, by mutation: removing `await drainPersistQueues()` left the whole spec GREEN | FIXED | the one line this PR exists for was unguarded. Three guards added (drain called / exit waits for it / cap verdict surfaced); all three red under mutation. |
| 5 | 2 | N — "warns and exits 0" too clean: `process.exit` re-runs the sidecar cleanup via the `exit` listener, unguarded | FIXED, though N does not block | a real second call, not prose. Stop is latched at most once per lifecycle; latch resets in `installShutdownHandlers`, without which three specs went red from module state leaking between tests. |
| 6 | 3 | axis C — a mutation Codex asked to be run: make the `exit` listener a no-op | RUN, result as predicted | the normal-return test fails, the idempotence test still passes — the pair covers different halves. |
| 7 | 3 | axis E — "the exit listener is redundant now" would be a wrong change if believed | NO CHANGE NEEDED | a warning about a belief, not a defect: `shutdown.ts`'s header and the spec both already say the `exit` listener covers the normal-return path. |

### Why three rounds, since more than two asks for an explanation

Round 1 found one thing; rounds 2 and 3 are the rounds those fixes owed. Nothing was trickled — every round carried `FINDINGS COMPLETE` and each new finding was about the PREVIOUS round's fix, which is the loop working rather than a reviewer holding back:

- round 1 → the stale spec (real, and my miss)
- round 2 → the async exit I wrote in round 1 re-ran the sidecar cleanup through the `exit` listener
- round 3 → clean

The round-reducing devices did their job in one place worth naming: round 1's batch carried **four** changes (the finding, plus three I found myself), so those did not become rounds 2, 3 and 4.

### What the mutations found that reading did not

Two of the most important facts in this PR came from mutation, not from review or from me:

1. **Removing `await drainPersistQueues()` left every test in the shutdown spec green.** The single line this PR exists for was unguarded until a mutation said so.
2. **A test that flushed one microtask could not tell "waits for all queues" from "waits for the first".** It passed against that mutation. It drains a macrotask now — and the same trap reappeared in the shutdown spec, where I knew to avoid it the second time.

And one from the tests themselves: adding the sidecar latch turned three specs red immediately, because module state leaked between tests. The reset lives in `installShutdownHandlers` for that reason.

### Recorded so it is not re-derived

A broad `test/server/{session,infra}` run is red on this machine at load 16-20 — `setup-temp-dirs.ts`'s `afterAll` timing out at 10s. Three runs a side at the same base: `projectRoot.spec.ts` fails 3/3 on BOTH, `sharedAppTool.spec.ts` flakes on both, `sharedAppWatch.spec.ts` fails at base. A **single** run each way had said "base passes, branch fails", which reads as a regression and is not one.

PR #2197, opened 2026-09-19
fix(git-status): coalesce the key lookup so concurrent reads of one cwd join (#2196)

isamu:
## Summary

`gitStatus` resolved its coalescing key *before* registering with the coalescer:

```ts
const top = await gitTopLevel(cwd);          // a real `rev-parse` process
return coalesce(top, () => readGitStatus(cwd), opts);
```

`coalesceByKey` registers synchronously, so a caller **that already has its key** always joins one in flight. The gap is the `await` above it: for the whole key lookup the caller is not in the in-flight map and cannot be seen. If the first caller's entire read then finishes before the second caller's `rev-parse` answers, the second finds nothing to join and starts another full read of the worktree.

This PR coalesces the key lookup too, keyed by **cwd** — the only thing that identifies a caller before `rev-parse` answers. Callers sharing a cwd now share one key promise, resume together, and register before any of them can finish.

The read below stays keyed by the **top level**. Per-cwd keying there is exactly what #2164 rejected: one cell's cwd can be a subdirectory of another's, and per-cwd those two would stack reads of one worktree again.

**Partially addresses #2196** — deliberately NOT `Closes`. The issue describes reads of one worktree failing to join; this closes that for callers sharing a cwd and leaves it open for two different cwds, for the reason in item 1 below. #2196 stays open to carry the remainder.

## Items to Confirm / Review

1. **The scope of the fix is smaller than the issue proposed, and it was measured rather than argued.** Two *different* cwds of one worktree still run two reads — they cannot be known to share a worktree before `rev-parse` answers for each. I wrote a spec that defers the key lookup per call so the interleaving is chosen rather than waited for: before the change both the same-cwd and different-cwd pairs ran two reads; after it the same-cwd pair runs one and the different-cwd pair still runs two. **Please confirm you are happy shipping the partial fix** rather than holding out for the whole thing.

2. **What closing the rest would cost, and why I did not do it.** It needs a remembered cwd-to-root mapping. That is a correctness decision, not a tidy-up: the read is keyed by root but *run* with the first registrant's cwd, so a stale mapping would serve one worktree's status for another. It wants its own issue and its own argument about invalidation. I did not take it unilaterally.

3. **`fresh` is forwarded to the key lookup as well as to the read — corrected during review.** The first draft did not forward it, arguing that a worktree root is not what a turn changes. That is false for exactly the turns a forced read exists to observe: a turn is arbitrary work and can run `git init` or `git worktree add`, after which a forced read joining a lookup that started earlier would answer about the directory as it was. Found in my own review pass before round 1, not by a reviewer. A spec now pins it, and `test/server/routes/git-status-fresh.spec.ts` still passes.

4. **A flaky assertion was changed, not deleted.** `serves two different cwds of ONE worktree from a single run` asserted `toBe` — object identity — on the pair now known not to be joinable on demand. It passed on timing and failed on a loaded runner. It now asserts the answer and says why in place. The same-cwd spec keeps its identity assertion, which this change makes deterministic.

## User Prompt

- git-status の flake を issue にして
- （issue の枠組みの確認に対し)「製品バグとして作り直す」
- つづけて

## Approach and key decisions

**The defect is real in production, and the path was traced rather than assumed.** `useGitStatus.ts` polls `/api/git-status` per terminal, served by `dir-routes.ts` calling `gitStatus(cwd)`. Several cells on one repository is ordinary use. Their reads coincide on tab focus and on first mount: `usePollWhileVisible` registers `focus` and `visibilitychange` per cell, and one event dispatches to all of them. (The intervals themselves are per component and drift apart, so those two moments are the ones that align — I had assumed a shared timer and checked.)

Missing the join costs a duplicate read, and a read is several git processes over the whole worktree. The window widens exactly when process spawning is contended, so #2164's mitigation weakened under the load it exists for.

**The behaviour that looked wrong was correct.** `coalesce-by-key.ts` documents itself as deliberately not a cache — a caller arriving after the previous run settled *should* start a fresh one. The thing in the wrong place was the key lookup, not the coalescer.

**Verification.** The interleaving is forced deterministically in a new spec file instead of being waited for, which is how this reached CI as a flake rather than as a test. Break-verified: putting the key lookup back outside the coalescing reddens the two coalescing specs, while the boundary spec stays green either way — which is what makes it a boundary. The repo-backed specs next door were re-run standalone with raised timeouts, because their `beforeEach` builds a temp git repo and times out on a loaded machine for reasons unrelated to this change.

Gates: `yarn format` leaves every touched file unchanged, `eslint` on them is clean, `yarn typecheck` exits 0, and `npx vitest run test/server/git/ test/server/routes/git-status-fresh.spec.ts` is green.

work in mulmoterminal4

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Git status handling for concurrent requests, reducing duplicate repository lookups.
  - Preserved efficient status reads for callers accessing the same worktree.
  - Ensured fresh requests correctly bypass in-flight lookup reuse.
  - Maintained consistent Git status results for different paths within the same worktree.

- **Tests**
  - Added coverage for concurrent lookups, fresh requests, shared working directories, and separate working directories.


isamu:
## Cross-review tier: **C**

Three separate lines of the tier-C list, none of them about diff size:

- **Concurrency and async ordering.** The entire change is *when* a caller becomes visible to other callers relative to an `await`. That is the class of bug where reading the diff and being convinced proves nothing.
- **Coalescing and invalidation.** A second coalescer is introduced, with a different key from the existing one, and a deliberate decision not to forward `fresh` to it.
- **A shared path every caller passes through.** `gitStatus` serves `/api/git-status`, which every terminal cell polls.

What does **not** force C, stated so the tier is auditable: the defect was reproduced deterministically before the fix, so the *"root cause not proven"* trigger does not apply. No auth, secrets, wire format or on-disk format is touched.

Round 1 names those axes explicitly and requires a per-axis answer.

**A specific thing I want the second reviewer to attack:** the PR ships a *partial* fix and says so. Two different cwds of one worktree still run two reads. I measured that rather than assumed it, and I declined to close it because doing so needs a remembered cwd-to-root mapping — and the read is keyed by root but *run* with the first registrant's cwd, so a stale mapping would serve one worktree's status for another. If that reasoning is wrong, this round is where it should break.

test/server/git/git-status-coalesce.spec.ts
```diff
@@ -0,0 +1,103 @@
...
+
+  // A forced read must not join the key lookup either. `fresh` is the post-turn read, and a turn is
+  // arbitrary work — one that ran `git init` here changes the answer, so joining a lookup that
+  // started before it would describe the directory as it was.
+  it("does not join an in-flight key lookup when the caller asked for a fresh read", async () => {
+    const polling = gitStatus("/repo");
+    await settle();
+    const forced = gitStatus("/repo", { fresh: true });
+    await settle();
+    expect(gitTopLevel).toHaveBeenCalledTimes(2);
```
isamu, line 77: P3: This spec can leave the imported `gitStatus` module with an unresolved `coalesceTopLevel` entry for `/repo` if it fails before the deferred lookups are resolved. For example, the intended mutation that stops forwarding `{ fresh: true }` to `coalesceTopLevel` makes this assertion fail before line 69 runs; `beforeEach` then clears `pendingTopLevel`, losing the only resolvers, while the module-level coalescer still has the pending `/repo` promise. Later tests using `/repo` can then join that never-settling promise and time out instead of reporting the mutation cleanly. I grepped the repo with `rg "coalesceByKey" server test`: the other coalescer specs create a fresh coalescer per test, and this new file is the only same-pattern site with an imported module-level coalescer plus shared deferred state. Smallest fix: add an `afterEach` that resolves/splices any pending top-level gates and awaits `settle()`, or make each test use a unique cwd so a stranded key cannot poison later tests. What would change my mind: showing the module graph/coalescer state is reset between tests, or a mutation run where this assertion fails without leaving later cases wedged.

isamu:

- P3: `test/server/git/git-status-coalesce.spec.ts` can strand an unresolved module-level coalescer entry when a spec fails before resolving its deferred `gitTopLevel` calls, causing later same-file tests to join a promise that can no longer be settled.

| Axis | Findings |
| --- | --- |
| CONCURRENCY AND ASYNC ORDERING | none in shipped `gitStatus`: same-cwd callers now share the top-level lookup before registering the root-keyed read; `fresh` bypasses both maps. P3 in tests: a failed interleaving assertion can leave the shared coalescer state wedged. |
| COALESCING / INVALIDATION | none in production. `gitTopLevel` rejection/null does not wedge because `coalesceByKey` wraps sync throws and deletes in `finally`; `gitTopLevel` itself resolves null on git failure. `fresh` replaces the cwd-keyed lookup and the root-keyed read, with stale-run deletion guarded. |
| BLAST RADIUS | none. `/api/git-status` remains the only polling route affected; the response path and shape are unchanged. |
| correctness and edge cases (non-repo dirs, a null top level, concurrent mixed fresh/non-fresh callers) | none. Non-repo/null top-level returns the existing `NOT_REPO`; mixed fresh/non-fresh callers get the fresh replacement semantics from `coalesceByKey`; two different cwds in one worktree remain a documented limitation. |
| error handling and failure paths | none in production. `git()` does not reject for normal git failures, and a thrown/rejected lookup/read still settles the coalescer entry through `finally`. |
| tests: do the new specs go RED on the right mutations, and do they assert the OBSERVABLE property rather than a timing coincidence? The specs use module-level coalescer state shared across tests in one file — is that sound, or can one test leak an in-flight entry that wedges a later one? | P3 inline. The specs assert observable call count/object identity for same-cwd coalescing and value equality for the timing-dependent repo-backed case, but the new file needs cleanup or unique keys so a failed mutation cannot leak an unresolved `/repo` lookup into later tests. |
| API / wire compatibility (the `/api/git-status` response shape and the `fresh` query) | none. `GitStatus` is unchanged and `fresh` remains `fresh=1` only; route coverage confirms `{ fresh: true }` vs `{ fresh: false }`. |
| consistency with the rest of the codebase | none. The change uses the existing `coalesceByKey` abstraction and preserves root-keying for the expensive read. `rg "coalesceByKey" server test` found no other production sites of this pattern. |
| security — anything reachable through a cwd the user controls | none. `/api/git-status` still goes through `workspaceForRoute`, which canonicalizes existing absolute directories and rejects malformed/unusable explicit cwd values before spawning git via argv. |
| docs/plan accuracy. Report a sentence ONLY if believing it would cause a WRONG CHANGE; file anything merely behind as severity N. | none. The plan accurately describes the partial fix and the stale cwd-to-root mapping risk; I do not think the limitation spec ossifies a bug as long as it is framed as a boundary for this partial fix. |

Verification: `git diff 16bf31ae...HEAD` read in full; `git diff --check 16bf31ae...HEAD` passed; `yarn vitest run test/server/git/git-status-coalesce.spec.ts test/server/infra/coalesce-by-key.spec.ts test/server/routes/git-status-fresh.spec.ts --hookTimeout=120000` passed; `yarn vitest run test/server/git/git-status.spec.ts --hookTimeout=120000` passed; `yarn typecheck` passed.

isamu:

**Codex: CHANGES REQUESTED, one P3. Accepted and fixed in `321d07e6`.**

### The finding

The spec file shares the imported module's coalescers across its tests. A test that failed *before* resolving its deferred lookup stranded that key, and later tests joined a promise that never settled — so they timed out. The real result was one clean assertion failure; what you saw was a wedge.

### Why this one is worth recording

**I had already seen this symptom and failed to diagnose it.** During my own break-verification the mutation produced one clean failure plus two fifteen-second hangs, and I wrote the hangs off as "a consequence of a mutated build" and moved on. Codex named the mechanism. That is the second reviewer doing exactly what it is for — the evidence was in my terminal and I had explained it away.

### The fix went further than the finding proposed

Codex offered two remedies; I took the per-test-key one. But a unique **cwd** alone is not sufficient: the *read* coalescer is keyed by the **worktree root**, which these tests share, so a stranded read entry would still cross tests. Each test now gets its own root, so both keys are unique.

### Verified the way the finding asked for

Its "what would change your mind" was a mutation run where the assertion fails without leaving later cases wedged. Re-running the same mutation (stop forwarding `fresh`) now fails **one** assertion in milliseconds with the other cases passing, where before it failed one and wedged two at 15s.

### Class sweep

Codex grepped `coalesceByKey` across `server` and `test`: the other coalescer specs construct a fresh coalescer per test, and this new file was the only site with an imported module-level coalescer plus shared deferred state. Nothing else to fix.

Gates on the new head: `eslint` clean, `yarn typecheck` exits 0, and `npx vitest run test/server/git/ test/server/routes/git-status-fresh.spec.ts` is green. Round 2 dispatched against `321d07e6` without waiting for CI.

```text
Review PR #2197 at https://github.com/receptron/mulmoterminal/pull/2197
Head: fe6cba06. Base: 16bf31ae. You are IN the checkout at that head.
Read the whole diff: git diff 16bf31ae...HEAD

time in the round AFTER it is pushed, so a finding raised late is not a later finding,
it is a repeat of this entire exchange. Read the WHOLE diff before you post anything.
Do not stop at the first problem. Do not save anything for "a follow-up review".

This PR is TIER C. The first three are why:

  * CONCURRENCY AND ASYNC ORDERING. The whole change is WHEN a caller becomes visible
    to other callers relative to an `await`. Reading it and being convinced proves
    nothing — construct the interleavings.
  * COALESCING / INVALIDATION. A second coalescer is added with a DIFFERENT key (cwd)
    from the existing one (worktree top level). Is the key choice right in both
    directions? Can either map leak an entry that never settles, wedging that key
    forever? What happens when `gitTopLevel` REJECTS rather than resolves?
  * BLAST RADIUS. `gitStatus` serves `/api/git-status`, which every terminal cell polls.
  * correctness and edge cases (non-repo dirs, a null top level, concurrent mixed
    fresh/non-fresh callers)
  * error handling and failure paths
  * tests: do the new specs go RED on the right mutations, and do they assert the
    OBSERVABLE property rather than a timing coincidence? The specs use module-level
    coalescer state shared across tests in one file — is that sound, or can one test
    leak an in-flight entry that wedges a later one?
  * API / wire compatibility (the `/api/git-status` response shape and the `fresh` query)
  * consistency with the rest of the codebase
  * security — anything reachable through a cwd the user controls
  * docs/plan accuracy. Report a sentence ONLY if believing it would cause a WRONG
    CHANGE; file anything merely behind as severity N.

── THE DECISION I MOST WANT ATTACKED ────────────────────────────────────────
This PR ships a PARTIAL fix and says so. Two DIFFERENT cwds of one worktree still run
two reads, because they are not known to share a worktree until `rev-parse` answers for
each. I measured that rather than assumed it, and there is a spec asserting the boundary
(`does not pretend to join two different cwds that are still resolving`).

I declined to close it. Closing needs a remembered cwd-to-root mapping, and the read is
keyed by root but RUN with the first registrant's cwd — so a stale mapping would serve
one worktree's status for another. If that reasoning is wrong, break it here.
Also: is a spec that asserts a LIMITATION a good idea, or does it ossify a bug?

1. SEVERITY — P1 / P2 / P3 / N. N is prose-only and does not block the verdict.
2. EVERY SITE of the same pattern in the repo — grep, do not guess.

── ALREADY SETTLED — do not re-raise in the same form ───────────────────────
row 0  | Sandbox pre-flight ALL GREEN: scoped vitest 0, typecheck 0, socket 0, curl 200,
         gh 0, deletions permitted. Use plain `rm`, NOT `rm -f` (refused by policy).
         You MAY run mutations.
row 0b | Do NOT run `yarn test` — the full suite deadlocks on this machine. Scoped runs only.
row 0c | A HOOK TIMEOUT in test/server/git/git-status.spec.ts is the machine, not the code:
         its beforeEach builds a temp git repo with execFileSync and exceeds the 10s hook
         budget under load. Re-run with --hookTimeout=120000 before calling it a finding.
row 1  | (MINE, fixed BEFORE this round) `fresh` was not forwarded to the key lookup, so a
         forced post-turn read could join a lookup started before the turn. A turn is
         arbitrary work and can run `git init`. Now forwarded; break-verified; claims swept
         across the comment, the plan and the PR body. Do NOT re-raise as new — but DO say
         if the fix is wrong or incomplete.

If a resolution above is wrong, say so and mark it 'REOPENING #<row>' with evidence.

Line 1, on its own:
  'CODEX VERDICT: LGTM'  or  'CODEX VERDICT: CHANGES REQUESTED' + bulleted summary.
Then, on its own line:

Post inline comments via gh api repos/receptron/mulmoterminal/pulls/2197/comments
and exactly ONE top-level summary via gh pr comment 2197.
```

</details>

<details><summary>Round 1 — Codex reply (verbatim stdout)</summary>

```text
Posted the review.

I left one inline P3 on `test/server/git/git-status-coalesce.spec.ts`: a failed spec can strand the module-level `/repo` coalescer promise and wedge later tests in the same file. The production coalescing change itself looked sound to me across the requested axes.

Verification run:
- `git diff --check 16bf31ae...HEAD`
- scoped vitest for coalescer/git-status/fresh route specs with `--hookTimeout=120000`
- `yarn typecheck`
EXIT=0
```

</details>

Codex also posted its own inline finding and verdict comment; this is the half of the exchange that would otherwise exist only in a terminal.

isamu:
CODEX VERDICT: LGTM

| Axis | Findings |
| --- | --- |
| CONCURRENCY AND ASYNC ORDERING | none. I still hold the round-1 production verdict: registering the top-level lookup synchronously through `coalesceTopLevel(cwd, ..., opts)` closes the same-cwd gap without changing the read coalescer's root-keyed semantics. `fresh` correctly replaces both the lookup and read entries, so a post-turn caller does not join work that may have sampled pre-turn state. |
| COALESCING / INVALIDATION | none. Both coalescers delete their entry in `finally`, guarded so an older replaced run cannot evict its successor. In production, ordinary `gitTopLevel` failures resolve to `null` via the git wrapper and then return `NOT_REPO`; if `gitTopLevel` ever did reject unexpectedly, `coalesceByKey` would still run `finally`, delete the top-level entry, and propagate the rejection rather than leaking a permanently pending map entry. |
| BLAST RADIUS | none. `/api/git-status` is polled by every cell, and the change reduces duplicate work for same-cwd bursts while preserving the existing one-read-per-worktree boundary after roots are known. No wire shape or route behavior changes. |
| correctness and edge cases | none. The accepted partial boundary is reasonable: the spec asserting that two different unresolved cwds in one worktree still run two reads documents a real limitation, not an ossified bug for this PR. Closing it would require a remembered cwd-to-root mapping; the stated concern is sound because a stale mapping could key the read by an old root while executing it from the first registrant's cwd, potentially serving one worktree's status for another after `git init`, worktree changes, deletion/recreation, or path reuse. |
| error handling and failure paths | none. `NOT_REPO` behavior is preserved for non-repos and failed `rev-parse`; coalesced rejection cleanup is covered by the shared coalescer tests. I am not reopening row 1: forwarding `fresh` to the lookup is present and break-verifies correctly. |
| tests | none. The per-test-root fix is complete for the module-level coalescers: sharing only a unique cwd would not isolate the read coalescer because its key is the resolved worktree root, so making the root unique per test is the right fix. The mocks are cleared, `pendingTopLevel` is emptied, and the imported module graph's persistent maps no longer create cross-test joins. Scoped tests passed. Mutation checks: removing `opts` from `coalesceTopLevel` fails exactly the intended `fresh` assertion quickly with the other cases passing; moving the key lookup outside coalescing makes the deterministic same-cwd guard go red, with the expected extra unresolved lookup in one mutation-only test. |
| API / wire compatibility | none. `gitStatus` and `/api/git-status` keep the same parameters and response contract; `fresh=1` semantics are preserved and tightened. |
| consistency with the rest of the codebase | none. The change uses the existing `coalesceByKey` helper and the same best-effort git wrapper behavior as adjacent git-status code. |
| security | none. No new user-controlled shell surface is introduced; cwd validation remains at the route boundary, and git calls still go through argv-based `spawn`. |
| docs/plan accuracy | none. The plan prose is aligned with the implementation and would not steer a wrong follow-up change; the limitation is described as measured and intentionally partial. |

isamu:
## Cross-review round 2 — CLEAN, converged

**Codex: LGTM**, with the axis table answered line by line and `FINDINGS COMPLETE` present. Nothing was pushed after the round was dispatched, so the reviewed head `321d07e6` is the head that would merge.

### It answered the question this PR most needed answering

I asked it twice to attack the decision to ship a *partial* fix, and to say whether a spec asserting a limitation is right or whether it ossifies a bug. Its answer:

> the spec asserting that two different unresolved cwds in one worktree still run two reads **documents a real limitation, not an ossified bug for this PR**. Closing it would require a remembered cwd-to-root mapping; the stated concern is sound because a stale mapping could key the read by an old root while executing it from the first registrant's cwd, potentially serving one worktree's status for another after `git init`, worktree changes, deletion/recreation, or path reuse.

It reached that independently and added two failure modes I had not listed — **path reuse**, and **delete/recreate**.

### It re-ran the measurements rather than taking mine

- removing `opts` from `coalesceTopLevel` fails exactly the `fresh` assertion quickly, with the other cases passing;
- moving the key lookup back outside the coalescing reddens the deterministic same-cwd guard.

Both match what I measured. It also confirmed it restored its own edits and left no PR-file diffs.

### Loop summary

Two rounds. Round 1 found one P3 — shared module-level coalescer state across tests in one spec file — which was a symptom I had already seen in my own break-verification and explained away. Round 2 was clean. One further finding was mine, found and fixed *before* round 1: `fresh` was not being forwarded to the key lookup.

```text
Round 2 on PR #2197 at https://github.com/receptron/mulmoterminal/pull/2197
NEW HEAD: 321d07e6 (round 1 reviewed fe6cba06). Base: 16bf31ae. You are IN the checkout.
Read the whole diff: git diff 16bf31ae...HEAD

Your round-1 P3 was ACCEPTED and FIXED. Review the fix and everything around it.

Every finding you hold back costs a full extra round. Read the WHOLE diff before
posting. Do not stop at the first problem. Do not save anything for a follow-up.

── WHAT CHANGED SINCE YOU LOOKED ────────────────────────────────────────────
Each test in test/server/git/git-status-coalesce.spec.ts now gets its OWN root
(`/repo-<n>`, bumped in beforeEach). I went further than your proposed remedy on
purpose: a unique **cwd** alone would have left the READ coalescer's key — the worktree
root — shared across tests, so the root itself has to differ too. Check that reasoning.

Verified as your finding asked: re-running the mutation that stops forwarding `fresh`
now fails ONE assertion in milliseconds with the other cases passing, where before it
failed one and wedged two at 15s.

Still TIER C.
  * CONCURRENCY AND ASYNC ORDERING — the production change is unchanged since round 1,
    but say whether you still hold that verdict having looked again.
  * COALESCING / INVALIDATION — both maps: can either still leak an entry that never
    settles in PRODUCTION (not just in tests)? What happens if `gitTopLevel` REJECTS?
  * BLAST RADIUS — /api/git-status is polled by every cell.
  * correctness and edge cases
  * error handling and failure paths
  * tests — is the per-test-root fix complete, or does some other shared state survive
    between tests in that file (the mocks, the module graph, the deferred array)? Do the
    specs still go RED on the right mutations?
  * API / wire compatibility
  * consistency with the rest of the codebase
  * security
  * docs/plan accuracy — report prose ONLY if believing it would cause a WRONG CHANGE;
    anything merely behind is severity N and does not block the verdict.

── STILL THE DECISION I WANT ATTACKED ───────────────────────────────────────
This PR ships a PARTIAL fix: two DIFFERENT cwds of one worktree still run two reads,
and there is a spec asserting that boundary. You did not challenge it in round 1.
Please answer it explicitly this round: is a spec that asserts a LIMITATION right, or
does it ossify a bug? And is my reason for declining the full fix sound — that closing
it needs a remembered cwd-to-root mapping, while the read is keyed by root but RUN with
the first registrant's cwd, so a stale mapping would serve one worktree's status for
another?

── ALREADY SETTLED — do not re-raise in the same form ───────────────────────
row 0  | Sandbox pre-flight ALL GREEN. Use plain `rm`, not `rm -f`. You MAY run mutations.
row 0b | Do NOT run `yarn test` — the full suite deadlocks on this machine. Scoped only.
row 0c | A HOOK TIMEOUT in test/server/git/git-status.spec.ts is the machine, not the code
         (its beforeEach builds a temp git repo). Re-run with --hookTimeout=120000 first.
row 1  | (mine, fixed before round 1) `fresh` was not forwarded to the key lookup. Now
         forwarded, break-verified, claims swept. Say if the fix is wrong or incomplete.
row 2  | (yours, round 1, P3) shared module-level coalescer state across tests in one file.
         FIXED by a per-test root — see "what changed" above.

If a resolution above is wrong, say so and mark it 'REOPENING #<row>' with evidence.

Line 1, on its own:
  'CODEX VERDICT: LGTM'  or  'CODEX VERDICT: CHANGES REQUESTED' + bulleted summary.
Then, on its own line:

Post inline comments via gh api repos/receptron/mulmoterminal/pulls/2197/comments
and exactly ONE top-level summary via gh pr comment 2197. Do not apply fixes yourself.
```

</details>

<details><summary>Round 2 — Codex reply (verbatim stdout)</summary>

```text
Posted the round-2 review comment:

https://github.com/receptron/mulmoterminal/pull/2197#issuecomment-5745258898

Verdict: `CODEX VERDICT: LGTM`

I also ran the scoped specs and the requested mutation checks, restored the temporary edits, and confirmed the worktree has no PR-file diffs left.
EXIT=0
```

</details>

PR #2198, opened 2026-09-19
fix(tmux): resolve every session target exactly, and check the phone's id at the door (#2192)

### LINKED ISSUE #2192: bug(tmux): `-t` が session 名を前方一致で解決している — 電話は id の前方部分だけで他セッションの画面を読める
## 測ったこと

`tmux has-session -t NAME` は、**名前が NAME で始まるだけのセッションにヒットします**。
`=` を付けると完全一致になります。tmux 3.6a、独立した socket で実測:

```
sessions: mt-abcdef
  has-session -t 'mt-abcdef'  -> exit 0
  has-session -t 'mt-abc'     -> exit 0   ← 存在しない名前が「ある」になる
  has-session -t 'mt-ab'      -> exit 0
  has-session -t 'mt-zzz'     -> exit 1
  has-session -t '=mt-abc'    -> exit 1   ← `=` で完全一致
  has-session -t '=mt-abcdef' -> exit 0
```

`server/infra/tmux.ts` の `-t tmuxSessionName(id)` は**全部この形**です:

| 行 | 操作 | 前方一致で当たると何が起きるか |
|---|---|---|
| 315 | `has-session` | 存在しないセッションが「ある」になる |
| **325** | **`kill-session`** | **別人のセッションを殺す** |
| 341 | `capture-pane` | 別のセッションの画面を返す |
| 354 / 397 / 473 | `display-message` | 別のセッションの状態を答える |
| 434 | `list-clients` | 別のセッションのクライアントを数える |

## いまは到達不能。ただし「たまたま」です

`tmuxHasSession` の呼び出し側を全部見ました:

- `ws-routes.ts:132 / 214 / 406` — `requested` は `SESSION_ID_RE.test(raw) ? raw : null` で検証済み（`wsConnectionContext`）
- `pty-exit.ts:43` / `pty-spawn.ts:99` / `scheduled-sessions-boot.ts:49` / `app-routes.ts:452` — いずれも id は uuid
- `hostBindings.ts:78` — **電話由来で未検証だった**。PR #2190 で `SESSION_ID_RE.test(id) &&` を前置して閉じた

つまり**全経路で id が固定長 uuid**なので、ある uuid が別の uuid の真の前方部分になることはなく、衝突しません。
**守っているのは呼び出し側の検証であって、ヘルパ自身ではありません。**

これは次に足す1つの呼び出しで壊れます。room id、launcher id、ログから拾った短縮 id、
どれか1つが検証なしで入った瞬間に `kill-session` が他人のセッションを殺します。
`tmuxHasSession` は「この**ちょうどこの**セッションはあるか」という問いに使われているのに、
プリミティブは別の問いに答えている、という形です。

## 提案

`tmux.ts` に完全一致のターゲット組み立てを1つ置き、`-t` を取る全箇所をそれに通す:

```ts
/** `=` は tmux に完全一致を要求する。付けないと名前が前方一致し、
 *  `kill-session -t mt-abc` が `mt-abcdef` を殺す。 */
export const tmuxTarget = (id: string): string => `=${tmuxSessionName(id)}`;
```

**これは挙動の変更で、影響範囲が広い**（`tmuxHasSession` だけで production の呼び出しが8箇所、
再アタッチ判定を含む）ので、独立した PR で、実際にサーバを起動して
再アタッチ・spawn・reap を通す確認まで必要だと考えます。だから #2181 の PR には入れていません。

## 由来

Codex が PR #2190 のレビュー round 2 で指摘。指摘自体は正しく、上の実測で裏を取りました。
到達可能性の分析（＝いまは踏めない）はこちらで追加したものです。

関連しそうなもの: #1533「terminal view reattaches to the wrong backend terminal process」。
再アタッチ判定が `tmuxHasSession` を使っているので**見に行く価値はあります**が、
上のとおり ws-routes 側は id を検証しているので、これが原因だとは確認できていません。

isamu:
## Summary

`tmux -t NAME` resolves by **prefix** when nothing matches exactly. `server/infra/tmux.ts` built every target as the bare `mt-<id>`, so an id that is merely the beginning of a live session's name reached that other session — on `kill-session` as much as on `capture-pane`.

It was reachable today. `getTerminalScreen` took the phone's `sessionId` with no shape check and handed it to `captureTerminalScreen` → `tmuxCaptureStyledPane` → `capture-pane -t mt-<id>`, so a phone could send a leading fragment of an id and read a session it did not name. Its sibling transcript handler checked the shape; this one did not.

Two changes:

- **`server/infra/tmux.ts`** — `tmuxSessionTarget(id)` and `tmuxPaneTarget(id)`, and every `-t` that names a session goes through the builder for its kind. Nothing else about the commands moves.
- **`server/backends/remoteHost/handlers/terminalSession.ts`** — one reader, `sessionIdOf`, checks presence *and* shape, and every handler that takes the phone's id uses it.

## Items to Confirm / Review

1. **The one-liner in the issue would have broken five call sites, silently.** The issue proposed `=${tmuxSessionName(id)}` for every `-t`. Measured on tmux 3.6a: a **pane** target given the bare `=NAME` resolves nothing *even when the session exists exactly* — `capture-pane` exits 1, and `display-message` **exits 0 and prints an empty line**. Every parser here reads an empty line as "tmux cannot say", so pane command, terminal modes, window size and attached-client count would all have gone permanently unknown with no error anywhere. Hence two builders, not one. **Please sanity-check the session/pane split** in the table below.

2. **`tmuxAttachedClientCount` is read by a kill decision.** Its own comment says a failure must not read as "nobody is attached". Under the fix a *prefix* id now yields `null` (was: the other session's count). `parseAttachedClientCount("")` already returns `null`, so the "cannot say" path is the one taken — but this is the call site where a wrong answer is most expensive, and worth a second pair of eyes.

3. **Handlers that did not strictly need the check now have it.** Tracing them: `sendTerminalInput` reaches a `Map` lookup, and `getOpenQuestion` / `answerQuestion` reach a store that already guards with `SESSION_ID_RE`. So only `getTerminalScreen` closed a hole. I routed all of them through one reader anyway, because the per-handler judgement call is what produced this bug. Say if you would rather have kept it narrow.

4. **`launchTerminal` deliberately does not use the reader** — its id is checked further in by `sessionExistsHere` (#2190), and `decideLaunchTerminal` owns the wording of every refusal the phone sees there. Routing it through would answer one refusal twice, in two vocabularies.

5. **Two error strings changed.** `answerQuestion` said "sessionId and toolUseId are required" for either; it now says which one. A malformed (not merely missing) id is now refused with "sessionId is not a session id".

6. **Not changed, both measured rather than assumed**: `new-session -A -s NAME` is already exact (asked to attach to `mt-<uuid>` while only `mt-<uuid>-suffix` was held, it created a new session), and `refresh-client -t <tty>` takes a target-client, not a session.

## What was measured

tmux 3.6a on an isolated socket (`-L`), holding only `mt-<uuid>-suffix`, asked for the bare `mt-<uuid>`:

| subcommand | target kind | bare `-t NAME` | `=NAME` | `=NAME:` |
|---|---|---|---|---|
| `has-session` | session | exit 0 (**wrong session**) | exit 1 | — |
| `kill-session` | session | exit 0 (**ends it**) | exit 1 | — |
| `list-clients` | session | exit 0 | exit 1 | — |
| `capture-pane` | pane | exit 0 (**its screen**) | exit 1 *even when exact* | exit 1 here, exit 0 when exact |
| `display-message` | pane | exit 0, **the other session's name** | exit 0 + **empty**, even when exact | empty here, the name when exact |

### Old code beside new, over a real tmux server

`origin/main`'s `tmux.ts` and this branch's, both pointed at an isolated socket, over the three states an id can be in. Whole answers compared per function:

- **id names a session exactly** — `hasSession`, `paneCommand`, `terminalModes`, `attachedClients`, `capture`, `windowSize`: **identical on both**.
- **nothing is held** — **identical on both**.
- **id is a strict prefix of a live session** — main answers with the other session's data, the branch refuses. `kill-session`: main **ended the wrong session**; the branch ended nothing and both survived.

So behaviour is preserved in the two states the app actually reaches, and only the prefix state changes. (`terminalModes` matched in all three because the probe's shell had no modes set — that row proves nothing either way.)

### Running the app

A real server, the real tmux socket, an isolated `HOME` (so the live config was never touched — confirmed `cwdPresets: 0` before anything was clicked), and the sweep off:

- opened a Shell cell, typed, and the pane rendered it;
- **killed the server and restarted it**: no new tmux session appeared on reload, `#{session_attached}` went 1 → 0 → 1, and the pane carried both the pre-restart line and one typed after — so `tmuxHasSession` still reattaches;
- ran the production `tmuxKillSession` against a leading fragment of a real id: refused, ended nothing; against the exact id: ended it;
- the nine sessions the machine already held were byte-identical before and after.

No uncaught console errors in any browser run.

### What holds it

- A **source-derived guard** over `tmux.ts`: every `-t` is followed by the builder matching that subcommand's kind, or is the allow-listed `refresh-client` line. A new call site spelling its own target fails the spec.
- A **source-derived guard** over `terminalSession.ts`: no handler reads `params.sessionId` raw except the documented `launchTerminal`.
- Mutation-verified, restores byte-identical (`cmp -s`): six mutations on `tmux.ts` (wrong builder per kind, either builder losing its marker, reverting a site to the bare name) and four on the handler — all ten go red.
- The handler specs now use a real UUID; they were passing on `"a"`, which every handler would now reject.

**Not verified locally:** the full `yarn test` run. The machine was at a load average above 20 throughout, where this suite produces a large number of unrelated timeout failures; CI is the ground truth for it. `format`, `lint`, `typecheck` and `build` were all run locally and are clean, and the 21 spec files touching the changed surface pass.

## User Prompt

- （順序）#2192 を次に回す（2件目に前倒し）
- （マージ方針）収束 + CI green で自動マージ

Closes #2192

work in mulmoterminal6

## Summary by CodeRabbit

* **Bug Fixes**
  * Terminal sessions are now addressed exactly, preventing similarly named sessions from being confused.
  * Terminal screen, transcript, input, and question actions now reject malformed or invalid session IDs.
  * Pane-based terminal operations now resolve the correct pane reliably.
  * Missing session and tool identifiers return clearer validation errors.


isamu:
CODEX VERDICT: LGTM

| Axis | Result |
|---|---|
| correctness and edge cases | none. I re-measured the target semantics on private tmux sockets: bare `-t NAME` prefix-matches for the relevant commands, target-session exact spelling is `=NAME`, target-pane exact spelling is `=NAME:`, bare `=NAME` is not a usable pane target, `new-session -A -s NAME` created an exact session rather than attaching to a suffix, and `refresh-client -t <tty>` remains outside the session/pane target split. I also checked the exact-plus-suffix-neighbor state; both old bare lookup and new exact targets resolve the exact session there. |
| SECURITY / AUTH BOUNDARY: what a remote phone client may name, and what the host must resolve for itself. Is `sessionIdOf` the complete door? Are there other remote-reachable paths (routes, ws, other handler files) that hand an unchecked id to a tmux target or a file path? | none. `getTerminalScreen`, transcript, typing, open-question, and answer-question now all use `sessionIdOf`; `launchTerminal` delegates to `sessionExistsHere`, which checks `SESSION_ID_RE` before consulting PTY/tmux state. I checked sibling remote-host handlers and the browser HTTP/ws routes that take session ids; the other reachable paths either validate with `SESSION_ID_RE`, resolve opaque project/repo ids host-side, or do not pass a phone-supplied session id to tmux/filesystem. |
| BLAST RADIUS of a shared helper: tmux.ts targets are on the path of every session operation. Name any caller whose behaviour changes in a way the PR does not claim. | none. The helper change affects `tmuxHasSession`, `tmuxKillSession`, `tmuxCaptureStyledPane`, `tmuxPaneCommand`, `tmuxTerminalModes`, `tmuxRedrawClient` via `list-clients`, `tmuxWindowSize`, and `tmuxAttachedClientCount`. That also covers browser websocket reattach paths through `tmuxHasSession`. I did not find an unclaimed behavior change beyond the intended prefix-refusal. |
| BEHAVIOUR PRESERVATION: the PR claims exact matching changes nothing for ids that name a session exactly, and nothing for ids naming nothing. Judge the evidence given (an old-vs-new differential over a real tmux server across three id states). Is the state space complete? What state is missing? | none. The missing state is “exact session exists while a longer same-prefix neighbor also exists”; I measured that too, and tmux resolves the exact session for bare targets while `=NAME` / `=NAME:` also resolve the exact session. So the evidence plus this extra state supports the preservation claim. |
| ROOT CAUSE: is the defect real and is this where it lives, or is this a symptom fix? | none. The defect is real: bare tmux targets prefix-match, including `capture-pane` and `kill-session`. Fixing both construction (`tmux.ts`) and the remote phone entry point (`sessionIdOf`) addresses the root behavior and the reachable unchecked path. |
| tests: does a test go red if the new behaviour breaks — including the two source-derived guards. Can either guard pass while checking nothing? | none. The builder assertions pin `=NAME` vs `=NAME:`; the tmux source sweep has a non-empty call-site assertion and validates subcommand-to-builder pairing; the handler source sweep has a reader/use-count assertion, so it cannot pass by scanning nothing. I ran `npm test -- --run server/backends/remoteHost/handlers.spec.ts test/server/infra/tmux.spec.ts` and got 2 files / 102 tests passing. |
| API / schema / wire / on-disk compatibility (the phone's error strings changed) | none. The changed errors are refusal strings on failed commands, not schema/on-disk changes. The new “which id is missing” wording and malformed-id rejection look compatible with the command-error display model. |
| consistency with the rest of the codebase | none. The central reader follows existing `SESSION_ID_RE` practice in routes/stores, and the tmux builders match the existing `tmuxSessionName` helper style without changing `new-session` name construction. |
| accessibility and i18n lockstep (if this repo has multi-locale dicts) | none. This diff changes server-side command errors/comments/tests and does not touch UI copy in a locale dictionary; I did not find a multi-locale lockstep requirement for these strings. |

isamu:
## Codex exchange — iteration 1 (tier C)

**Sandbox pre-flight** (ledger row 0) — every claim in this PR was independently re-measurable by the reviewer:

```text
1. `npx vitest run test/server/infra/tmux.spec.ts` EXIT CODE: `0`
2. `npx vue-tsc -b` EXIT CODE: `0`
3. `node -e 'require("net").createServer().listen(0,()=>process.exit(0))'` EXIT CODE: `0`
4. `curl -sS -o /dev/null -w '%{http_code}' https://api.github.com` EXIT CODE: `0`
5. `tmux -V ; tmux -L probe-preflight new-session -d -s zzz sleep 60 ; tmux -L probe-preflight kill-server` EXIT CODE: `0`

can you RUN vitest: yes  
can you RUN vue-tsc: yes  
can you BIND a socket: yes  
can you REACH api.github.com: yes, HTTP `200`  
can you RUN tmux and start a session on a private socket: yes  
are file deletions permitted: yes; requested `rm -f` form was rejected before execution, plain `touch /tmp/x2198 && rm /tmp/x2198` EXIT CODE: `0`
```

<details><summary>Prompt sent (verbatim)</summary>

```text
Review PR #2198 at https://github.com/receptron/mulmoterminal/pull/2198.

── SAFETY, READ FIRST ───────────────────────────────────────────────────────
This machine has a LIVE tmux server on the socket named `mulmoterminal` holding the
user's real work sessions. NEVER run any tmux command against `-L mulmoterminal`,
and never run a bare `tmux` (it would hit the default socket). Use your own
`-L <something-unique>` for every tmux command, and kill only that server.

  gh api repos/receptron/mulmoterminal/pulls/2198/comments
or top-level comments via:
  gh pr comment 2198

── WHAT THE CHANGE CLAIMS ───────────────────────────────────────────────────
`tmux -t NAME` prefix-matches. Two builders now construct every session target in
server/infra/tmux.ts: `tmuxSessionTarget(id)` = `=mt-<id>` for subcommands taking a
target-SESSION (has-session, kill-session, list-clients), and `tmuxPaneTarget(id)` =
`=mt-<id>:` for subcommands taking a target-PANE (capture-pane, display-message).
Separately, every handler in server/backends/remoteHost/handlers/terminalSession.ts
now reads the phone's `sessionId` through one `sessionIdOf` that checks presence AND
shape (SESSION_ID_RE), closing an unchecked path from the phone to `capture-pane`.

YOU CAN AND SHOULD VERIFY THE TMUX CLAIMS YOURSELF — your sandbox runs tmux on a
private socket. The entire design rests on measurements I took; re-take them.
In particular I claim, on tmux 3.6a:
  (a) bare `-t NAME` prefix-matches on has-session, kill-session, capture-pane,
      display-message and list-clients;
  (b) a target-PANE given the bare `=NAME` resolves NOTHING even when the session
      exists exactly — capture-pane exits 1, display-message exits 0 with EMPTY output;
  (c) `=NAME:` is the pane spelling that both resolves an exact session and refuses a
      prefix;
  (d) `new-session -A -s NAME` is already exact and needs no change;
  (e) `refresh-client -t <tty>` is a target-CLIENT and needs no change.
If any of (a)-(e) is wrong, that is a P1 and the most useful thing you can report.

  correctness and edge cases
  SECURITY / AUTH BOUNDARY: what a remote phone client may name, and what the host
    must resolve for itself. Is `sessionIdOf` the complete door? Are there other
    remote-reachable paths (routes, ws, other handler files) that hand an unchecked
    id to a tmux target or a file path?
  BLAST RADIUS of a shared helper: tmux.ts targets are on the path of every session
    operation. Name any caller whose behaviour changes in a way the PR does not claim.
  BEHAVIOUR PRESERVATION: the PR claims exact matching changes nothing for ids that
    name a session exactly, and nothing for ids naming nothing. Judge the evidence
    given (an old-vs-new differential over a real tmux server across three id states).
    Is the state space complete? What state is missing?
  ROOT CAUSE: is the defect real and is this where it lives, or is this a symptom fix?
  tests: does a test go red if the new behaviour breaks — including the two
    source-derived guards. Can either guard pass while checking nothing?
  API / schema / wire / on-disk compatibility (the phone's error strings changed)
  accessibility and i18n lockstep (if this repo has multi-locale dicts)

   N is for a finding whose only remedy is editing PROSE — a comment, a docblock
   sentence, a plan file, the PR description. It does NOT block the verdict. A prose
   finding is P3 or higher only when believing the sentence would lead someone to make
   a WRONG CHANGE. Merely behind is N. Counts are the recurring instance of that.
   occurrence NOW — grep for it.

Your sandbox refuses `rm -f` but allows plain `rm` and `cp`. If you mutate a file,
back it up with `cp` and restore with `cp`, and verify the restore with `cmp -s`.

| 0 | pre-flight | Codex sandbox capability | N/A | vitest 0, vue-tsc 0, socket bind 0, api.github.com 200, tmux on a private socket 0. `rm -f` refused; plain `rm` works. |

_No findings yet — this is round 1._

  Then, on its own line:
  not reach and why.

```

</details>

<details><summary>Codex stdout (verbatim)</summary>

```text
Posted the review comment on PR #2198: https://github.com/receptron/mulmoterminal/pull/2198#issuecomment-5745357526

Verdict: `CODEX VERDICT: LGTM`.

I verified the tmux claims on private sockets, checked the broader remote/session surfaces, read every diff hunk, and ran the targeted changed specs: `2 passed`, `102 tests passed`.
```

</details>

The verdict itself was posted by Codex as its own comment: https://github.com/receptron/mulmoterminal/pull/2198#issuecomment-5745357526

isamu:
## Cross-review triage — round 1 (tier C), and the loop's close

**Tier C**, stated before round 1, for four reasons on the C-list: an exposure boundary (a remote client reading a session it did not name), a shared helper every session operation passes through, an explicit behaviour-preservation claim, and a bug fix whose root cause had to be proved rather than argued.

**Verdict: `CODEX VERDICT: LGTM`, `FINDINGS COMPLETE`, all nine axes answered individually.** No finding at any severity, including N. Nothing has been pushed since the head it reviewed (`f1af5c94`), so this is a clean round on the commit that will merge.

### What made a round-1 LGTM credible here rather than premature

The sandbox pre-flight is the reason, and it is worth stating because it does not usually go this way. Codex could run `tmux` on a private socket — so **every measurement this design rests on was re-taken by the reviewer rather than read from my PR body**. It confirmed independently that bare `-t NAME` prefix-matches, that `=NAME` is the session spelling and `=NAME:` the pane spelling, that a bare `=NAME` is not a usable pane target, that `new-session -A -s` is already exact, and that `refresh-client -t <tty>` sits outside the split. A reviewer that can re-run the premise is not agreeing with my reasoning; it is agreeing with tmux.

### Two things I did not take on trust

1. **Codex named a state my differential had missed** — the exact session live *at the same time* as a longer same-prefix neighbour. That is the state a real collision would live in, so I re-measured it myself on an isolated socket rather than accepting the report: bare, `=NAME` and `=NAME:` all resolve the **exact** session there. The preservation claim holds in the fourth state too, and the differential in the PR body is now the weaker evidence of the two.

2. **A gap neither of us had listed:** `=NAME:` means "session NAME, *current* window", so if the trailing colon pinned window 0 it would silently capture the wrong pane in any session with more than one window. Measured with two windows: both spellings return the active window's pane, and both follow it across a `select-window`. No change needed — but it was the one way the pane spelling could have been quietly wrong, and nothing in the diff would have shown it.

### Loop

One round, one clean verdict, nothing pushed in it. CI is green on `f1af5c94` (18/18).

<details><summary>Ledger</summary>

| 0 | pre-flight | Codex sandbox capability | N/A | vitest 0, vue-tsc 0, socket bind 0, api.github.com 200, tmux on a private socket 0. `rm -f` refused by policy; plain `rm` works. Consequence: every tmux claim here was re-measurable by the reviewer, and was re-measured. |
| 1 | 1 | No finding above N on any of nine axes | — | LGTM + FINDINGS COMPLETE; the four tier-C axes answered individually. |
| 2 | 1 | Codex named an uncovered state: exact session AND longer same-prefix neighbour both live | VERIFIED MYSELF | Re-measured rather than accepted: bare and both exact spellings resolve the exact session. |
| 3 | 1 | (mine, unprompted) does `=NAME:` agree with the bare name in a multi-window session | VERIFIED, no change | Both return the active window's pane, and both follow `select-window`. |

</details>

PR #2199, opened 2026-09-19
feat(settings): report the sweep cadence this server ARMED, not just the saved one (#2184)

isamu:
## Summary

`sessionReapIntervalHours` is read once, at boot, and the timer is deliberately **not** re-armed when the config is POSTed — a stream of edits would reset the countdown forever (#2167). So from a save until the next restart, the saved number describes a **future** server and the running one describes **this** server.

The browser only ever saw the saved one. #2189 handled that honestly by removing every present-tense claim: the row names the **event** ("the next sweep ends it"), the hints say what is **saved**, and one standing line says the cadence is read at startup. All true whatever was armed — and none of it answers the question the row raises, which is *when*.

This adds the missing half. The server reports what it armed, and the standing line says it.

- `reap-schedule.ts` keeps the armed cadence **paired with the timer handle in one value**, so both are set and cleared in the same assignment. `armedReapIntervalHours()` reads it.
- `/api/tmux/sessions` carries it beside the list — no new route, since the section already calls this one.
- The line states what this process is doing, and says the saved cadence applies from the next start **only when the two differ**.

**Partially addresses #2184** — deliberately not `Closes`, because the issue also sketches alternatives I did not take; see item 3.

## Items to Confirm / Review

1. **The unknown state is a first-class case, not a default.** The value is `null` — and the line falls back to #2189's general sentence — in every state where the cadence is not *confirmed*: an older server that does not send the field, a read that failed, and the window between asking and being answered. That last one was found by the reviewer in round 1 and reproduced before being accepted: a reload that goes on to fail held the stale claim for the whole fetch timeout. A value the server could not have meant (`NaN`, an infinity, a negative) goes to the same place, because the loose `typeof x === "number"` I first wrote would have rendered an infinity into the sentence and dropped `NaN` silently to "does not repeat". Found in my own pass before round 1. It is **never** defaulted to the saved number: that is precisely the substitution this area exists to stop, and it would read as fact while being a guess. Please confirm you want the fallback rather than hiding the line.

2. **Pairing the cadence with the timer handle is the main design choice.** Every bug in this area has been a pair that drifted, so there is now no state in which a cadence is reported and nothing is ticking, or the reverse. The alternative — a second module-level variable — is what I deliberately avoided.

3. **The wording changed on a line #2189 had just settled.** #2189 made that line unconditional precisely because no clock could be named; its own spec comment says "the armed cadence is not sent to it (#2184)". I kept the invariant (the line is never missing) and made it precise. If you would rather the armed value went somewhere else on the screen, this is the thing to say so about.

4. **What I did NOT resurrect.** An earlier branch (`feat/2184-armed-reap-interval`) implemented this before #2189, #2193 and #2194 landed. It bundled the timer-cancellation work that has since shipped as #2194, and its UI half was written against wording #2189 replaced. Re-deriving on current main was cheaper and safer than merging it; that branch is now superseded and can be deleted.

## User Prompt

- #2184やって

## Approach and key decisions

**The route value is asked for per request, not captured at mount.** A value read once would be the boot value forever, which is the exact mistake being corrected. There is a spec pinning that it is called on every request.

**The getter is pinned in both directions**, including the one that used to lie: a schedule that arms nothing must stop *reporting* as well as stop sweeping.

**Break-verified.** Substituting the saved value for the armed one in the component reddens the three pending cases **and** #2189's three "always says" cases — because that substitution also destroys the unknown-state fallback. The pre-existing specs in the section all pass untouched, which makes them the regression test for the older-server path.

i18n is in lockstep across all five locales; `vue-tsc` enforces that.

Gates: `eslint` clean on every touched file, `yarn typecheck` exits 0, and the affected specs (`test/server/session/`, `test/server/infra/tmux-routes.spec.ts`, the section spec, `test/server/config/`) are green.

work in mulmoterminal4

## Summary by CodeRabbit

- **New Features**
  - Settings now show the server’s currently active session-cleanup schedule.
  - Indicates whether cleanup repeats, runs only at startup, or a saved change takes effect after the next restart.
  - The schedule status is reported across supported languages.

- **Documentation**
  - Updated configuration and settings guidance to clarify active versus saved cleanup intervals.

- **Tests**
  - Added coverage for schedule reporting, pending changes, unavailable values, and older server responses.


isamu:
## Cross-review tier: **C**

Three C-list items, none of them about diff size:

- **A wire-format change.** `/api/tmux/sessions` grows a field. The response shape is a contract between two halves of this repo, and the browser half decides what to *say to the user* from it.
- **Module-level lifecycle state, paired with a timer.** The armed cadence is stored beside the `setInterval` handle. Every bug this area has produced was a pair that drifted, so the pairing is the design and it deserves to be attacked as one.
- **A shared read path.** Every open of the Settings section calls this route.

Plus i18n lockstep across five locales, which `vue-tsc` enforces but which is worth an explicit axis.

What does **not** force C, stated so the tier is auditable: nothing here is destructive, there is no auth or secret surface, and no behaviour-preservation claim is being made — this adds a field and a sentence rather than promising the old ones are unchanged.

**What I most want attacked:** the **unknown state**. When a reply does not carry the field the UI falls back to the general sentence and never substitutes the saved number. I believe defaulting to the saved value would re-create exactly the lie #2189 removed. If that reasoning is wrong, or if the fallback has a state I have not considered, this round is where it should break.

src/composables/useSurvivingSessions.ts
```diff
@@ -44,15 +56,17 @@ export function useSurvivingSessions() {
       // A malformed row is dropped rather than asserted: the alternative is a stop button whose
       // key is undefined, posting to `/api/session/undefined/terminate`.
       sessions.value = isUnknownArray(body.sessions) ? body.sessions.filter(isSurvivingSession) : [];
+      armedReapIntervalHours.value = isArmedCadence(body.armedReapIntervalHours) ? body.armedReapIntervalHours : null;
```
isamu, line 65: After one successful load, a later `reload()` leaves the previous `armedReapIntervalHours` in place while the new request is in flight. The sessions list is hidden by the loading branch, but the sweep note remains visible, so a user can still see “This server repeats…” as fact during the exact unknown window before the replacement reply arrives (or for up to the timeout before a failed reload clears it in `catch`). The smallest fix is to set `armedReapIntervalHours.value = null` when starting a reload, next to `loading.value = true`, so the always-visible note falls back to the generic sentence until the current response proves the armed cadence again.

isamu:
- P3: after a successful `/api/tmux/sessions` load, a later reload keeps the previous `armedReapIntervalHours` visible while the replacement request is still in flight, so the always-visible sweep note can state an old process cadence as fact during an unknown state.

Finding details:
- P3 / every site: this pattern exists in `src/composables/useSurvivingSessions.ts` only; grep for `armedReapIntervalHours` found the route producer, this composable state owner, the component reader, tests, locales, and docs, with no second client-side state owner. Smallest change: set `armedReapIntervalHours.value = null` at the start of `reload()`, next to `loading.value = true`, before `fetchWithTimeout`. What would change my mind: if the sweep note were hidden/marked non-authoritative during `loading`, or if a reload after a prior success could not occur while the section remains mounted.

| Axis | Result |
| --- | --- |
| WIRE / API COMPATIBILITY | No blocking wire finding. Older browser against new server ignores the extra field. New browser against older server gets no field, the guard maps it to `null`, and the generic startup sentence renders. Absence is distinguishable from the valid off value `0`, but not from malformed `false`; that seems right because the server could only mean a finite nonnegative number, and `false` should be unknown rather than “off.” |
| MODULE-LEVEL LIFECYCLE STATE | none. `armed` pairs the handle and cadence, `stopArmedTimer()` clears both together, `startReapSchedule()` stops before fallible sweep work, and a disabled replacement reports off and has no timer. I did not find a path that reports cadence without a timer or a timer without a cadence in the Node server path. |
| A SHARED READ PATH | P3 above. Because every open/reload of the Settings section uses this shared route and the note remains visible through loading, stale armed state can leak into the shared read path while a fresh read is pending. |
| correctness and edge cases | P3 above. The unknown-state choice is right for absent/malformed replies and after failed reload completion, but it is not entered during an in-flight reload after a previous success. |
| error handling and failure paths | P3 above. A failed reload eventually clears to `null` in `catch`, but until the request fails or times out the old armed cadence is still rendered. |
| tests: do the new specs go RED on the right mutations? | Targeted command passes clean after restoring mutations: `npx vitest run test/server/session/reap-schedule.spec.ts test/server/infra/tmux-routes.spec.ts test/src/components/settings/survivingSessionsSection.spec.ts` => 61/61. Saved-for-armed mutation reddened the three pending cases and the three pre-existing “always says” cases, plus the six unusable-wire fallback cases because it bypasses null. Loosening the guard to `typeof x === "number"` reddened exactly NaN / infinity / negative. |
| i18n lockstep across en, ja, ko, zh-CN, zh-TW | none. All five locales have `sweepRunning`, `sweepRunningOff`, and `sweepPending`; `{hours}` is present in all `sweepRunning` messages, and the off/pending meanings match English. |
| accessibility | P3 above. This line is read as operational state, so showing the previous cadence during loading can mislead a user deciding whether a session is about to die. Otherwise the wording is explicit and not color-only. |
| security | none. The route adds a numeric process cadence only; row validation still drops malformed session rows, and the stricter cadence guard prevents invalid values from becoming UI claims. |
| consistency with the rest of the codebase | none. The route continues the injected-deps pattern, the composable follows existing row validation style, and the Settings component keeps existing copy/stepper structure. |
| docs/plan accuracy | N: `src/components/settings/SurvivingSessionsSection.vue` still has a stale historical comment saying “#2184 would let this say what IS armed”; grep found that one runtime-adjacent stale sentence. Smallest change is to update that comment to describe the current `sweepNote` behavior. What would change my mind: if that paragraph is intentionally preserving pre-#2184 reasoning rather than documenting the current component. This does not block the verdict. |

isamu:

**Codex: CHANGES REQUESTED, one P3. Accepted and fixed in `17e58e4c`.**

### The finding

After a successful load, a later `reload()` left the previous armed cadence in place while the new request was in flight. The sessions list is hidden by the loading branch — but the sweep note is not, so the screen kept stating last-known process state as fact. The sharpest part of the finding is the failure path: a reload that goes on to **fail** held that claim for the whole fetch timeout before `catch` cleared it.

This is the state I asked the round to hunt for in the brief, and it found it.

A deferred second fetch, held unanswered: the composable still reported the old cadence. That is now a permanent spec rather than a scratch file.

### The fix

The value is dropped when the read **starts**, not merely replaced when it finishes — so it is `null` in every state where the cadence is not *confirmed*: an older server that does not send the field, a read that failed, and the window between asking and being answered.

The sessions beside it may stay on screen while they refresh; a sentence asserting what this process is doing may not.

### The composable got its own spec file

Deliberate: the thing being pinned is a **window**, which a mounted component cannot express — the section does not expose `reload`, and the note it renders is one frame of the answer rather than the answer itself.

### Break-verification

Removing the start-of-read clear reddens exactly the new window spec. Reproduce with `npx vitest run test/src/composables/useSurvivingSessions.spec.ts`.

### Also fixed before round 1, by my own pass

The new wire field was validated with `typeof x === "number"`, which admits `NaN`, `Infinity` and negatives, while the row validator one line above drops malformed data on principle. An infinity would have rendered into the sentence; `NaN` would have fallen silently to "does not repeat". Anything unusable now goes to the unknown state. Break-verified in both directions.

### One qualification on this round's sandbox, recorded rather than re-derived

Codex's `vitest` exits 1 in **its** sandbox on `session-settings.spec.ts`, `spawn-antigravity.spec.ts` and `spawn-grok-mcp-sync.spec.ts` — `EPERM ... mkdir '/test/dir/.agents'`, because the sandbox refuses writes outside the workspace and those specs write under the real `HOME`. All three are byte-identical to `origin/main`, none is in this diff, and my own run of the same command is exit 0. Rounds are handed a scoped command covering the diff's own specs so the verdict means something.

```text
Review PR #2199 at https://github.com/receptron/mulmoterminal/pull/2199
Head: 8059f0ca. Base: cfc7dbfb. You are IN the checkout at that head.
Read the whole diff: git diff cfc7dbfb...HEAD

time in the round AFTER it is pushed. Read the WHOLE diff before you post anything.
Do not stop at the first problem. Do not save anything for a follow-up review.

── HOW TO RUN THE TESTS IN YOUR SANDBOX ─────────────────────────────────────
Use exactly this, and nothing wider:
  npx vitest run test/server/session/reap-schedule.spec.ts test/server/infra/tmux-routes.spec.ts test/src/components/settings/survivingSessionsSection.spec.ts
It passes 55/55 here. Do NOT run `yarn test` (the full suite deadlocks on this machine)
and do NOT run all of `test/server/session/` — see row 0 below for why that is red in
YOUR sandbox and not in mine. Use plain `rm`, not `rm -f`.

This PR is TIER C. The first three are why:

  * WIRE / API COMPATIBILITY. `/api/tmux/sessions` grows a field. An older BROWSER
    against a new server, and a new browser against an OLDER server, are both real —
    say what each does. Is the field's absence distinguishable from its being false?
  * MODULE-LEVEL LIFECYCLE STATE. The armed cadence is stored PAIRED with the
    `setInterval` handle in one value, set and cleared in the same assignment. Attack
    that: is there any path that sets one without the other, or reports a cadence when
    nothing is ticking, or the reverse? What about a throw partway through arming?
  * A SHARED READ PATH — every open of the Settings section calls this route.
  * correctness and edge cases
  * error handling and failure paths
  * tests: do the new specs go RED on the right mutations? I measured two: substituting
    the SAVED value for the ARMED one in the component reddens the three pending cases
    AND the three pre-existing "always says" cases; loosening the wire guard back to
    `typeof x === "number"` reddens exactly NaN / infinity / negative. Verify both.
  * i18n lockstep across en, ja, ko, zh-CN, zh-TW — `vue-tsc` enforces key parity, but
    check the MESSAGES: does each locale's text say what the English says, and are the
    interpolations present in all five?
  * accessibility — the line is read by a user deciding whether a session is about to die.
  * security
  * consistency with the rest of the codebase
  * docs/plan accuracy. Report a sentence ONLY if believing it would cause a WRONG
    CHANGE; file anything merely behind as severity N, which does not block the verdict.

── THE DECISION I MOST WANT ATTACKED ────────────────────────────────────────
The UNKNOWN state. When the reply does not carry the field, or carries one the server
could not have meant, the value is null and the line falls back to the general sentence
("The cadence is read when the server starts…"). It is NEVER defaulted to the saved
number. My reasoning: defaulting would re-create exactly the lie #2189 removed — it
reads as fact while being a guess. Is that right? Is there a state of the fallback I
have not considered — for instance while the fetch is still in flight, or after a failed
reload following a successful one?

1. SEVERITY — P1 / P2 / P3 / N (N is prose-only and does not block the verdict).
2. EVERY SITE of the same pattern in the repo — grep, do not guess.

── ALREADY SETTLED — do not re-raise in the same form ───────────────────────
row 0  | Sandbox pre-flight: typecheck 0, socket 0, curl 200, gh 0, deletions permitted.
         vitest exited 1 in YOUR sandbox on session-settings.spec.ts (14), 
         spawn-antigravity.spec.ts (1), spawn-grok-mcp-sync.spec.ts (1), all with
         `EPERM ... mkdir '/test/dir/.agents'` — your sandbox refuses writes outside the
         workspace and those specs write under the real HOME. All three are
         BYTE-IDENTICAL to origin/main and none is in this diff; my own run of the same
         command is exit 0. Do not report those as findings. If you think one IS mine,
         say so with evidence.
row 0b | Do NOT run `yarn test` — the full suite deadlocks on this machine.
row 0c | A hook timeout in a spec that builds a temp git repo is the load, not the code.
row 1  | (MINE, fixed BEFORE this round, in 8059f0ca) the new wire field was validated with
         `typeof x === "number"`, which admits NaN / Infinity / negative — an infinity
         rendered into the sentence and NaN fell silently to "does not repeat". Now a named
         guard sends anything unusable to the unknown state. Break-verified both ways.
         Do NOT re-raise as new — but DO say if the guard is wrong or incomplete.

If a resolution above is wrong, say so and mark it 'REOPENING #<row>' with evidence.

Line 1, on its own:
  'CODEX VERDICT: LGTM'  or  'CODEX VERDICT: CHANGES REQUESTED' + bulleted summary.
Then, on its own line:

Post inline comments via gh api repos/receptron/mulmoterminal/pulls/2199/comments
and exactly ONE top-level summary via gh pr comment 2199. Do not apply fixes yourself.
```

</details>

<details><summary>Round 1 — Codex reply (verbatim stdout)</summary>

```text
Posted the review on PR #2199.

Filed one inline P3 on `useSurvivingSessions.reload()`: after a successful load, a later reload keeps the previous armed cadence visible while the new fetch is still in flight, so the always-visible sweep note can temporarily state stale process state as fact.

Also posted the required top-level axis table and completion line. Verified with the exact command: 3 files, 61/61 passing after restoring mutation checks.
EXIT=0
```

</details>

isamu:
CODEX VERDICT: LGTM

| Axis | Result |
| --- | --- |
| WIRE / API COMPATIBILITY — old browser/new server and new browser/old server | none — old browsers ignore the added field; new browsers treat an absent or unusable field as unknown and fall back to the general sentence. |
| MODULE-LEVEL LIFECYCLE STATE — the armed cadence PAIRED with the interval handle | none — the single `armed` value is set only after a timer is created, cleared before fallible replacement work, and reports `0` when no interval handle is stored. |
| A SHARED READ PATH — every open of the Settings section calls this route | none — the existing `/api/tmux/sessions` path carries the cadence and asks the getter per request. |
| correctness and edge cases | none — dropping the cadence at reload start creates only the intended temporary fallback; I did not find a path that reports a cadence the server has not confirmed. |
| error handling and failure paths | none — failed reads clear sessions and the armed cadence, preserving the unknown-state fallback. |
| tests — do the specs go RED on the right mutations? | none — requested slice passed: `npx vitest run test/src/composables/useSurvivingSessions.spec.ts test/src/components/settings/survivingSessionsSection.spec.ts test/server/session/reap-schedule.spec.ts test/server/infra/tmux-routes.spec.ts` = 4 files / 64 tests. Mutation checks: loosening the wire guard reddened exactly NaN / infinity / negative; removing the start-of-read clear reddened exactly the new window spec; substituting the saved value for the confirmed armed value was caught by the pending and older-server fallback specs, with the exact component-level mutation also catching the malformed-wire fallback cases. |
| i18n lockstep across en, ja, ko, zh-CN, zh-TW — check the MESSAGES, not just key parity | none — the new running/off/pending messages are present and semantically aligned across all five locale files. |
| accessibility | none — the changed note remains plain text; no interactive affordance or ARIA contract changed. |
| security | none — the new field exposes only already-local cadence state on an existing local read route; malformed wire values are not rendered as facts. |
| consistency with the rest of the codebase | none — the route dependency injection, composable validation, and Settings rendering follow nearby patterns. |
| docs/plan accuracy — report a sentence ONLY if believing it would cause a WRONG CHANGE; file anything merely behind as N, which does not block the verdict. | N only — one nearby code comment still reads as pre-#2184 background (“#2184 would let this say what IS armed”), but it is behind rather than a wrong-change risk. |

isamu:
- REOPENING #4: `docs/guide/en/config.md:1991` still says the Settings section names “only what is saved.” After #2184, the section also states what the running server actually armed, so that remaining all-keys sentence is now false. This is prose-only; I did not find a behavior, assertion, key, or app-rendered-string regression.
- Commit `6c44a137` itself is genuinely prose/comment/docs-only versus `17e58e4c`: the changed files are `docs/guide/en/config.md`, `docs/guide/ja/config.md`, `src/components/settings/SurvivingSessionsSection.vue`, and `test/src/components/settings/survivingSessionsSection.spec.ts`; no runtime code, test assertion, config key, or app UI string changed. The docs rendered prose changed as intended.
- Exact requested test command passed: `npx vitest run test/src/composables/useSurvivingSessions.spec.ts test/src/components/settings/survivingSessionsSection.spec.ts test/server/session/reap-schedule.spec.ts test/server/infra/tmux-routes.spec.ts` → 4 files, 64 tests.

| Axis | Result | Notes |
| --- | --- | --- |
| Round-3 delta from `17e58e4c` | Checked | Prose/comment/docs only; no behavior changes. |
| Four corrected sentences | Checked | The component comment, component spec comment, and both guide overview sentences now match the code: the row names the event, the stepper edits/describes the saved value, and the note reports the armed cadence plus pending saved changes. |
| Further claim exposed by the fix | Finding | `docs/guide/en/config.md:1991` still says the section shows “only what is saved,” which is no longer true. |
| `#2184` / `armed` sweep across `server/`, `src/`, `test/`, `docs/` | Checked | No remaining relevant claim says #2184 is future work or that the armed cadence is not sent to the browser. The remaining problem above was found by the broader adjacent wording sweep, not by an exact `#2184` hit. |
| Server schedule state | Unchanged since round 2, checked | `armed` pairs timer handle and interval in one value; `startReapSchedule(0)` clears both reporting and ticking. |
| Route payload | Unchanged since round 2, checked | `/api/tmux/sessions` asks `armedReapIntervalHours()` per request and returns it beside `sessions`. |
| App route wiring | Unchanged since round 2, checked | `mountSessionFacingRoutes` injects the live getter from `reap-schedule.ts`. |
| Composable validation | Unchanged since round 2, checked | Missing, non-number, `NaN`, infinity, and negative values become unknown (`null`), not screen claims. |
| In-flight reload behavior | Unchanged since round 2, checked | The composable drops the armed cadence at read start, so stale cadence is not reported during an unanswered or failing reload. |
| Component rendering | Unchanged since round 2, checked | Confirmed armed values render as running cadence; differing saved values add the pending-next-start sentence; unknown falls back to the generic startup-read sentence. |
| Saved-value UI semantics | Unchanged since round 2, checked | Row and stepper still describe saved config/event semantics rather than pretending to know an exact next sweep time. |

isamu:

- REOPENING #5: `server/skills/mulmoterminal-config/SKILL.md:315-322` still carries the old premise: “the browser only sees the saved one.” That is no longer true on this head, because `/api/tmux/sessions` now returns `armedReapIntervalHours` and `SurvivingSessionsSection` renders the armed cadence when present. The next bullet in the same skill correctly says #2184 added that reporting, so the skill now contradicts itself inside the `sessionReapIntervalHours` guidance. This is the same stale-prose class, found by meaning rather than by issue-number grep, and it lives under `server/`, which was in the requested sweep.

| Axis | Status | Notes |
|---|---|---|
| Head/base | checked | HEAD is `a139ca6539d524bf0692f50d4415b2b36106152a`; base `cfc7dbfb3868fea35eb2ecd15852a30e45952d2a` is present. |
| Diff since round 3 | checked | `git diff 6c44a137...HEAD` touches only `docs/guide/en/config.md`, `docs/guide/ja/config.md`, `server/config/config-routes.ts`, and `src/components/settings/SurvivingSessionsSection.vue`; the delta is prose/comments only. No behavior, assertion, config key, or app-rendered string changed in this round. |
| Full PR diff | checked | I read the hunks across server, src, tests, docs, skill prose, and the plan file. The behavioral shape from earlier rounds is unchanged except for the stale skill prose finding above. |
| Shape change | partially holds | The “describe beside you, point at the section rather than speak for it” rule is durable in the guides and source comments changed this round. It did not reach the bundled config skill prose, which still speaks for the browser’s visibility. |
| Rendered section names a next-sweep time | ok | Rendered strings name events/cadences only: “due to be ended,” “the next sweep ends it,” saved cadence, armed cadence, pending next start. I did not find a timestamp/countdown/time for when the next sweep will run. |
| Browser sees only saved cadence claim | failed | Stale claim remains in `server/skills/mulmoterminal-config/SKILL.md:315-318`. |
| “Section cannot say what is running” claim | checked | I did not find a remaining rendered/comment/doc claim in `server/`, `src/`, `test/`, or `docs/` saying the section cannot say what is running, apart from the contradictory skill paragraph above implying it by saying the browser only sees saved. |
| “#2184 is future work” claim | checked | No remaining `#2184 would let...` / future-work wording in `server/`, `src/`, `test/`, or `docs/`. Older `plans/` files still contain historical planning text, but that directory was outside the requested sweep. |
| Round-2/3 behavior axes | unchanged-since-round-3 | Because the round-4 diff is prose/comment only, the prior armed-cadence behavior, in-flight reload clearing, malformed cadence fallback, per-request API read, and timer pairing axes are unchanged. I re-read the relevant hunks while checking this. |
| Tests | passed | Ran exactly: `npx vitest run test/src/composables/useSurvivingSessions.spec.ts test/src/components/settings/survivingSessionsSection.spec.ts test/server/session/reap-schedule.spec.ts test/server/infra/tmux-routes.spec.ts` → 4 files passed, 64 tests passed. |
| Whitespace | checked | `git diff --check cfc7dbfb...HEAD` produced no output. |

isamu:
CODEX VERDICT: LGTM

| Axis | Result |
| --- | --- |
| HEAD / base / round delta | checked — HEAD is `57bb40c7b7718bb43e592129f817174f87009bb2`; base `cfc7dbfb3868fea35eb2ecd15852a30e45952d2a` is present. The delta from `a139ca65` to `57bb40c7` is one file, `server/skills/mulmoterminal-config/SKILL.md`, and it is prose-only. |
| Round-5 prose-only confirmation | checked — no runtime code, tests, assertions, app UI strings, config keys, route shapes, or generated docs changed since round 4; only the bundled config skill wording changed. |
| Stale phrase / meaning sweep | clean — I searched by exact phrases and nearby meanings for the stale premise: “browser only sees/only ever sees the saved one,” “only what is saved,” “armed cadence is not sent,” “#2184 would…,” future-work wording, “cannot say what is running,” and clock/cadence wording. No current docs, source comments, tests, skill prose, or rendered strings still say the browser sees only the saved cadence, the section cannot say what is running, or #2184 is future work. The remaining `plans/` hits are historical pre-change context, not current guidance. |
| WIRE / API COMPATIBILITY — old browser/new server and new browser/old server | unchanged-since-round-4, checked — old browsers ignore the added field; new browsers treat absent or unusable `armedReapIntervalHours` as unknown and render the generic startup-read sentence. Absence is not confused with valid `0`. |
| MODULE-LEVEL LIFECYCLE STATE — armed cadence paired with interval handle | unchanged-since-round-4, checked — `armed` stores timer + interval together, `stopArmedTimer()` clears both together, disabled schedules report `0`, and I did not find a path reporting a cadence without a timer or a timer without a cadence. |
| A SHARED READ PATH — every open of the Settings section calls this route | unchanged-since-round-4, checked — `/api/tmux/sessions` still asks `armedReapIntervalHours()` per request and returns it beside `sessions`; no new route or mount-time capture. |
| correctness and edge cases | unchanged-since-round-4, checked — confirmed armed values render as running cadence; differing saved values add the next-start pending sentence; unknown, malformed, absent, in-flight, and failed-read states fall back rather than guessing. |
| error handling and failure paths | unchanged-since-round-4, checked — the composable clears the armed cadence at read start and on catch, so a failed reload does not keep asserting the last confirmed process cadence. |
| tests — requested slice and mutation coverage | checked — exact command passed: `npx vitest run test/src/composables/useSurvivingSessions.spec.ts test/src/components/settings/survivingSessionsSection.spec.ts test/server/session/reap-schedule.spec.ts test/server/infra/tmux-routes.spec.ts` -> 4 files, 64 tests. Earlier mutation-axis results still apply because the round-5 delta is prose-only: saved-for-armed, loose numeric guard, and missing start-of-read clear are covered by the existing specs. |
| i18n lockstep across en, ja, ko, zh-CN, zh-TW | unchanged-since-round-4, checked — all five locales carry running/off/pending/generic note messages with the `{hours}` interpolation where required, and the meanings match the English. |
| accessibility | unchanged-since-round-4, checked — the operational state is plain text and no interactive or ARIA contract changed; the stale-state window remains closed by clearing at read start. |
| security | unchanged-since-round-4, checked — the new field exposes only the local server’s sweep cadence on an existing local read route, and unusable wire values are not rendered as facts. |
| consistency with the rest of the codebase | unchanged-since-round-4, checked — route dependency injection, composable validation, Settings rendering, and skill/doc wording follow the surrounding patterns. |
| docs / plan / skill accuracy | clean for current guidance — guides, source comments, tests, and `server/skills/mulmoterminal-config/SKILL.md` now state the current behavior: event wording for rows, saved wording for the stepper, armed cadence for the note, fallback when unconfirmed. Historical `plans/` prose remains historical and does not read as current instructions. |

CODEX CHECKPOINT: round 5

1. Yes. The diff still accomplishes the PR’s purpose: the server reports the cadence it actually armed, the existing sessions route carries it, the composable treats unknown/unusable states as unknown, and the Settings section distinguishes running cadence from saved-next-start cadence.

2. Bluntly: this loop did drift into prose. Rounds 2, 3, and 4 were all wording/comments/docs. It was still worth it through round 4 because the stale claims lived in active guidance and could cause a future wrong change, especially the bundled skill. Past this point, continuing would be reviewing the review unless a new behavioral fact appears.

3. I do not see any remaining diff that exists only because a reviewer asked for it and lacks its own merit. The prose fixes stand on their own because active comments/docs/skills are part of how this codebase preserves intent.

4. No split, redirect, or close. This is still one coherent PR: expose armed cadence, render it honestly, and align the docs/tests around that distinction. It should converge.

5. The single largest remaining risk is conceptual drift later: someone may again collapse “saved cadence,” “armed cadence,” and “next sweep time” into one idea. The current code and prose now guard that distinction; I do not see a concrete remaining defect in this diff.

isamu:
## Cross-review rounds 2–5 — converged

Four rounds in one comment rather than four near-identical ones; each round's prompt and verbatim reply is under its own fold below.

**Round 5: `CODEX VERDICT: LGTM`, `FINDINGS COMPLETE`, checkpoint answered.** Nothing was pushed after it was dispatched, so `57bb40c7` is the head that would merge.

### What each round found

| Round | Verdict | Finding |
|---|---|---|
| 2 | LGTM + one **N** | a comment still reading as pre-#2184 background |
| 3 | CHANGES REQUESTED | **REOPENED** its own row 4 — a guide line still said the section names "only what is saved" |
| 4 | CHANGES REQUESTED | the same false premise in `SKILL.md`, in the paragraph above the one I had updated |
| 5 | **LGTM** | clean; checkpoint says converge |

### Rounds 2, 3 and 4 were all prose, and the fault was mine each time

Worth recording plainly, because it is the useful part of this loop:

- **Round 3**: my round-2 sweep grepped `#2184`. The offending guide line does not cite it, so the sweep failed at the boundary the tool worked within. Enumerating sentences by *meaning* instead found two more Codex had not named — including a **source comment** in `config-routes.ts` that no docs sweep could reach.
- **Round 4**: I had *deleted* the sentence "the browser only ever sees the saved one" from the component in round 3 and never grepped for copies of it. A phrase you remove in one place is exactly the phrase to search for. Separately, I had edited that same `SKILL.md` in the first commit and fixed the bullet citing `#2184` while the bullet **above** it carried the same premise without citing the issue.

### The shape changed twice, rather than a third case being patched

Per the third-finding gate:

1. Those sentences all rotted for one reason — each asserted what **another file's UI** displays, so each went false whenever the UI changed. They now describe the behaviour of the thing they sit beside and *point at* the section instead of speaking for it.
2. The skill doc was contradicting itself one bullet apart because I had **layered** an "#2184 added…" update on top of the stale paragraph. A document that narrates its own history does this every time one of its claims changes. The two are now a single statement of what the section *does*.

And fixing those exposed the next claim, as the method predicts: the component's own block still rested on "the browser only ever sees the saved one" — the premise this PR falsified.

### The checkpoint's own answer on whether that was worth it

> Bluntly: this loop did drift into prose. Rounds 2, 3, and 4 were all wording/comments/docs. It was still worth it through round 4 because the stale claims lived in active guidance and could cause a future wrong change, especially the bundled skill. Past this point, continuing would be reviewing the review unless a new behavioral fact appears.

It also named the largest remaining risk as conceptual rather than concrete: that someone later re-collapses "saved cadence", "armed cadence" and "next sweep time" into one idea.

### Verification carried through every round

Codex ran the three named mutations itself and reproduced my measurements: substituting the saved value for the armed one reddens the pending and older-server fallback specs; loosening the wire guard reddens exactly NaN / infinity / negative; removing the start-of-read clear reddens exactly the window spec.

```text
Round 2 on PR #2199 at https://github.com/receptron/mulmoterminal/pull/2199
NEW HEAD: 17e58e4c (round 1 reviewed 8059f0ca). Base: cfc7dbfb. You are IN the checkout.
Read the whole diff: git diff cfc7dbfb...HEAD

Your round-1 P3 was ACCEPTED, REPRODUCED and FIXED. Review the fix and everything around it.

Every finding you hold back costs a full extra round. Read the WHOLE diff before
posting. Do not stop at the first problem. Do not save anything for a follow-up.

── HOW TO RUN THE TESTS IN YOUR SANDBOX ─────────────────────────────────────
Use exactly this and nothing wider:
  npx vitest run test/src/composables/useSurvivingSessions.spec.ts test/src/components/settings/survivingSessionsSection.spec.ts test/server/session/reap-schedule.spec.ts test/server/infra/tmux-routes.spec.ts
Do NOT run `yarn test` and do NOT run all of `test/server/session/` — row 0 says why.
Use plain `rm`, not `rm -f`.

── WHAT CHANGED SINCE YOU LOOKED ────────────────────────────────────────────
`armedReapIntervalHours` is now dropped when the read STARTS, not merely replaced when it
finishes, so it is null in every state where the cadence is not confirmed: an older server,
a failed read, and the window between asking and being answered. The composable also gained
its own spec file, because the thing being pinned is a WINDOW the mounted component cannot
express — the section does not expose `reload`.

Attack the fix, not only the original defect:
  * Does dropping it at the start introduce a flicker that matters? The note falls back to
    the general sentence during EVERY reload, including the one the stop button triggers.
    Is that the right trade, or is there a state where it is worse than the staleness it replaces?
  * Is there still a path that reports a cadence the server has not confirmed?
  * Is the new spec file asserting the window it claims to, or does it pass for another reason?

Still TIER C.
  * WIRE / API COMPATIBILITY — old browser/new server and new browser/old server.
  * MODULE-LEVEL LIFECYCLE STATE — the armed cadence PAIRED with the interval handle.
    Any path that sets one without the other, or a throw partway through arming?
  * A SHARED READ PATH — every open of the Settings section calls this route.
  * correctness and edge cases
  * error handling and failure paths
  * tests — do the specs go RED on the right mutations? I measured three: substituting the
    SAVED value for the ARMED one reddens the three pending cases and the three pre-existing
    "always says" cases; loosening the wire guard reddens exactly NaN / infinity / negative;
    removing the new start-of-read clear reddens exactly the new window spec. Verify them.
  * i18n lockstep across en, ja, ko, zh-CN, zh-TW — check the MESSAGES, not just key parity.
  * accessibility
  * security
  * consistency with the rest of the codebase
  * docs/plan accuracy — report a sentence ONLY if believing it would cause a WRONG CHANGE;
    file anything merely behind as N, which does not block the verdict.

── ALREADY SETTLED — do not re-raise in the same form ───────────────────────
row 0  | Sandbox: typecheck 0, socket 0, curl 200, gh 0, deletions permitted. vitest exits 1
         in YOUR sandbox on session-settings.spec.ts (14), spawn-antigravity.spec.ts (1),
         spawn-grok-mcp-sync.spec.ts (1) — `EPERM ... mkdir '/test/dir/.agents'`, because your
         sandbox refuses writes outside the workspace and those specs write under the real
         HOME. All three are BYTE-IDENTICAL to origin/main and none is in this diff; my own
         run is exit 0. Not findings.
row 0b | Do NOT run `yarn test` — the full suite deadlocks on this machine.
row 0c | A hook timeout in a spec that builds a temp git repo is the load, not the code.
row 1  | (mine, fixed before round 1) the wire field was validated with `typeof x === "number"`,
         admitting NaN / Infinity / negative. Now a named guard sends anything unusable to the
         unknown state. Break-verified both ways.
row 2  | (YOURS, round 1, P3) the in-flight reload kept reporting the previous armed cadence.
         FIXED — see "what changed" above. Say if the fix is wrong or incomplete.

If a resolution above is wrong, say so and mark it 'REOPENING #<row>' with evidence.

Line 1, on its own:
  'CODEX VERDICT: LGTM'  or  'CODEX VERDICT: CHANGES REQUESTED' + bulleted summary.
Then, on its own line:

Post inline comments via gh api repos/receptron/mulmoterminal/pulls/2199/comments
and exactly ONE top-level summary via gh pr comment 2199. Do not apply fixes yourself.
```

</details>

<details><summary>Round 2 — Codex reply (verbatim stdout)</summary>

```text
Posted the round-2 review comment on PR #2199: LGTM, no blocking findings.

I read the full diff, checked the surrounding code and i18n messages, ran the exact requested Vitest slice successfully, and verified the named mutation behaviors. No inline comments were needed.
EXIT=0
```

</details>

<details><summary>Round 3 — prompt sent (verbatim)</summary>

```text
Round 3 on PR #2199 at https://github.com/receptron/mulmoterminal/pull/2199
NEW HEAD: 6c44a137 (round 2 reviewed 17e58e4c and returned LGTM). Base: cfc7dbfb.
Read the whole diff: git diff cfc7dbfb...HEAD

── WHY THERE IS A ROUND 3 AT ALL ────────────────────────────────────────────
Your round 2 was LGTM with one N (prose): a comment still reading as pre-#2184
background. I fixed it — and a sweep for `#2184` across the repo found THREE MORE
surfaces you did not name:

  1. src/components/settings/SurvivingSessionsSection.vue — the comment you named.
  2. test/src/components/settings/survivingSessionsSection.spec.ts — a comment
     justifying its assertion with "the armed cadence is not sent to it", which is no
     longer true. The assertion itself is still right, for a different reason: the
     STEPPER's hint describes what is SAVED because that is what the control edits.
  3. + 4. docs/guide/en/config.md and docs/guide/ja/config.md — both described the row
     naming an event rather than a clock and stopped there; they now also say the
     section names what the server actually armed.

That commit is PROSE ONLY — no behaviour changed. But it is a push, so the head you
approved is not the head that would merge, which is the whole reason for this round.

── WHAT TO DO THIS ROUND ────────────────────────────────────────────────────
Short and targeted, but do not assume:

  * Read commit 6c44a137 and confirm it is genuinely prose-only — no behaviour,
    no assertion, no key, no rendered string changed. If any of it changes behaviour,
    say so; that is the one thing that would make this round expensive.
  * Confirm the four corrected sentences are now TRUE of the code, and that correcting
    them has not exposed a further claim that is wrong (a fix to a claim often reveals
    the next one).
  * Sweep `#2184` and the phrase "armed" yourself across server/, src/, test/, docs/
    and say whether anything ELSE still describes this as future work or still says the
    cadence is not sent to the browser.
  * Re-confirm the round-2 axis results still hold on this head. You may state them as
    unchanged if the diff since 17e58e4c is prose only — but say that you checked.

Run the tests with exactly this and nothing wider:
  npx vitest run test/src/composables/useSurvivingSessions.spec.ts test/src/components/settings/survivingSessionsSection.spec.ts test/server/session/reap-schedule.spec.ts test/server/infra/tmux-routes.spec.ts
Do NOT run `yarn test`; do NOT run all of test/server/session/ (row 0). Use plain `rm`.

── ALREADY SETTLED — do not re-raise in the same form ───────────────────────
row 0  | Sandbox: typecheck 0, socket 0, curl 200, gh 0, deletions permitted. vitest exits 1
         in YOUR sandbox on session-settings.spec.ts (14), spawn-antigravity.spec.ts (1) and
         spawn-grok-mcp-sync.spec.ts (1) — `EPERM ... mkdir '/test/dir/.agents'`; your sandbox
         refuses writes outside the workspace and those specs write under the real HOME. All
         three are byte-identical to origin/main and none is in this diff. Not findings.
row 0b | Do NOT run `yarn test` — the full suite deadlocks on this machine.
row 0c | A hook timeout in a spec that builds a temp git repo is the load, not the code.
row 1  | (mine, fixed pre-round-1) the wire field was validated with `typeof x === "number"`,
         admitting NaN / Infinity / negative. Named guard now sends anything unusable to the
         unknown state. Break-verified both ways.
row 2  | (yours, round 1, P3) the in-flight reload kept reporting the previous armed cadence.
         FIXED by dropping the value when the read STARTS. Break-verified.
row 3  | Round 2: LGTM, FINDINGS COMPLETE, all 11 axes answered, three mutations reproduced.
row 4  | (yours, round 2, N) the stale pre-#2184 comment. FIXED, plus the three surfaces above.

If a resolution is wrong, say so and mark it 'REOPENING #<row>' with evidence.

Line 1, on its own:
  'CODEX VERDICT: LGTM'  or  'CODEX VERDICT: CHANGES REQUESTED' + bulleted summary.
Then the axis table (you may mark axes unchanged-since-round-2 where the prose-only diff
makes that true, but say you checked).
Then, on its own line:

Post via gh pr comment 2199 (ONE top-level summary). Do not apply fixes yourself.
```

</details>

<details><summary>Round 3 — Codex reply (verbatim stdout)</summary>

```text
Posted the round-3 PR comment:

https://github.com/receptron/mulmoterminal/pull/2199#issuecomment-5745984601

Verdict was `CHANGES REQUESTED` for one remaining prose-only stale claim at `docs/guide/en/config.md:1991`. The requested vitest slice passed: 4 files, 64 tests.
EXIT=0
```

</details>

<details><summary>Round 4 — prompt sent (verbatim)</summary>

```text
Round 4 on PR #2199 at https://github.com/receptron/mulmoterminal/pull/2199
NEW HEAD: a139ca65 (round 3 reviewed 6c44a137). Base: cfc7dbfb.
Read the whole diff: git diff cfc7dbfb...HEAD

── WHAT HAPPENED IN ROUND 3, AND WHAT I DID WITH IT ─────────────────────────
You REOPENED row 4 and you were right: docs/guide/en/config.md still said the section
names "only what is saved". My round-2 sweep grepped `#2184`; that line does not cite it,
so the sweep failed at the boundary the tool worked within.

I enumerated every sentence that DESCRIBES the behaviour instead, and it found two more
you did not name — including one you could not have reached with a docs sweep:

  1. docs/guide/en/config.md — the line you named.
  2. docs/guide/ja/config.md — its counterpart, kept in sync.
  3. server/config/config-routes.ts:187 — a SOURCE comment carrying the same false claim
     ("nothing in the Settings section names when the next sweep is").

This was the third round on one class, so I changed the SHAPE rather than patching a third
case. Every one of these rotted for the same reason: it asserted what ANOTHER file's UI
displays, so it went false whenever the UI changed. They now describe the behaviour of the
thing they sit beside and POINT at the section instead of speaking for it.

Fixing them exposed the next claim, as it usually does. The component's own block still
said "the browser only ever sees the saved one" — the premise the whole paragraph rested
on, and precisely what this PR stopped being true. I removed it and kept the surviving
rule (no sentence names a TIME the next sweep will run) with the reason it still holds:
a cadence is not a countdown, and nothing here knows when the current interval started.

── WHAT TO DO THIS ROUND ────────────────────────────────────────────────────
  * Confirm a139ca65 is genuinely prose-only — no behaviour, assertion, config key or
    app-rendered string changed. That is the one thing that would make this expensive.
  * Judge the SHAPE CHANGE, not just the sentences: is "describe the behaviour beside you,
    point at the section rather than speaking for it" actually durable, or have I written a
    new claim that will rot in its own way? Name it if so.
  * Is the surviving rule TRUE — that nothing in that section names a time the next sweep
    will run? Check the rendered strings, not the comment.
  * Sweep for the class yourself, by MEANING rather than by the issue number: any sentence
    anywhere in server/, src/, test/, docs/ that still says the browser sees only the saved
    cadence, or that the section cannot say what is running, or that #2184 is future work.
  * Re-confirm the round-2/3 axis results still hold on this head; you may mark them
    unchanged if the diff since 6c44a137 is prose only, but say that you checked.

Run the tests with exactly this and nothing wider:
  npx vitest run test/src/composables/useSurvivingSessions.spec.ts test/src/components/settings/survivingSessionsSection.spec.ts test/server/session/reap-schedule.spec.ts test/server/infra/tmux-routes.spec.ts
Do NOT run `yarn test`; do NOT run all of test/server/session/ (row 0). Use plain `rm`.

── ALREADY SETTLED — do not re-raise in the same form ───────────────────────
row 0  | Sandbox: typecheck 0, socket 0, curl 200, gh 0, deletions permitted. vitest exits 1 in
         YOUR sandbox on session-settings.spec.ts (14), spawn-antigravity.spec.ts (1),
         spawn-grok-mcp-sync.spec.ts (1) — `EPERM ... mkdir '/test/dir/.agents'`; your sandbox
         refuses writes outside the workspace. All byte-identical to origin/main, none in this
         diff, my own run exit 0. Not findings.
row 0b | Do NOT run `yarn test` — the full suite deadlocks on this machine.
row 0c | A hook timeout in a spec that builds a temp git repo is the load, not the code.
row 1  | (mine, pre-round-1) wire field validated with `typeof x === "number"`, admitting
         NaN/Infinity/negative. Named guard now sends anything unusable to the unknown state.
row 2  | (yours, round 1, P3) in-flight reload kept reporting the previous armed cadence.
         FIXED by dropping the value when the read STARTS. Break-verified.
row 3  | Round 2: LGTM, FINDINGS COMPLETE, 11 axes, three mutations reproduced.
row 4  | (yours, round 2, N) stale pre-#2184 comment — FIXED, plus three surfaces you had not
         named. REOPENED by you in round 3 because one guide line survived; see row 5.
row 5  | (yours, round 3) the surviving guide line. FIXED, plus two more found by enumerating
         meaning rather than grepping the issue number — one of them a SOURCE comment.
row 6  | Third round on one class → the shape changed rather than the cases.
row 7  | (mine, round 3) "the browser only ever sees the saved one" — the exposed next claim.
         FIXED.

If a resolution is wrong, say so and mark it 'REOPENING #<row>' with evidence.

Line 1, on its own:
  'CODEX VERDICT: LGTM'  or  'CODEX VERDICT: CHANGES REQUESTED' + bulleted summary.
Then the axis table (mark axes unchanged-since-round-3 where the prose-only diff makes that
true, but say you checked).
Then, on its own line:

Post via gh pr comment 2199 (ONE top-level summary). Do not apply fixes yourself.
```

</details>

<details><summary>Round 4 — Codex reply (verbatim stdout)</summary>

```text
Posted the review comment on PR #2199.

Verdict: `CHANGES REQUESTED`

Main finding: `server/skills/mulmoterminal-config/SKILL.md:315-318` still says “the browser only sees the saved one,” which is stale now that this PR reports `armedReapIntervalHours` to the browser. Tests passed with the exact requested vitest command: 4 files, 64 tests.
EXIT=0
```

</details>

```text
Round 5 on PR #2199 at https://github.com/receptron/mulmoterminal/pull/2199
NEW HEAD: 57bb40c7 (round 4 reviewed a139ca65). Base: cfc7dbfb.
Read the whole diff: git diff cfc7dbfb...HEAD

── WHAT I DID WITH YOUR ROUND-4 FINDING ─────────────────────────────────────
You were right, and the cause was mine twice over:
  (a) In round 3 I DELETED "the browser only ever sees the saved one" from the component
      and never grepped for copies of it. A phrase removed in one place is exactly the
      phrase to search for.
  (b) In the FIRST commit I edited that same SKILL.md and updated the bullet citing
      #2184, while the bullet above it carried the same premise WITHOUT citing the
      issue — so neither an `#2184` grep nor my by-meaning sweep reached it.

I also found the deeper fault: I had LAYERED an "#2184 added…" bullet on top of the stale
one, so the document contradicted itself — one bullet said the browser sees only the saved
value, the next said it now sees the armed one. The two are now ONE statement of what the
section does, rather than a narration of what changed.

An exhaustive phrase sweep this round says that was the last copy. Check that claim.

── THIS IS ROUND 5: THE CHECKPOINT RIDES IN THIS CALL ───────────────────────
As well as this round's findings, read the PR WHOLE — title, body, full diff, ledger — and
answer, under a heading 'CODEX CHECKPOINT: round 5', after the axis table:

  1. Does the diff still accomplish what the PR says it does?
  2. Has the review conversation drifted onto something other than the PR's purpose?
     Be blunt: rounds 2, 3 and 4 were ALL prose. Was that worth it, or did this loop
     start reviewing itself?
  3. Is any part of the diff there only because a reviewer asked for it, without standing
     on its own merit?
  4. Should this PR be split, redirected, or CLOSED rather than converged?
  5. What is the single largest remaining risk in this change?

── WHAT ELSE TO DO THIS ROUND ───────────────────────────────────────────────
  * Confirm 57bb40c7 is prose-only.
  * Sweep the class yourself, by MEANING and by PHRASE: anything anywhere still saying the
    browser sees only the saved cadence, that the section cannot say what is running, or
    that #2184 is future work. If it is clean, say so plainly — that is the answer that
    ends this.
  * Re-confirm earlier axis results hold; mark them unchanged-since-round-4 if the diff
    since a139ca65 is prose only, but say you checked.

Run the tests with exactly this and nothing wider:
  npx vitest run test/src/composables/useSurvivingSessions.spec.ts test/src/components/settings/survivingSessionsSection.spec.ts test/server/session/reap-schedule.spec.ts test/server/infra/tmux-routes.spec.ts
Do NOT run `yarn test`; do NOT run all of test/server/session/ (row 0). Use plain `rm`.

── ALREADY SETTLED — do not re-raise in the same form ───────────────────────
row 0  | Sandbox: typecheck 0, socket 0, curl 200, gh 0, deletions permitted. vitest exits 1 in
         YOUR sandbox on session-settings.spec.ts (14), spawn-antigravity.spec.ts (1),
         spawn-grok-mcp-sync.spec.ts (1) — EPERM writing outside the workspace. All
         byte-identical to origin/main, none in this diff, my own run exit 0. Not findings.
row 0b | Do NOT run `yarn test` — the full suite deadlocks on this machine.
row 0c | A hook timeout in a spec that builds a temp git repo is the load, not the code.
row 1  | (mine, pre-round-1) wire field admitted NaN/Infinity/negative. Named guard now sends
         anything unusable to the unknown state. Break-verified.
row 2  | (yours, r1, P3) in-flight reload reported the previous armed cadence. FIXED by dropping
         the value when the read STARTS. Break-verified.
row 3  | Round 2: LGTM, FINDINGS COMPLETE, 11 axes, three mutations reproduced.
row 4  | (yours, r2, N) stale pre-#2184 comment. FIXED plus three more. REOPENED in r3.
row 5  | (yours, r3) surviving guide line. FIXED plus two more, one a SOURCE comment.
row 6  | Third round on one class → shape changed: sentences describe the behaviour beside them
         and point at the section rather than speaking for it.
row 7  | (mine, r3) "the browser only ever sees the saved one" in the component. FIXED.
row 8  | (yours, r4) the same premise in SKILL.md. FIXED — see above.
row 9  | (mine, r4) the doc contradicted itself because an update was LAYERED on a stale
         paragraph. Merged into one statement of what IS.

If a resolution is wrong, say so and mark it 'REOPENING #<row>' with evidence.

Line 1, on its own:
  'CODEX VERDICT: LGTM'  or  'CODEX VERDICT: CHANGES REQUESTED' + bulleted summary.
Then the axis table.
Then 'CODEX CHECKPOINT: round 5' with the five answers.
Then, on its own line:

Post via gh pr comment 2199 (ONE top-level summary). Do not apply fixes yourself.
```

</details>

<details><summary>Round 5 — Codex reply (verbatim stdout)</summary>

```text
Posted the round-5 top-level comment: https://github.com/receptron/mulmoterminal/pull/2199#issuecomment-5746036387

Verdict: `CODEX VERDICT: LGTM`

Confirmed `57bb40c7` is prose-only since round 4, ran the exact requested vitest slice with 4 files / 64 tests passing, and found no remaining current stale claim that the browser only sees saved cadence, the section cannot say what is running, or `#2184` is future work.
EXIT=0
```

</details>

PR #2200, opened 2026-09-19
test(pty): cover writing into a live PTY and reading it back (#2170)

### LINKED ISSUE #1595: fix: PTY の master fd がリークし、ptmx 上限到達で新規セッションが起動不能になる（kill() のみで destroy() 未呼び出し）
## What happened

サーバを 5 日連続稼働させたあと、新しいセッションが一切起動できなくなった。

```
[forkpty: Device not configured]
[新しいプロセスを作成して擬似ttyを開くことができませんでした。]
```

原因は PTY の master fd リーク。サーバプロセスが `/dev/ptmx` を 498 本掴んだままで、macOS の
上限に達していた。

```
$ lsof /dev/ptmx | awk 'NR>1{print $1}' | sort | uniq -c | sort -rn
 498 node      ← server/index.ts (稼働 5日07時間)
  13 tmux      ← セッション数と一致、正常
-----
 511 合計

$ sysctl kern.tty.ptmx_max
kern.tty.ptmx_max: 511    ← ぴったり満杯

$ ps -e -o tty= | grep -v '^ *??' | sort -u | wc -l
26                        ← 実際に制御端末として使われているのは 26 本だけ
```

プロセスは死んでいるのに fd だけが残る、という形のリーク（472 本前後が死んだ PTY の残骸）。

### 機構

`kill()` はプロセスにシグナルを送るだけで、master fd は閉じない。node-pty 1.1.0 の実装:

```js
// node_modules/node-pty/lib/unixTerminal.js:93
_this._socket = new tty.ReadStream(term.fd);   // master fd を保持

// node_modules/node-pty/lib/terminal.js:169  ← プロセス終了時に呼ばれる経路
Terminal.prototype._close = function () {
    this._socket.readable = false;
    this.write = function () { };
    this.end = function () { };
    this._writable = false;
    this._readable = false;                    // フラグを落とすだけ。fd は開いたまま
};

// node_modules/node-pty/lib/unixTerminal.js:215  ← fd を解放する唯一の経路
UnixTerminal.prototype.destroy = function () {
    this._close();
    this._socket.once('close', function () { _this.kill('SIGHUP'); });
    this._socket.destroy();                    // ← ここで初めて fd が閉じる
    this._writeStream.dispose();
};
```

つまり `_socket.on('close') → _close()` の自動経路ではフラグしか落ちず、**`destroy()` を明示的に
呼ばない限り master fd は永久に残る**。

そして本リポジトリには node-pty の PTY に対する `.destroy()` 呼び出しが 1 箇所も無い:

```
$ grep -rn "\.destroy()" server common src --include="*.ts"
server/backends/streamFile.ts:23    else res.destroy();        # http response
server/routes/ws-routes.ts:387      socket.destroy();          # websocket
src/components/cmEditor.ts:69       destroy: () => view.destroy();  # CodeMirror
```

PTY 側は全て `kill()` のみ:

- `server/session/lifecycle.ts:152` — `entry.term.kill()`
- `server/routes/ws-routes.ts:136` — `term.kill()` （ephemeral run）
- `server/agents/rate-limit-probe.ts:76` — `pty?.kill()`
- `server/infra/sandbox.ts:213` — `proc.kill()`

（`server/session/headlessScreen.ts:48` の `dispose()` は xterm.js のヘッドレス
エミュレータで、node-pty とは無関係。）

### 主なリーク源は rate-limit プローブ

`ProbePty` インターフェース（`server/agents/rate-limit-probe.ts:28-31`）が構造的に destroy を
呼べない形になっている:

```ts
export interface ProbePty {
  write(data: string): void;
  kill(): void;          // ← destroy が無いので、呼び出し側に解放手段がない
}
```

`server/index.ts:292` で `spawn: (args, cwd) => spawnPty(CLAUDE_BIN, args, cwd)` として本物の
`IPty` を渡しているが、この狭いインターフェース越しでは `destroy()` に手が届かない。

プローブは `RATE_LIMIT_STALE_MS = 10 * 60_000`（`server/agents/rate-limit-store.ts:24`）に従い、
ブラウザが見ている間 10 分ごとに 1 本ずつ PTY を作って kill する。実測リーク速度と一致する:

```
498 本 ÷ 127 時間 ≒ 3.9 本/時
10 分間隔 = 6 本/時 × ブラウザ稼働率 65% ≒ 3.9 本/時   ← 整合
```

ただし欠陥自体は `kill()` を呼ぶ全経路に共通なので、プローブだけを直しても通常セッションの
終了で同じリークが（より低速に）続く。

## What I expected

PTY を kill したら、そのプロセスが握っていた master fd も解放され、`lsof /dev/ptmx` の本数が
実際に生きているセッション数の近傍で安定してほしい。少なくとも、サーバを数日動かしただけで
OS の PTY 上限を使い切って新規セッションが起動不能になることはないでほしい。

## Steps to reproduce

1. tmux のある macOS で MulmoTerminal を起動する（`npm run server`）
2. ブラウザでグリッドを開いたまま放置する（rate-limit プローブが 10 分ごとに走る条件）
3. 本数を定期的に観測する:
   ```
   lsof /dev/ptmx | grep -c node
   ```
   単調増加し、減ることがない。実セッション数を増減させても戻らない。
4. `sysctl kern.tty.ptmx_max` の値（デフォルト 511）に達すると、以降の新規セッション作成が
   `forkpty: Device not configured` (ENXIO) で失敗する。

現在の環境では 5 日 7 時間で 498 本に到達した。

回避策はサーバプロセスの再起動（tmux セッションは親が PID 1 で独立しているため生存する）。
`sudo sysctl -w kern.tty.ptmx_max=999` は上限を上げるだけで、同じペースで約 5 日後に再発する。

## 直し方（案）

1. `IPty` を確実に破棄する共通ヘルパを 1 つ用意し、`kill()` の全呼び出し箇所を置き換える。
   `destroy()` は内部で `_socket` の close 後に `kill('SIGHUP')` を呼ぶので、二重 kill に
   ならない順序（kill → onExit → destroy、あるいは destroy 単独）に統一する必要がある。
2. `ProbePty` に `destroy(): void` を足し、`stop()` で `kill()` の後に呼ぶ。
   プローブは 10 分ごとに確実に走る経路なので、ここだけでもリーク速度は大きく下がる。
3. リグレッションテスト: fake PTY で `stop()` が destroy まで到達することを検証する
   （`server/agents/rate-limit-probe.spec.ts` に既に fake spawn の枠組みがある）。

## Environment

- mulmoterminal: 2.7.0 — ローカル開発、`npm run server`、`main` @ `e746cb6`
  （`/api/update-status` は `Update available: e746cb6 → origin` と表示。origin 側の
  最新でも上記コードパスが変わっていなければ同じはず）
- node-pty: 1.1.0
- Node: v25.2.1
- OS: macOS (Darwin 25.2.0, x86_64)
- shell: zsh
- tmux: 3.6a（13 セッション稼働中）
- gh: 2.83.2
- claude: 2.1.226 / codex-cli: 0.142.3
- ブラウザ: サーバ側の fd リークのため未取得（症状は描画とは無関係）

## Attachments

なし（本文中のコマンド出力が全て）。

### LINKED ISSUE #2170: test(windows): 生きた PTY への write を覆うテストが無い — ピン留めした node-pty にその経路の未対応報告がある
## What happens

**Windows のテストは、セルに文字を打ち込む経路を一度も通っていません。**

製品がセルで使うのは「シェルを起こしたまま `write()` で流し込む」形です。

```ts
server/session/pty-connection.ts:70    term.write(msg.data);
server/session/pty-connection.ts:162   entry.term.write(msg.data);
server/session/draft-injection.ts:103  entry.term.write(`\x1b[200~${line}\x1b[201~`);
```

一方、Windows の spec 2本はどちらも**引数を渡して `onExit` を待つ一発コマンド型**です。

```ts
// test/server/session/shell-spawn-win.spec.ts:30
const term: IPty = spawnPty(shell, args, cwd);
return new Promise((resolve) => term.onExit(({ exitCode }) => resolve(...)));

// test/server/session/pty-spawn-win.spec.ts:64
const term = spawnPty(SHIM, args, dir);
```

**生きた PTY に `write()` して読み返すテストが無い。** つまり Windows で最も使われる経路に、CI の目が届いていません。

## なぜ今これが問題か

`package.json` は `node-pty` を **`1.2.0-beta.15`** に固定しています（`^` なし）。#1595 の fd リークを直すために 1.1.0 から上げた版です（PR #1597 / 4.8.1）。

その版に対して、上流に**未対応の報告**があります。

> [microsoft/node-pty#955](https://github.com/microsoft/node-pty/issues/955)
> Windows: 1.2.0-beta.15 terminal process exits before the first write (beta.14 is fine)

報告は bisect 済みで、`beta.14` まで通り `beta.15` だけが落ちる、と書かれています。落ちる形は `Error: terminal process has exited` — **最初の write の時点**です。2026-08-18 起票、**コメント0**。

**Windows で実際に壊れているとは言っていません。** 動かして確かめていないからです。言えるのは、**壊れていても気づけない状態にある**ことです。

## 提案

Windows の spec に、**永続セッションの往復**を1本足す。

1. 引数なしでシェルを `spawnPty` する
2. `onData` を張る
3. `write()` でコマンドと改行を送る
4. 期待する文字列が返るまで待つ（タイムアウトつき）
5. 片付ける

これで、無事なら**無事だという証拠**が残り、壊れていれば CI が見つけます。

## 壊れていた場合

`beta.14` に下げるのが候補になりますが、**その前に確かめることがあります** — #1595 の fd リーク修正が `beta.14` に入っているか。入っていなければ、リークとの二択になります。

## 確認していないこと

- **Windows で実際に壊れるかどうか**。Windows 環境で動かしていません
- `beta.14` に fd リーク修正が入っているか。node-pty の 1.1.0 と beta.15 では `onexit` の
  socket 破棄経路が同一だったので、リーク修正は別の場所にあるはずですが、どの beta で
  入ったかは追っていません

isamu:
## Summary

Nothing in this repo has ever written into a PTY that stays alive. Every spec that spawns one hands it argv and waits for `onExit` — a one-shot command — while the shape a cell actually uses is a shell started once and fed with `write()`:

```
server/session/pty-connection.ts:70    term.write(msg.data)
server/session/pty-connection.ts:162   entry.term.write(msg.data)
server/session/draft-injection.ts:103  entry.term.write(`\x1b[200~${line}\x1b[201~`)
```

This adds one spec covering that. **No production code changes.**

The issue frames the gap as Windows-only. It is not — grepping `test/` and `server/**/*.spec.ts` for a pty `write()` finds none on any platform. Windows is the leg with an **open upstream report**: `node-pty` is pinned to an exact `1.2.0-beta.15` (no caret — the version that fixed #1595's fd leak), and [microsoft/node-pty#955](https://github.com/microsoft/node-pty/issues/955) reports that exact version exiting with *"terminal process has exited"* **at the first write**, bisected, beta.14 clean, no reply since 2026-08-18. Nobody knew whether this repo was affected and nothing would have said so. `windows-pr.yaml` runs `yarn test` on every PR, so **this PR's own Windows check is the answer.**

The spec starts the **Shell cell's own invocation** — `launchInvocation(defaultShellTarget(...))` — rather than an ad-hoc shell. **Which** shell that is belongs to the environment, not to the spec: `$SHELL` if set, else `%ComSpec%` (cmd.exe) on Windows or `/bin/sh` on POSIX. So the commands are written in a form all of them parse the same, `node -e "…"`.

## Items to Confirm / Review

1. **It runs on every platform, not only Windows.** The gap is platform-independent and a cross-platform spec is one I could verify locally, where a Windows-only one I could only guess at. Say if you would rather it were gated to `win32` like its two neighbours.

2. **The echo trap is the core of the design.** A terminal repeats what it is sent, so a command whose source contains the token it prints would satisfy every assertion **while the shell sat idle** — the spec would pass against a pty that reads and never runs, which is precisely the failure it exists to detect. Two guards: each command assembles its token from two halves (`node -e "process.stdout.write('MT'+'OK-live')"`), and one shared assertion requires the token to appear **exactly once**. Both are themselves tested — one case checks the construction, another drives the assertion with a fabricated stream carrying a doubled token, because the two guards only ever fail together and a mutation over the file cannot otherwise reach the second.

3. **The waiter races the pty's exit.** It settles on `output`, `exited` or `timeout`. Without that, a pty that DIED and a runner that was merely SLOW are indistinguishable. Measured by mutating the pty to exit at the first write — node-pty#955's own shape: **with** the distinction the suite reports in ~6s saying `expected 'exited' to be 'output'`; **without** it, ~125s saying `expected 'timeout'`, which reads as a loaded runner.

4. **The separate liveness case** asserts the pty is up *before* the write as well as after. "Never started" and "the write killed it" are different bugs with different owners; nothing else here tells them apart.

5. **The paste case asserts what this repo owns, not what the shell does** — please check you agree with where that line is drawn. It originally asserted the shell *ran* the pasted command, which passes under zsh and **times out under bash and under `/bin/sh`** — what ubuntu CI would have hit. Codex found it; I had only ever run the passing configuration. Asserting it was also the wrong subject: draft injection targets an agent's TUI, which enables bracketed paste, while an arbitrary `$SHELL` may not have the mode.

   It now asserts that the escape-wrapped bytes **reached the child**, evidenced by the child's own echo of them, plus that the session still answers a command afterwards. The echo half is load-bearing: Codex deleted *both* paste writes from the earlier version and it still passed, because "the session works afterwards" is true of a write that never happened.

6. **Commands do not assume a shell.** They branched on `process.platform`, which does not tell you the shell — a Windows runner sets `%ComSpec%`, so PowerShell text would have gone into cmd.exe. They are now `node -e "…"`, identical in sh, bash, zsh, PowerShell and cmd. Node is installed wherever the suite runs; what `ptyEnv` strips from the child's PATH is yarn's shim directory and `node_modules/.bin`, not a runner's own node directory — and a shell that cannot find node fails loudly. Matching is done on escape-stripped text, because a shell that highlights as you type splits its echo with SGR sequences and PSReadLine does.

7. **Readiness is observed, not slept through.** The shell's first output (its prompt), then a short grace for the terminal modes it sets immediately after. A fixed sleep is either too long on every run or too short on the one loaded runner that matters.

8. **No interaction with `pty-fd-leak.spec.ts`**, which counts this process's fds — vitest's default pool runs each file in its own process (no `pool`/`isolate` override), confirmed independently by Codex. Flagging it because a shared process would have made that spec flaky in a way that looked like a real leak.

## Verification

- Passes under **zsh, bash, `/bin/sh`, and with `SHELL` unset** — the P2 above is why that matrix is listed rather than a single run. `format`, `lint`, `typecheck` clean.
- **Break-verified**, restores byte-identical (`cmp -s`):

  | mutation | result |
  |---|---|
  | the pty exits at the first write (node-pty#955's shape) | red, fast, naming `exited` |
  | the exactly-once half of the shared assertion dropped | red (via the fabricated-stream case) |
  | `write()` made a no-op | red |
  | a command's token put into its own source | red |
  | the pty exits at spawn, before any write | red |
  | both bracketed-paste writes deleted (Codex's own mutation) | red |
  | the bracketed-paste write deleted alone | red |
  | commands written in PowerShell syntax, run under bash — the Windows failure in miniature | red |

  Three further mutations did **not** go red, and I am not claiming they are covered: removing the exit/timeout distinction (still red, but slowly and with the wrong reason — demonstrated by the 6s/125s measurement instead), removing the readiness wait, and removing the teardown's awaited exit. The last two are guards for a slow runner and for Windows `EBUSY`; neither is observable on a fast macOS shell, which is the honest limit of what a local sweep can say here.

- **What I could not exercise locally: Windows.** That is the leg the issue is about, and this PR's `Windows (PR)` check is the ground truth for it. If it goes red at the first write, the report applies here and the next question is whether beta.14 carries #1595's fd-leak fix — deliberately out of scope until we know.

## User Prompt

- （マージ方針）収束 + CI green で自動マージ

Closes #2170

work in mulmoterminal6


isamu:

- P3: the echo-trap defense is only fully enforced in the first round-trip test; the other output-bearing cases can be satisfied by `waitFor(token)` without proving the token appeared exactly once as command output.
- P3: `afterEach` kills the interactive shell but does not wait for `onExit`, so a slow shell/conpty can still have `dir` as its cwd when `afterAll` tries to remove it, especially on Windows.
- Verified baseline locally: `yarn vitest run test/server/session/pty-live-write.spec.ts --reporter=dot` passed, 5 tests in ~10s. Extra mutation sweep: removing the bracketed-paste terminator was caught by the paste test; restored from `cp` backup and verified byte-identical with `cmp -s`.

| Axis | Result |
|---|---|
| Can this spec PASS while the thing it covers is broken? | P3 finding below: not for the current first round-trip path, but the suite does not apply its stated “exactly once” proof to every command it writes. Also, the liveness case by itself can pass a no-op-but-alive write because it ignores the `waitFor` result; the suite’s other cases currently catch a total no-op write. |
| Is the token-appears-exactly-once assertion sound on all three platforms? | P3 finding below: the assertion is sound where it exists, and the token length/120-col PTY make conpty wrapping unlikely to split the command output token. Prompt redraw/scrollback repaint should append duplicate bytes rather than satisfy exactly-once. Syntax-highlighted echo can split the source with escapes, but the source does not contain the contiguous token. The problem is that exact-once is missing from several cases. |
| FLAKINESS / TIMING | P3 teardown finding below. The fixed 800ms settle writes before readiness on very slow shells, but queued PTY input should still reach the shell; the polling deadlines fail rather than pass spuriously. 60s per test / 30s per command wait looks consistent with the existing Windows real-PTY spec’s Defender rationale, though a very overloaded Windows runner could still make the 30s inner wait red. |
| CROSS-PLATFORM | none. POSIX `echo "MT""OK-live"` works under bash/zsh/dash/sh. Windows `Write-Output ("MT" + "OK-live")` is valid PowerShell. `launchInvocation(defaultShellTarget(...))` is the right shipped path to exercise for Shell cells. |
| RESOURCE LEAKS | P3 finding below: `afterEach` calls `kill()` on every assertion path, but does not await exit/release. Nothing intentionally outlives the run, but the cleanup is not synchronized. |
| INTERACTION with `test/server/session/pty-fd-leak.spec.ts` | none. This repo does not set `pool`, and local Vitest 5.0.1 resolves the default pool to `forks`; `fileParallelism` defaults true. That means test files run in child processes, so the macOS fd-leak spec counts fds in its own worker process, not the live-PTY spec’s worker. |
| correctness and edge cases | P3 exact-once finding and P3 teardown finding below. Otherwise the command construction, live-shell path, multiple writes, and bracketed paste shape cover the intended edge cases. |
| consistency with existing real-PTY specs | mostly consistent with `shell-spawn-win.spec.ts` on 60s timeout and plain-text/conpty concerns. Different from `pty-fd-leak.spec.ts`, which waits for `onExit` after `kill()` before asserting fd cleanup; this new spec should do the same style of synchronized teardown. |
| tests: would this go red for the RIGHT reason if node-pty stopped accepting writes? | yes for total no-op writes: the output-bearing tests time out. The liveness test alone would pass no-op-but-alive writes, but the suite would still fail. If node-pty accepts writes but corrupts bracketed paste termination, the extra mutation I ran confirms the paste case goes red. |

### P3: Apply the exact-once proof to every command output

Every site: `test/server/session/pty-live-write.spec.ts:139`, `test/server/session/pty-live-write.spec.ts:152`, `test/server/session/pty-live-write.spec.ts:154`, and `test/server/session/pty-live-write.spec.ts:170` wait for a token without also asserting `occurrences(token) === 1`; line 139 also ignores the `waitFor` boolean entirely.

Smallest change that resolves it: introduce a helper like `expectCommandOutput(live, token)` that asserts both `await live.waitFor(token, PTY_TIMEOUT_MS / 2)` and `live.occurrences(token) === 1`, then use it in all four output-bearing tests. In the multi-write test, check the first token after the first write and the second token after the second write; each should still be exactly one occurrence in the accumulated stream.

What would change my mind: if the intended contract is that only the first round-trip test proves “echo cannot satisfy output” and the other cases are deliberately weaker smoke tests. That would require softening the plan/spec comments that currently describe exact-once as the defense for each command.

### P3: Wait for the killed shell to exit before removing its cwd

Every site: `test/server/session/pty-live-write.spec.ts:104`-`106` is the only cleanup for these live interactive shells.

Smallest change that resolves it: make `afterEach` async, register/keep an exit promise for `live.term`, call `kill()`, and await exit with a short timeout before clearing `live`. That matches the spirit of `pty-fd-leak.spec.ts`, which kills and awaits `onExit` before checking cleanup. This also prevents `afterAll` from racing `rmSync(dir, { recursive: true, force: true })` while the child still has `dir` as its cwd.

What would change my mind: concrete evidence from node-pty on all three CI platforms that `term.kill()` synchronously waits until the child exits and releases its cwd, or evidence that Windows can always remove a directory that is still the cwd of the just-killed-but-not-yet-reaped shell.

isamu:

- P2: the bracketed-paste case is not cross-platform for the Shell-cell path; it passes under zsh here, but times out under bash and under the POSIX `/bin/sh` fallback.
- P3: teardown waits only until either exit or the cap, but does not fail when the cap wins, so a shell that will not die can be hidden on platforms that allow removing an in-use cwd.
- P3: the PR body still contains the round-1 false liveness/measurement claim and stale 5/5 verification, even though the plan/spec were updated.

| Axis | Result |
|---|---|
| Round-1 finding #1, shared exact-once assertion | fixed. `expectCommandOutput` asserts both `settleOn(...) === "output"` and `occurrences(token) === 1`, and every output-bearing real-shell case goes through it: live, alive, first write, second write, and paste. No bypass found. |
| Round-1 finding #2, awaited teardown | fixed in shape but not complete: `afterEach` now awaits exit with a cap, but P3 below because the cap winning is treated as success. |
| `settleOn` race/verdict order | none. Output-before-exit is the right order for “token printed, then shell died”; a merely slow pty cannot become `exited` without `onExit`; and the source guard plus exactly-once assertion keep echoed input from satisfying output. |
| Fabricated doubled-token case | none. It honestly exercises the part of the `LiveShell` contract used by `expectCommandOutput`, and it fails if the exactly-once assertion is removed. |
| Observed readiness | P2 finding below. Waiting for first output plus grace is better than a blind sleep, but it does not make a non-bracketed-paste shell accept `\x1b[200~...\x1b[201~`. |
| Awaited teardown | P3 finding below. It runs on assertion failures, but the timeout cap does not fail the test when it expires. |
| FLAKINESS | P2: bash or `/bin/sh` shells fail the paste case without a node-pty regression. P3: a stuck shell can pass teardown spuriously if the cwd removal does not object. Slow startup can still spend the 10s readiness deadline and then the 30s command deadline, but I do not see that passing spuriously. |
| CROSS-PLATFORM correctness (ubuntu/dash, macOS/zsh, Windows/powershell) | P2: macOS/zsh passes locally; bash and `/bin/sh` fail the paste test locally. Ubuntu can hit bash or dash/fallback, and `defaultShellPath("linux", {})` explicitly falls back to `/bin/sh`. I could not execute Windows here; PowerShell syntax itself looks valid. |
| RESOURCE LEAKS | P3 teardown finding below. The real shells are killed on every path, but a non-exiting child is not made red by the cap. |
| correctness and edge cases | P2 and P3 below. Otherwise the live write, multiple write, exact-once, and exit-vs-timeout cases line up with the intended behavior. |
| consistency with existing real-PTY specs | Mostly consistent on using real `spawnPty` and 60s timeouts. The paste case is less consistent with the shipped draft path because draft injection targets agent TUIs that enable bracketed paste, while this spec targets arbitrary `$SHELL`. |
| CLAIMS | P3 stale PR-body claim below. Plan/spec comments are mostly aligned; minor N: the plan still says “Five cases” while the spec now has six including the fabricated-stream guard. |

### P2: Bracketed paste times out under bash and `/bin/sh`

Every site: `test/server/session/pty-live-write.spec.ts:28`-`35` claims first output plus `READY_GRACE_MS` covers bracketed paste mode; `test/server/session/pty-live-write.spec.ts:232`-`245` writes bracketed-paste markers to the Shell-cell shell and requires the token; `plans/test-2170-live-pty-write.md:25`-`31` and `plans/test-2170-live-pty-write.md:59`-`64` describe this as one spec running on every platform and waiting for prompt/grace to cover bracketed paste.

Smallest change that resolves it: do not run the bracketed-paste assertion against an arbitrary Shell-cell shell unless the test first proves that shell has enabled/understands bracketed paste. Either gate/skip that case for shells without the mode, or run the paste-shape test against a deterministic helper/TUI/readline setup that enables bracketed paste before the markers are sent. Keep the ordinary live-shell write cases on the shipped Shell-cell invocation.

What would change my mind: evidence that CI and supported Shell-cell configurations always use a shell that accepts bracketed paste markers in this exact setup. My local probes say otherwise: `SHELL=/bin/zsh yarn vitest run test/server/session/pty-live-write.spec.ts` passed 6/6, but both `SHELL=/bin/bash ...` and `env -u SHELL ...` failed only `runs a bracketed paste followed by a separate submit` with `expected 'timeout' to be 'output'` after ~30s.

### P3: Make teardown fail when the kill wait cap expires

Every site: `test/server/session/pty-live-write.spec.ts:151`-`158` says a shell that will not die “must fail the run”, but the code only `await Promise.race([dying.whenExited(), settle(KILL_GRACE_MS)])` and then continues; `test/server/session/pty-live-write.spec.ts:160` removes the cwd afterward.

Smallest change that resolves it: make the race return which branch won and assert/throw when the timeout branch wins, e.g. race `whenExited().then(() => "exited")` against `settle(KILL_GRACE_MS).then(() => "timeout")` and `expect(result).toBe("exited")` before `afterAll` removes `dir`.

What would change my mind: proof that `term.kill()` plus this `whenExited()` promise cannot miss or hang on all supported node-pty platforms, or that every platform running this spec will fail `rmSync(dir, ...)` if the killed shell remains alive in that cwd. On macOS/POSIX, removing a directory that is another process's cwd can succeed, so the current cap can hide the leak it says it exposes.

### P3: Update the PR body’s stale liveness/verification claims

Every site: PR body “Items to Confirm / Review” item 3 still says the liveness case failed in 815ms while each round trip timed out; PR body “Verification” still says “5/5 pass locally” and lists “the pty exits at spawn, before any write (node-pty#955's symptom)”. The spec now has six tests, and the corrected diagnosis is the `settleOn` output/exited/timeout race for the first-write-exit shape.

Smallest change that resolves it: rewrite those PR-body bullets to match the new code and plan: six tests, `settleOn` is what distinguishes `exited` from `timeout`, and node-pty#955's relevant shape is exit at/when the first write happens rather than the old spawn-time measurement.

What would change my mind: if GitHub is showing me a stale body through `gh pr view` and the visible PR body already contains the corrected six-test / `settleOn` wording.

isamu:
- P2: `pty-live-write.spec.ts` starts whatever `defaultShellTarget(process.platform, process.env)` chooses, but `commandFor()` picks syntax from `process.platform`. On a normal Windows runner `defaultShellPath()` prefers `ComSpec`, so this starts `cmd.exe` and then writes `Write-Output (...)`, which will not produce the token. Sites: `test/server/session/pty-live-write.spec.ts:52-56` and `:84-87`; claim sites `test/server/session/pty-live-write.spec.ts:23-25`, `:85-86`, `plans/test-2170-live-pty-write.md:25-28`, and the PR body summary item saying Windows is `powershell.exe`. Smallest change: either force this spec to launch PowerShell with an explicit test env and document that narrowed subject, or choose the probe command from the actual launched shell/target and cover the Windows `ComSpec`/`SHELL` variants the production path can start. What would change my mind: proof that the Windows PR process has neither `ComSpec` nor `SHELL`, or code that intentionally bypasses those env vars for this test.
- P3: the replacement paste case is worth keeping as a non-crash/liveness case, but it does not prove the escape-wrapped write or separate submit reached the child. I backed up with `cp`, mutated both writes out, ran `SHELL=/bin/bash yarn test test/server/session/pty-live-write.spec.ts`, and the file still passed 6/6; restored with `cp` and verified with `cmp -s`. Sites: `test/server/session/pty-live-write.spec.ts:239-263`, `plans/test-2170-live-pty-write.md:67-69`, and PR body item 5. Smallest change: after the paste write/submit, assert a portable observable from the pasted input before the follow-up command, e.g. wait for the echoed pasted source/fragments under zsh, bash, dash/sh and PowerShell, then keep the after-command liveness assertion. What would change my mind: explicitly narrowing the case and all claims to “escape-wrapped bytes do not kill the session; later plain input still works,” with a reason that no portable child-visible echo can be required.
- Round-2 status: the three accepted findings are fixed in substance except for the replacement paste assertion being narrower than its new prose claims; the teardown race now asserts the winner, and the PR/plan stale-count issue is cleaned up.
- Verification I ran: `SHELL=/bin/zsh`, `/bin/bash`, `/bin/sh`, and `env -u SHELL yarn test test/server/session/pty-live-write.spec.ts` all passed locally before mutation; one old Vitest flag I tried first was rejected before tests ran.

| Axis | Result |
|---|---|
| Are all three round-2 findings actually fixed, or fixed in appearance? | The teardown and PR-body/count fixes are real. The paste fix is real for portability, but fixed in appearance for the claim that the paste write and submit “reach the child”; the code only proves later usability. |
| The REPLACEMENT paste assertion strength | It is strong enough to be worth a liveness/non-crash case, not strong enough for the stated “both reach the child” invariant. See P3. |
| Other local-shell-only cases | P2: all `commandFor()`-driven real-shell cases depend on the Windows default shell being PowerShell, but production default-shell selection can launch `cmd.exe` or `SHELL` instead. POSIX zsh/bash/dash paths passed locally. |
| Asserted teardown in `afterEach` | No finding. Vitest attributes `afterEach` failures to the test being torn down; it can add a second failure/noise after a real assertion, but it should not make a broken test pass or report against a different test. |
| FLAKINESS | Slow-fail risks: first shell output exceeding 10s, command output exceeding 30s, kill/reap exceeding 5s, or the fixed 300ms paste beat being too short on a loaded runner. Spurious pass: the paste case passes even when both paste writes are removed. |
| CROSS-PLATFORM correctness | P2 on Windows `ComSpec`/`SHELL`; no new Ubuntu dash or macOS zsh/bash finding from the current diff. |
| RESOURCE LEAKS | No finding. The async teardown now waits for `onExit`, asserts the kill cap, and `afterAll` removes the temp dir after tests. Residual risk is only the intentional failure path where an unkillable shell fails the test. |
| correctness and edge cases | P2 shell/command mismatch; P3 paste reach not asserted. No additional correctness finding. |
| consistency with existing real-PTY specs | Mostly consistent with the existing raised-timeout real PTY style. The inconsistency is that nearby Windows shell specs distinguish PowerShell command tests from `defaultShellTarget("win32", { SHELL })`; this new spec conflates platform with actual shell syntax. |
| CLAIMS: comments, plan file, PR body | P2/P3 claim findings above. The specific wrong-change sentence is “`powershell.exe` on Windows and `bash -lc "exec '<shell>'"` on POSIX” because it hides `ComSpec`/`SHELL` and leads directly to PowerShell syntax being sent to non-PowerShell shells. “escape-wrapped write and a separate submit both reach the child” is also wrong as written. |

isamu:
CODEX VERDICT: LGTM

- No blocking findings in round 4.
- Both round-3 findings are fixed in substance: the command text no longer assumes the platform-selected shell, and the paste case now proves the pasted payload reached the child via echo before checking later usability.
- Verification I ran on this head locally: `SHELL=/bin/zsh`, `SHELL=/bin/bash`, `SHELL=/bin/sh`, and `env -u SHELL yarn test test/server/session/pty-live-write.spec.ts` all passed 6/6.

| Axis | Result |
|---|---|
| Are both round-3 findings actually fixed, or fixed in appearance? | none. The shell/platform mismatch is fixed by using one `node -e "process.stdout.write('MT'+'OK-x')"` form for every selected shell. The paste case is fixed in substance: deleting the paste write would remove the only source of `PASTED_MARKER`, so the case no longer proves only later liveness. |
| The new command form through cmd.exe, PowerShell/PSReadLine, sh/dash, bash, and zsh | none. `cmd.exe` strips the outer double quotes and passes the JS to node; PowerShell treats the double-quoted argument as one native argument, with the single-quoted JS strings intact; POSIX sh/bash/zsh do the same. The exact output token is not present in the typed source because it is split as `'MT'+'OK-*'`. |
| Is there a shell where `node` is not on PATH inside the pty although it is outside? | none as a finding. `spawnPty` does sanitize PATH, so the PR-body phrase “by definition” is slightly stronger than the code contract: `process.execPath` could exist while sanitized child PATH lacks `node`. But the targeted CI/user-shell shapes should have setup-node/user PATH entries visible to the shell, and my local zsh/bash/sh/fallback runs all found `node`. I would change this only with evidence of a supported runner where `node` is available to Vitest but intentionally absent from Shell-cell PATH. |
| The new paste assertion on conpty echo | none. The assertion only requires the visible payload bytes to be echoed, not the bracketed-paste controls or command execution. That is the portable evidence available from an interactive child; on conpty, cmd/PowerShell/PSReadLine may render input through the console host, but the payload text should still appear in the output stream. What would change my mind: a Windows run where typed printable payload characters are accepted but not surfaced through node-pty `onData`. |
| `visibleText` escape coverage | none. The regex covers the relevant CSI forms that can split prompt/input echo on Windows and PSReadLine: SGR (`m`), cursor movement/erase (`G`, `K`, etc.), and private `?` forms such as bracketed paste/cursor visibility. It does not strip OSC/DCS, but I do not see a realistic shell echo path that would inject those inside these short printable tokens. Keeping line breaks is the right call; these tokens are far below the 120-column conpty width, and removing line breaks could create false tokens across lines. |
| FLAKINESS: slow or loaded runner failures and spurious passes | none. Slow shell startup can still consume the 10s readiness window plus a 30s command wait and fail red, and a stuck kill fails after 5s; those are real slow-runner residual risks, not hidden passes. I do not see a current spurious pass for no-op writes, first-write exit, missing paste write, or doubled token. |
| CROSS-PLATFORM correctness (ubuntu/dash, macOS/zsh, Windows/cmd and PowerShell) | none. POSIX zsh/bash/sh passed locally, including the `/bin/sh` fallback. The command text is shell-neutral for cmd and PowerShell, and matching is now done on escape-stripped visible text for PSReadLine-style highlighting. |
| RESOURCE LEAKS | none. `afterEach` kills the live shell, waits for `onExit`, and asserts the kill cap winner before `afterAll` removes the temp dir. A shell that will not die is now a test failure rather than a silent leftover. |
| correctness and edge cases | none. The spec covers the live shell path, output vs exit vs timeout, liveness across first write, multiple writes on one pty, exact-once token defense, and escape-wrapped payload arrival plus post-submit usability. |
| consistency with existing real-PTY specs | none. The 60s real-PTY timeout matches the Windows shell-spawn rationale, `visibleText` follows the same idea as `plainText()` while deliberately preserving line breaks, and teardown is now synchronized like the fd-leak spec’s cleanup expectations. |
| CLAIMS: comments, plan, and PR body | none. The old “powershell.exe on Windows” claim is gone from the spec header, plan, and PR body. The paste prose now matches the code: payload arrival is proved by echo, and command execution of bracketed paste is not claimed. Minor non-blocking note: “node present by definition” is a little broad because child PATH is sanitized, but it does not point to a wrong implementation change for this CI-targeted test. |

isamu:
CODEX VERDICT: LGTM
- No blocking findings in the executable spec or the round-5 prose change.
- The round-3 fixes are substantive: the command is one shell-neutral `node -e` form, and the paste case now proves arrival plus subsequent usability.
- N-level claims drift remains in the PR/plan prose, listed below, but it does not change my merge verdict.

| Axis | Result |
|---|---|
| Are both round-3 findings actually fixed, or fixed in appearance? | Fixed in substance. The command no longer branches on `process.platform`, so `%ComSpec%`/cmd.exe is not fed PowerShell syntax; the paste test now fails if the escape-wrapped payload is not echoed and still checks the session answers after submit. |
| THE NEW COMMAND FORM | No blocking finding. `node -e "process.stdout.write('MT'+'OK-x')"` survives sh/dash, bash, zsh, cmd.exe, and PowerShell/PSReadLine: double quotes group the JS snippet, the inner single quotes are passed as JS string quotes, and the token is never present contiguously in the typed source. `node` can be missing inside the child only when it was available solely through stripped launcher PATH entries; that is no longer claimed as impossible in the spec/plan, and it would fail loudly as `node` not found. |
| THE NEW PASTE ASSERTION | No finding. On conpty, the visible input echo for cmd/PowerShell line editing comes from the console host rather than POSIX tty line discipline, but it is still exactly the observable surface node-pty returns. The assertion is therefore a valid “bytes reached the child’s terminal” check, not a shell-executed-command check. |
| `visibleText` | No finding. The regex covers the relevant SGR, cursor movement, and private `?` CSI forms that can split PSReadLine/conpty-visible tokens here. It is narrower than the repo’s full `stripPtyEscapes` helper (for example OSC and CSI with intermediates are outside it), but I do not see a missed form that would split these short `MTOK-*` tokens on the targeted shells. Keeping line breaks is the right call; conpty’s 120-column wrapping does not affect these short tokens. |
| FLAKINESS | No blocking finding. Slow/loaded runners can still fail without a product regression at readiness, command-output, or kill-grace deadlines; those failures are explicit (`timeout`, readiness delay, or teardown still running). Spurious pass surface looks covered: command tokens are split plus exactly-once, paste requires the payload echo, and then a later command must answer. |
| CROSS-PLATFORM correctness (ubuntu/dash, macOS/zsh, Windows/cmd and PowerShell) | No finding. I re-ran the spec locally with default shell, `SHELL` unset, `/bin/sh`, Homebrew bash, and Homebrew zsh; all passed. Windows parsing is consistent with cmd.exe and PowerShell native-command argument rules, with CI as the conpty ground truth. |
| RESOURCE LEAKS | No finding. Each live shell is killed in `afterEach`, the exit is awaited with an asserted cap, and the temp cwd is removed after all tests. I do not see a newly introduced orphan path. |
| correctness and edge cases | No blocking finding. The waiter distinguishes output, exit, and timeout; output wins if it arrived before exit; repeated writes are covered; and the paste test is scoped to what this repo owns. |
| consistency with the existing real-PTY specs | No finding. Timeout sizing and escape stripping are consistent with the existing Windows PTY specs, while this file covers the previously missing live-pty-write shape rather than duplicating one-shot spawn coverage. |
| CLAIMS: check every comment, the plan and the PR body against the code | N: stale examples/PR text only. Severity: N. Sites: PR body item 2 says commands split as `echo "MT""OK-live"`; `plans/test-2170-live-pty-write.md:46` says the same; PR body item 6 still says node is “present by definition since it runs the suite.” Smallest change: update those prose examples to the actual `node -e ... ('MT'+'OK-live')` construction and mirror the new child-PATH caveat from `test/server/session/pty-live-write.spec.ts:61-65`. What would change my mind: if the PR body is intentionally allowed to lag as historical review notes, or if those sentences are updated. This is not P3 because the spec and plan’s changed contract now say the important PATH caveat correctly, and the stale `echo` text is behind rather than behavior-shaping. |

Verification: read the full PR diff (`3988096..44207cff`) and the round-5-only diff (`8af4789c..44207cff`). Local runs passed for `yarn vitest run test/server/session/pty-live-write.spec.ts`, `env -u SHELL ...`, `SHELL=/bin/sh ...`, `SHELL=/opt/homebrew/bin/bash ...`, and `SHELL=/opt/homebrew/bin/zsh ...`.

isamu:
## Codex exchange — the prompts, rounds 1-5

Codex posted its own verdict each round (linked below). This is the other half of the record: what was asked to produce them. Posted as one comment rather than five, to keep the thread readable and stay well clear of the comment rate limit.

**Sandbox pre-flight (ledger row 0)** — what the reviewer could actually run, which is what stops a round being silently blind:

```text
reused from PR #2198, same machine and session: vitest exit 0, vue-tsc exit 0, socket bind exit 0, api.github.com 200, tmux on a private socket exit 0. `rm -f` refused by policy; plain `rm` and `cp` work, so mutation sweeps back up and restore with `cp` and verify with `cmp -s`.
```

```text
Review PR #2200 at https://github.com/receptron/mulmoterminal/pull/2200.

── SAFETY ───────────────────────────────────────────────────────────────────
This machine has a LIVE tmux server on socket `mulmoterminal` holding the user's real
work sessions. Never run tmux against `-L mulmoterminal` and never run a bare `tmux`.
(This PR does not touch tmux, but the machine is shared.)
The spec under review SPAWNS REAL SHELLS. Running it is fine; it kills each pty.

time in the round AFTER it is pushed. Read the WHOLE diff before you post anything.
Do not stop at the first problem. Do not save anything for a follow-up review.

Use the gh CLI to inspect the diff, then post inline comments via
  gh api repos/receptron/mulmoterminal/pulls/2200/comments
or a top-level comment via
  gh pr comment 2200

── WHAT THIS PR IS ──────────────────────────────────────────────────────────
ONE new spec, test/server/session/pty-live-write.spec.ts, plus a plan file. NO
production code changes. It closes a coverage gap: nothing in this repo has ever
written into a PTY that stays alive, which is the shape server/session/pty-connection.ts
and server/session/draft-injection.ts actually use. It runs on every platform.

THE DELIVERABLE IS THE TEST'S CORRECTNESS. A test that passes for the wrong reason is
the exact failure mode here, so review it as you would review production code.

The central hazard, and the thing to attack hardest: a terminal ECHOES what it is sent.
A command whose source text contains the token it prints would satisfy every assertion
in the file WHILE THE SHELL SAT IDLE — the spec would pass against a pty that reads and
never runs, which is precisely the failure it exists to detect. The defence is that each
command assembles its token from two halves and the token must appear EXACTLY ONCE.

  Can this spec PASS while the thing it covers is broken? Enumerate the ways. This is
    the most important axis: name any path by which an assertion is satisfied without
    the shell having executed the command.
  Is the token-appears-exactly-once assertion sound on all three platforms? Consider
    conpty line wrapping, a shell that redraws its prompt, a shell echoing with syntax
    highlighting that splits the token with escape sequences, and scrollback repaint.
  FLAKINESS / TIMING: fixed sleeps (SHELL_SETTLE_MS 800, AFTER_PASTE_MS 300) and polling
    with deadlines. Where can a slow or loaded runner make this fail without a regression?
    Where can it PASS spuriously? Is 60s per test right on a Windows runner with Defender on?
  CROSS-PLATFORM: it runs on ubuntu, macOS AND Windows. Will the POSIX command
    `echo "MT""OK-live"` behave under dash/sh (ubuntu's default if SHELL is unset) and
    the Windows one under powershell.exe? Is defaultShellTarget/launchInvocation the right
    production path to exercise?
  RESOURCE LEAKS: every test starts a real shell. Are they all killed on every path,
    including when an assertion throws? Does anything outlive the run?
  INTERACTION with test/server/session/pty-fd-leak.spec.ts, which counts THIS PROCESS's
    file descriptors. I claim vitest's default pool gives each file its own process so
    there is no interaction. Verify or refute.
  correctness and edge cases
  consistency with the rest of the codebase (compare with shell-spawn-win.spec.ts and
    pty-spawn-win.spec.ts, the existing real-pty specs)
  tests: would this go red for the RIGHT reason if node-pty stopped accepting writes?

An axis with no findings is a RESULT and has to be stated. 'No findings' on an axis you
did not look at is the answer that makes this loop longer.

1. SEVERITY — P1 / P2 / P3 / N. N is prose-only and does not block the verdict; a prose
   finding is P3+ only when believing the sentence would cause a WRONG CHANGE.
2. EVERY SITE — grep for the same mistake elsewhere and list all occurrences now.

Your sandbox refuses `rm -f` but allows plain `rm` and `cp`. Back up with `cp`, restore
with `cp`, verify with `cmp -s`. I already ran three mutations, all caught, restores
byte-identical: (1) the pty exits at spawn before any write — the liveness case failed in
815ms while the three round trips took the full ~31s timeout; (2) write() made a no-op;
(3) a command's token put into its own source. Find a mutation I did not.

── ALREADY SETTLED ──────────────────────────────────────────────────────────
| 0 | pre-flight | Codex sandbox capability | N/A | vitest 0, vue-tsc 0, socket bind 0, api.github.com 200, tmux on a private socket 0. `rm -f` refused; plain `rm` works. |

_No findings yet — this is round 1._

  Line 1: 'CODEX VERDICT: LGTM' or 'CODEX VERDICT: CHANGES REQUESTED' + a bulleted summary.
  Then, on its own line:
  Omit that line if untrue and say what you did not reach.

```

</details>

```text
Review PR #2200 at https://github.com/receptron/mulmoterminal/pull/2200, round 2, at head ea5971e6.

── SAFETY ───────────────────────────────────────────────────────────────────
Never run tmux against `-L mulmoterminal` and never run a bare `tmux` — that socket holds
the user's real sessions. The spec under review spawns real shells; running it is fine.

Every finding you hold back costs a full extra round. Read the WHOLE diff before posting.
Do not stop at the first problem. Do not save anything for a follow-up review.

Post inline comments via gh api repos/receptron/mulmoterminal/pulls/2200/comments
or a top-level comment via gh pr comment 2200.

── WHAT CHANGED SINCE ROUND 1 ───────────────────────────────────────────────
Both of your P3 findings were accepted and fixed, and the first one exposed a false claim
of mine that is also fixed. The new commit:

1. `expectCommandOutput(shell, token)` — one shared assertion used by EVERY output-bearing
   case, asserting both that the wait settled on "output" AND that the token appeared
   exactly once. Your finding was that only the first case did this.
2. A new case drives that assertion with a FABRICATED stream carrying a doubled token,
   because a mutation over this file cannot otherwise reach it: the construction guard and
   the exactly-once guard only ever fail together. (I ran that mutation in round 1 and it
   survived; it is now caught.)
3. `afterEach` awaits the killed shell's exit, capped, before `afterAll` removes the cwd.
4. NEW, and the part to review hardest: the waiter now races the token against the pty's
   EXIT and settles on "output" | "exited" | "timeout". The reason is a claim of mine that
   was false. The liveness case said it diagnosed node-pty#955 quickly; I had measured that
   against a pty killed at SPAWN, which the pre-write assertion catches, while the report
   describes an exit at the FIRST WRITE — where the old waiter burned the whole timeout and
   reported "no output". Measured on the reported shape: 6s saying `expected 'exited' to be
   'output'`, against 125s saying `expected 'timeout'` without the distinction.
5. The fixed 800ms pre-write sleep is replaced by observed readiness — wait for the shell's
   first output (its prompt), then a short grace for the terminal modes it sets immediately
   after, bracketed paste among them. That was your flakiness axis.

  Are findings #1 and #2 from round 1 ACTUALLY fixed, or fixed in appearance? Check the
    shared assertion is used by every output-bearing case with no path around it.
  The new race in `settleOn`: is the verdict order right? I put "output" ahead of "exited"
    so a pty that sent the token and then died still counts as success. Argue the opposite
    if you disagree. Can it resolve "exited" for a pty that is merely slow, or "output" from
    something other than the command?
  The new fabricated-stream case: does it actually pin the exactly-once assertion, or can it
    pass while the real assertion is gone? Is the stub honest about the LiveShell contract?
  Observed readiness: is "first output, then a grace" sound on all three platforms? Is there
    a shell that prints nothing until it is written to, which would make this wait the full
    deadline? Does the grace reliably cover bracketed-paste mode being enabled?
  The awaited teardown: correct on every path, including when an assertion throws? Does the
    cap hide a shell that will not die?
  FLAKINESS: where can a slow or loaded runner make this fail without a regression, and
    where can it PASS spuriously?
  CROSS-PLATFORM correctness (ubuntu/dash, macOS/zsh, Windows/powershell)
  RESOURCE LEAKS
  correctness and edge cases
  consistency with the existing real-PTY specs
  CLAIMS: does any comment, the plan file, or the PR body now say something the code does
    not do? I rewrote the plan's reasoning and the PR body after the change; check them
    against the code. Report a sentence that would cause a WRONG CHANGE if believed as P3+;
    anything merely behind is N.

1. SEVERITY — P1 / P2 / P3 / N (N is prose-only and does not block the verdict).
2. EVERY SITE — grep and list all occurrences now.

`rm -f` is refused; use `cp` to back up and restore, and `cmp -s` to verify.
My sweep on this head, for the record — 2 caught, 3 not, and I am not claiming the three
are covered:
  CAUGHT  pty exits at the first write (node-pty#955's shape)      red in 9s
  CAUGHT  the exactly-once half of the shared assertion dropped     red in 19s
  MISSED  settleOn stops distinguishing exit from timeout           — still red, but slowly
          and with the wrong reason; demonstrated by measurement instead (6s vs 125s)
  MISSED  readiness no longer waits for the prompt                  — a slow-runner guard,
          not observable on a fast local shell
  MISSED  teardown stops awaiting the shell's exit                  — a Windows EBUSY guard,
          not observable on macOS where a cwd can be removed while in use
Find a mutation I did not, or show one of the three MISSED ones is observable after all.

| 0 | pre-flight | Codex sandbox capability | N/A | vitest 0, vue-tsc 0, socket bind 0, api.github.com 200, tmux on a private socket 0. `rm -f` refused; plain `rm` works. |
| 1 | 1 | exactly-once proof applied to only the first case | FIXED | shared `expectCommandOutput` in every output-bearing case; commit ea5971e6. |
| 2 | 1 | teardown kills but does not await exit before rmSync | FIXED | async `afterEach` awaits `whenExited()` capped at KILL_GRACE_MS; commit ea5971e6. |
| 3 | 1 | (mine) the liveness case's "fast diagnosis" claim was measured against the wrong failure shape | FIXED | waiter races token vs exit; 6s/"exited" against 125s/"timeout". |

  Then the axis table: one row per axis, findings or 'none'.
  Then, on its own line:

```

</details>

<details><summary>Round 3 — prompt sent (verbatim)</summary>

```text
Review PR #2200 at https://github.com/receptron/mulmoterminal/pull/2200, round 3, at head a00de8d8.

── SAFETY ───────────────────────────────────────────────────────────────────
Never run tmux against `-L mulmoterminal` and never a bare `tmux` — that socket holds the
user's real sessions. The spec spawns real shells; running it is fine.

Every finding you hold back costs a full extra round. Read the WHOLE diff before posting.

Post inline via gh api repos/receptron/mulmoterminal/pulls/2200/comments, or
gh pr comment 2200.

── WHAT CHANGED SINCE ROUND 2 ───────────────────────────────────────────────
All three of your findings accepted and fixed.

1. P2 (the important one — you were right and it would have gone red on ubuntu). The paste
   case no longer asserts that the SHELL runs the pasted command. It now asserts the
   portable invariant: an escape-wrapped write and a separate submit reach the child and
   leave the session usable, checked by a following plain command that must round-trip.
   Renamed to "stays usable after an escape-wrapped write and a separate submit".
   I reproduced your failure first (bash: `expected 'timeout' to be 'output'` after ~30s)
   and then verified the new form under zsh, bash, /bin/sh and with SHELL unset: 6/6 each.
2. P3: the teardown's kill cap is now asserted — the race returns which branch won and
   `expect(ending).toBe("exited")`.
3. P3: the PR body is rewritten — six cases, the settleOn race, the corrected failure shape,
   and an explicit list of the three mutations that did NOT go red.
4. N: the plan's case count is deleted rather than corrected, per this repo's CLAUDE.md
   rule against putting figures in prose.

  Are all three round-2 findings actually fixed, or fixed in appearance?
  The REPLACEMENT paste assertion: is "the session is still usable afterwards" actually
    strong enough to be worth the case, or is it now so weak it proves nothing? If you
    think it is too weak, say what it should assert instead THAT IS PORTABLE across zsh,
    bash, dash/sh and powershell. I would rather have a weaker portable assertion than a
    stronger one that is red on ubuntu, but argue the other side if you disagree.
  Is there any OTHER case in this file that passes only because of my local shell? I tested
    zsh, bash, /bin/sh and SHELL-unset on macOS. Windows I cannot run. Name anything whose
    correctness depends on the shell rather than on node-pty.
  The asserted teardown: can `expect()` inside `afterEach` mask or replace a real test
    failure, or report against the wrong test?
  FLAKINESS: where can a slow or loaded runner make this fail without a regression, and
    where can it PASS spuriously?
  CROSS-PLATFORM correctness (ubuntu/dash, macOS/zsh, Windows/powershell)
  RESOURCE LEAKS
  correctness and edge cases
  consistency with the existing real-PTY specs
  CLAIMS: check every comment, the plan file and the PR body against the code. Report a
    sentence that would cause a WRONG CHANGE if believed as P3+; merely behind is N.

1. SEVERITY — P1 / P2 / P3 / N.  2. EVERY SITE.  3. THE SMALLEST CHANGE.  4. WHAT WOULD
CHANGE YOUR MIND.

`rm -f` is refused; back up and restore with `cp`, verify with `cmp -s`.

| 0 | pre-flight | Codex sandbox capability | N/A | vitest 0, vue-tsc 0, socket bind 0, api.github.com 200, tmux on a private socket 0. `rm -f` refused; plain `rm` works. |
| 1 | 1 | exactly-once proof applied to only the first case | FIXED | shared `expectCommandOutput`; commit ea5971e6. |
| 2 | 1 | teardown kills but does not await exit before rmSync | FIXED | async `afterEach` awaits `whenExited()`; ea5971e6. |
| 3 | 1 | (mine) the liveness "fast diagnosis" claim was measured against the wrong failure shape | FIXED | waiter races token vs exit; 6s/"exited" vs 125s/"timeout". |
| 4 | 2 | paste case not cross-platform — passes zsh, times out bash and /bin/sh | FIXED | asserts the portable invariant instead; verified on 4 shell configurations; a00de8d8. |
| 5 | 2 | teardown cap winning was treated as success | FIXED | race reports the winner and it is asserted; a00de8d8. |
| 6 | 2 | PR body carried the round-1 false claim and a stale case count | FIXED | body rewritten; plan's count deleted rather than corrected. |

  Then the axis table: one row per axis, findings or 'none'.
  Then, on its own line:

```

</details>

<details><summary>Round 4 — prompt sent (verbatim)</summary>

```text
Review PR #2200 at https://github.com/receptron/mulmoterminal/pull/2200, round 4, at head 8af4789c.

── SAFETY ───────────────────────────────────────────────────────────────────
Never run tmux against `-L mulmoterminal` and never a bare `tmux`. The spec spawns real
shells; running it is fine.

Every finding you hold back costs a full extra round. Read the WHOLE diff before posting.
Post via gh api repos/receptron/mulmoterminal/pulls/2200/comments or gh pr comment 2200.

── WHAT CHANGED SINCE ROUND 3 ───────────────────────────────────────────────
Both findings accepted; you were right on both and both would have been red on Windows.

1. P2 (platform != shell). Commands no longer branch on `process.platform`. They are now
   `node -e "process.stdout.write('MT'+'OK-x')"` — one text that sh, bash, zsh, PowerShell
   and cmd all parse the same, with node present by definition since it runs the suite.
   I verified under zsh, bash, /bin/sh and SHELL-unset (6/6 each), and confirmed the
   mutation that puts PowerShell syntax back, run under bash, goes red — the Windows
   failure in miniature.
2. P3 (the paste case proved nothing). It now asserts the child ECHOED the pasted payload,
   which is the only portable evidence the bytes arrived. YOUR mutation — deleting both
   paste writes — now goes red, and so does deleting the paste write alone. I confirmed
   both.
3. Matching moved onto escape-stripped text (`visibleText`), because PSReadLine and other
   highlighting shells split their echo with SGR sequences. Line breaks are deliberately
   KEPT, since stripping them could join two lines into a token neither contains.
4. Claims: "powershell.exe on Windows" is gone from the spec header, the plan and the PR body.

  Are both round-3 findings actually fixed, or fixed in appearance?
  THE NEW COMMAND FORM is the thing to attack. `node -e "process.stdout.write('MT'+'OK-x')"`
    is written once and typed into whatever shell the environment picked. Walk it through
    cmd.exe, PowerShell (including PSReadLine), sh/dash, bash and zsh: does the quoting
    survive each one intact, and does the token still NOT appear in the source as typed?
    Is there a shell where `node` is not on PATH inside the pty although it is outside?
  THE NEW PASTE ASSERTION: is the child's echo actually guaranteed on a conpty, where the
    echo is produced by the console host rather than a POSIX tty line discipline? If it is
    not, this is red on Windows and I need to know now rather than from CI.
  `visibleText`: does the regex cover the escape forms a shell's echo actually emits — SGR,
    cursor moves, and the private `?`-prefixed sequences? Anything it misses that could
    split a token on Windows? Is keeping line breaks the right call given conpty wraps at
    120 columns and these tokens are short?
  FLAKINESS: where can a slow or loaded runner fail this without a regression, and where
    can it PASS spuriously?
  CROSS-PLATFORM correctness (ubuntu/dash, macOS/zsh, Windows/cmd and PowerShell)
  RESOURCE LEAKS
  correctness and edge cases
  consistency with the existing real-PTY specs
  CLAIMS: check every comment, the plan and the PR body against the code. A sentence that
    would cause a WRONG CHANGE if believed is P3+; merely behind is N.

1. SEVERITY  2. EVERY SITE  3. THE SMALLEST CHANGE  4. WHAT WOULD CHANGE YOUR MIND.

`rm -f` is refused; back up and restore with `cp`, verify with `cmp -s`.
Caught so far on this head: pty exits at first write; pty exits at spawn; write() no-op;
exactly-once dropped; token in its own source; both paste writes deleted; paste write
deleted alone; PowerShell syntax under bash.
Still NOT caught, and I am not claiming otherwise: removing the exit/timeout distinction
(red but slow and mis-reasoned — shown by measurement, 6s vs 125s), removing the readiness
wait, removing the teardown's awaited exit. The last two are a slow-runner guard and a
Windows EBUSY guard, neither observable on a fast macOS shell.

| 0 | pre-flight | Codex sandbox capability | N/A | vitest 0, vue-tsc 0, socket 0, github 200, tmux private socket 0; `rm -f` refused, plain `rm` works. |
| 1 | 1 | exactly-once applied to only the first case | FIXED | shared `expectCommandOutput`; ea5971e6. |
| 2 | 1 | teardown did not await exit before rmSync | FIXED | ea5971e6. |
| 3 | 1 | (mine) liveness "fast diagnosis" measured against the wrong shape | FIXED | settleOn races token vs exit. |
| 4 | 2 | paste case not cross-platform (zsh ok, bash/sh time out) | FIXED | asserts the portable invariant; a00de8d8. |
| 5 | 2 | teardown cap winning treated as success | FIXED | the winner is asserted; a00de8d8. |
| 6 | 2 | PR body carried a false claim and a stale count | FIXED | body rewritten, count deleted not corrected. |
| 7 | 3 | commands branched on platform, but %ComSpec% means cmd.exe | FIXED | one `node -e` form for every shell; 8af4789c. |
| 8 | 3 | paste case passed with both writes deleted | FIXED | asserts the child's echo of the payload; 8af4789c. |

  Then the axis table: one row per axis, findings or 'none'.
  Then 'FINDINGS COMPLETE: I read every hunk of the diff and this is every finding I have.'

```

</details>

```text
Review PR #2200 at https://github.com/receptron/mulmoterminal/pull/2200, round 5, at head 44207cff.

── SAFETY ───────────────────────────────────────────────────────────────────
Never run tmux against `-L mulmoterminal` and never a bare `tmux`. The spec spawns real
shells; running it is fine.

Every finding you hold back costs a full extra round. Read the WHOLE diff before posting.
Post via gh api repos/receptron/mulmoterminal/pulls/2200/comments or gh pr comment 2200.

── WHAT CHANGED SINCE ROUND 4 ───────────────────────────────────────────────
You returned LGTM with FINDINGS COMPLETE on 8af4789c. One thing changed since, and it is
the non-blocking note you attached to that LGTM:

  Your note: "node present by definition" is broader than the code contract, because the
  child's PATH is sanitized.

You were right. `ptyEnv` strips yarn's shim directory, `node_modules/.bin` and
`node-gyp-bin` from the child's PATH, so "node is installed" is not "node is on the PATH
the shell searches". The comment in the spec header and the sentence in the plan now say
that, and say what actually holds: a runner's own node directory survives the stripping,
all four local shell configurations resolved it, and a shell that cannot find node fails
loudly. NO behaviour changed — comment and plan text only, commit 44207cff.

This round exists because that was a push, and a pushed change is unreviewed until the
next round. It should be short. Please confirm the new sentences are true and that nothing
else moved, and re-state the axes.

  Are both round-3 findings actually fixed, or fixed in appearance?
  THE NEW COMMAND FORM is the thing to attack. `node -e "process.stdout.write('MT'+'OK-x')"`
    is written once and typed into whatever shell the environment picked. Walk it through
    cmd.exe, PowerShell (including PSReadLine), sh/dash, bash and zsh: does the quoting
    survive each one intact, and does the token still NOT appear in the source as typed?
    Is there a shell where `node` is not on PATH inside the pty although it is outside?
  THE NEW PASTE ASSERTION: is the child's echo actually guaranteed on a conpty, where the
    echo is produced by the console host rather than a POSIX tty line discipline? If it is
    not, this is red on Windows and I need to know now rather than from CI.
  `visibleText`: does the regex cover the escape forms a shell's echo actually emits — SGR,
    cursor moves, and the private `?`-prefixed sequences? Anything it misses that could
    split a token on Windows? Is keeping line breaks the right call given conpty wraps at
    120 columns and these tokens are short?
  FLAKINESS: where can a slow or loaded runner fail this without a regression, and where
    can it PASS spuriously?
  CROSS-PLATFORM correctness (ubuntu/dash, macOS/zsh, Windows/cmd and PowerShell)
  RESOURCE LEAKS
  correctness and edge cases
  consistency with the existing real-PTY specs
  CLAIMS: check every comment, the plan and the PR body against the code. A sentence that
    would cause a WRONG CHANGE if believed is P3+; merely behind is N.

1. SEVERITY  2. EVERY SITE  3. THE SMALLEST CHANGE  4. WHAT WOULD CHANGE YOUR MIND.

`rm -f` is refused; back up and restore with `cp`, verify with `cmp -s`.
Caught so far on this head: pty exits at first write; pty exits at spawn; write() no-op;
exactly-once dropped; token in its own source; both paste writes deleted; paste write
deleted alone; PowerShell syntax under bash.
Still NOT caught, and I am not claiming otherwise: removing the exit/timeout distinction
(red but slow and mis-reasoned — shown by measurement, 6s vs 125s), removing the readiness
wait, removing the teardown's awaited exit. The last two are a slow-runner guard and a
Windows EBUSY guard, neither observable on a fast macOS shell.

| 0 | pre-flight | Codex sandbox capability | N/A | vitest 0, vue-tsc 0, socket 0, github 200, tmux private socket 0; `rm -f` refused, plain `rm` works. |
| 1 | 1 | exactly-once applied to only the first case | FIXED | shared `expectCommandOutput`; ea5971e6. |
| 2 | 1 | teardown did not await exit before rmSync | FIXED | ea5971e6. |
| 3 | 1 | (mine) liveness "fast diagnosis" measured against the wrong shape | FIXED | settleOn races token vs exit. |
| 4 | 2 | paste case not cross-platform (zsh ok, bash/sh time out) | FIXED | asserts the portable invariant; a00de8d8. |
| 5 | 2 | teardown cap winning treated as success | FIXED | the winner is asserted; a00de8d8. |
| 6 | 2 | PR body carried a false claim and a stale count | FIXED | body rewritten, count deleted not corrected. |
| 7 | 3 | commands branched on platform, but %ComSpec% means cmd.exe | FIXED | one `node -e` form for every shell; 8af4789c. |
| 8 | 3 | paste case passed with both writes deleted | FIXED | asserts the child's echo of the payload; 8af4789c. |

  Then the axis table: one row per axis, findings or 'none'.
  Then 'FINDINGS COMPLETE: I read every hunk of the diff and this is every finding I have.'

```

</details>

isamu:
## Cross-review triage and ledger — PR #2200, tier S, five rounds

**Tier S**, stated before round 1. Test-only with no production change would normally be tier T, but the deliverable here **is the test's correctness**, and a test that passes for the wrong reason is the exact failure mode — so it was reviewed as production code. That call was right: rounds 2 and 3 each found a P2 that would have turned a CI leg red.

### What the loop actually bought

Three of the five rounds found something my own verification could not have. All three come from the same root — **I tested one configuration and reasoned about the rest**:

| round | finding | why my local run could not see it |
|---|---|---|
| 2 | the paste case asserted the shell RAN the pasted command — passes under zsh, **times out under bash and `/bin/sh`** | my shell is zsh. `defaultShellPath` falls back to `/bin/sh` when `SHELL` is unset, which is what ubuntu CI hits |
| 3 | commands chose their syntax from `process.platform`, but the **platform does not tell you the shell** — a Windows runner sets `%ComSpec%`, so PowerShell text would have been typed into **cmd.exe** | I cannot run Windows, and I had written the claim "powershell.exe on Windows" into three places |
| 3 | the replacement paste case **proved nothing**: Codex deleted both paste writes and it still passed | "the session works afterwards" is true of a write that never happened — a mutation I did not think to run |

The second one is the instructive one. I had reasoned "Windows means PowerShell" from reading `defaultShellTarget`, and that reasoning was wrong two lines further down in the same function. Both P2s were **verification gaps wearing the appearance of coverage**, which is precisely what a second reviewer is for.

### Ledger

| # | iter | finding | sev | disposition | why / where |
|---|------|---------|-----|-------------|-------------|
| 0 | pre-flight | reviewer sandbox capability | — | N/A | vitest, vue-tsc, socket bind, api.github.com and tmux-on-a-private-socket all exit 0. `rm -f` refused by policy; `cp`/`rm` work, so sweeps restore with `cp` + `cmp -s`. |
| 1 | 1 | the exactly-once proof applied to only the first case | P3 | FIXED | one shared `expectCommandOutput`, used by every output-bearing case. `ea5971e6` |
| 2 | 1 | teardown killed the shell but did not await its exit before `rmSync` | P3 | FIXED | `rmSync`'s `force` only swallows ENOENT, so a live cwd is EBUSY on Windows. `ea5971e6` |
| 3 | 1 | **(mine, not Codex's)** the liveness case's fast-diagnosis claim was measured against the wrong failure shape | P2 | FIXED | I had killed the pty at SPAWN; the report describes an exit at the first WRITE, where the old waiter burned the whole timeout. The wait now races token against exit. `ea5971e6` |
| 4 | 2 | the paste case is not cross-platform: zsh passes, bash and `/bin/sh` time out | P2 | FIXED | asserts the portable invariant instead; verified on four shell configurations. `a00de8d8` |
| 5 | 2 | the teardown cap winning was treated as success | P3 | FIXED | the race reports its winner and it is asserted. `a00de8d8` |
| 6 | 2 | the PR body still carried the round-1 false claim and a stale case count | P3 | FIXED | body rewritten; the plan's count **deleted** rather than corrected, per this repo's rule against figures in prose. `a00de8d8` |
| 7 | 3 | commands branched on `process.platform`; `%ComSpec%` means cmd.exe | P2 | FIXED | one `node -e` form every shell parses the same. `8af4789c` |
| 8 | 3 | the paste case passed with both paste writes deleted | P3 | FIXED | asserts the child's echo of the payload — the only portable evidence of arrival. `8af4789c` |
| 9 | 4 | "node present by definition" is broader than the contract: the child's PATH is sanitized | N | FIXED | `ptyEnv` strips yarn's shim dir and `node_modules/.bin`. `44207cff` |
| 10 | 5 | the plan still showed the old `echo` example | N | FIXED | prose only. `74123773` |

### Mutation coverage, stated honestly

Caught, each restoring byte-identical (`cmp -s`): the pty exiting at the first write; exiting at spawn; `write()` made a no-op; the exactly-once assertion dropped; a token placed in its own source; both paste writes deleted; the paste write deleted alone; PowerShell syntax run under bash.

**Not caught, and not claimed as covered:** removing the exit/timeout distinction (still red, but slowly and for the wrong reason — shown by measurement instead: ~6s saying `exited` against ~125s saying `timeout`), removing the readiness wait, and removing the teardown's awaited exit. The last two guard a slow runner and Windows `EBUSY`; neither is observable on a fast macOS shell, which is the honest limit of a local sweep here.

### Exit

Round 4 returned `LGTM` + `FINDINGS COMPLETE`; round 5 confirmed it on the prose fix. Both remaining notes were severity N and both are now fixed. The last commit is plan text only, with no code change.

**Ubuntu's three shards are green** — that is the leg the round-2 P2 would have turned red. The Windows leg is the one the issue exists for, and it is the ground truth this PR was opened to obtain.

isamu:
## The Windows answer

All three `Windows (PR)` shards are green on `74123773`.

That is the result this PR existed to obtain. [microsoft/node-pty#955](https://github.com/microsoft/node-pty/issues/955) reports the exact pinned version, `1.2.0-beta.15`, exiting with *"terminal process has exited"* at the first write; bisected, `beta.14` clean, unanswered since 2026-08-18. **It does not reproduce in this setup** — a live PTY on Windows takes a write, answers it, takes a second one, and survives an escape-wrapped payload.

So no downgrade to `beta.14`, and the question of whether `beta.14` carries #1595's fd-leak fix does not have to be opened. What changes is that the claim is now held by a test instead of by nobody: the next `node-pty` bump reports on this path rather than shipping past it.

Worth recording that the suite went red on the way here for two reasons that had nothing to do with node-pty — a shell-dialect assumption and a platform-is-not-a-shell assumption — and both were caught by review rather than by me. Details in the ledger above.

PR #2202, opened 2026-09-20
perf(files): restore the remembered expansions by level, not one at a time (#2148)

isamu:
## Summary

Stages 1 and 2 of #2148 shipped and are released. What remained in the issue is the **structure of the wait**, which neither stage touched:

> `restore()`: 記憶している展開ディレクトリ**ごとに** `/list` を1回、`for` ループで**直列**。つまり展開を N 個覚えていれば N+2 回の往復が順番待ちになります。

That is what this changes. **Nothing is cached and no request is removed** — the same listings are fetched, in fewer waits.

The ordering constraint is **between depths, not between siblings**: a directory cannot be looked up until the level above it has been read, which is why `restoreOrder` existed. But it returned a flat list and `restore()` awaited it one entry at a time, applying the depth rule to siblings that do not need it.

So `restoreOrder` becomes `restoreLevels`, returning `string[][]`. The grouping puts the constraint in the **type** — going back to serial now requires flattening on purpose — and `restore()` awaits a `Promise.all` per level, with the levels in order.

## Items to Confirm / Review

1. **The cache the issue also sketches is deliberately not here, and I would like that decision checked.** `toggleDir` reads `node.loaded` as "these children are real", so cached children either lie to it and are never refreshed, or are not marked and get fetched again anyway. Doing it safely means splitting `loaded` into "painted from cache" and "read from the server" — the same split stage 1 made for `roots` — plus a storage bound and a rule for clicking a since-deleted file. That is a larger change with a staleness surface; this one has none and removes most of the wait. The cache can still be built on top. Costed in the plan rather than attempted.

2. **`restoreOrder` → `restoreLevels` is a signature change** with one production caller and one spec. I changed the shape rather than adding a sibling helper, so that the flat list cannot be reintroduced by accident — say if you would rather keep both.

3. **Concurrency within a level is safe because siblings touch disjoint state**: each `toggleDir` mutates only its own node's `expanded`, `children` and `loaded`. Worth a second look, since it is the one thing the change actually relaxes.

4. **A failed parent still skips its children**, unchanged: the child's `findNode` returns null and it is passed over, exactly as before.

## Verification

- `format`, `lint`, `typecheck`, `build` clean, and `npx vitest run test/src/` is green.
- **End to end against a real server**, on a tree built for the purpose (four directories at depth 1, one nested), driving the real pane in a browser. The restore's `/list` requests:

  ```
  <root>@+0ms   alpha@+95ms   bravo@+95ms   charlie@+95ms   delta@+95ms   alpha/deep@+120ms
  ```

  The four depth-1 directories leave **together**; the depth-2 one follows only after that level is done. The restored tree is row-for-row identical to the tree before the reload, there is no "Empty directory." flash, and no console errors.

- **Break-verified**, each restoring byte-identical (`cmp -s`), against a baseline confirmed green first:

  | mutation | result |
  |---|---|
  | `restore()` awaits one directory at a time (the shape this replaces) | red |
  | every level fired at once, so a child rides with its parent | red |
  | levels collapsed into one | red |
  | one directory per level again | red |
  | levels returned deepest-first | red |
  | a path dropped on the way into the levels | red |

- The pane spec watches **concurrency, not duration** — how many `/list` calls are in flight at once. A timing assertion would be a flake; "three were in flight together" is exactly the difference between the two shapes, and serial never exceeds one.

**What I could not exercise:** a remote/high-latency host, which is where the issue's "数百 ms〜数秒" is felt. The win there is the same arithmetic with a larger constant — the number of waits drops from one per directory to one per level, and the localhost run above confirms the mechanism rather than the magnitude.

## User Prompt

- （マージ方針）収束 + CI green で自動マージ
- （このPR）https://github.com/receptron/mulmoterminal/issues/2148

Partially addresses #2148 — deliberately not `Closes`, because the expanded-directory cache in the issue is costed above rather than implemented.

work in mulmoterminal6

## Summary by CodeRabbit

- **Performance**
  - Restored expanded directories more efficiently by loading sibling directories concurrently while preserving parent-before-child ordering.
  - Improved performance across remembered directory structures with multiple branches and depths.

- **Tests**
  - Added coverage for concurrent sibling loading, sequential parent-child loading, depth grouping, ordering, and duplicate handling.

- **Documentation**
  - Added a performance plan describing the updated restoration approach and concurrency requirements.


isamu:

- P3: the regression tests can pass while non-root same-depth siblings are serialized, so the full “one wait per depth” contract is not pinned.
- N: one code comment overstates the depth shape as if the 200 remembered-path cap implies two levels.

| Axis | Result |
|---|---|
| CONCURRENCY | No production finding. Exact duplicate paths are deduped before work starts; prefix-only paths such as `ab`/`ab-c` stay distinct; same-level restore calls mutate the node they resolved, and `toggleDir` catches per-directory failures. A re-root can still let an older restore continue after an await because `toggleDir` has no tree generation check, but the visible stale-child write is to old node objects and the follow-on restore behavior is unchanged from the serial loop. Test coverage finding P3-T1 below is related. |
| ORDERING | No finding. `restoreLevels` sorts by slash-count depth first, so a child is not grouped with its slash-delimited parent for normal stored paths. Trailing/double-slash strings do not get normalized, but those strings are not produced by `expandedPaths`; if hand-injected, they fail lookup and are skipped rather than causing early child fetches. Prefix-only strings (`ab` vs `ab-c`) remain siblings, not ancestors. |
| IS THE DEPTH RULE EVEN RIGHT? | No production finding. Depth is not the same as “my parent is remembered”: input like `['a/b']` produces a depth-2 level, `findNode('a/b')` cannot see through unloaded `a`, and the path is skipped. That is the same behavior the old flat depth sort had, so this PR does not remove a fetch or alter the restored tree for that malformed/truncated remembered set. |
| BEHAVIOUR PRESERVATION | No finding. For valid snapshots produced by `expandedPaths`, the same distinct paths are attempted, parents still precede children, failed/missing paths are skipped, and `toggleDir` still fetches the same listings. |
| Signature change `restoreOrder` → `restoreLevels` | No finding. Grep shows one production caller (`FilesPane.vue`) and the pure spec updated; no production `restoreOrder` reference remains. Historical plan docs still mention the old name, but they are not part of this PR’s behavioral surface. |
| Tests | P3-T1. The tests assert root-level sibling concurrency and the pure helper’s root same-depth grouping, but a same-depth descendant split can still pass. |
| Correctness and edge cases; consistency with the codebase | No production finding. The helper keeps the existing path model: it does not normalize persisted strings, while the actual persisted paths come from tree nodes created without trailing or duplicate separators. |
| Error handling: one directory 404s | No finding. `toggleDir` catches a failed child listing and collapses only that node; because the promise resolves, `Promise.all` does not fail-fast and other siblings continue. |
| CLAIMS: comments, plan, PR body | N-C1. The concurrency/cache/no-request-removed claims match the code. One comment’s “cap’s worth ... is two levels deep” wording is overbroad. |

## Findings

### P3-T1: Add a regression for same-depth descendants under different parents

**Every site:**
- `test/src/components/filesTreeState.spec.ts:46` / `:47` only proves same-depth grouping with root paths (`src`, `server`, `test`, `docs`).
- `test/src/components/FilesPane.spec.ts:391` / `:399` only proves in-flight overlap for root siblings (`a`, `b`, `c`) before one deeper child (`a/deep`).

**Why it matters:** the PR’s performance contract is one wait per depth, not one wait per parent-at-depth. I mutation-tested a version of `restoreLevels` that still kept root siblings together, still kept parents before children, still kept every path, but split non-root same-depth directories by parent, e.g. `a/x` and `b/y` became separate sequential groups. The touched specs stayed green (`46 passed`). That mutation would preserve the restored tree but lose part of the concurrency this PR is meant to protect.

**Smallest change that resolves it:** add a pure helper assertion like:

```ts
expect(restoreLevels(['a', 'b', 'a/x', 'b/y'])).toEqual([['a', 'b'], ['a/x', 'b/y']]);
```

or extend the pane spec so `a/x` and `b/y` are both fetched concurrently after `a` and `b` have loaded.

**What would change my mind:** a test that fails if same-depth descendants under different parents are split into separate restore levels or otherwise restored serially.

### N-C1: Narrow the “two levels deep” claim

**Every site:**
- `src/components/filesTreeState.ts:38`: `Restoring the cap's worth of directories is two levels deep and was two hundred waits.`

**Why it matters:** `MAX_EXPANDED_PATHS` is 200, but I did not find a depth cap in `filesPaneStore`; a remembered set can theoretically be a 200-deep chain. The implementation handles that correctly, so this is only wording, not a behavior bug.

**Smallest change that resolves it:** reword to avoid tying the 200-path cap to a two-level shape, for example “A common two-level remembered set used to cost one wait per directory; grouping makes it one wait per level.”

**What would change my mind:** a code-level invariant that limits remembered expanded paths to two slash depths before `restoreLevels` sees them.

Verified locally: baseline `npm test -- test/src/components/filesTreeState.spec.ts test/src/components/FilesPane.spec.ts test/src/composables/useFilesTree.spec.ts` is green (68 tests). Mutation sweep used `cp` backup/restore and `cmp -s`; final `cmp_exit=0` and the working tree was clean.

isamu:
## Cross-review triage — round 1 (tier C)

**Tier C**: it relaxes ordering into concurrency on a path every pane open takes, and it changes a helper's signature. Both round-1 findings were accepted.

### The finding worth reading

Codex did not argue that my coverage was thin — it built the mutation that proves it. A `restoreLevels` that

- still grouped root siblings together,
- still put every parent before its children,
- still dropped nothing,

but split same-depth directories under **different parents** into separate sequential groups — `a/x` and `b/y` becoming two waits instead of one — **passed both specs**.

That mutation preserves the restored tree exactly. It gives back a large part of the concurrency this PR exists to create, and nothing would have said so. My tests only ever watched *root* siblings, so the contract I was claiming ("one wait per depth") was pinned only at depth one.

It is now pinned at both levels: the pure helper asserts the grouping, and a new pane case asserts `a/x` and `b/y` are in flight **together** after their parents load — counting only paths containing `/`, so the root level cannot satisfy it. I reconstructed Codex's mutation against a verified-green baseline and it goes red.

### The other one

A comment tied the remembered-path cap to a two-level shape. Nothing bounds how deep a remembered set goes, so the claim was simply wrong. The figures are **deleted rather than narrowed** — this repo's rule is that a number in prose is checked by a person and rots silently, and this one had already rotted before anyone read it. I removed the same kind of figure from the PR body for the same reason.

### Ledger

| # | iter | finding | sev | disposition | why / where |
|---|------|---------|-----|-------------|-------------|
| 0 | pre-flight | reviewer sandbox capability | — | N/A | vitest, vue-tsc, socket bind, api.github.com and tmux-on-a-private-socket all exit 0. `rm -f` refused by policy; `cp`/`rm` work. |
| 1 | 1 | same-depth siblings under different parents could be serialized with every test green | P3 | FIXED | pinned in the pure helper and the pane spec; the mutation now goes red. `1554dbc7` |
| 2 | 1 | a comment tied the path cap to a two-level shape | N | FIXED | figures deleted, not narrowed. `1554dbc7` |

Round 1 found **no production defect** — the concurrency, ordering, depth-rule, behaviour-preservation, signature and error-handling axes all came back clean, including the cases I asked it to attack (duplicate paths, `ab` vs `ab-c`, a remembered path whose parent is missing, a 404 in one sibling). What it found was a hole in the evidence, which is the more useful answer here.

isamu:

- P3: a new mutation still keeps the targeted tests green while weakening the “one wait per depth” contract: restore can chunk wide same-depth levels into multiple sequential batches.
- N: the stale two-level/path-cap claim was removed from the code comment and PR body, but it still exists in the new plan file.

| Axis | Result |
|---|---|
| Are both round-1 findings actually fixed, or fixed in appearance? | Fixed for the code/test surface that was called out. I reconstructed the round-1 “group by depth AND parent” mutation in `restoreLevels`; it now fails both `test/src/components/filesTreeState.spec.ts:56` and `test/src/components/FilesPane.spec.ts:439`. The new pane test cannot pass if deeper same-depth siblings are serialized by parent: `deepestTogether` only reaches `1`, so the `>= 2` assertion fails. It also cannot pass vacuously if no deeper fetch happens at all: then `deepestTogether` remains `0`, and the same assertion fails. No finding on this axis. |
| Find another mutation that keeps every test green while weakening the “one wait per depth” contract. | **P3-T2. Tests still allow chunking one depth into multiple sequential waits.** **Every site:** production contract is at `src/components/FilesPane.vue:199`-`205`; the pane assertions at `test/src/components/FilesPane.spec.ts:360`-`399` require only three root siblings in flight, and `test/src/components/FilesPane.spec.ts:408`-`439` requires only two deeper siblings in flight; the helper assertions in `test/src/components/filesTreeState.spec.ts` prove the returned grouping but do not prove the caller awaits the whole returned level as one batch. **Smallest change:** add a pane test with at least four directories in the same level, preferably same-depth descendants under different parents such as `a/x`, `b/y`, `c/z`, `d/w`, and assert that all four slash-containing paths are in flight together after their parents load. **What would change my mind:** that test, or equivalent coverage, goes red if `restore()` processes each `level` in sequential chunks such as `level.slice(i, i + 3)`. I proved the gap locally: after a green baseline, mutating `restore()` to await batches of three kept `npm test -- test/src/components/filesTreeState.spec.ts test/src/components/FilesPane.spec.ts test/src/composables/useFilesTree.spec.ts` green (`70 passed`), then I restored from `cp` backup and verified byte equality with `cmp -s`. |
| CONCURRENCY, again | No production finding. The current code does await one `Promise.all` over the whole returned level, so the implementation itself satisfies the wide-level case; P3-T2 is a missing regression guard, not a current behavior bug. The re-root / `toggleDir` generation point is still the same class as before: an old restore can continue after an await and mutate old node objects, and in a same-path new tree it may find current nodes on later iterations, but this was already possible in the serial loop and this PR does not add a new stale-write mechanism beyond more requests being in flight at once. |
| ORDERING and the depth rule | None. Parents stay in shallower groups than children, levels are awaited sequentially, and the accepted parent-split mutation now goes red. The code still uses slash-count depth rather than full ancestry, which preserves the old behavior for malformed/truncated remembered sets: if a parent is absent, the child is skipped because `findNode` cannot see it. |
| BEHAVIOUR PRESERVATION | None. For snapshots produced by `expandedPaths`, the same distinct directory paths are attempted, parents precede children, missing paths are skipped, duplicate stored paths are still deduped, and the restored visible tree is preserved. |
| tests: any remaining way this breaks with nothing going red | P3-T2 above. A caller can split each returned level into batches of three, preserving all existing behavior assertions and both new round-2 assertions while losing the “one wait per depth” property for wider levels. |
| correctness and edge cases; consistency with the codebase | None beyond P3-T2 as test coverage. `restoreLevels` follows the existing path model without normalization; persisted paths come from tree nodes, and hand-injected odd strings are bounded by existing parse caps and skipped if not found. The helper signature change is consistent with its single production caller. |
| error handling | None. `toggleDir` catches per-directory listing failures and collapses only that node, so `Promise.all` does not fail fast for a sibling 404; this matches the previous skip/continue behavior. |
| CLAIMS: every comment, the plan file and the PR body against the code | **N-C2. The stale “cap means two levels” claim remains in the plan.** **Every site:** `plans/perf-2148-restore-expansions-by-depth.md:22`-`24` says “a set of top-level directories is one level; the remembered-path cap is two.” That is the same unsupported depth claim as round 1, just in the plan instead of the code comment/PR body. **Smallest change:** delete that sentence or replace it with shape-based wording, e.g. “a set of top-level directories is one level; deeper remembered sets save according to their shape.” **What would change my mind:** a code-level invariant that limits the remembered expanded path set to two slash depths before `restoreLevels` sees it. The code comment at `src/components/filesTreeState.ts:38`-`39` and the PR body now use shape-dependent wording, so they are fine. |

isamu:
## Cross-review triage — round 2: the rule is inverted, not extended

Round 2 found a **second** way to weaken the "one wait per depth" contract with every test still green — chunking a level into batches of three, after round 1's split of same-depth directories by parent. Two rounds, two different breaches, both walking past a lower bound like *"at least three were in flight"*.

A third lower bound would have bought a fourth round. So the rule is inverted rather than extended, which is what this loop's own guidance says to do when the same class of finding keeps arriving: **state what is permitted instead of enumerating what is not.**

The pane spec no longer asserts a lower bound. It counts **waves** — how many times the in-flight `/list` set goes from empty to non-empty — and asserts that equals the root listing plus exactly one per depth:

```
expect(waves).toBe(3);   // the root, then depth 1, then depth 2
```

over `["a","b","c","d","a/x","b/y","c/z","d/w"]` — four wide at each depth, the deeper four under four different parents.

That is the contract itself rather than a proxy for it. Every way of breaking it produces **more** waves, so it does not matter whether anyone thought of that particular way:

| mutation | before | now |
|---|---|---|
| a level chunked into batches of three (round 2's finding) | green | **red** |
| same-depth directories split by parent (round 1's finding) | green until round 1's fix | **red** |
| back to the serial loop | red | **red** |
| every level fired at once | red | **red** |
| levels returned deepest-first | red | **red** |

One assertion replaces the two lower bounds and covers strictly more. Zero requests is zero waves, so it cannot pass vacuously either.

I have asked Codex, in round 3, for a form that gets past the wave count — that is now the entire remaining risk on this PR, and it is answerable in one pass.

### The other finding

`N-C2`: the "cap is two levels" claim **survived in the plan file** after I removed it from the code comment and the PR body. That is the claims-sweep failure in its textbook form — a claim living in more places than the one I was shown, and I fixed the two I was shown. All four surfaces swept this time.

### Ledger

| # | iter | finding | sev | disposition | why / where |
|---|------|---------|-----|-------------|-------------|
| 0 | pre-flight | reviewer sandbox capability | — | N/A | vitest, vue-tsc, socket bind, api.github.com, tmux on a private socket all exit 0; `rm -f` refused. |
| 1 | 1 | same-depth siblings under different parents could be serialized, all green | P3 | FIXED → subsumed | now covered by the wave count. `1554dbc7` / `4dec3473` |
| 2 | 1 | a comment tied the path cap to a two-level shape | N | FIXED | figures deleted rather than narrowed. `1554dbc7` |
| 3 | 2 | a level could be chunked into sequential batches, all green | P3 | FIXED | the wave count makes any extra wait red. `4dec3473` |
| 4 | 2 | the same stale claim survived in the plan file | N | FIXED | all four surfaces swept. `4dec3473` |

Still **no production defect** in two rounds: concurrency, ordering, the depth rule, behaviour preservation, the signature change and error handling have all come back clean, twice. Both findings were about the strength of the evidence, which on a change whose whole point is a timing property is the right thing for the loop to be arguing about.

isamu:

- Found one remaining test-contract hole: batching a level into groups of four keeps the suite green but restores wider same-depth levels in multiple request waves.
- Baseline checked first: `npx vitest run test/src/` passed, and the focused restored-head run passed (`test/src/components/FilesPane.spec.ts`, `test/src/components/filesTreeState.spec.ts`: 47 tests).
- Mutation hygiene: backed up with `cp`, restored with `cp`, and verified byte identity with `cmp -s`; no source changes left behind.

| Axis | Result |
|---|---|
| Can the wave count be defeated? | finding below |
| Is the wave count itself sound / flaky? | no flake finding. The stub’s two microtask yields are enough for the current `Promise.all(level.map(...))` shape because all fetch calls are started synchronously before any stubbed request can leave `inFlight`. I did find a microtask-stagger mutation that stays green, but it only delays request starts by microtasks and does not add a per-directory round trip, so I am not counting that as the latency-contract breach. |
| Are both round-2 findings actually fixed? | fixed in appearance for the named cases: per-parent splitting and batch-of-three chunking both produce extra waves against this fixture. The remaining issue is a new batch size equal to the fixture width. |
| CONCURRENCY / ORDERING / the depth rule | one finding: width-limited batching can pass while weakening same-depth concurrency beyond four siblings. Parent-before-child ordering itself is preserved. |
| BEHAVIOUR PRESERVATION | none. The production code still restores parents before children, skips missing paths as before, and preserves the open-file generation guard. |
| tests: any remaining way this breaks with nothing going red | finding below: batch-of-four mutation keeps focused specs green. |
| correctness and edge cases; consistency with the codebase | none beyond the test hole. `restoreLevels` groups by slash depth, de-dupes, orders predictably, and matches the existing path representation. |
| error handling | none. `toggleDir` still collapses a failed directory read and `restore()` continues with later paths/levels as before. |
| CLAIMS: comments, plan file, PR body against code | no stale two-level/cap claim found across the code, tests, plan, or PR body. The remaining overclaim is the test comment/name saying “however wide the level” while the fixture is only four-wide. |

## Findings

1. Severity: P2. Site(s): `test/src/components/FilesPane.spec.ts` lines 392-400, especially the four-wide `expanded` fixture and hard-coded `expect(waves).toBe(3)`; the vulnerable production mutation point is `src/components/FilesPane.vue` lines 199-205. The wave assertion can still be defeated by chunking each level into batches of four. I changed the restore loop to `for (let i = 0; i < level.length; i += 4) await Promise.all(level.slice(i, i + 4).map(...))`; the focused suite stayed green (`2 passed, 47 passed`). That implementation would do two same-depth waits for five remembered siblings, so it violates “one wave per depth” for wider real trees while the current four-wide test still observes root + depth 1 + depth 2 = 3 waves. Smallest change: make the pane fixture wider than any plausible hidden chunk by adding at least a fifth root sibling and a fifth depth-2 child under a fifth parent, e.g. `e` and `e/v`, and keep the expected wave count at 3 while asserting all expanded paths were requested. What would change my mind: the batch-of-four mutation above goes red, or an equivalent property-style test proves one wave for widths greater than four.

isamu:
## Cross-review triage — round 3: stop moving the constant, state the property

Round 3 found the **third** distinct way past this contract, and it was the sharpest one: it chunked each level into batches of **four**, and my fixture was four wide. Identical wave count. Green.

That is the lesson of the round. A wave count is only ever as strong as the level it is measured on, and the same was true of the lower bound before it. Three breaches, three rounds:

| round | the breach | what it defeated |
|---|---|---|
| 1 | same-depth directories split by parent | "at least three were in flight" |
| 2 | a level chunked into batches of three | the same lower bound, widened |
| 3 | a level chunked into batches of **four** | the wave count, on a four-wide fixture |

Codex's suggested fix was to widen the fixture — add a fifth sibling. That would have worked for a batch of four and bought a fifth round for a batch of five. **Widening only moves the constant**, so the assertion now states the property instead:

> **The peak number of requests in flight at each depth must EQUAL that depth's width.**

"The whole level went at once", said directly. A chunk of `k` on a level of width `W` peaks at `k`, so only `k >= W` passes — and `k >= W` means that level *was* fetched in one go, which is the correct behaviour rather than a hole. There is no constant left to move.

The fixture is eight wide at each depth, the deeper eight under eight different parents. Against a baseline confirmed green first, every batch size goes red:

| chunk size | 2 | 3 | 4 | 5 | 7 |
|---|---|---|---|---|---|
| result | red | red | **red** | red | red |

The batch of four — the one that beat the wave count — is among them. Restored byte-identical each time.

Codex's own "what would change my mind" was *"the batch-of-four mutation goes red, or an equivalent property-style test proves one wave for widths greater than four."* Both hold.

It also caught an overclaim I had written: the test was named *"however wide the level"* while the fixture was four wide. The name and comment now say what is actually asserted, including what `k >= W` means and why it is not a gap.

### A note on the record

**Round 3's process exited before posting**, and for a while this PR had a round that looked like it had never happened. I recovered the finding from the run's stderr transcript and started on it from there; the verdict did land afterwards, and the two agree. Flagging it because the loop's own rule is that a silent round and a clean round are indistinguishable — the monitor watching this now treats "the process is gone with no verdict" as an event rather than as silence.

### Not treated as a breach

Codex found a microtask-stagger mutation that stays green, and said itself it is not a latency breach: it delays request *starts* without adding a round trip, and the requests still overlap. I agree, and asked it to say in round 4 if that judgement is wrong.

### Ledger

| # | iter | finding | sev | disposition | why / where |
|---|------|---------|-----|-------------|-------------|
| 0 | pre-flight | reviewer sandbox capability | — | N/A | all green; `rm -f` refused, `cp`/`rm` fine. |
| 1 | 1 | same-depth siblings under different parents could be serialized | P3 | FIXED → subsumed | peak-equals-width. |
| 2 | 1 | a comment tied the path cap to a two-level shape | N | FIXED | figures deleted. |
| 3 | 2 | a level chunked into batches of three | P3 | FIXED → subsumed | peak-equals-width. |
| 4 | 2 | the stale claim survived in the plan file | N | FIXED | four surfaces swept. |
| 5 | 3 | the wave count died to a batch of four, the fixture's own width | P2 | FIXED | peak-equals-width, eight wide; 2/3/4/5/7 red. `16e757f8` |
| 6 | 3 | the test name overclaimed "however wide the level" | N | FIXED | restated. `16e757f8` |

Three rounds, **still no production defect**: concurrency, ordering, the depth rule, behaviour preservation, the signature change and error handling have come back clean every time. Every finding has been about the strength of the evidence — which, on a change whose entire point is a timing property, is the right argument for the loop to be having.

isamu:
- I could not find a change to `restore()` or `restoreLevels()` that makes restore add sequential network waits while keeping the touched specs green. Serial-in-level, chunked-by-7, per-parent grouping, and all-levels-at-once all went red after a green baseline.
- I agree the microtask-stagger case is not a breach: adding `await Promise.resolve()` before each sibling lookup stayed green because the `/list` requests still overlap; it delays scheduling, not round trips.
- One stale live comment still claims `restoreOrder` exists after this PR removes it.

| Axis | Result |
|---|---|
| Can the peak-equals-width assertion be defeated? | none for the requested contract. I tried serializing each level, chunking a level into batches of 7, splitting by parent in `restoreLevels()`, and flattening all levels at once. The first three dropped the peak below width; flattening skipped children before parents existed. The only green probe was microtask staggering, which does not add a request round trip. |
| Is it sound - can it miscount, or be flaky? | none. Counting only paths at the same depth avoids root/depth cross-talk, and the stub records the peak immediately after `fetch` starts, before its two microtask yields resolve, so same-tick starts overlap deterministically. A child-with-parent overlap is separately tracked. What would change my mind: a scheduler/runtime case where same-tick `Promise.all` fetch calls can resolve and remove themselves before the later callbacks in the same map run; I do not see that with this stub shape. |
| Is the round-3 finding actually fixed, or fixed in appearance? | fixed. A batch of four no longer passes: the fixture width is 8 and the assertion requires peak `=== WIDTH`, so any `k < W` batch peaks below width. I also tested `k = 7`, which failed at peak 7 vs 8. |
| CONCURRENCY / ORDERING / the depth rule | none. The production loop still awaits one `Promise.all` per level, so parents create child nodes before the child level is looked up. Same-depth siblings share one wait. All-level flattening went red because depth-2 paths never started. |
| BEHAVIOUR PRESERVATION | none. Requests are unchanged in identity and order by depth; deleted/missing paths still skip via `findNode` returning null; remembered open file and scroll restoration remain after directory restore. |
| tests: any remaining way this breaks with nothing going red | none found for added sequential network waits. Baseline `npx vitest run test/src/` was green: 328 files, 4756 tests. Final focused run of `FilesPane.spec.ts` and `filesTreeState.spec.ts` was green: 47 tests. |
| correctness and edge cases; consistency with the codebase | Finding P3. Severity: P3. Every site: `src/composables/useFilesReveal.ts:62`. Smallest change: replace the stale `restoreOrder` reference with `restoreLevels`, or reword it to say it follows the same parent-before-child rule as restore. What would change my mind: if comments are intentionally allowed to reference removed historical helper names, but that seems inconsistent with this PR renaming the helper and updating the adjacent comment in `filesTreeState.ts`. |
| error handling | none. `toggleDir` still collapses on listing failure, `restore()` still ignores missing/non-dir nodes, and `Promise.all` does not introduce a new rejection path because `toggleDir` catches listing errors. |
| CLAIMS: every comment, the plan file and the PR body against the code. The test comment now makes a specific claim about k >= W - check it. | Finding P3, same site as above: `src/composables/useFilesReveal.ts:62` says `restoreOrder` exists, but the new head removes it. New plan file, PR body, and the new test comment match the code: the `k >= W` claim is accurate because a chunk with `k >= W` covers the whole level of width `W` in this fixture, and anything `k < W` peaks below width. |

isamu:
## Cross-review triage — round 4: the inverted rule held

I asked Codex for a change to `restore()` or `restoreLevels()` that adds sequential waiting while keeping the suite green. It answered plainly:

> I could not find a change to `restore()` or `restoreLevels()` that makes restore add sequential network waits while keeping the touched specs green.

What it tried, each against a baseline confirmed green first and restored byte-identical: serializing within a level, chunking a level into batches of **seven**, splitting by parent inside `restoreLevels`, and flattening every level into one. All red — the first three by dropping the peak below the width, the last by starting children before their parents existed.

That is the answer the inversion was for. Three rounds of enumerating breaches bought three more breaches; stating the property ended it in one.

It also confirmed the two things I could have got wrong about the new assertion:

- **It cannot miscount.** Counting only paths at the same depth avoids root/depth cross-talk, and the peak is recorded before the stub's yields resolve, so same-tick starts overlap deterministically.
- **The `k >= W` claim in the comment is accurate** — a chunk that covers the whole level is not a split of it.

And it agreed the microtask-stagger case is not a breach: it delays scheduling, not round trips, and the requests still overlap.

### The one finding

`useFilesReveal.ts` explained its outermost-first ordering by pointing at **`restoreOrder`**, which this PR removes. A reader would have gone looking for a helper that no longer exists. Fixed.

I left the other surviving mentions deliberately: this PR's own plan describes the rename, and two older plan files record what was true when they were written. Rewriting those to match today's code would be falsifying a record, not fixing a claim.

### Ledger

| # | iter | finding | sev | disposition | where |
|---|------|---------|-----|-------------|-------|
| 0 | pre-flight | reviewer sandbox capability | — | N/A | all green; `rm -f` refused, `cp`/`rm` fine |
| 1 | 1 | same-depth siblings under different parents could be serialized | P3 | FIXED → subsumed | peak-equals-width |
| 2 | 1 | a comment tied the path cap to a two-level shape | N | FIXED | figures deleted |
| 3 | 2 | a level chunked into batches of three | P3 | FIXED → subsumed | peak-equals-width |
| 4 | 2 | the stale cap claim survived in the plan file | N | FIXED | four surfaces swept |
| 5 | 3 | the wave count died to a batch of four, the fixture's own width | P2 | FIXED | `16e757f8` |
| 6 | 3 | the test name overclaimed "however wide the level" | N | FIXED | `16e757f8` |
| 7 | 4 | a live comment still named the removed `restoreOrder` | P3 | FIXED | `384dcbbc` |

**Four rounds, no production defect.** Concurrency, ordering, the depth rule, behaviour preservation, the signature change and error handling came back clean in every round. Every finding was about the strength of the evidence — which is the right argument to have about a change whose entire deliverable is a timing property, and which is why the assertion that ships is a statement of the contract rather than a sample of it.

isamu:
## Codex round 5 — LGTM, relayed verbatim

**Codex could not post this one.** Its process wrote the verdict to stdout and exited before reaching the API, the same way round 3 nearly went unrecorded. Pasting it unedited rather than summarising, because an unposted verdict is a round whose only copy is my terminal:

```text
CODEX VERDICT: LGTM

- Round-4 finding is fixed: the live `useFilesReveal.ts` comment now names `restoreLevels`.
- Round-5 commit is prose-only; I saw no behavior movement.
- Baseline green before mutation trust: `npx vitest run test/src/components/filesTreeState.spec.ts test/src/components/FilesPane.spec.ts` passed, 47 tests.
- Mutation sweeps used `cp` backups and `cmp -s` restores, no `rm -f`: serial level, batch-of-7, per-parent split, and flattened levels all went red; final restored baseline passed again.

| Axis | Findings |
|---|---|
| Is the round-4 finding fixed, and did anything else move? | none. Fixed at `src/composables/useFilesReveal.ts:62`; only that comment changed in `384dcbbc`. |
| Can the peak-equals-width assertion be defeated? | none. I could not defeat it: serial peak `1/8`, batch-7 peak `7/8`, per-parent split depth-2 peak `1/8`, flattened levels skipped descendants and failed. |
| CONCURRENCY / ORDERING / the depth rule | none. `restore()` awaits one `Promise.all` per depth; children are only looked up after parent depth completes. |
| BEHAVIOUR PRESERVATION | none. Same listings are requested; no cache introduced; missing nodes remain skipped as before. |
| tests: any remaining way this breaks with nothing going red | none found. The peak assertion covers the surviving concurrency regressions I tried, and the helper specs cover grouping, ancestor separation, duplicates, and no drops. |
| correctness and edge cases; consistency with the codebase | none. `toggleDir` mutates only its own node, matching the sibling-concurrency claim. |
| error handling | none. Failed directory fetch still collapses that node; failed parent still means child `findNode` misses and is skipped, unchanged. |
| CLAIMS: every comment, the plan file and the PR body against the code. | none. Live code comments, new plan, and PR body match. I agree with leaving `restoreOrder` mentions in this PR’s plan and the two older plans as historical context. |

```

```text
Review PR #2202 at https://github.com/receptron/mulmoterminal/pull/2202, round 5, at head 384dcbbc.

── SAFETY ───────────────────────────────────────────────────────────────────
Never run tmux against `-L mulmoterminal` and never a bare `tmux`.

── WHAT CHANGED SINCE ROUND 4 ───────────────────────────────────────────────
Round 4 found one thing and I took it: `src/composables/useFilesReveal.ts:62` explained its
ordering by pointing at `restoreOrder`, which this PR removes. It now says `restoreLevels`.
Prose only, no behaviour, one line.

I deliberately left the other surviving mentions:
  - `plans/perf-2148-...md` — this PR's own plan, describing the rename itself
  - `plans/feat-910-...md`, `plans/feat-2099-...md` — older plans recording what was true
    when they were written
Say if you disagree with that line.

This round exists because that was a push, and a pushed change is unreviewed until the next
round. It should be short. Round 4 already reported that it could not defeat the
peak-equals-width assertion — serializing a level, chunking by seven, splitting by parent
and flattening every level all went red — and that the claims matched the code.

  Is the round-4 finding fixed, and did anything else move?
  Can the peak-equals-width assertion be defeated? (unchanged since round 4; re-state)
  CONCURRENCY / ORDERING / the depth rule
  BEHAVIOUR PRESERVATION
  tests: any remaining way this breaks with nothing going red
  correctness and edge cases; consistency with the codebase
  error handling
  CLAIMS: every comment, the plan file and the PR body against the code.

1. SEVERITY  2. EVERY SITE  3. THE SMALLEST CHANGE  4. WHAT WOULD CHANGE YOUR MIND.

`rm -f` is refused; back up and restore with `cp`, verify with `cmp -s`. Confirm the
baseline is green before trusting any result.

| # | iter | finding | disposition |
|---|------|---------|-------------|
| 0 | pre-flight | sandbox capability | N/A — all green; `rm -f` refused |
| 1 | 1 | same-depth siblings under different parents could be serialized | FIXED, subsumed by peak-equals-width |
| 2 | 1 | a comment tied the path cap to a two-level shape | FIXED, figures deleted |
| 3 | 2 | a level chunked into batches of three | FIXED, subsumed |
| 4 | 2 | the stale cap claim survived in the plan file | FIXED |
| 5 | 3 | the wave count died to a batch of four, the fixture's own width | FIXED — peak-equals-width, eight wide; 2/3/4/5/7 red |
| 6 | 3 | the test name overclaimed "however wide the level" | FIXED |
| 7 | 4 | a live comment still named the removed `restoreOrder` | FIXED — 384dcbbc |

  Then the axis table: one row per axis, findings or 'none'.
  Then 'FINDINGS COMPLETE: I read every hunk of the diff and this is every finding I have.'

```

</details>

### Loop closed

Five rounds on `384dcbbc`, nothing pushed since the head it reviewed. My own evaluation of that head has no MUST-FIX left: `format`, `lint`, `typecheck` and `build` are clean, `test/src/` is green, the change is verified end to end against a real server, and every mutation I or Codex could devise against the concurrency contract goes red.

**What made this loop five rounds** is worth recording, because it was not the diff — the production change is one helper and one loop, and no round ever found a defect in it. It was that the deliverable is a *timing property*, and my first three attempts to pin it were samples rather than statements: a lower bound, a wider lower bound, then a wave count. Each was defeated by a shape nobody had listed — a per-parent split, a batch of three, a batch of four that happened to equal my fixture width. Stating the contract instead (`peak in flight at each depth === that depth width`) ended it in one round, and round 4 then failed to defeat it on purpose.

